PR #4200, opened 2023-11-03
feat: add testnet profile to Cargo.toml

dandanlen:
This simply adds an optional testnet profile (`cargo build --profile=testnet`) that keeps debug assertions, debug symbols and overflow checks. 

We could use this for Sisyphos/Perseverance, and could use the `debug_assertions` to enable/disable testnet-only functionality. 

(i'm not entirely convinced that debug_assertions is a good idea, there are a few scenarios where this could cause a runtime panic...).

This is the same approach that is used by parity for polkadot testnets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,10 +106,15 @@ panic = "unwind"
 
 [profile.production]
 inherits = "release"
-
 # Sacrifice compile speed for execution speed by using optimization flags:
-
 # https://doc.rust-lang.org/rustc/linker-plugin-lto.html
 lto = "fat"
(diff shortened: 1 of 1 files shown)


kylezs:
Can we close this as stale and/or merge it? I'm not convinced that we should deploy this profile on persa - what we deploy on sisyphos, persa and mainnet should be exactly the same to avoid cases of compiler / library bugs that are introduced by adding/removing these flags. But I think this could be useful on devnets or localnets.

dandanlen:
Yeah not convinced this is useful.

PR #4202, opened 2023-11-03
feat/pro-609/order-validity-range

Janislav:
# Pull Request

Closes: PRO-615

## Checklist

## Summary


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -472,6 +532,26 @@ pub mod pallet {
 					}
 				}
 			}
+			let scheduled_limit_orders = ScheduledLimitOrders::<T>::take(current_block);
+			for order_update in scheduled_limit_orders {
```
dandanlen, line 536: Use `for OrderUpdate { lp, id, call } ` and you can avoid some unnecessary clones.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -472,6 +532,26 @@ pub mod pallet {
...
+				if let Err(err) = order_update
+					.clone()
+					.call
```
dandanlen, line 539: I think we might want to wrap the dispatch in a storage transaction to ensure we roll back state if it fails.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -217,11 +218,65 @@ pub mod pallet {
...
+		pub option_tick: Option<Tick>,
+		pub sell_amount: AssetAmount,
+	}
+
+	/// Represents the operational details of a scheduled limit order.
+	#[derive(Clone, Debug, TypeInfo, PartialEq, Eq, Encode, Decode, MaxEncodedLen)]
+	#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+	pub struct OrderScheduleDetails<BlockNumber: PartialOrd + Copy> {
+		/// The block number range during which the order is valid on entering state-chain.
+		pub valid_at: Option<Range<BlockNumber>>,
```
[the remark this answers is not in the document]
Janislav: Okay, I was thinking this is for checking when the transaction is actual entering the state-chain. When we craft the transaction we don't know when it finally gets picked up from the mem pool so we need some kind of validity checking if the transaction is still in the validity range. I can remeber we talked about that we want to keep it for that use case 🤔

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -140,6 +141,7 @@ pub trait Rpc {
 		id: OrderIdJson,
 		tick: Option<Tick>,
 		amount_change: IncreaseOrDecrease<NumberOrHex>,
+		details: Option<OrderScheduleDetails<BlockNumber>>,
```
AlastairHolmes, line 144: Shouldn't this be called something like scheduling, or schedule_details..., details is abit unclear, given these are visible to external users via the jsonrpcs.

api/lib/src/lp.rs
```diff
@@ -241,20 +243,39 @@ pub trait LpApi: SignedExtrinsicApi {
...
-				option_tick,
-				amount_change,
-			})
-			.await
-			.until_in_block()
-			.await?;
-
+		let events = if let Some(details) = details {
+			let (_tx_hash, events, ..) = self
+				.submit_signed_extrinsic(pallet_cf_pools::Call::schedule {
```
AlastairHolmes, line 251: You would dedup some codeby moving the update_limit_order construction outside the `if let Some()`

AlastairHolmes: You could then dedup more code, by introducing a function scheduled_or_immediate (or something like that) that takes a call, and an Option<OrderScheduleDetail>, and either schedules it or does the call immediately.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -472,6 +532,26 @@ pub mod pallet {
...
+						order_id: order_update.id,
+						error: err.error,
+					});
+				} else {
+					Self::deposit_event(Event::<T>::SuccessfullyExecutedLimitOrder {
+						lp: order_update.lp,
+						order_id: order_update.id,
+					});
+				}
+			}
```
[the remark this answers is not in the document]
Janislav: Or we remove all the manual calculation and just create a benchmark for it.

api/bin/chainflip-lp-api/src/main.rs
Janislav: Good point 👍

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1116,31 +1071,25 @@ pub mod pallet {
...
 		/// - [BadOrigin](frame_system::BadOrigin)
 		/// - [UnsupportedCall](pallet_cf_pools::Error::UnsupportedCall)
-		/// - [OrderScheduleDetailsInconsistent](pallet_cf_pools::Error::OrderScheduleDetailsInconsistent)
 		#[pallet::call_index(8)]
 		#[pallet::weight(T::WeightInfo::schedule())]
 		pub fn schedule(
 			origin: OriginFor<T>,
 			call: Box<Call<T>>,
-			details: OrderScheduleDetails<BlockNumberFor<T>>,
+			dispatch_at: BlockNumberFor<T>,
```
dandanlen, line 1082: We don't handle expired blocks (in the past) or if it's the current block.

dandanlen (commented):
I added checks to handle dispatch_at when it's in the past (error) or when it's the current block (dispatch immediately). 

I also did some renaming to avoid suggesting order executions. 


Btw. this will need a companion PR in the docs repo.

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: (fixed)

martin-chainflip:
This looks ok, but it only applies to limit orders and only supports a starting time. In the corresponding ticket PRO 615 we were discussing any LP order (including range orders) and also an endpoint on which the order expires automatically. These features are not in the PR?

martin-chainflip (approved):

Janislav:
Added docs PR here: https://github.com/chainflip-io/chainflip-docs/pull/114

martin-chainflip:
LGTM?

dandanlen:
Yep let's get this merged.

PR #4207, opened 2023-11-06
feat: broadcast barrier

ramizhasan111:
# Pull Request

Closes: PRO-949

## Checklist

## Summary

This includes the following changes:

- Broadcast pause is introduced with the broadcast of the rotation tx which pauses and queues all future new broadcast requests and only allows the retry of already initiated broadcasts upto and including the rotation tx.
- removes the key locking mechanism from the threshold signature pallet. The new key is immediately active and unlocked  and the vault is rotated immediately (optimistic rotation now for all chains) as soon as key activation is initiated.
- the ```vault_key_rotated``` extrinsic is removed since there are no tasks that need to be done on the successful broadcast of the key activation tx.


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -779,19 +816,35 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-					api_call,
-					RequestCallbacks::<T, I>::get(broadcast_id),
+				// Only storage to be removed, the other storages corresponding to this broadcast
+				// are either still valid or will get overwritten on re-threshold sign and
+				// broadcast.
+				TransactionOutIdToBroadcastId::<T, I>::remove(broadcast_attempt.transaction_out_id);
+
+				// We update the initiated_at here since as the tx is resigned and broadcast, it is
+				// not possible for it to be successfully broadcasted before this point.
+				let initiated_at = T::ChainTracking::get_block_height();
```
dandanlen, line 826: What does this mean for witnessing? For example, if one of the previous broadcasts is still pending in the witnesser, does this have any effect?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -874,16 +927,24 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
...
+	) -> BroadcastId {
 		Self::threshold_sign_and_broadcast(api_call, Some(callback))
 	}
+
+	fn threshold_sign_and_broadcast_rotation_tx(api_call: Self::ApiCall) -> BroadcastId {
+		let broadcast_id = Self::threshold_sign_and_broadcast(api_call, None);
+		let mut broadcast_barriers = <<<T as pallet::Config<I>>::TargetChain as Chain>::ChainCrypto as ChainCrypto>::maybe_broadcast_barriers_on_rotation(
+			broadcast_id,
+		);
+		broadcast_barriers.reverse();
```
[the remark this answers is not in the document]
ramizhasan111: > I agree the whole reverse/last is a bit weird. What's wrong with just using `first`?

The reason to do the reverse is so that we can use ```pop()```. `there is no pop_front in Vec. I guess it is better to just use VecDeque and use ```pop_front```

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -793,19 +830,35 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
 				log::info!(
 					"Signature is invalid -> rescheduled threshold signature for broadcast id {}.",
 					broadcast_id
 				);
-				Self::deposit_event(Event::<T, I>::ThresholdSignatureInvalid {
-					broadcast_id,
-					retry_broadcast_id,
-				});
+				Self::deposit_event(Event::<T, I>::ThresholdSignatureInvalid { broadcast_id });
```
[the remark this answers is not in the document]
ramizhasan111: will do

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -400,9 +408,29 @@ pub mod pallet {
...
+						} else {
+							true
+						}
+					});
+				}
+
 				if retries.len() > num_retries_that_fit {
-					BroadcastRetryQueue::<T, I>::put(retries.split_off(num_retries_that_fit));
+					paused_broadcasts.append(&mut retries.split_off(num_retries_that_fit));
 				}
```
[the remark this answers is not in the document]
ramizhasan111: > IMO the follow reads quite nicely, there is no mutable Vec, no if. It reads like we want: From the retry queue, extract any value below the limit, take at most `num_retries_that_fit`.
> 
> ```rust
> 				let retries = BroadcastRetryQueue::<T, I>::mutate(|retry_queue| {
> 					let id_limit =
> 						BroadcastPause::<T, I>::get().unwrap_or(BroadcastId::max_value());
> 					retry_queue
> 						.extract_if(|broadcast| {
> 							broadcast.broadcast_attempt_id.broadcast_id < id_limit
> 						})
> 						.take(num_retries_that_fit)
> 						.collect::<Vec<_>>()
> 				});
> ```

Looks cleaner, will make the change.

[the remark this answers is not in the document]
ramizhasan111: As a discussion point though, I dont see how in this case a min/max is a better choice over a simple if statement. In its implementation, at the innermost level of the function, the min/max function would have an if statement to do the comparison to find the min/max and the compiler would still have to compile that.

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: @kylezs maybe you can answer this - will this cause any side effects in the engine?

state-chain/runtime/src/migrations.rs
```diff
@@ -0,0 +1,50 @@
...
+		if System::runtime_version().spec_version == SPEC_VERSION {
+			U::pre_upgrade()
+		} else {
+			Ok(vec![])
+		}
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(state: Vec<u8>) -> Result<(), DispatchError> {
+		if System::runtime_version().spec_version == SPEC_VERSION {
```
dandanlen, line 46: Is this correct (`System::runtime_version().spec_version == SPEC_VERSION` in each of these methods)? Shouldn't some of them be `<` or `>` ?

state-chain/pallets/cf-vaults/src/vault_rotator.rs
```diff
@@ -152,38 +152,18 @@ impl<T: Config<I>, I: 'static> VaultRotator for Pallet<T, I> {
...
-							PendingVaultRotation::<T, I>::put(
-								VaultRotationStatus::<T, I>::AwaitingActivation { new_public_key },
-							);
-						}
+						T::Broadcaster::threshold_sign_and_broadcast_rotation_tx(activation_call);
+
+						Self::activate_new_key(
+							new_public_key,
+							T::ChainTracking::get_block_height(),
+						);
```
dandanlen, line 166: Just thinking about what this means for Ethereum:

We activate the new key, so if a threshold ceremony needs to be retried, we would sign with the new key? Isn't this a bug? 

If we have the broadcast barrier, then only old ceremonies will be retried, but we will be retrying with the new key (?).

state-chain/runtime/src/chainflip/broadcaster.rs
```diff
@@ -38,10 +38,12 @@ impl Broadcaster<AnyChain> for GovKeyBroadcaster {
 			AnyChainApi::Ethereum(eth_api_call) =>
 				<EthereumBroadcaster as Broadcaster<Ethereum>>::threshold_sign_and_broadcast(
 					eth_api_call,
+					false,
 				),
 			AnyChainApi::Polkadot(dot_api_call) =>
 				<PolkadotBroadcaster as Broadcaster<Polkadot>>::threshold_sign_and_broadcast(
 					dot_api_call,
+					false,
```
dandanlen, line 46: This impl seems wrong/incomplete... I can't find a matching trait definition anywhere. What am I missing?

dandanlen: Looks like this module is never actually declared! Let's delete the file.

state-chain/runtime/src/migrations.rs
ramizhasan111: should be correct. This migration will only run if the post runtime upgrade spec version matches exactly to SPEC_VERSION for which this migration is created for as says in the comment. we then initialise this versioned migration with the spec version we want it to apply to (one plus the current) in the runtime lib file.

state-chain/runtime/src/chainflip/broadcaster.rs
ramizhasan111: yeah, it is unused. will deleter the file

state-chain/pallets/cf-broadcast/src/lib.rs
kylezs: It's possible it does impact witnessing yeah, will have to look a bit deeper to confirm. Relevant code is in `engine/src/witness/common/chunked_chain_source/chunked_by_vault/deposit_addresses.rs` . Otherwise I'll have a look in a bit.

kylezs: Ok, so there might be an issue here - if we're resetting the initiated at this scenario could occur:
1. sign tx with initiated at 20
2. finalised witnessing is at block 22
3. tx gets into block 24 - we have witnessed, since we're slightly behind
4. timeout on SC occurs verification fails
5. block head is at 26, so we resign and set initiated at to block 26
6. Cfe gets to finalised witnessing at block 24, however when it queries storage for what it should witness it filters out the transaction it should because it sees that it's only potentially valid from block 26, so we don't witness it.

so the correct thing to do would be if there already exists a TransactionOutId of this type, we use the lower initiated_at. 

Deleting the `TransactionOutIdToBroadcastId` as is done above here will mean we won't witness it if the transaction is on the external chain in the meantime, which is basically the same issue.

kylezs: also this is a decent chunk of dup'd code

dandanlen: So basically we just don't need to do this resetting of the `initiated_at`, right?

kylezs: Yep. But we set it for a new one ofc.

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -810,14 +806,11 @@ where
...
-		let request_type = if <T::TargetChainCrypto as ChainCrypto>::sign_with_specific_key() {
-			T::KeyProvider::active_epoch_key().defensive_map_or_else(
-				|| RequestType::CurrentKey,
-				|EpochKey { key, epoch_index, .. }| RequestType::SpecificKey(key, epoch_index),
-			)
-		} else {
-			RequestType::CurrentKey
-		};
+		let request_type = T::KeyProvider::active_epoch_key().defensive_map_or_else(
+			|| RequestType::SpecificKey(Default::default(), Default::default()),
```
dandanlen, line 810: If we use `RequestType::CurrentKey` it will fail more gracefully (it will emit `CurrentKeyUnavailable` and retry with CurrentKey until one is available).

state-chain/pallets/cf-broadcast/src/lib.rs
ramizhasan111: The case you are describing above is similar to the one discussed in https://github.com/chainflip-io/chainflip-backend/pull/4262. This can only happen in the unlikely case of the tx witness delay anomaly and the polkadot Runtime version update happening at the same time which themselves own their own, rare events. Moreover, as dan said if in that case we do witness the dot runtime version update, it means the witnessing has been done uptil that point and if the tx hasnt gone through until then, it will fail.

I think it is safe to reset the initiated_at when re signing the tx.

state-chain/pallets/cf-threshold-signature/src/lib.rs
ramizhasan111: I have removed the ```CurrentKey``` variant in the ```RequestType``` since we dont use it anymore. The ceremonies will always be signed with key active at the time of threshold signature request since do optimistic rotation for all chains now and every ceremony has a key associated to it.

dandanlen (approved):

PR #4222, opened 2023-11-08
feat: rpc to get list of supported assets.

marcellorigotti:
# Pull Request

Closes: PRO-953

## Checklist

## Summary

Added custom RPC to retrieve all the available Assets as a Vec[RpcAsset]

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -458,7 +458,7 @@ pub trait CustomApi {
 	) -> RpcResult<Option<Vec<AssetAmount>>>;
 
 	#[method(name = "supported_assets")]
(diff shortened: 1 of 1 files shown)


state-chain/custom-rpc/src/lib.rs
```diff
@@ -456,6 +456,9 @@ pub trait CustomApi {
 		to_asset: RpcAsset,
 		at: Option<state_chain_runtime::Hash>,
 	) -> RpcResult<Option<Vec<AssetAmount>>>;
+
+	#[method(name = "supported_assets")]
+	fn cf_supported_assets(&self) -> RpcResult<HashMap<ForeignChain, Vec<Asset>>>;
```
dandanlen, line 461: Please make sure this works at runtime - sometimes serde fails to serialize HashMaps (I think it's ok in this case because ForeignChain serializes to json String, but best to be sure).

PR #4228, opened 2023-11-10
feat: CFE processes SC events in the initial block on startup

j4m1ef0rd:
# Pull Request

Closes: PRO-918

## Checklist

## Summary

- Before the SCO starts it adds the cached block to the front of the stream so it will process the events in it.
- The multisig clients now start with the ceremony id counter set to what it would have been at the start of the initial block instead of the end.
	- To avoid accessing  the prev block, I created a function that checks for the events in a block that would change the ceremony id and calculates what the ceremony id counter would be at the start of the block.
- Moved the p2p registered event monitoring to a function so it can be called from within the p2p module instead of the SCO. So it uses a stream that does not include the initial block.


engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -286,7 +285,15 @@ where
 
         info!("Sending heartbeat every {blocks_per_heartbeat} blocks");
 
-        let mut sc_block_stream = Box::pin(sc_block_stream);
+        // Add the initial (cached) block to the stream so we can process the events in it.
+        let mut sc_block_stream =
+        Box::pin(
+            futures::stream::iter(vec![BlockInfo{
```
AlastairHolmes, line 291: The cache's type is already BlockInfo so you can just clone/copy it.

AlastairHolmes: Also you can use std::iter::once, or stream::once()

engine/multisig/src/crypto.rs
```diff
@@ -35,7 +35,7 @@ pub const MAX_SCALAR_SIZE: usize = 32;
 /// Used as a unique identifier when serializing/deserializing chain specific data.
 /// The values are explicitly given and should never be changed.
 #[repr(u16)]
-#[derive(Clone, Copy, Debug, FromPrimitive)]
+#[derive(Clone, Copy, Debug, FromPrimitive, PartialEq, Eq, PartialOrd, Ord)]
```
AlastairHolmes, line 38: What is this for? Particularly PartialOrd and Ord seem undesirable to impl.

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -619,3 +612,163 @@ where
...
+		Ok(events) => {
+			for event_record in events {
+				match event_record.event {
+					state_chain_runtime::RuntimeEvent::EthereumVault(
+						pallet_cf_vaults::Event::KeygenRequest { .. },
+					) |
+					state_chain_runtime::RuntimeEvent::EthereumThresholdSigner(
+						pallet_cf_threshold_signature::Event::ThresholdSignatureRequest { .. },
+					) => {
+						ceremony_id_counters.ethereum -= 1;
```
AlastairHolmes, line 677: I think it would make more sense to simply look at the first event and sub one from its ceremony id (if not zero). And if there are no events get from the storage.

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -297,29 +304,13 @@ where
...
-                                            .unwrap();
-                                    }
+                                        pallet_cf_validator::Event::PeerIdRegistered(..),
+                                    ) |
                                     state_chain_runtime::RuntimeEvent::Validator(
-                                        pallet_cf_validator::Event::PeerIdUnregistered(
-                                            account_id,
-                                            ed25519_pubkey,
-                                        ),
+                                        pallet_cf_validator::Event::PeerIdUnregistered(..),
```
AlastairHolmes, line 310: You can delete these branches from the match

AlastairHolmes: If you want to log something move the macro into a shared location and use the macro, in the monitor_p2p_registration_events function.

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -619,3 +612,163 @@ where
...
+		},
+		Err(error) => {
+			bail!("Failed to decode events at block hash {block_hash}. {error}");
+		},
+	}
+
+	Ok(ceremony_id_counters)
+}
+
+pub async fn monitor_p2p_registration_events<StateChainClient, BlockStream: StateChainStreamApi>(
```
AlastairHolmes, line 715: Shouldn't this be with the p2p code?

engine/multisig/src/crypto.rs
j4m1ef0rd: Left over from some experimenting that i did, forgot that i didn't need it anymore.

engine/src/state_chain_observer/sc_observer/mod.rs
j4m1ef0rd: The problem is that the macro logs un-matched events with `Ignoring event`.So if used in 2 places, it will log deceptive ignore messages. Was not sure what to do about this.

engine/src/state_chain_observer/sc_observer/mod.rs
j4m1ef0rd: I couldn't decide. Thought it was nice to keep all the event matching (observing) in the SCO. But if you think it's better in p2p i will move it.

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -529,3 +524,80 @@ where
...
+				.storage_value::<pallet_cf_vaults::CeremonyIdCounter<
+					state_chain_runtime::Runtime,
+					state_chain_runtime::PolkadotInstance,
+				>>(block_hash)
+				.await
+				.context("Failed to get Polkadot CeremonyIdCounter from SC")?
+		},
+		bitcoin: if let Some(ceremony_id) = events.iter().find_map(|event| match event {
+			CfeEvent::BtcThresholdSignatureRequest(req) => Some(req.ceremony_id),
+			CfeEvent::BtcKeygenRequest(req) => Some(req.ceremony_id),
```
msgmaxim, line 589: I think you need to include key handover request here too.

j4m1ef0rd: Good catch. I changed the unit test to cover this sort of mistake.

msgmaxim (approved):

j4m1ef0rd:
Going to merge now because I have another PR that will affect this code. I will make any changes retroactively in that next PR. @AlastairHolmes

PR #4252, opened 2023-11-15
Fix: avoid breaking change when submitting Node version

marcellorigotti:
# Pull Request

Closes: PRO-970

## Checklist

## Summary

- Removed the NodeVersionUpdated event, which is not used and is only introducing a breaking change
- Added back old extrinsic to only report the cfe version to ensure compatibility of old engines (not sure if this is required since we are not going to use it, because this extrinsic gets called only when a new version of the cfe is used, hence the new cfe knows already to use the new extrinsic)

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -312,11 +312,6 @@ pub mod pallet {
 			old_version: Version,
 			new_version: Version,
 		},
-		NodeVersionUpdated {
-			account_id: ValidatorIdOf<T>,
-			old_version: Version,
-			new_version: Version,
-		},
 		/// An authority has register her current PeerId \[account_id, public_key, port,
 		/// ip_address\]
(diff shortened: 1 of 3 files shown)


kylezs:
What's the status here?

marcellorigotti:
Decided to postpone this since there is not currently need to have the node version rn. When/If we'll need it we can come back to this.

dandanlen:
For visibility: this is a tricky one to make backwards-compatible (or even compatible with our current upgrade mechansim) because the version info needs to be submitted *before* the engine has determined if it is compatible. However, the new version info (engine+node) is only valid if we *are already* compatible (if the upgrade has already happened). So the problem is not just the event.

Ultimately the most pain-free way to add this may be to create a new storage item with a new extrinsic, and some custom if/else code in the engine to conditionally submit just the engine version if we're on the old runtime, or the new engine version if we've updated... and then remove the dead code later on. But this is still a pain.

I don't really see a clear reason why we need to introduce this in the first place tbh, so for now i would suggest we revert the original change.

kylezs:
Think we can close as stale @marcellorigotti ?

PR #4257, opened 2023-11-17
Test: added tests to swapping pallet

marcellorigotti:
# Pull Request

Closes: PRO-966

## Checklist

## Summary

Added tests:
- check broker fee calculation is correct
- check broker fee cannot exceed amount swapped
- check broker fee is correctly subtracted from the amount swapped
- check swap fails if the amount swapped is too low

### DIFF
--- a/state-chain/pallets/cf-swapping/src/tests.rs
+++ b/state-chain/pallets/cf-swapping/src/tests.rs
@@ -1,3 +1,4 @@
+use super::*;
 use crate::{
 	mock::{RuntimeEvent, *},
 	CcmFailReason, CcmGasBudget, CcmIdCounter, CcmOutputs, CcmSwap, CcmSwapOutput,
@@ -2084,8 +2085,6 @@ fn swap_with_custom_broker_fee(
 	amount: AssetAmount,
 	broker_fee: BasisPoints,
 ) {
-	const ALICE: u64 = 2_u64;
-
 	<Pallet<Test> as SwapDepositHandler>::schedule_swap_from_channel(
 		ForeignChainAddress::Eth([2; 20].into()),
 		Default::default(),
@@ -2102,11 +2101,10 @@ fn swap_with_custom_broker_fee(
 #[test]
 fn swap_broker_fee_calculated_correctly() {
 	new_test_ext().execute_with(|| {
-		const ALICE: u64 = 2_u64;
 		let fees: [BasisPoints; 12] =
 			[1, 5, 10, 100, 200, 500, 1000, 1500, 2000, 5000, 7500, 10000];
 		const AMOUNT: AssetAmount = 100000;
-		const BASIS_POINTS_PER_MILLION: u32 = 100;
+
 		// calculate broker fees for each asset available
 		Asset::all().iter().for_each(|asset| {
 			let total_fees: u128 =
@@ -2123,7 +2121,6 @@ fn swap_broker_fee_calculated_correctly() {
 #[test]
 fn swap_broker_fee_cannot_exceed_amount() {
 	new_test_ext().execute_with(|| {
-		const ALICE: u64 = 2_u64;
 		swap_with_custom_broker_fee(Asset::Usdc, Asset::Flip, 100, 15000);
 		assert_eq!(EarnedBrokerFees::<Test>::get(ALICE, cf_primitives::Asset::Usdc), 100);
 	});
@@ -2134,7 +2131,7 @@ fn swap_scheduled_event_witnessed(
 	source_asset: Asset,
 	deposit_amount: AssetAmount,
 	destination_asset: Asset,
-	broker_commission: Option<AssetAmount>,
+	broker_commission: AssetAmount,
 ) {
 	System::assert_last_event(RuntimeEvent::Swapping(Event::<Test>::SwapScheduled {
 		swap_id,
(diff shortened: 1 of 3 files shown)


kylezs:
Itertools is the first place to look for fancy iterator operations: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.cartesian_product

marcellorigotti:
> Itertools is the first place to look for fancy iterator operations: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.cartesian_product

Yeah I saw that, I was wondering if maybe there was a straightforward way without relying to third party libraries

kylezs:
We already use this library in some places, and you can just import it in dev depeendencies, so doesn't increase size of the main binary.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2075,3 +2077,156 @@ fn can_swap_ccm_below_max_amount() {
...
+		broker_commission,
+	}));
+}
+#[test]
+fn swap_broker_fee_subtracted_from_swap_amount() {
+	new_test_ext().execute_with(|| {
+		const ALICE: u64 = 2_u64;
+		let amounts: [AssetAmount; 6] = [50, 100, 200, 500, 1000, 10000];
+		let fees: [BasisPoints; 4] = [100, 1000, 5000, 10000];
+		const BASIS_POINTS_PER_MILLION: u32 = 100;
```
kylezs, line 2160: this const already exists, we can just import the one in the pallet

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2075,3 +2077,156 @@ fn can_swap_ccm_below_max_amount() {
...
+		assert_eq!(EarnedBrokerFees::<Test>::get(ALICE, cf_primitives::Asset::Usdc), 100);
+	});
+}
+
+fn swap_scheduled_event_witnessed(
+	swap_id: u64,
+	source_asset: Asset,
+	deposit_amount: AssetAmount,
+	destination_asset: Asset,
+	broker_commission: Option<AssetAmount>,
```
kylezs, line 2137: nit: we always pass in a `Some()` to this, so we can dedup the wrapping and just wrap it in some in here

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2075,3 +2077,156 @@ fn can_swap_ccm_below_max_amount() {
...
+
+		assert_eq!(MinimumSwapAmount::<Test>::get(asset), amount);
+		assert_eq!(Swapping::minimum_swap_amount(asset), amount);
+
+		System::assert_last_event(RuntimeEvent::Swapping(Event::<Test>::MinimumSwapAmountSet {
+			asset,
+			amount,
+		}));
+
+		const ALICE: u64 = 2_u64;
```
kylezs, line 2215: We can just pull the const out of the `swap_with_custom_broker_fee` and then use it here - they need to be linked, so we should create that link explicitly by using the same const

same in the other place/s we do this too ofc

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2075,3 +2077,156 @@ fn can_swap_ccm_below_max_amount() {
...
+			channel_id: 1,
+			deposit_block_height: Default::default(),
+		},
+	}));
+}
+#[test]
+fn swap_fail_if_below_minimum_swap_amount() {
+	new_test_ext().execute_with(|| {
+		let asset = Asset::Usdc;
+		let amount = 1_000u128;
```
kylezs, line 2201:
```suggestion
		let minimum_amount = 1_000u128;
```

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2075,3 +2077,156 @@ fn can_swap_ccm_below_max_amount() {
...
+
+		const ALICE: u64 = 2_u64;
+		//swap with amount >= MinimumAmount is scheduled
+		swap_with_custom_broker_fee(Asset::Usdc, Asset::Flip, 1500, 1000);
+		assert_eq!(EarnedBrokerFees::<Test>::get(ALICE, cf_primitives::Asset::Usdc), 150);
+		swap_scheduled_event_witnessed(1, Asset::Usdc, 1500, Asset::Flip, Some(150));
+
+		//Swap with amount == MinimumAmount completes (even after subtracting brokerFee, netAmount
+		// is < minAmount)
+		swap_with_custom_broker_fee(Asset::Usdc, Asset::Flip, 1000, 1000);
```
kylezs, line 2223:
```suggestion
		swap_with_custom_broker_fee(Asset::Usdc, Asset::Flip, minimum_amount, 1000);
```
Looks like a magic value otherwise, but actually they're directly related to the value we set above - for the other tests we can do `minimum_amount + 500` and `minimum_amount - 500` - the test is then telling the story that these are relative to this value.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2075,3 +2077,156 @@ fn can_swap_ccm_below_max_amount() {
...
+}
+#[test]
+fn swap_broker_fee_subtracted_from_swap_amount() {
+	new_test_ext().execute_with(|| {
+		const ALICE: u64 = 2_u64;
+		let amounts: [AssetAmount; 6] = [50, 100, 200, 500, 1000, 10000];
+		let fees: [BasisPoints; 4] = [100, 1000, 5000, 10000];
+		const BASIS_POINTS_PER_MILLION: u32 = 100;
+
+		let combinations = amounts.iter().cartesian_product(fees);
```
kylezs, line 2162: 🔥  nice!

state-chain/pallets/cf-swapping/src/tests.rs
marcellorigotti: My bad here we already have a global ALICE we can use for the tests

kylezs (approved):

kylezs:
Some slick lookin' tests 👌

PR #4260, opened 2023-11-20
feat: store witness safety margins on SC

msgmaxim:
# Pull Request

Closes: PRO-983

## Checklist

## Summary

As discussed, used storage itmes instead of constants so we can have different values depending on the network. Everything expect for Perseverance is set to 2 blocks (it is 5 blocks on Perseverance as requested in PRO-983)


msgmaxim:
Added assertions so that the migration can be tested, and checked it with `try_runtime_upgrade.ts` scirpt. @kylezs

engine/src/witness/btc.rs
```diff
@@ -134,9 +132,20 @@ where
...
 		.spawn(scope);
 
+	let btc_safety_margin = state_chain_client
+		.storage_value::<pallet_cf_ingress_egress::WitnessSafetyMargin<
+			state_chain_runtime::Runtime,
+			state_chain_runtime::BitcoinInstance,
+		>>(state_chain_stream.cache().hash)
+		.await?
+		// Default to berghain in case the value is missing (e.g. during initial upgrade)
+		.unwrap_or(BITCOIN_SAFETY_MARGIN);
```
kylezs, line 142: Hmm, this has the same issue of which network we're on I guess - perhaps we should do the same thing as we did in the migrations and use the NetworkEnvironment, and then default based on that, else we risk some issues when the update is down on perseverance, 2/3 block reorgs on bitcoin testnet are fairly common

kylezs (approved):

kylezs (approved):

PR #4264, opened 2023-11-20
fix/PRO-982/minimum-swap-amount-check

Janislav:
# Pull Request

Closes: PRO-982

## Checklist

## Summary

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -622,6 +622,7 @@ pub mod pallet {
 				from,
 				to,
 				deposit_amount,
+				0,
 				destination_address_internal.clone(),
 				&swap_origin,
 			) {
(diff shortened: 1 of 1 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1044,13 +1048,14 @@ pub mod pallet {
...
 				amount,
+				net_amount,
 				destination_address.clone(),
 				&swap_origin,
 			) {
 				Self::deposit_event(Event::<T>::SwapScheduled {
 					swap_id,
 					source_asset: from,
-					deposit_amount: amount,
+					deposit_amount: net_amount,
```
dandanlen, line 1058:
```suggestion
					deposit_amount: amount,
```

The event's deposit_amount should match the amount deposited.

dandanlen (commented):
When reviewing this I noticed that the broker's fee is earned even if the swap is rejected. I think we should change this. I'll raise it on Linear.

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: I don't think this is right. We want to emit the same amount we did before, right? Or was this also wrong before? `amount` is now the raw amount and `net_amount` is `amount - fees`.

Janislav:
Not 100% sure if this is really an issue though. The broker has done his job. Why should he not get his earned fees even if we swallow the swap amount.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: I think it makes more sense to emit the amount deposited in `deposit_amount` (ie. it was wrong before). But maybe this also depends on how it's used in teh swapping app. @acdibble ?

dandanlen:
> Not 100% sure if this is really an issue though. The broker has done his job. Why should he not get his earned fees even if we swallow the swap amount.

Because part of the broker's job is spam prevention.

dandanlen (approved):

PR #4286, opened 2023-11-29
test: polkadot runtime update bouncer test

j4m1ef0rd:
# Pull Request

Closes: PRO-632

## Checklist

## Summary

- If the current runtime is the default, the test uses a precompiled runtime file that is in the repo. If it needs a different runtime, it will compile it (takes longer).
- Moved the code for the`alice` polkadot account to file so it can be common, this stops nonce issues.
- Added an argument to the `testSwap` function that can be used to stop logging. So the test is not flooded with swap logs.
- Created new code to decoding a polkadot dispatch errors because the existing code did not work, why is it different from the SC?.


bouncer/tests/polkadot_runtime_update.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  await testSwap(sourceAsset, destAsset, undefined, undefined, undefined, undefined, false);
+  swapsComplete++;
+}
+
+async function doPolkadotSwaps(): Promise<void> {
+  console.log('Running polkadot, new swap every 1 second');
+  while (!runtimeUpgradeComplete) {
+    randomPolkadotSwap();
```
kylezs, line 195: Got this error:
```
Preimage hash: 0x7d5245955a66c95156942a2a8f4c7a3067942021c9970c9796aa74229ea542c1
2023-11-29 09:22:12        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1014: Priority is too low: (12396 vs 2): The transaction has too low priority to replace another transaction already in the pool.
2023-11-29 09:22:13        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1014: Priority is too low: (12396 vs 2): The transaction has too low priority to replace another transaction already in the pool.
```
and I'm guessing it's from here - haven't looked super deeply but this will make two swaps in succession within the space of a block, and I don't think it checks the mempool before submitting.

kylezs: passed this another time I tried though - which would also check out - if the timing lines up it can work.

j4m1ef0rd:
Should we make a command that will run a polkadot runtime upgrade? easy enough to do now that we have the code.

j4m1ef0rd:
I'm still paying around with running it in parallel with the other tests. But you can start your review now.

bouncer/shared/polkadot_runtime_update.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+async function randomPolkadotSwap(): Promise<void> {
+  const assets: Asset[] = [Assets.BTC, Assets.ETH, Assets.USDC, Assets.FLIP];
+  const randomAsset = assets[Math.floor(Math.random() * assets.length)];
+
+  let sourceAsset: Asset;
+  let destAsset: Asset;
+
+  if (Math.random() < 0.5) {
```
kylezs, line 194: I like this randomisation aspect, I think we should introduce more randomness like this in the bouncer tests

bouncer/shared/utils/bump_spec_version.ts
```diff
@@ -1,59 +1,69 @@
...
-    console.error(`An error occurred: ${error.message}`);
-  }
+  console.log(`Successfully updated spec_version to ${incrementedVersion}.`);
 }
 
 // Bump the spec version in the runtime file, using the spec version of the network.
-export async function bumpSpecVersionAgainstNetwork(projectRoot: string): Promise<number> {
-  const currentSpecVersion = await getCurrentSpecVersion();
+export async function bumpSpecVersionAgainstNetwork(
+  filePath: string,
```
kylezs, line 60:
```suggestion
  runtimeLibPath: string,
```

bouncer/tests/polkadot_runtime_update.ts
```diff
@@ -0,0 +1,28 @@
...
+async function observeBroadcastAborted(): Promise<void> {
+  for (;;) {
+    await observeEvent(':BroadcastAborted', await getChainflipApi());
+    console.log('Broadcast aborted');
+    broadcastAbortedCount++;
+  }
+}
+
+async function main(): Promise<void> {
+  observeBroadcastAborted();
```
kylezs, line 16: I think we should write this test such that if we get an aborted, we throw an error and fail the test. When I tried it seemed to just log "Broadcasts aborted"

bouncer/tests/test_data/polkadot_runtime_10001.wasm
kylezs, line 1: 👌 was speedy with this

kylezs (commented):
This is looking good, besides the minor comments, I'd create a separate `run.sh` file called `run-3-node.sh` or something, that runs run.sh and any other 3 node tests. There might be a better to way to achieve this - for example, having the concurrent tests take a a parameter for the number of nodes so this could be in the concurrent tests? Will let you decide. 
We should probably think about using 3 node localnet on CI instead of 1-node too - ofc, this is out of scope here

bouncer/shared/fund_redeem.ts
```diff
@@ -29,7 +29,7 @@ async function redeemAndObserve(
   await redeemFlip(seed, redeemEthAddress, redeemAmount);
 
   const newBalance = await observeBalanceIncrease('FLIP', redeemEthAddress, initBalance);
-  const balanceIncrease = newBalance - parseInt(initBalance);
+  const balanceIncrease = newBalance - parseFloat(initBalance);
```
j4m1ef0rd, line 32: FYI: Fixed a bug here.

kylezs: I don't know if this will completely fix it, I also added an amount of precision, here:  https://github.com/chainflip-io/chainflip-backend/pull/4274/commits/364b3c6df05f24c6b7ba6d64ee07fc1077bc0449

kylezs (approved):

PR #4291, opened 2023-11-29
fix: don't remove awaiting broadcast before retry

dandanlen:
Fixes a race condition whereby witnessing failure to sign at the same block as the broadcast success prevents broadcast from ever succeeding.


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -746,6 +746,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 	fn start_next_broadcast_attempt(broadcast_attempt: BroadcastAttempt<T, I>) {
 		let broadcast_id = broadcast_attempt.broadcast_attempt_id.broadcast_id;
+		Self::take_awaiting_broadcast(broadcast_attempt.broadcast_attempt_id);
```
kylezs, line 749: This is not required, in both places we call start_next_broadcast_attempt, we already take

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -407,7 +407,9 @@ pub mod pallet {
 				let retries_len = retries.len();
 
 				for retry in retries {
-					Self::start_next_broadcast_attempt(retry);
+					if Self::take_awaiting_broadcast(retry.broadcast_attempt_id).is_some() {
```
kylezs, line 410: Is there still an issue here? 

Scenario:
1. We take this attempt and start next broadcast
2. Signature is invalid, so we re-threshold sign (and don't yet insert the new attempt)
4. Transaction success witnessed for the retried/timedout broadcast

In success we would do: 
```
AwaitingBroadcast::<T, I>::get(BroadcastAttemptId {
						broadcast_id,
						attempt_count: BroadcastAttemptCount::<T, I>::get(broadcast_id),
```
But we're taking this here.

kylezs: Will write a test to confirm this.

state-chain/chains/src/mocks.rs
```diff
@@ -196,9 +196,9 @@ pub struct MockTransaction;
 impl FeeRefundCalculator<MockEthereum> for MockTransaction {
 	fn return_fee_refund(
 		&self,
-		_fee_paid: <MockEthereum as Chain>::TransactionFee,
+		fee_paid: <MockEthereum as Chain>::TransactionFee,
 	) -> <MockEthereum as Chain>::ChainAmount {
-		<MockEthereum as Chain>::ChainAmount::default()
+		fee_paid.effective_gas_price * fee_paid.gas_used
```
kylezs, line 201: The test that was testing refunds before was actually just testing default = default 😬 ... fortunately, it does work.

kylezs (approved):

martin-chainflip (approved):

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -625,3 +625,72 @@ fn retry_and_success_in_same_block() {
...
+		Broadcaster::on_idle(0, LARGE_EXCESS_WEIGHT);
+		assert_eq!(
+			BroadcastAttemptCount::<Test, Instance1>::get(broadcast_attempt_id.broadcast_id),
+			1
+		);
+
+		MockTransactionBuilder::<MockEthereum, RuntimeCall>::set_invalid_for_rebroadcast();
+
+		// Retry will be triggered on timeout, and it'll take the invalid signature code path.
+		Broadcaster::on_initialize(expected_expiry_block);
```
dandanlen, line 672: If possible it would be better to use the `execute_at_block` test util since it also runs the on_idle hook - which is what triggers retries. In this test we run on_idle before on_initialize, which only reflects the real order of execution if we assume these things happen on subsequent blocks.

dandanlen:
Rebased to main and resolved conflicts with #4277

PR #4296, opened 2023-11-30
fix: ignore extra btc vouts

kylezs:
# Pull Request

Closes: PRO-1007

## Checklist

## Summary

Ignores extra BTC vouts in a single transaction.

### DIFF
--- a/engine/src/witness/btc/btc_deposits.rs
+++ b/engine/src/witness/btc/btc_deposits.rs
@@ -81,30 +81,26 @@ fn deposit_witnesses(
 	for tx in txs {
 		let tx_hash = tx.txid().as_raw_hash().to_byte_array();
 
-		let mut deposits_for_this_tx = vec![];
-		for (vout, tx_out) in (0..).zip(&tx.output) {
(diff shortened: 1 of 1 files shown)


AlastairHolmes (approved):

AlastairHolmes (approved):

engine/src/witness/btc/btc_deposits.rs
```diff
@@ -80,18 +80,28 @@ fn deposit_witnesses(
 	let mut deposit_witnesses = Vec::new();
 	for tx in txs {
 		let tx_hash = tx.txid().as_raw_hash().to_byte_array();
-		for (vout, tx_out) in (0..).zip(&tx.output) {
-			if tx_out.value > 0 {
+
+		if let Some(deposit) = (0..)
```
martin-chainflip, line 84: A valid usecase would be this: You want to swap your bitcoin into DOT and ETH. You generate two deposit channels, and send funds to both in a single transaction. We should allow this, but it wouldn't work with the suggested change. We should only filter out outputs in the same TX to the same destination address

PR #4297, opened 2023-11-30
fix: lp logs in localnet

AlastairHolmes:
Just copied this from the broker start script


AlastairHolmes:
@tomjohnburton I'm guessing this problem was duplicated in the mainnet, as there have been similar problems of no lp logs on mainnet.

kylezs (approved):

PR #4301, opened 2023-12-01
fix: don't block unfinalised dot source when can't find events

kylezs:
# Pull Request

Closes: PRO-1010

## Checklist

## Summary

Previously in the case of a dot reorg it was possible that the dot source would wait forever to try to get events from a hash that didn't exist, causing dot chain tracking to stop. Now it just makes 2 attempts, and fails out, allowing the chain tracking to continue in such a case.

### DIFF
--- a/engine/src/witness/dot/dot_source.rs
+++ b/engine/src/witness/dot/dot_source.rs
@@ -1,6 +1,6 @@
 use std::{pin::Pin, time::Duration};
 
-use crate::retrier::RetryLimit;
+use crate::retrier::NoRetryLimit;
 use cf_chains::dot::PolkadotHash;
 use cf_primitives::PolkadotBlockNumber;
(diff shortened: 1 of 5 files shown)


engine/src/witness/dot/dot_source.rs
```diff
@@ -155,7 +155,8 @@ impl<
 	async fn stream_and_client(
 		&self,
 	) -> (BoxChainStream<'_, Self::Index, Self::Hash, Self::Data>, Self::Client) {
-		polkadot_source!(self, subscribe_finalized_heads, NoRetryLimit, |raw_events| { raw_events })
+		let unwrap_events = |raw_events| raw_events;
```
AlastairHolmes, line 158: you could dedup this by putting the var inside the macro.

PR #4303, opened 2023-12-04
fix: CLI redeem amount rounding error

j4m1ef0rd:
# Pull Request

Closes: PRO-1003

## Checklist

## Summary

limits the precision of the amount of flip the user can specify to 6 decimal places. But now we don't get bad rounding errors.
We can set the number of decimal places to whatever we want, I think it starts to get rounding errors if it's >11. Hard to test, so i just left it at 6, seems like plenty for now.

```sh
./target/debug/chainflip-cli redeem --exact 199995 0x835012efffe34fabf82f26a9712d3eff1b2d5145
Connecting to state chain node at: `ws://localhost:9944` and using private key located at: `/Users/jamieford/...`
Submitting redemption with amount `199995` FLIP (`199995000000000000000000` Flipperinos) to ETH address `0x835012efffe34fabf82f26a9712d3eff1b2d5145`.
```

### DIFF
--- a/api/bin/chainflip-cli/src/main.rs
+++ b/api/bin/chainflip-cli/src/main.rs
@@ -144,6 +144,20 @@ async fn run_cli() -> Result<()> {
 	.await
 }
 
+/// Turns the amount of FLIP into a RedemptionAmount in Flipperinos.
+fn flip_to_redemption_amount(amount: Option<f64>) -> RedemptionAmount {
+	// Using a set number of decimal places of accuracy to avoid floating point rounding errors
+	const DECIMAL_PLACES: u32 = 6;
+	match amount {
+		Some(amount_float) => {
+			let atomic_amount = ((amount_float * 10_f64.powi(DECIMAL_PLACES as i32)) as u128) *
+				10_u128.pow(18 - DECIMAL_PLACES);
+			RedemptionAmount::Exact(atomic_amount)
+		},
+		None => RedemptionAmount::Max,
+	}
+}
+
 async fn request_redemption(
 	api: StateChainApi,
 	amount: Option<f64>,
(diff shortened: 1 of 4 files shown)


kylezs (approved):

api/bin/chainflip-cli/src/main.rs
```diff
@@ -144,6 +144,20 @@ async fn run_cli() -> Result<()> {
...
 }
 
+/// Turns the amount of FLIP into a RedemptionAmount in Flipperinos.
+fn flip_to_redemption_amount(amount: Option<f64>) -> RedemptionAmount {
+	// Using a set number of decimal places of accuracy to avoid floating point rounding errors
+	const DECIMAL_PLACES: u32 = 6;
+	match amount {
+		Some(amount_float) => {
+			let atomic_amount = ((amount_float * 10_f64.powi(DECIMAL_PLACES as i32)) as u128) *
+				10_u128.pow(18 - DECIMAL_PLACES);
```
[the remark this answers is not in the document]
kylezs: done, and added some other minor changes

api/bin/chainflip-cli/src/main.rs
```diff
@@ -412,3 +423,36 @@ fn generate_keys(json: bool, path: Option<PathBuf>, seed_phrase: Option<String>)
...
+
+	assert_eq!(
+		flip_to_redemption_amount(Some(69420.123456)),
+		RedemptionAmount::Exact(69420123456000000000000)
+	);
+
+	// specifying more than the allowed precision ignores the extra precision
+	assert_eq!(
+		flip_to_redemption_amount(Some(69420.123456789)),
+		RedemptionAmount::Exact(69420123456000000000000)
```
dandanlen, line 450: This is a bit weird - I know it makes little difference but I'd prefer if we either rounded to the nearest decimal, or were just super strict and refused anything with more than the supported precision. 

ie. 69420.123456789 should either Error or give 69420.123457.

api/bin/chainflip-cli/src/settings.rs
```diff
@@ -112,7 +112,7 @@ pub enum CliCommand {
 	)]
 	Redeem {
 		#[clap(
-			help = "Amount to redeem in FLIP (omit this option to redeem all available FLIP)",
+			help = "Amount to redeem in FLIP (omit this option to redeem all available FLIP). Only accepts up to 6 decimal places of precision.",
```
dandanlen, line 115: > Only accepts up to 6 decimal places of precision.

This is not true in the current impl: we accept more than 6 and then silently ignore it...

kylezs: The user doesn't really care better to make it clear for them

kylezs: with the rounding I updated this to state that beyond 6, it's rounded.

dandanlen (approved):

PR #4304, opened 2023-12-04
feat: Handle failed vault transfers

syan095:
# Pull Request

Closes: PRO-789

## Checklist

## Summary

Added pathway to handle failed vault transfers.
Basic workflow:
* When one transfer fails in BatchAll call, the Vault contract emits an event, which is observed by Engine and the Statechain is notified:
    - pallet_cf_ingress_egress::Call::vault_transfer_failed()
* Here the TransferFallback EthApi call is constructed with the transfer details included.
* The call is delegated to the Broadcast pallet to be signed, but NOT broadcasted. 
* After threshold signature is completed, the details are saved in the Broadcast pallet, the same way Failed CCM are handled.
* In the next epoch, the call is re-threshold signed, using the same mechanism as Failed CCM.
* In the epoch after that, all call data are culled from on-chain storage.

### DIFF
--- a/state-chain/cf-integration-tests/src/lib.rs
+++ b/state-chain/cf-integration-tests/src/lib.rs
@@ -27,8 +27,8 @@ use sp_consensus_grandpa::AuthorityId as GrandpaId;
 use sp_core::crypto::Pair;
 use state_chain_runtime::{
 	constants::common::*, opaque::SessionKeys, AccountId, BitcoinVault, Emissions, EthereumVault,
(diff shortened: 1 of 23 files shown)


albert-llimos:
I've just had a call with Roy and he has enlightened me on how the `transferFallback` metadata won't be stored for refund purposes, as it a broadcast is not requested for that transaction. Therefore, the problem I described above is not an issue, by default a `transferFallback` is not eligible for a refund. Great stuff!

dandanlen:
LGTM, I just rebased against main + resolved conflicts, fixed some typos.

dandanlen (approved):

PR #4310, opened 2023-12-05
fix: connections can become stale when reconnecting

msgmaxim:
# Pull Request

## Checklist

## Summary

Previously in some cases it was possible for connections to never become "stale" (preventing reconnections when inactive). Specifically, when we fail to connect to a node due to some authentication error, this would result us reconnecting to the node manually (rather than this being done by zmq internally), but this also reset activity for that node to "now" on every restart, preventing us from detecting the connection as stale. I changed the code to carry over any previous activity state when reconnecting, so inactive connections should now always enter "stale" state.

Writing a proper unit test for this is a bit cumbersome and I'm not sure if worth it, but I have tested this manually by changing some parameters and observing logs that the bug was present initially and that the new code fixes the incorrect behaviour.

### DIFF
--- a/engine/src/p2p/core.rs
+++ b/engine/src/p2p/core.rs
@@ -521,7 +521,7 @@ impl P2PContext {
 				state: ConnectionState::Connected(connected_socket),
 				info: peer,
 				last_activity: Cell::new(
-					previous_activity.unwrap_or_else(|| tokio::time::Instant::now()),
+					previous_activity.unwrap_or_else(tokio::time::Instant::now),
 				),
 			},
 		) {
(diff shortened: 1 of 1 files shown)


engine/src/p2p/core.rs
```diff
@@ -539,13 +541,17 @@ impl P2PContext {
 			return
 		}
 
+		let mut previous_activity = None;
```
j4m1ef0rd, line 544: Could we set this to `tokio::time::Instant::now` here instead of using an option in `connect_to_peer`?

j4m1ef0rd (approved):

PR #4312, opened 2023-12-05
fix: threshold sig migrations

kylezs:


dandanlen (approved):

dandanlen:
The move_prefix calls are still required I think (they were from Roy's PR).

PR #4313, opened 2023-12-05
feat: New Lp Interfaces

AlastairHolmes:


state-chain/amm/src/tests.rs
```diff
@@ -41,25 +38,21 @@ fn test_basic_swaps() {
...
-				pool_state.swap(asset, Order::Sell, Amount::MAX, None),
-				(amount, Amount::MAX - amount)
-			);
+			assert_eq!(pool_state.swap(order, 0.into(), None), (0.into(), 0.into()));
+			assert_eq!(pool_state.swap(order, Amount::MAX, None), (amount, Amount::MAX - amount));
 		}
 
 		{
-			let initial_sqrt_price = match asset {
-				common::Side::Zero => MAX_SQRT_PRICE - 1,
```
martin-chainflip, line 61: did we want to remove the '-1' ?

state-chain/amm/src/tests.rs
```diff
@@ -85,19 +78,19 @@ fn test_basic_swaps() {
...
+					Amount::MAX - minted_amounts[!order.to_sold_side()]
 				)
 			);
 		}
 
 		{
-			let initial_sqrt_price = match asset {
-				Side::Zero => MAX_SQRT_PRICE - 1,
+			let initial_sqrt_price = match order.to_sold_side() {
+				Side::Zero => MAX_SQRT_PRICE,
```
martin-chainflip, line 93: same here

state-chain/amm/src/tests.rs
```diff
@@ -129,29 +122,31 @@ fn test_basic_swaps() {
...
+							range_order_minted_amounts[!order.to_sold_side()]) -
+						1 /* -1 is due to rounding down */
 				)
 			);
 		}
 
 		{
-			let initial_sqrt_price = match asset {
-				Side::Zero => MAX_SQRT_PRICE - 1,
+			let initial_sqrt_price = match order.to_sold_side() {
```
martin-chainflip, line 148: and here

state-chain/amm/src/tests.rs
AlastairHolmes: yes (should have been done before in previous pr, saw I hadn't before)

AlastairHolmes: I increase the vlid sqrt_prices to include the MAX_SQRT_PRICE in all cases, previously so some of internal functions where safer to use. As a result MAX_SQRT_PRICE is a valid starting price. Where it wasn't before.

martin-chainflip (approved):

state-chain/pallets/cf-pools/src/migrations/v1.rs
```diff
@@ -21,21 +26,25 @@ mod old {
...
 impl<T: Config> OnRuntimeUpgrade for Migration<T> {
 	fn on_runtime_upgrade() -> frame_support::weights::Weight {
-		Pools::<T>::translate::<old::Pool<T>, _>(|_key, old_pool: old::Pool<T>| {
-			Some(Pool::<T> {
-				range_orders_cache: old_pool.range_orders_cache,
-				limit_orders_cache: old_pool.limit_orders_cache,
-				pool_state: old_pool.pool_state.into(),
-			})
-		});
+		for (key, pool) in old::Pools::<T>::iter().filter_map(|(old_key, old_pool)| {
```
dandanlen, line 36: I think this should use `drain` otherwise it won't remove the old pools (they will remain stored under the old key).

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1894,25 +1863,37 @@ impl<T: Config> Pallet<T> {
 		pool: &mut Pool<T>,
 		asset_pair: &AssetPair,
 		lp: &T::AccountId,
-		side: Side,
-		order: OrderId,
+		order: Order,
```
dandanlen, line 1866: nit: I think Side was a better name. Order suggests it's the actual order. (but I can see `Side` already has another meaning). 

How about OrderSide or BuyOrSell or something less ambiguous?

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -40,182 +40,205 @@ mod tests;
...
+	Encode,
+	Decode,
+	TypeInfo,
+	MaxEncodedLen,
+	PartialEq,
+	Eq,
+	Deserialize,
+	Serialize,
+)]
+pub enum Assets {
```
dandanlen, line 95: This is also kind of a confusing name...

dandanlen: We have have OrderSide (buy/sell), PoolSide (base/quote), AmmSide (one,zero) to disambiguate all of this a bit more clearly?

dandanlen: I did give some better names 😅 
But ok to leave it if it's not exposed. What about order? This is one is in the Api.

dandanlen: I know we want to get this out but once we settle on external-facing terminology it will be harder to change it.

state-chain/pallets/cf-pools/src/lib.rs
AlastairHolmes: OrderType?

dandanlen: Order types are generally used to refer to things like limit order / market order / fill or kill etc. Ie. it's a collection of attributes or criteria.

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Not sure what you're suggesting here...

AlastairHolmes: image

GitHUb decided to not show your comments

PR #4318, opened 2023-12-06
feat: LP API allows user to specify what level of confirmation they want

kylezs:


api/lib/src/lp.rs
```diff
@@ -193,21 +235,29 @@ pub trait LpApi: SignedExtrinsicApi {
...
+					id,
+					option_tick_range,
+					size_change,
+				},
+				wait_for,
+			)
 			.await?;
 
-		Ok(collect_range_order_returns(events))
+		Ok(match xt_result {
```
AlastairHolmes, line 255: this map could be cut off into a function, one that Takes a WaitForResult and turns it into a ApiWaitForResult and takes a closure for the Details case?

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -96,7 +96,8 @@ pub trait Rpc {
 	async fn request_liquidity_deposit_address(
 		&self,
 		asset: RpcAsset,
-	) -> Result<String, AnyhowRpcError>;
+		wait_for: WaitFor,
```
AlastairHolmes, line 99: Not sure the rpc macro, will generate this as an optional arg, just because it impls default. If that was what you intended

api/lib/src/lp.rs
```diff
@@ -17,6 +17,31 @@ use state_chain_runtime::RuntimeCall;
 use std::ops::Range;
 use utilities::rpc::NumberOrHex;
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub enum ApiWaitForResult<T> {
+	TransactionHash(H256),
+	Details { tx_hash: H256, response: T },
```
AlastairHolmes, line 24: Also I think maybe different variant names, as they will show up in the jsonrpc response would be good, maybe "TxHash" and "TxResult"

AlastairHolmes:
This has been put into the subscribe_order_fills sub PR

PR #4323, opened 2023-12-07
Broadcaster pallet improvements

syan095:
# Pull Request

Closes: PRO-1020
Closes: PRO-1021
Closes: PRO-116
Closes: PRO-1001
Closes: PRO-1081

## Checklist

## Summary

* Removed BroadcastAttemptCount, and used FailedBroadcasters's len() instead to count the number of attempts
* Time-outs count as broadcast failures and therefore can cause broadcasts to abort.
* Anyone can report Broadcast failures.
* Refactored pallet storage - broadcast_id is used to index many storages.
* Added storage migrations
* Added migration tests and pre-/post- checks

* Added broadcast retry delays
* Delay policy for Eth and Polkadot: Delay by 10 blocks (1 min)
* Delay policy for Bitcoin: 0-24 attempts: 0 blocks, starting from 25, increase by power of 2 (1, 2, 4, 8..) for each failure. Caps at 1200 blocks (2 hours)
* Retries will not happen if the broadcast has already succeeded or aborted.
* Added integration test

* fixed a bug where during safe mode, retries and Timeouts replaces future storage instead of appending to it


state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +1011,28 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+		let broadcast_attempt_id = start_mock_broadcast();
+		let expiry = System::block_number()
+			.saturating_add(<Test as crate::Config<Instance1>>::BroadcastTimeout::get());
+		let nominee =
+			AwaitingBroadcast::<Test, Instance1>::get(broadcast_attempt_id).unwrap().nominee;
+
+		assert_eq!(
+			FailedBroadcasters::<Test, Instance1>::get(broadcast_attempt_id.broadcast_id),
+			None
+		);
```
kylezs, line 1027: nit
```suggestion
		assert!(
			FailedBroadcasters::<Test, Instance1>::get(broadcast_attempt_id.broadcast_id).is_none()
		);
```

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +1011,28 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+
+		// Let the broadcast attempt time out
+		Broadcaster::on_initialize(expiry);
```
kylezs, line 1030: Could we use the block processing pipeline methods like `execute_at_block` instead of this - that we ensure there's also not weird interaction with the other hooks in reality (which we've had in the past)

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -409,9 +409,11 @@ pub mod pallet {
 						Self::deposit_event(Event::<T, I>::BroadcastAttemptTimeout {
 							broadcast_attempt_id: *attempt_id,
 						});
-						if let Some(broadcast_attempt) = Self::take_awaiting_broadcast(*attempt_id)
+						if let Some((broadcast_attempt, nominee)) =
+							Self::take_awaiting_broadcast(*attempt_id)
 						{
-							Self::start_next_broadcast_attempt(broadcast_attempt);
+							FailedBroadcasters::<T, I>::append(attempt_id.broadcast_id, nominee);
```
kylezs, line 415: I think we should use the `transaction_signing_failure` logic here, that way if everyone is attempted and everyone times out we abort. 

Currently, what happens here, is if the last broadcast attempt fails due to a timeout, adding the final authority to failed broadcasters, we fill up FailedBroadasters storage... then we can't nominate anyone, and we enter an infinite retry loop, a retry every block trying to nominate someone, but everyone's listed as Failed. 

As part of this PR, I think we should probably look at using `FailedBroadcasters.len()` instead of `attempt_count` for 
```rust
if signing_attempt.broadcast_attempt.broadcast_attempt_id.attempt_count >=
				T::EpochInfo::current_authority_count()
					.checked_sub(1)
					.expect("We must have at least one authority")
```

which is currently in the `transaction_signing_failure` logic, but eventually will be shared.

We should also add a test for this case, what happens when the *last* authority fails, and ensure that it gets aborted.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -888,36 +747,39 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			broadcast_id,
+			Self::attempt_count(broadcast_id),
+			broadcast_attempt.transaction_payload.clone(),
+		)
+			.encode();
 		if let Some(nominated_signer) = T::BroadcastSignerNomination::nominate_broadcaster(
 			seed,
-			&FailedBroadcasters::<T, I>::get(broadcast_attempt.broadcast_attempt_id.broadcast_id)
-				.unwrap_or_default(),
+			&FailedBroadcasters::<T, I>::get(broadcast_id).into_iter().collect::<Vec<_>>(),
```
kylezs, line 778: can do the same as above, `nominate_broadcaster` could take an IntoIter in order to avoid this allocation, we call `iter()` on it inside this call anyway...

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -930,29 +792,23 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 				transaction_payload: broadcast_attempt.transaction_payload,
 				transaction_out_id: broadcast_attempt.transaction_out_id,
 			});
 		} else {
-			log::warn!(
-				"Failed to select a signer for broadcast {:?}. Scheduling Retry",
-				broadcast_attempt.broadcast_attempt_id
-			);
-			Self::schedule_for_retry(&broadcast_attempt);
+			Self::abort_broadcast(broadcast_id);
```
kylezs, line 805: This is quite different to what was mentioned in the issue. The reason we want to check the length of failed broadcasters to decide if we should abort or not and *not* whether we can nominate is because people can be excluded for reasons *other* than just the fact they failed to broadcast this particular transaction. If you look inside `nominate_broadcaster` nodes can be excluded for other reasons:
```rust
let mut all_excludes = Reputation::validators_suspended_for(&[
			Offence::FailedToBroadcastTransaction,
			Offence::MissedHeartbeat,
		]);
```

These are temporary suspensions. After some time these suspensions will wear off and then this nomination can succeed - so there's a bit of an issue here.

I think using the approach as the code was before, where we abort when the failure actually occurs is a bit better. Doing it this way also means we don't do unnecessary extra things before aborting like this:

```rust
		T::TransactionBuilder::refresh_unsigned_data(&mut broadcast_attempt.transaction_payload);
		TransactionMetadata::<T, I>::insert(
			broadcast_id,
			<<T::TargetChain as Chain>::TransactionMetadata>::extract_metadata(
				&broadcast_attempt.transaction_payload,
			),
		);		
```

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -144,7 +106,6 @@ pub mod pallet {
 	#[derive(Clone, RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo)]
 	#[scale_info(skip_type_params(T, I))]
 	pub struct BroadcastAttempt<T: Config<I>, I: 'static> {
-		pub broadcast_attempt_id: BroadcastAttemptId,
```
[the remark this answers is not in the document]
syan095: I was trying to reduce dedundant storage, but its cool, I'll revert this change.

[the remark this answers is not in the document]
syan095: Change reverted!

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -713,10 +594,11 @@ pub mod pallet {
 			}
 
 			// Report the people who failed to broadcast this tx during its whole lifetime.
-			if let Some(failed_signers) = FailedBroadcasters::<T, I>::take(broadcast_id) {
+			let failed_signers = FailedBroadcasters::<T, I>::take(broadcast_id);
+			if !failed_signers.is_empty() {
 				T::OffenceReporter::report_many(
```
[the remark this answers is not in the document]
syan095: Done! 
I've change it to IntoIterator<Item = ValidatorId> to prevent dealing with  <'a> lifetime. Is this OK?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -397,21 +344,18 @@ pub mod pallet {
...
-				for attempt_id in expiries.iter() {
-					if PendingBroadcasts::<T, I>::get().contains(&attempt_id.broadcast_id) {
-						Self::deposit_event(Event::<T, I>::BroadcastAttemptTimeout {
-							broadcast_attempt_id: *attempt_id,
-						});
-						if let Some(broadcast_attempt) = Self::take_awaiting_broadcast(*attempt_id)
+				for broadcast_id in expiries.into_iter() {
+					if PendingBroadcasts::<T, I>::get().contains(&broadcast_id) {
+						Self::deposit_event(Event::<T, I>::BroadcastTimeout { broadcast_id });
+						if let Some(signing_attempt) = AwaitingBroadcast::<T, I>::get(broadcast_id)
```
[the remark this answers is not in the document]
syan095: Refactored

state-chain/pallets/cf-broadcast/src/lib.rs
syan095: Done

state-chain/pallets/cf-broadcast/src/lib.rs
syan095: Acknowledged. I've put the abort checking logic into "handle_broadcast_failure".
Note with this, if a broadcast has not reported failure, cannot be nominated (due to failed heartbeat for example) and cannot recover, then we will keep retrying forever (atleast until the next epoch?)

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -930,28 +791,25 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 
 			Self::deposit_event(Event::<T, I>::TransactionBroadcastRequest {
-				broadcast_attempt_id: broadcast_attempt.broadcast_attempt_id,
+				broadcast_id,
 				nominee: nominated_signer,
 				transaction_payload: broadcast_attempt.transaction_payload,
 				transaction_out_id: broadcast_attempt.transaction_out_id,
 			});
 		} else {
-			log::warn!(
```
kylezs, line 943: could we please leave this log message here, I have used it recently to debug the exact issue we're fixing now haha

state-chain/pallets/cf-broadcast/src/lib.rs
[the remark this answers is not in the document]
kylezs: Yeah perfect, no clones anyway 👌

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -899,25 +756,29 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			broadcast_id,
+			Self::attempt_count(broadcast_id),
+			broadcast_attempt.transaction_payload.clone(),
+		)
+			.encode();
 		if let Some(nominated_signer) = T::BroadcastSignerNomination::nominate_broadcaster(
 			seed,
-			&FailedBroadcasters::<T, I>::get(broadcast_attempt.broadcast_attempt_id.broadcast_id)
-				.unwrap_or_default(),
+			FailedBroadcasters::<T, I>::get(broadcast_attempt.broadcast_id),
```
kylezs, line 777:
```suggestion
			FailedBroadcasters::<T, I>::get(broadcast_id),
```

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -962,6 +820,83 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+					T::EpochInfo::current_authority_count()
+						.checked_sub(1)
+						.expect("We must have at least one authority")
+				{
+					Self::abort_broadcast(broadcast_id);
+				} else {
+					Self::schedule_for_retry(&signing_attempt.broadcast_attempt);
+				}
+			} else {
+				log::warn!("Broadcast failure already reported: Broadcast ID:{}", broadcast_id,);
```
kylezs, line 852:
```suggestion
				log::warn!("Broadcast failure by {reporter} already recorded for Broadcast ID:{}", broadcast_id,);
```

state-chain/pallets/cf-broadcast/src/migrations/v2.rs
```diff
@@ -2,47 +2,168 @@ use crate::*;
...
-use sp_std::marker::PhantomData;
 #[cfg(feature = "try-runtime")]
 use sp_std::prelude::Vec;
+use sp_std::{collections::btree_map::BTreeMap, marker::PhantomData};
 
 mod old {
-
 	use super::*;
 
+	/// A unique id for each broadcast attempt
```
kylezs, line 12: These new migrations need to be in a v3 file, and the appropriate pallet version bump. v2 has already been released.

state-chain/pallets/cf-broadcast/src/migrations/v2.rs
```diff
@@ -2,47 +2,168 @@ use crate::*;
...
+
+		let new_retry_queue = BroadcastRetryQueue::<T, I>::get()
+			.into_iter()
+			.map(|attempt| attempt.broadcast_id)
+			.collect::<Vec<_>>();
+
+		broadcast_retry_queue
+			.into_iter()
+			.for_each(|broadcast_id| assert!(new_retry_queue.contains(&broadcast_id)));
+
```
kylezs, line 166: we can add some unit tests for these migrations too - a good habit to get into

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -200,68 +189,43 @@ fn test_abort_after_number_of_attempts_is_equal_to_the_number_of_authorities() {
 	});
 }
 
+// Helper function: make a broadcast to be aborted upon the next failure.
+fn ready_to_abort_broadcast(broadcast_id: BroadcastId) {
```
kylezs, line 193: this would be slightly better if it returned the nominee right? that way we can use the nominee who has not yet failed to submit the transaction, and guarantee it in the test

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -200,68 +189,43 @@ fn test_abort_after_number_of_attempts_is_equal_to_the_number_of_authorities() {
...
+	validators.remove(&nominee);
+	FailedBroadcasters::<Test, Instance1>::insert(broadcast_id, validators);
+}
+
 #[test]
-fn broadcasts_aborted_after_all_report_failures_after_retry() {
-	new_test_ext()
-		.execute_with(|| {
-			let (_tx_out_id1, api_call1) = api_call(1);
+fn broadcasts_aborted_after_all_report_failures() {
```
kylezs, line 203: we should also add a test for the case where it timesout, that it should also result in aborted

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -58,29 +63,16 @@ impl MockCfe {
...
 			RuntimeEvent::Broadcaster(broadcast_event) => match broadcast_event {
 				BroadcastEvent::TransactionBroadcastRequest {
-					broadcast_attempt_id,
+					broadcast_id,
 					nominee,
 					transaction_payload: _,
 					transaction_out_id: _,
 				} => {
 					match scenario {
 						Scenario::SigningFailure => {
```
kylezs, line 72: we can rename this now to be `BroadcastFailure`

state-chain/pallets/cf-broadcast/src/tests.rs
kylezs: We should also add a test that after we have aborted, we can still successfully do `transaction_succeeded` - this case may happen where the transaction was actually broadcast but was stuck in the mempool and so was just witnessed late

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -458,22 +388,16 @@ fn test_transmission_request_expiry() {
 		// Simulate the expiry hook for the expected expiry block.
 		let expected_expiry_block = current_block + BROADCAST_EXPIRY_BLOCKS;
 		Broadcaster::on_initialize(expected_expiry_block);
+		Broadcaster::on_idle(expected_expiry_block, 1_000_000_000_000_000u64.into());
```
kylezs, line 391: Could we use the test utilitie instead of this, like `then_execute_at_block` for example, this way we can ensure the hooks are executed in the correct order - more similar to how the chain actually runs, and we don't test cases that might not actually be possible

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -485,32 +409,34 @@ fn test_transmission_request_expiry() {
...
-		assert!(AwaitingBroadcast::<Test, Instance1>::get(broadcast_attempt_id).is_some());
-		assert_eq!(BroadcastAttemptCount::<Test, Instance1>::get(broadcast_id), 0);
+		assert!(AwaitingBroadcast::<Test, Instance1>::get(broadcast_id).is_some());
+		assert_eq!(Broadcaster::attempt_count(broadcast_id), 0);
 
 		MockTransactionBuilder::<MockEthereum, RuntimeCall>::set_requires_refresh();
 
 		// If invalid on retry then we should re-threshold sign
 		Broadcaster::on_initialize(BROADCAST_EXPIRY_BLOCKS + 1);
+		Broadcaster::on_idle(BROADCAST_EXPIRY_BLOCKS + 1, 1_000_000_000_000_000u64.into());
```
kylezs, line 433: We should use `LARGE_EXCESS_WEIGHT` - but may not be required if we use the test utils

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -769,10 +694,7 @@ fn retry_with_threshold_signing_still_allows_late_success_witness_second_attempt
 				MOCK_TX_METADATA,
 			));
 
-			// This is a bug. If this bug didn't exist, this would be assert_eq!().
-			// Leaving this assert here so when the code changes, whoever changes it, thinks about
-			// this case.
```
kylezs, line 774: 👌

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +895,25 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
 		MOCK_TX_METADATA,
 	));
 }
-enum TxType {
-	Normal,
-	Rotation,
+
+#[test]
+fn time_out_broadcaster_are_reported() {
```
kylezs, line 900:
```suggestion
fn timed_out_broadcaster_are_reported() {
```

state-chain/pallets/cf-vaults/src/tests.rs
```diff
@@ -211,30 +211,32 @@ fn handover_success_triggers_handover_verification() {
 	});
 }
 
-fn keygen_failure(bad_candidates: &[<Test as Chainflip>::ValidatorId]) {
+fn keygen_failure(
+	bad_candidates: impl IntoIterator<Item = <Test as Chainflip>::ValidatorId> + Clone,
```
kylezs, line 215: 👌

state-chain/runtime/src/migrations/threshold_signature_callbacks.rs
```diff
@@ -1,11 +1,13 @@
-use crate::{BitcoinInstance, Box, EthereumInstance, PolkadotInstance, Runtime, RuntimeCall};
+use crate::{
+	BitcoinBroadcaster, BitcoinInstance, Box, EthereumBroadcaster, EthereumInstance,
+	PolkadotBroadcaster, PolkadotInstance, Runtime, RuntimeCall,
+};
```
kylezs, line 4: This file shouldn't be changed, it's a migration for a particular version - I think we can delete it instead of changing here.

state-chain/test-utilities/src/rich_test_externalities.rs
```diff
@@ -72,8 +72,8 @@ impl<Runtime: HasAllPallets> RichExternalities<Runtime> {
 			frame_system::Pallet::<Runtime>::set_block_number(block_number);
 			Runtime::on_initialize(block_number);
 			let context = f();
-			Runtime::on_idle(block_number, Weight::MAX);
 			Runtime::on_finalize(block_number);
+			Runtime::on_idle(block_number, Weight::MAX);
```
kylezs, line 76: https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.Hooks.html#method.on_idle

On idle runs before on_finalize

state-chain/pallets/cf-broadcast/src/lib.rs
kylezs: Yep this is how it should behave. It won't retry forever, but until a suspension has elapsed for one of the authorities.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -962,6 +820,83 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
+		let broadcast_id = signing_attempt.broadcast_attempt.broadcast_id;
+		let reporter = match maybe_reporter {
+			Some(reporter) => reporter,
+			None => signing_attempt.nominee,
+		};
+		FailedBroadcasters::<T, I>::mutate(broadcast_id, |failed_broadcasters| {
+			if failed_broadcasters.insert(reporter) {
+				// We will abort the broadcast if all validators reported failure.
+				if FailedBroadcasters::<T, I>::decode_len(broadcast_id).unwrap_or_default() as u32 >=
```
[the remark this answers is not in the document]
syan095: Yeah man forgot the code I wrote myself  >: o

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: great idea!

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -962,6 +820,83 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+					T::EpochInfo::current_authority_count()
+						.checked_sub(1)
```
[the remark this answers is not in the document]
syan095: Found out that the -1 is actually accounting for adding the current reported broadcaster. 
Anyways it's fine, we add the new broadcaster to the set and re-check the length now.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -962,6 +820,83 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		Ok(())
+	}
+
+	/// Called when we cannot select a broadcaster, because all validators have failed to broadcast
+	/// this call. We abort to prevent infinite retries.
+	/// The failed callback is dispatched (if any), and data is kept in storage for potential future
+	/// governance functions.
+	fn abort_broadcast(broadcast_id: BroadcastId) {
+		log::warn!(
+			"Failed to select a signer for broadcast {:?}. Broadcast is aborted.",
```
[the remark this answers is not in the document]
syan095: done

state-chain/pallets/cf-broadcast/src/lib.rs
syan095: Added logging here

state-chain/pallets/cf-broadcast/src/migrations/v2.rs
syan095: Removed  old migration, advanced version from v2 -> v3

state-chain/pallets/cf-broadcast/src/migrations/v2.rs
syan095: Added unit tests for migration. I didn't even know we could unit test this but having these test makes me much more confidant about the migration!

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: Added test for when the last broadcaster timed out.

Added test for `transaction_succeeded` on aborted broadcast

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: Renamed

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: Refactored this out. 
Also changed to other on_idle and on_initialize when it's possible (or when it makes sense)

state-chain/runtime/src/migrations/threshold_signature_callbacks.rs
syan095: Old migration removed. 
There's a separate linear ticket for removing stale migrations. I'll leave the other pallets to be done in a separate PR

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+			{
+				Some(attempt) =>
+					if attempt_count > *attempt {
+						latest_attempt.insert(broadcast_id, attempt_count);
+					},
+				None => {
+					latest_attempt.insert(broadcast_id, attempt_count);
+				},
+			},
+		);
```
kylezs, line 71: ⛳ 
```suggestion
		old::AwaitingBroadcast::<T, I>::iter_keys().for_each(
			|old::BroadcastAttemptId { broadcast_id, attempt_count }| {
				if match latest_attempt.get(&broadcast_id) {
					Some(attempt) => attempt_count > *attempt,
					None => true,
				} {
					latest_attempt.insert(broadcast_id, attempt_count);
				}
			},
		);
```

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+			);
+			old::FailedBroadcasters::<Test, Instance1>::insert(2, vec![1u64, 2u64, 3u64]);
+
+			// Perform runtime migration.
+			crate::migrations::v3::Migration::<Test, Instance1>::on_runtime_upgrade();
+
+			// Verify data is correctly migrated into new storage.
+			let failed_1 = FailedBroadcasters::<Test, Instance1>::get(1);
+			let failed_2 = FailedBroadcasters::<Test, Instance1>::get(2);
+			assert!(failed_1.contains(&1u64));
```
kylezs, line 226: we also want to check the len = 3 - could also just do this be doing `vec.into_iter().collect::<BTreeSet<_>>()` and comparing the btreesets, directly

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+			}
+		});
+
+		// Failed broadcaster storage: Vec<ValidatorId> -> BTreeSet<ValidatorId>
+		old::FailedBroadcasters::<T, I>::drain().for_each(|(broadcast_id, failed_broadcasters)| {
+			FailedBroadcasters::<T, I>::insert(
+				broadcast_id,
+				BTreeSet::from_iter(failed_broadcasters),
+			)
+		});
```
kylezs, line 102:
```suggestion
		FailedBroadcasters::<T, I>::translate(|_, failed_broadcasters: Vec<T::ValidatorId>| {
			Some(BTreeSet::from_iter(failed_broadcasters))
		});
```
we only really need the `old` version in this case for the tests

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+			.unwrap_or_default()
+			.into_iter()
+			.map(|attempt| BroadcastAttempt::<T, I> {
+				broadcast_id: attempt.broadcast_attempt_id.broadcast_id,
+				transaction_payload: attempt.transaction_payload,
+				threshold_signature_payload: attempt.threshold_signature_payload,
+				transaction_out_id: attempt.transaction_out_id,
+			})
+			.collect::<Vec<_>>();
+		BroadcastRetryQueue::<T, I>::put(retry_queue);
```
kylezs, line 114: can use translate here too

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+
+		Weight::zero()
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, DispatchError> {
+		let broadcasts_with_failed_broadcasters =
+			old::FailedBroadcasters::<T, I>::iter_keys().collect::<Vec<_>>();
```
kylezs, line 122: we could also check the values

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+#[cfg(test)]
+mod migration_tests {
+	use super::*;
+	use crate::mock::*;
+
+	#[test]
+	fn migration_works_v2_to_v3_awaiting_broadcast() {
+		new_test_ext().execute_with(|| {
+			// Insert mock data into old storage
+			for broadcast_id in 1..10 {
```
kylezs, line 168: 👌  these tests definitely increase confidence!

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,271 @@
...
+					};
+					old::AwaitingBroadcast::<Test, Instance1>::insert(
+						broadcast_attempt_id,
+						singing_attempt,
+					);
+				}
+			}
+
+			// Perform runtime migration.
+			crate::migrations::v3::Migration::<Test, Instance1>::on_runtime_upgrade();
```
kylezs, line 189: You can also run the pre-and post upgrade hooks in the test, before and after the upgrade respectively - just have to make the tests with the `try-runtime` flag too. I think that's worth doing

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +932,90 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+			ready_to_abort_broadcast(broadcast_id);
+
+			(broadcast_id, transaction_out_id)
+		})
+		.then_execute_at_block(expiry, |(broadcast_id, transaction_out_id)| {
+			(broadcast_id, transaction_out_id)
+		})
+		.then_execute_with(|(broadcast_id, transaction_out_id)| {
+			// Broadcast should be aborted
+			assert_eq!(FailedBroadcasters::<Test, Instance1>::decode_len(broadcast_id), None);
```
kylezs, line 998: We should add an assert to check that the AbortedBroadcasts has an item or check the event, just to confirm that it's aborted (and not just the FailedBroadcasters is empty - which would usually suggest it's not even been attempted yet and therefore not aborted

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +932,90 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+			));
+
+			// Storage should be cleaned, event emitted.
+			System::assert_last_event(RuntimeEvent::Broadcaster(
+				crate::Event::<Test, Instance1>::BroadcastSuccess {
+					broadcast_id,
+					transaction_out_id,
+				},
+			));
+			assert!(!AwaitingBroadcast::<Test, Instance1>::contains_key(broadcast_id));
```
kylezs, line 1016: can use assert_broadcast_storage_cleaned_up

kylezs (approved):
Nice, just small comments here

kylezs:
This will need to be rebased onto main, and the minor version needs to be bumped - there are breaking CFE changes here. So this PR can bump the version to 1.2.0 of all the bins, and bump the spec version to 120

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -476,72 +435,16 @@ pub mod pallet {
 		///
 		/// ## Errors
 		///
-		/// - [InvalidBroadcastAttemptId](Error::InvalidBroadcastAttemptId)
-		/// - [InvalidSigner](Error::InvalidSigner)
+		/// - [InvalidBroadcastId](Error::InvalidBroadcastId)
 		#[pallet::call_index(0)]
-		#[pallet::weight(T::WeightInfo::transaction_signing_failure())]
-		pub fn transaction_signing_failure(
+		#[pallet::weight(T::WeightInfo::transaction_failed())]
```
dandanlen, line 440: I would prefer if we remained a bit more defensive here: we can keep the old extrinsic (but have it just return an Error), and add a new one with the new functionality.

Will change.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -760,26 +655,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-		}
-	}
-
 	pub fn remove_pending_broadcast(broadcast_id: &BroadcastId) {
 		PendingBroadcasts::<T, I>::mutate(|pending_broadcasts| {
 			if !pending_broadcasts.remove(broadcast_id) {
-				cf_runtime_utilities::log_or_panic!(
-					"The broadcast_id should exist in the pending broadcasts list since we added it to the list when the broadcast was initated"
+				log::warn!(
+					"The broadcast_id already aborted, but is now reported as successful. broadcast_id: {}", broadcast_id
```
dandanlen, line 662: nit: this error message makes some strong assumptions about the calling context...

How do we know that it's not the other way round: a previously successful broadcast is now reported as failed? Or a failed broadcast was reported as failed again?

state-chain/pallets/cf-reputation/src/lib.rs
```diff
@@ -388,24 +388,27 @@ impl<T: Config> OffenceReporter for Pallet<T> {
 	type ValidatorId = T::ValidatorId;
 	type Offence = T::Offence;
 
-	fn report_many(offence: impl Into<Self::Offence>, validators: &[Self::ValidatorId]) {
+	fn report_many(
+		offence: impl Into<Self::Offence>,
+		validators: impl IntoIterator<Item = T::ValidatorId> + Clone,
```
dandanlen, line 393: nit: this could have been split into a separate PR - not really relevant and adds lot more line changes to this one (which is already complex enough).

state-chain/chains/src/lib.rs
```diff
@@ -488,3 +488,21 @@ impl<C: Chain> FeeEstimationApi<C> for () {
...
+	fn next_attempt_delay(retry_attempts: Self::AttemptCount) -> Option<Self::BlockNumber>;
+}
+
+pub struct DefaultRetryPolicy;
+impl RetryPolicy for DefaultRetryPolicy {
+	type BlockNumber = u32;
+	type AttemptCount = u32;
+
+	fn next_attempt_delay(_retry_attempts: Self::AttemptCount) -> Option<Self::BlockNumber> {
+		Some(10u32)
```
dandanlen, line 506: I want to review these numbers... Defaulting to 10 blocks doesn't seem right.

dandanlen (commented):
I made some changes, but the overall shape of this looks good to me. I want to have a closer look at the migration, didn't have time for that today. 

It's quite complex so I think it's worth @kylezs or @ramizhasan111 giving it another look over too after i'm done.

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: (4b62708001058a40fd1bee4d1bc2cb3af5daa42f)

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: (changed this)

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,283 @@
...
+							transaction_out_id: attempt.broadcast_attempt.transaction_out_id,
+						},
+						nominee: attempt.nominee,
+					},
+				);
+			}
+		});
+
+		// Failed broadcaster storage: Vec<ValidatorId> -> BTreeSet<ValidatorId>
+		FailedBroadcasters::<T, I>::translate(|_, failed_broadcasters: Vec<T::ValidatorId>| {
```
dandanlen, line 95: This shouldn't be required - BTreeSet and Vec have the same encoding.

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,283 @@
...
+					None => true,
+				} {
+					latest_attempt.insert(broadcast_id, attempt_count);
+				}
+			},
+		);
+
+		// Migrate data from old storage into new storage.
+		latest_attempt.into_iter().for_each(|(broadcast_id, attempt_count)| {
+			if let Some(attempt) = old::AwaitingBroadcast::<T, I>::get(old::BroadcastAttemptId {
```
dandanlen, line 73: The old AwaitingBroadcast storage needs to be removed. You could use `drain` to store all the old::SigningAttempts into a BTreeMap above and then write them to the new storage here.

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,283 @@
...
+			.collect::<Vec<_>>();
+		Ok(MigrationVerification::<T> {
+			broadcasts_with_failed_broadcasters,
+			broadcast_retry_queue,
+		}
+		.encode())
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(state: Vec<u8>) -> Result<(), DispatchError> {
```
dandanlen, line 139: The post_upgrade checks could ensure that the old storage values have been removed.

dandanlen (changes requested):

dandanlen:
As far as i'm concerned, only the migration needs to be fixed (removal of old storage) then this is good to go.

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
syan095: This translation will remove any duplicates in the vec to ensure correctness.

syan095: Tested via unit test no migration is needed - duplication is also removed

syan095: I've added this translate to remove any potential duplicate entries to ensure correctness.

Updated: Via unit test - without this translate(), the value obtained is correct, however the decode_len() will return incorrect result - (decode len returns length with duplicate, not the length of the set).

Though we don't expect there are dups in the old storage, I think it is worth it to add this line here to ensure the set is constructed correctly.

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
syan095: we now check the old AwaitingBroadcast storage is cleared.
However for `FailedBroadcasters` and `BroadcastRetryQueue`, the old storage is linked with the new one (clearing old storage also clears the new storage) - I suspect this is due to the use of the "translate" call to migrate.

kylezs:
There's really no need for this PR to be this big, many of these items could have been done in individual PRs, makes it painful to review and just more likely bugs are introduced. In the future let's refrain from adding more to a PR, and target PRs with more specific names than "improvements".

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -784,61 +709,64 @@ fn retry_with_threshold_signing_still_allows_late_success_witness_second_attempt
...
+			witness_broadcast(tx_out_id1);
 
-		// Now tx2 succeeds which should allow tx3 to be broadcast
-		witness_broadcast(tx_out_id2);
-		Broadcaster::on_idle(0, LARGE_EXCESS_WEIGHT);
+			// tx3 should still not be broadcasted because the blocking tx (tx2) has still not
+			// succeeded.
+			(tx_out_id2, tx_out_id3, broadcast_id_3)
+		})
+		.then_execute_at_next_block(|(tx_out_id2, tx_out_id3, broadcast_id_3)| {
```
kylezs, line 747: we should add a util that does this pass through without having to pass everything through manually - we do this in a number of places (would be in a separate PR ofc)

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +936,239 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+
+			BroadcastDelay::set(None);
+			// With no delay, retries are added to the normal queue, and is retried in the next
+			// block.
+			assert_ok!(Broadcaster::transaction_failed(RuntimeOrigin::signed(0u64), broadcast_id));
+			assert_eq!(BroadcastRetryQueue::<Test, Instance1>::get()[0].broadcast_id, broadcast_id);
+			System::assert_last_event(RuntimeEvent::Broadcaster(
+				crate::Event::<Test, Instance1>::BroadcastRetryScheduled {
+					broadcast_id,
+					retry_block: System::block_number() + 1,
```
kylezs, line 1045: where does this +1 come from? Is this from the retry policy? If so, we should just the `BroadcastDelay` here instead of the magic `1`

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +936,239 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+				},
+			));
+		})
+		.then_execute_at_next_block(|_| {})
+		.then_execute_with(|_| {
+			BroadcastDelay::set(Some(10));
+			// Set delay - retries will be added to the Delayed queue.
+			assert_ok!(Broadcaster::transaction_failed(RuntimeOrigin::signed(1u64), broadcast_id));
+			target = System::block_number() + 10;
```
kylezs, line 1054: same here, let's use a const instead of passing 10 into both places, otherwise it's not super clear they're related

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +936,239 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+	new_test_ext()
+		.execute_with(|| {
+			(broadcast_id, _) = start_mock_broadcast();
+
+			BroadcastDelay::set(Some(10));
+			target = System::block_number() + BROADCAST_EXPIRY_BLOCKS;
+		})
+		.then_execute_at_block(target, |_| {})
+		.then_execute_with(|_| {
+			target = System::block_number() + 10;
```
kylezs, line 1090: same as above

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +936,239 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+}
+
+#[test]
+fn aborted_broadcasts_will_not_retry() {
+	let mut target = 0;
+	let mut broadcast_id = 0;
+	new_test_ext()
+		.execute_with(|| {
+			(broadcast_id, _) = start_mock_broadcast();
+			BroadcastDelay::set(Some(100));
```
kylezs, line 1113: same

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +936,239 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+		});
+}
+
+#[test]
+fn succeeded_broadcasts_will_not_retry() {
+	let mut target = 0;
+	new_test_ext()
+		.execute_with(|| {
+			let (broadcast_id, transaction_out_id) = start_mock_broadcast();
+			BroadcastDelay::set(Some(100));
```
kylezs, line 1145: same

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1006,7 +936,239 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
...
+		.then_execute_at_block(target, |_| {})
+		.then_execute_with(|_| {
+			// assert no retry happened
+			assert!(FailedBroadcasters::<Test, Instance1>::get(broadcast_id).is_empty());
+			assert!(!PendingBroadcasts::<Test, Instance1>::get().contains(&broadcast_id))
+		});
+}
+
+#[test]
+fn succeeded_broadcasts_will_not_retry() {
```
kylezs, line 1140: what if there's a `transaction_failure` after success (again could leave this to another PR)

state-chain/runtime/src/migrations/threshold_signature_callbacks.rs
kylezs: This was added back? what happened here?

syan095: Re-removed. Might be re-added due to incorrect merge from main : (

state-chain/chains/src/lib.rs
syan095: yeah the number can be updated after the system is implemented in a separate PR. 

As I mentioned to dan:
"for the bitcoin, the delay caps off 35 failed attempts, and is gonna be 2 hours delay for each retry after wards, which means it's gonna take 115 * 2 hours (~10 days) to abort for 150 validators, assuming each validator immidiately report failure and not time out, not sure if this is what we want"

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -397,31 +355,36 @@ pub mod pallet {
...
-			// forever. It's possible that the reason for timeout could be something like a chain
-			// halt on the external chain. If the signature is valid then we expect it to succeed
-			// eventually. For outlying, unknown unknowns, these can be something governance can
-			// handle if absolutely necessary (though it likely never will be).
+			// We treat a time out here as a Broadcast Failure. This is handled the same way - the
+			// current broadcaster is reported as Failed to broadcast, and a new broadcaster is
+			// nominated. If there are no more broadcaster available, then the broadcast is aborted.
 			let expiries = Timeouts::<T, I>::take(block_number);
+			let pending_broadcasts = PendingBroadcasts::<T, I>::get();
+			let delayed_retries = DelayedBroadcastRetryQueue::<T, I>::take(block_number);
```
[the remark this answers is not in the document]
syan095: I tried to unify the two - but it as slightly more complicated than a very simple change - I will leave this to another PR.
We should get this next PR into the same release as this PR to avoid adding another round of migration for this pallet.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -467,82 +423,14 @@ pub mod pallet {
...
-		///
-		/// ## Events
-		///
-		/// - N/A
-		///
-		/// ## Errors
-		///
-		/// - [InvalidBroadcastAttemptId](Error::InvalidBroadcastAttemptId)
-		/// - [InvalidSigner](Error::InvalidSigner)
+		/// DEPRECATED. This call is no longer used.
```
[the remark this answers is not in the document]
syan095: Dan added this - I would prefer to replace the extrinsic too - I think this is more like a coding-style / maintenance rule type of discussion that we should make a decision on (this will undoubtedly happen more in the future)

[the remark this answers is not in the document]
syan095: (and by "we" I mean the dev managers, like Dan, Martin and you. I do not have a strong feeling against either way, except that I just think it's "ugly" to leave this deprecated call here. 
Also I think the "call_index" macro is specifically there to cover for this situation)

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,288 @@
...
+					},
+				)
+			});
+
+		//Failed broadcaster storage: Vec<ValidatorId> -> BTreeSet<ValidatorId> - dedup
+		FailedBroadcasters::<T, I>::translate(|_, failed_broadcasters: Vec<T::ValidatorId>| {
+			Some(BTreeSet::from_iter(failed_broadcasters))
+		});
+
+		let _ = BroadcastRetryQueue::<T, I>::translate(
```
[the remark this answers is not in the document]
syan095: This is a different "translate", and will generate a rust warning if the result is ignored

state-chain/pallets/cf-broadcast/src/lib.rs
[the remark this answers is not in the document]
kylezs: Ok, let's leave it for now, if Dan's added that, and we can look at removing it later

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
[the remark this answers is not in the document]
kylezs: Ah this is the Value translate - we can expect - at least that's what we do elsewhere

state-chain/pallets/cf-broadcast/src/lib.rs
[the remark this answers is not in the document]
kylezs: 100% can leave to another PR

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,288 @@
...
+#[cfg(test)]
+mod migration_tests {
+	use super::*;
+	use crate::mock::*;
+	#[cfg(feature = "try-runtime")]
+	use frame_support::assert_ok;
+
+	// Perform runtime migration. Run pre- and post- upgrade if "try-runtime"
+	fn do_upgrade() {
+		#[cfg(feature = "try-runtime")]
```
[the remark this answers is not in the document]
syan095: yeah I can confirm cargo cf-test and cf-test-ci both contain --try-runtime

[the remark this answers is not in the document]
syan095: I think it is worth it to still run the migration test with and without try-runtime, as the tests are written within the try-runtime in mind

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: this just means that the retry is scheduled to the next block. I've added an extra step to make this logic more clear. 
There's a comment above to explain

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: Added this case to the unit test

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -0,0 +1,288 @@
...
+		StorageMap<Pallet<T, I>, Twox64Concat, BroadcastAttemptId, TransactionSigningAttempt<T, I>>;
+
+	#[frame_support::storage_alias]
+	pub type FailedBroadcasters<T: Config<I>, I: 'static> =
+		StorageMap<Pallet<T, I>, Twox64Concat, BroadcastId, Vec<<T as Chainflip>::ValidatorId>>;
+
+	#[frame_support::storage_alias]
+	pub type BroadcastRetryQueue<T: Config<I>, I: 'static> =
+		StorageValue<Pallet<T, I>, Vec<BroadcastAttempt<T, I>>>;
+}
```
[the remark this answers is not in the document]
syan095: Added Timeouts migration and tests

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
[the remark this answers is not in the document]
syan095: Added dedup logic for this

state-chain/pallets/cf-broadcast/src/tests.rs
syan095: https://linear.app/chainflip/issue/PRO-1090/context-passing-in-test-harness

kylezs (approved):

state-chain/pallets/cf-broadcast/src/migrations/v3.rs
```diff
@@ -147,17 +178,17 @@ impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
...
-		broadcast_retry_queue
-			.into_iter()
-			.for_each(|broadcast_id| assert!(new_retry_queue.contains(&broadcast_id)));
+			.collect::<BTreeSet<_>>();
+		assert!(broadcast_retry_queue.difference(&new_retry_queue).next().is_none());
+
+		// Ensure Timeouts data are migrated
+		timeout_broadcasts.into_iter().for_each(|(block_number, attempts)|
+			// Assert the pre- and post- migrated data is identical.
+			assert!(attempts.difference(&Timeouts::<T, I>::get(block_number)).next().is_none()));
```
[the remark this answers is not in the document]
kylezs: I just pushed this along with some other small things.

dandanlen (approved):

PR #4331, opened 2023-12-08
Fix: CFE Witnessing, use parent block metadata when decoding events

j4m1ef0rd:
# Pull Request

Closes: PRO-1017

## Checklist

## Summary

The issue is caused by the `state_getRuntimeVersion` rpc returning the new runtime version when called at the given block, but the events in that block at encoded with the old runtime.
- Use parent block hash when getting polkadot runtime version & metadata
- Added a test that gets the events from the block on polkadot mainnet that had the runtime update in it and caused the error that we seen in the logs. Then parses the events and checks for any failure to decode. This test is set to ignore because it uses an external public endpoint.
- A few other small changes cleaning up code here and there.

### DIFF
--- a/engine/src/dot/http_rpc.rs
+++ b/engine/src/dot/http_rpc.rs
@@ -151,13 +151,21 @@ impl DotRpcApi for DotHttpRpcClient {
 		Ok(self.block(block_hash).await?.map(|block| block.block.extrinsics))
 	}
 
-	// TODO: When witnessing is catching up we query blocks in batches. It's posible that when
+	// TODO: When witnessing is catching up we query blocks in batches. It's possible that when
 	// a batch is made over a runtime boundary that the metadata will need to be queried more than
-	// necessary, as the order within the batch is not necessarily guaranteed. Beacause we limit
+	// necessary, as the order within the batch is not necessarily guaranteed. Because we limit
 	// Polkadot to 32 concurrent requests and runtime upgrades are infrequent this should not be an
 	// issue in reality, but probably worth solving at some point.
-	async fn events(&self, block_hash: PolkadotHash) -> Result<Option<Events<PolkadotConfig>>> {
-		let chain_runtime_version = self.runtime_version(Some(block_hash)).await?;
+	async fn events(
+		&self,
+		block_hash: PolkadotHash,
+		parent_hash: PolkadotHash,
+	) -> Result<Option<Events<PolkadotConfig>>> {
+		// We need to get the runtime version at the previous block instead the desired block
+		// because the events in the block are encoded using the previous block's runtime version,
+		// not the desired block's runtime version.
+		let chain_runtime_version = self.runtime_version(Some(parent_hash)).await?;
+		println!("using spec_version: {:?}", chain_runtime_version.spec_version);
 
 		let client_runtime_version = self.online_client.runtime_version();
 
@@ -168,7 +176,7 @@ impl DotRpcApi for DotHttpRpcClient {
 			chain_runtime_version.transaction_version !=
 				client_runtime_version.transaction_version
 		{
-			let new_metadata = self.metadata(block_hash).await?;
+			let new_metadata = self.metadata(parent_hash).await?;
 
 			self.online_client.set_runtime_version(subxt::rpc::types::RuntimeVersion {
 				spec_version: chain_runtime_version.spec_version,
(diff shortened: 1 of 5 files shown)


engine/src/dot/http_rpc.rs
```diff
@@ -224,4 +232,40 @@ mod tests {
...
+			.unwrap()
+			.unwrap();
+
+		// Calling iter() will cause the events to be decoded. None of the events should fail to
+		// decode if the correct metadata is used.
+		assert!(!events.iter().any(|event| event.is_err()));
+
+		// Check that mapping the events does not panic
+		let _mapped_events: Vec<_> =
+			events.iter().filter_map(crate::witness::dot::filter_map_events).collect();
```
kylezs, line 269: nit
```suggestion
		events.iter().filter_map(crate::witness::dot::filter_map_events).collect::<Vec<_>>();
```

engine/src/dot/rpc.rs
```diff
@@ -125,8 +129,13 @@ impl DotRpcApi for DotRpcClient {
 
 	/// Returns the events for a particular block hash.
 	/// If the block for the given block hash does not exist, then this returns `Ok(None)`.
-	async fn events(&self, block_hash: PolkadotHash) -> Result<Option<Events<PolkadotConfig>>> {
-		self.http_client.events(block_hash).await
+	/// The parent hash is used to determine the runtime version to decode the events.
```
kylezs, line 132: this should be a normal comment just on the parent_hash param

engine/src/dot/http_rpc.rs
```diff
@@ -151,13 +151,21 @@ impl DotRpcApi for DotHttpRpcClient {
...
-	async fn events(&self, block_hash: PolkadotHash) -> Result<Option<Events<PolkadotConfig>>> {
-		let chain_runtime_version = self.runtime_version(Some(block_hash)).await?;
+	async fn events(
+		&self,
+		block_hash: PolkadotHash,
+		parent_hash: PolkadotHash,
+	) -> Result<Option<Events<PolkadotConfig>>> {
+		// We need to get the runtime version at the previous block instead the desired block
+		// because the events in the block are encoded using the previous block's runtime version,
+		// not the desired block's runtime version.
```
kylezs, line 166: we should mention that this is because the state_getRuntimeVersion RPC returns the runtime version at the end of the block

engine/src/dot/http_rpc.rs
```diff
@@ -151,13 +151,21 @@ impl DotRpcApi for DotHttpRpcClient {
...
+		let chain_runtime_version = self.runtime_version(Some(parent_hash)).await?;
+		println!("using spec_version: {:?}", chain_runtime_version.spec_version);
```
kylezs, line 168: can be removed

engine/src/dot/http_rpc.rs
```diff
@@ -224,4 +232,40 @@ mod tests {
...
+	async fn test_parsing_events_in_runtime_update_block() {
+		use std::str::FromStr;
+
+		// Block hash of the block that the runtime update occurred in
+		let block_hash_of_runtime_update =
+			H256::from_str("0xa0b52be60216f8e0f2eb5bd17fa3c66908cc1652f3080a90d3ab20b2d352b610")
+				.unwrap();
+
+		// The hash of the block before the runtime update block
+		let block_hash_of_parent =
```
kylezs, line 247: nit: we really only need to hardcode the runtime update hash, and we can get the parent hash by calling block(), makes it a little more flexible

engine/src/dot/http_rpc.rs
```diff
@@ -224,4 +232,40 @@ mod tests {
 		let block_hash = dot_http_rpc.block_hash(1).await.unwrap();
 		println!("block_hash: {:?}", block_hash);
 	}
+
+	#[ignore = "Uses public mainnet polkadot endpoint"]
+	#[tokio::test]
+	async fn test_parsing_events_in_runtime_update_block() {
```
kylezs, line 238: 👍  good idea to leave this test here

engine/src/witness/dot.rs
```diff
@@ -143,12 +143,14 @@ pub async fn process_egress<ProcessCall, ProcessingFut>(
...
-					tracing::info!("Witnessing transaction_succeeded. signature: {signature:?}");
-					process_call(
+		match PolkadotUncheckedExtrinsic::decode(&mut xt_bytes) {
+			Ok(unchecked) =>
+				if let Some(signature) = unchecked.signature() {
+					if monitored_egress_ids.contains(&signature) {
+						tracing::info!(
+							"Witnessing transaction_succeeded. signature: {signature:?}"
+						);
+						process_call(
```
msgmaxim, line 153: Nit: this used to be properly indented.

msgmaxim (approved):

PR #4335, opened 2023-12-08
feat: shave fees on ingress

dandanlen:
Closes: PRO-1047


state-chain/chains/src/dot.rs
```diff
@@ -252,6 +252,40 @@ impl Default for PolkadotTrackedData {
...
+		const WEIGHT_FEE: u128 = 4_000_000;
+		/// This should be a minor over-estimate. It's the length in bytes of an extrinsic that
+		/// encodes a single fetch operation. In practice, multiple fetches and transfers might be
+		/// encoded in the extrinsic, bringing the per-fetch average down.
+		const EXTRINSIC_LENGTH: u128 = 80;
+
+		self.median_tip +
+			MULTIPLIER
+				.checked_mul_int(BASE_FEE * LENGTH_FEE * EXTRINSIC_LENGTH + WEIGHT_FEE)
+				.expect("TODO: Make sure this can't panic...")
```
dandanlen, line 278: TODO: add cost of egress here too.

state-chain/chains/src/dot.rs
```diff
@@ -252,6 +252,76 @@ impl Default for PolkadotTrackedData {
...
+/// See https://wiki.polkadot.network/docs/learn-transaction-fees
+mod fee_constants {
+	use frame_support::sp_runtime::FixedU128;
+
+	// See https://wiki.polkadot.network/docs/learn-DOT.
+	pub const MICRO_DOT: u128 = 10_000;
+	pub const MILLI_DOT: u128 = 1_000 * MICRO_DOT;
+
+	/// Taken from the Polkadot runtime.
+	pub const MULTIPLIER: FixedU128 = FixedU128::from_inner(100_000_000_000_000_000);
```
ramizhasan111, line 264: Isnt this a variable? This would cary according to the traffic on polkadot chain?

state-chain/chains/src/eth.rs
```diff
@@ -77,6 +77,43 @@ impl EthereumTrackedData {
...
+	) -> <Ethereum as Chain>::ChainAmount {
+		use fees::*;
+
+		// Note: this is taking the egress cost of the swap in the ingress currency (and basing the
+		// cost on the ingress chain).
+		let gas_cost_per_fetch = BASE_COST_PER_BATCH +
+			match asset {
+				assets::eth::Asset::Eth => Zero::zero(),
+				assets::eth::Asset::Flip | assets::eth::Asset::Usdc => GAS_COST_PER_FETCH,
+			};
```
ramizhasan111, line 100: As discussed, this is a type, should be: GAS_COST_PER_BATCH + GAS_COST_PER_FETCH

state-chain/chains/src/eth.rs
```diff
@@ -77,6 +77,43 @@ impl EthereumTrackedData {
 	}
 }
 
+mod fees {
+	// TODO: refine these constants.
```
albert-llimos, line 81: Difficult to set accurate values here, but I'd suggest some changes.
A signature verification in Ethereum will cost ~60-80k normally. And on top of that, the fetch or transfer will be 20-30k on top. It probably makes sense to lower that slightly since there will be some transactions batching multiple actions, so my proposal would be something along the lines of:
```
	pub const BASE_COST_PER_BATCH: u128 = 50_000;
	pub const GAS_COST_PER_FETCH: u128 = 30_000;
	pub const GAS_COST_PER_TRANSFER_NATIVE: u128 = 20_000;
	pub const GAS_COST_PER_TRANSFER_TOKEN: u128 = 40_000;
```
I have added a differentiation between native and token on the egress as explained in the other comment.

state-chain/chains/src/eth.rs
```diff
@@ -77,6 +77,43 @@ impl EthereumTrackedData {
...
+		(self.base_fee + self.priority_fee).saturating_mul(gas_cost_per_fetch)
+	}
+
+	fn estimate_egress_fee(
+		&self,
+		_asset: <Ethereum as Chain>::ChainAsset,
+	) -> <Ethereum as Chain>::ChainAmount {
+		use fees::*;
+
+		let gas_cost_per_transfer = BASE_COST_PER_BATCH + GAS_COST_PER_TRANSFER;
```
albert-llimos, line 111: I'd suggest to differentiate on the egress between tokens and ETH. While it might be a bit of micromanaging, it is just a small code change and the difference is like 20k gas, so it's not small.

state-chain/chains/src/dot.rs
dandanlen: It might be but we don't track it and I don't want to bloat this PR. It also seems to be static right now on Polkadot.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -942,15 +948,42 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 		let channel_id = deposit_channel_details.deposit_channel.channel_id;
 		Self::deposit_event(Event::<T, I>::DepositFetchesScheduled { channel_id, asset });
 
+		// Shave off the fetch fee from the deposit amount.
+		// For now, we take ingress and egress fees here. This is a simplification: we should
+		// really take the egress fee at the point of egress.
+		let tracked_data = T::ChainTracking::get_tracked_data();
+		let tx_fees = tracked_data
+			.estimate_ingress_fee(deposit_channel_details.deposit_channel.asset) +
+			tracked_data.estimate_egress_fee(deposit_channel_details.deposit_channel.asset);
```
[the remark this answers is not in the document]
ramizhasan111: The problem is that the error in this impl is huge because if we swap from cheap chain (e.g. polkadot) to expensive chain (like bitcoin), we basically deduct no fees for egress. 

I think the implementation would be still relatively not complex if we just move this calculation to the match statement below.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1082,6 +1120,42 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			IngressOrEgress::Ingress => tracked_data.estimate_ingress_fee(asset),
+			IngressOrEgress::Egress => tracked_data.estimate_egress_fee(asset),
+		};
+
+		let fee_estimate = if asset == <T::TargetChain as Chain>::GAS_ASSET {
+			fee_estimate
+		} else {
+			T::PriceOracle::convert_asset_value(
+				asset,
+				<T::TargetChain as Chain>::GAS_ASSET,
```
dandanlen, line 1139: Wrong way round..

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: As discussed I moved the egress fee shaving to when we do the egress.

state-chain/cf-integration-tests/src/mock_runtime.rs
```diff
@@ -198,8 +198,8 @@ impl ExtBuilder {
 				init_chain_state: ChainState::<Ethereum> {
 					block_height: 0,
 					tracked_data: EthereumTrackedData {
-						base_fee: 1000000u32.into(),
-						priority_fee: 100u32.into(),
+						base_fee: 0u32.into(),
+						priority_fee: 0u32.into(),
```
dandanlen, line 202: I had to change these to zero to make the integration tests pass.

state-chain/cf-integration-tests/src/mock_runtime.rs
```diff
@@ -218,7 +218,7 @@ impl ExtBuilder {
 			bitcoin_chain_tracking: BitcoinChainTrackingConfig {
 				init_chain_state: ChainState::<Bitcoin> {
 					block_height: 0,
-					tracked_data: BitcoinTrackedData { btc_fee_info: BitcoinFeeInfo::new(1000) },
+					tracked_data: BitcoinTrackedData { btc_fee_info: BitcoinFeeInfo::new(0) },
```
dandanlen, line 221: Same here.

state-chain/chains/src/dot.rs
```diff
@@ -1068,4 +1140,25 @@ mod test_polkadot_extrinsics {
...
+			median_tip: Default::default(),
+			runtime_version: Default::default(),
+		}
+		.estimate_egress_fee(assets::dot::Asset::Dot);
+
+		// The values are not important. This test serves more as a sanity check that
+		// the fees are valid, and a reference to compare against the actual fees. These values must
+		// be updated if we update the fee calculation.
+		assert_eq!(ingress_fee, 21004500u128);
+		assert_eq!(egress_fee, 24004500u128);
```
dandanlen, line 1162: I need to check again that these fees are reasonable. I think it might still be off by a factor of ~10...

For reference, a regular balance transfer on Polkadot costs 0.0158 DOT, which is 158_000_000 units.

state-chain/chains/src/mocks.rs
```diff
@@ -182,6 +184,22 @@ impl BenchmarkValue for MockTrackedData {
 	}
 }
 
+impl FeeEstimationApi<MockEthereum> for MockTrackedData {
+	fn estimate_ingress_fee(
+		&self,
+		_asset: <MockEthereum as Chain>::ChainAsset,
+	) -> <MockEthereum as Chain>::ChainAmount {
+		todo!()
```
dandanlen, line 192: I forgot this - but it appears to be unused 😅

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -363,6 +372,10 @@ pub mod pallet {
 	pub type WitnessSafetyMargin<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, TargetChainBlockNumber<T, I>, OptionQuery>;
 
+	#[pallet::storage]
+	pub type WithheldTransactionFees<T: Config<I>, I: 'static = ()> =
```
dandanlen, line 376: TODO: add doc comment.

state-chain/primitives/src/chains.rs
```diff
@@ -110,7 +110,7 @@ chains! {
 pub struct AnyChain;
 
 impl ForeignChain {
-	pub fn gas_asset(self) -> assets::any::Asset {
+	pub const fn gas_asset(self) -> assets::any::Asset {
```
dandanlen, line 113: TODO: remove this.

state-chain/primitives/src/chains/assets.rs
```diff
@@ -132,6 +132,12 @@ macro_rules! chain_assets {
 				)+
 			}
 
+			pub const GAS_ASSET: Asset = [
```
dandanlen, line 135: TODO: remove this.

dandanlen (commented):
I'll do another pass tomorrow.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1082,6 +1122,43 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			.unwrap_or_else(|| {
+				log::warn!(
+					"Unable to convert input to gas for input of {:?} ${:?}. Ignoring transaction fees.",
+					available_amount,
+					asset,
+				);
+				(available_amount, Zero::zero())
+			});
+
+		WithheldTransactionFees::<T, I>::mutate(<T::TargetChain as Chain>::GAS_ASSET, |fees| {
```
ramizhasan111, line 1156: we might as well just use StorageValue and not a Map since we only store Gas asset fees her so only one entry in the map.

state-chain/cf-integration-tests/src/mock_runtime.rs
AlastairHolmes: I assume this was because all the swap outputs where different, because of the subtraction?

AlastairHolmes: Same below?

AlastairHolmes: If we want this in today, then I understand but can you add a comment or an issue to go back and set these to none zero, and fix the tests.

state-chain/chains/src/btc.rs
```diff
@@ -112,6 +112,24 @@ impl Default for BitcoinTrackedData {
 	}
 }
 
+impl FeeEstimationApi<Bitcoin> for BitcoinTrackedData {
+	fn estimate_ingress_fee(
+		&self,
+		_asset: <Bitcoin as Chain>::ChainAsset,
+	) -> <Bitcoin as Chain>::ChainAmount {
+		// Include the min fee so we over-estimate the cost.
+		self.btc_fee_info.min_fee_required_per_tx + self.btc_fee_info.fee_per_input_utxo
```
AlastairHolmes, line 121: min_fee_required_per_tx is not an overestimate as we assume 10 bytes, but the maximum size of a transaction that isn't associated with an input/output, seems to be 26 bytes (https://en.bitcoin.it/wiki/Transaction). So we could under-estimate if there are large numbers of inputs or outputs.

AlastairHolmes: I'm not clear on the reason a fee numbers to be taken on ingress here. I think a comment would be help as the reasoning for the fees being needed is different for each chain.

In the Eth case, we take fees in the ingress case because we do a fetch into the vault.

But in the Btc case, we take fees to pay for the cost of including the UXTO, in some future egress.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -824,7 +838,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 						asset,
-						amount,
+						amount: Self::withhold_transaction_fee(
+							IngressOrEgress::Egress,
+							asset,
+							amount,
+						),
 						to: destination_address,
 					});
 					DepositBalances::<T, I>::mutate(asset, |tracker| {
```
AlastairHolmes, line 848: Believe you should be passing the result from withhold_transaction_fee here.

dandanlen: On reflection I'm not so sure - the fee is no longer in a deposit (ie. available to be spent), it's now in the withheld assets, which are earmarked for paying fees.

state-chain/chains/src/btc.rs
dandanlen: As discussed, I agree that we are double-counting the min_fee_required_per_tx but this is deliberate to ensure we overestimate the fees. I agree there is plenty of room for improvement. I'll add some comments that hopefully make some of this a bit clearer.

state-chain/pallets/cf-ingress-egress/src/lib.rs
ramizhasan111: I think if we do what alastair suggested, we also need to account for it in the other places where the deposit tracker is mutated. So basically the ```mark_as_fetched``` and the ```register_deposit``` functions should also take the net amount for accounting to be correct. Alternatively, we can leave it as as and include the fee in the accounting at all places DepositTracker is mutated.

ramizhasan111: I think leaving it as is probably the best option. It would still do its job of tracking the deposits (kind off). Doing it the right way would require non-trivial changes.

dandanlen: As mentioned in Discord - I think it's correct as is. The withheld assets should be considered 'spent' ie. they are no longer ours.

state-chain/chains/src/btc.rs
dandanlen: I will add these comments separately if that's ok - have them prepared but forgot to push the commit. And I don't want to push it now and delay the PR anoother 40 minutes while CI completes.

martin-chainflip (approved):

PR #4337, opened 2023-12-11
fix: sweeping before withdrawal

AlastairHolmes:
I'm not sure how this was happened, but sweeping was intended to occur before withdrawal automatically.

Also I fixed that error message you mentioned previously.


martin-chainflip (approved):

martin-chainflip:
Let's add a test case where a user has pending fees and withdraws more than their free balance. This should now work

PR #4338, opened 2023-12-12
feat: btc utxo consolidation

msgmaxim:
# Pull Request

Closes: PRO-1008

## Checklist

## Summary

~~This isn't final, just wanted to get some feedback.~~ 

- For `CONSOLIDATION_THRESHOLD`, I'm assuming we want to be able to change it with a governance extrinsic?
- I'm assuming if we consolidate all available UTXOs, then there will be some time window when no UTXO are available (until the change is added back in `on_broadcast_ready`) leading to some failed egress attempts. It looks like egress attempts will be retried in that case? Should we have a separate `CONSOLIDATION_SIZE` < `CONSOLIDATION_THRESHOLD`, so that we still have (say, half) of utxos available? For example, we could consolidate 100 utxos every time we reach 200.
- What kind of tests do we want for this? I have created some unit tests, and thinking of adding one to test BitcoinApi's implementation of ConsolidateCall. Together I think they would cover most of the new code. Would this be sufficient, or do we also want some kind of integration/bouncer test?

Update:

- Added ConsolidationParameters containing separate "size" and "threshold" as discussed
- Added governance extrinsics to update them + benchmarks
- Added migrations to initialize storage for ConsolidationParameters; tested migrations with try_runtime_upgrade script

I think there are still some tests that could be added (bouncer, notably), but other than that the code is ready for review.

Update 2:
- Added a test for updating parameters thourhg gov call
- Added a bouncer test (`btc_utxo_consolidation.ts`). It is not part of any test suite yet (i.e. not in run.sh) since running it in parallel to other tests (swaps) makes it hard to track utxos, and it is a little slow. (Have we introduced the concept of "nightly" tests yet?)


martin-chainflip:
if we remove all UTXOs for consolidation, egress will just wait until there are enough UTXOs available. But I think your proposal makes a lot of sense, so let's make it so that there is a threshold that triggers consolidation and the consolidation size is below that. having these thresholds configurable would be great. testing this on bouncer should be pretty easy.

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -92,6 +92,86 @@ fn test_btc_utxo_selection() {
...
+		// Dust utxo does not count:
+		add_utxo_amount(dust_amount);
+		assert_eq!(
+			Environment::select_and_take_bitcoin_utxos(UtxoSelectionType::SelectForConsolidation),
+			None
+		);
+		assert_eq!(crate::BitcoinAvailableUtxos::<Test>::decode_len(), Some(2));
+
+		add_utxo_amount(10000);
+		add_utxo_amount(10000);
```
dandanlen, line 140:
```suggestion
		add_utxo_amount(10000);
		
		assert_eq!(crate::BitcoinAvailableUtxos::<Test>::decode_len(), Some(4));
```

?

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -92,6 +92,86 @@ fn test_btc_utxo_selection() {
...
+
+		// Any utxo that didn't get consolidated should still be available:
+		assert_eq!(
+			Environment::select_and_take_bitcoin_utxos(UtxoSelectionType::SelectForConsolidation),
+			Some((vec![utxo(10000), utxo(10000)], 15980))
+		);
+		assert_eq!(crate::BitcoinAvailableUtxos::<Test>::decode_len(), Some(2));
```
dandanlen, line 147: It would be good to ensure that the remaining utxos are the right ones... Maybe use different utxos, UTXO1, UTXO2, UTXO3, UTXO4, each with different amounts, and assert that the remaining and extracted together form the same set, ie. we didn't accidentally duplicate or delete any utxos.

dandanlen (commented):
LGTM, just some nitty comments.

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -58,6 +58,27 @@ pub fn select_utxos_from_pool<UTXO: GetUtxoAmount + Clone>(
...
 
+pub fn select_utxos_for_consolidation<UTXO: GetUtxoAmount + Clone>(
+	available_utxos: &mut Vec<UTXO>,
+	fee_per_utxo: u64,
+	params: ConsolidationParameters,
+) -> Vec<UTXO> {
+	let mut all_utxos = sp_std::mem::take(available_utxos);
+
+	let (mut spendable, mut dust) =
+		all_utxos.drain(..).partition::<Vec<_>, _>(|utxo| utxo.amount() > fee_per_utxo);
```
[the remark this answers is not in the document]
msgmaxim: Yeah, I think it was necessary in some previous iteration of this code. Removed now.

state-chain/pallets/cf-environment/src/tests.rs
msgmaxim: Sure. I will compare directly as `Vec`s since the logic is deterministic and to keep things simple.

state-chain/chains/src/btc/utxo_selection.rs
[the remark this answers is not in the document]
msgmaxim: Also, I added a few lines putting utxos back into `available_utxos` as a precaution. I know it is not necessary, but it is not immediately clear that it is safe, and it doesn't hurt to have multiple layers of defence.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -508,6 +509,20 @@ pub mod pallet {
 				log::error!("Ingress-Egress failed to send BatchAll. Error: {e:?}");
 			}
 
+			if let Ok(egress_transaction) =
+				<T::ChainApiCall as ConsolidateCall<T::TargetChain>>::consolidate_utxos()
+			{
+				let broadcast_id = T::Broadcaster::threshold_sign_and_broadcast_with_callback(
```
kylezs, line 515: Can just call `threshold_sign_and_broadcast`

state-chain/chains/src/lib.rs
```diff
@@ -344,6 +344,16 @@ pub enum AllBatchError {
 	Other,
 }
 
+#[derive(Debug)]
+pub enum ConsolidationError {
```
kylezs, line 348: Looks like we don't differentiate between these errors, anyway? are two variants required in that case?

bouncer/tests/btc_utxo_consolidation.ts
```diff
@@ -0,0 +1,97 @@
...
+  );
+
+  // Setting the threshold to current utxo count + 2 allows us to test the more
+  // general case of consolidationSize != consolidationThreshold even when there
+  // is only 1 UTXO available initially
+  const consolidationSize = initialUtxos.count + 1;
+  const consolidationThreshold = initialUtxos.count + 2;
+
+  await provideLiquidity('BTC', 2);
+  await provideLiquidity('BTC', 3);
```
kylezs, line 45: wasn't clear to me initially that these are related to the +2 above - a little comment would help

bouncer/tests/btc_utxo_consolidation.ts
```diff
@@ -0,0 +1,97 @@
...
+      // Ensure our balance is mostly unchanged after consolidation (taking tx fees into account):
+      assert(
+        amountAfterConsolidation <= amountBeforeConsolidation &&
+          amountAfterConsolidation > amountBeforeConsolidation - ERROR_MARGIN,
+      );
+      break;
+    }
+
+    console.log(`Waiting util UTXO count is 2 (currently: ${utxos.count})`);
+  }
```
kylezs, line 81: I think an important part of the consolidation which is not tested here is that the transactions actually succeed. Waiting for transaction success here I think would be a valuable addition. Plus, we could also pull out the witnessed fees from the tx fee deficit event?, then we don't need a rough error margin, but we have the exact value that was used - though this last part about the fees isn't as important I'd say.

kylezs (commented):
👍  looks good, just some minor/testing things

state-chain/chains/src/lib.rs
msgmaxim: I would say it is still better to provide more information to the caller, and it servees as a form of documentation (it highlights the fact that there is a case other than "not required" which we might want to handle separately).

bouncer/tests/btc_utxo_consolidation.ts
msgmaxim: Added a check for `BroadcastSuccess` and `TransactionFeeDeficitRecorded`. Now using the fee when asserting on the amount. I also added the UtxoConsolidation event containing broadcastId, so it is easier to track.

kylezs (approved):

dandanlen (approved):

PR #4339, opened 2023-12-12
feat: add ingress/egress fees to cf_ingress_egress_environment rpc

dandanlen:


kylezs (approved):
Just made a small commit to use the named instances

PR #4342, opened 2023-12-12
feat: API Bins check SC compatibility

j4m1ef0rd:
# Pull Request

Closes: PRO-972

## Checklist

## Summary

Added the existing version check to the API lib. This means the Broker, LP API and CLI will all return an error when trying to do a command on the SC using an older version then what's on chain.
- Version check is on all 3 bins, not just the CLI. Let me know if this is a bad idea.
- The check uses the API package version, **NOT** the individual bins version. Is this what we want? I think so.
	- Updated the API version to 1.1.0, like the others. We will need to make sure to update this version when we update the others.
- The error message does **not** let the user know they need to download an update or how to do that. This is because the message general, the CFE uses it.
	- Added block number to the error message.

```sh
Error: This version '0.1.0' is incompatible with the current release '1.1.0' at block 2117: 0xebd4e5c9c477360da764b20bcc4c0e61c723e6641aa422ebce9d0a07120598ae.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1411,7 +1411,7 @@ dependencies = [
 
 [[package]]
 name = "chainflip-api"
-version = "0.1.0"
+version = "1.1.0"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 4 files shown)


AlastairHolmes (approved):

dandanlen (approved):

dandanlen:
@ahasna FYI the api/lib/Cargo.toml  will need to be added to the new CI checks.

PR #4360, opened 2023-12-19
fix: keep utxos even if swallowed (PRO-1068)

dandanlen:
also: register net deposit amount, not gross.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -971,6 +971,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			deposit_amount,
 		);
 
+		// Add the deposit to the balance.
+		T::DepositHandler::on_deposit_made(
+			deposit_details.clone(),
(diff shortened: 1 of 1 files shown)


msgmaxim (approved):
Seems like a straightforward change, so LGTM

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -971,6 +971,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 		);
 
+		// Add the deposit to the balance.
+		T::DepositHandler::on_deposit_made(
+			deposit_details.clone(),
+			deposit_amount,
+			deposit_channel_details.deposit_channel,
+		);
+		DepositBalances::<T, I>::mutate(asset, |deposits| {
+			deposits.register_deposit(net_deposit_amount)
```
dandanlen, line 981: This is worth noting - instead of registering a deposit of the full deposit amount, we register a deposit of just the net amount. The extracted fees are accounted for in the WithheldFees storage.

PR #4365, opened 2023-12-21
fix: can timeout when updating CFE version

msgmaxim:
# Pull Request

Closes: PRO-1064

## Checklist

## Summary

As discussed, this seems like the most appropriate solution at this time. I have tested this on localnet, both the happy path, and with the request timing out (by setting a very short timeout).

### DIFF
--- a/engine/src/state_chain_observer/client/mod.rs
+++ b/engine/src/state_chain_observer/client/mod.rs
@@ -41,6 +41,9 @@ const SUBSTRATE_BEHAVIOUR: &str = "Unexpected state chain node behaviour";
 
 const SYNC_POLL_INTERVAL: Duration = Duration::from_secs(4);
 
+/// Enough time for a state chain transaction to make it into a (unfinalised) block
+const CFE_VERSION_SUBMIT_TIMEOUT: Duration = Duration::from_secs(30);
+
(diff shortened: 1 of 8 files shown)


engine/src/state_chain_observer/client/mod.rs
```diff
@@ -753,26 +756,30 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
...
-									("patch", this_version.patch),
-								],
-							)],
-						),
-						&subxt_signer,
-					)
-					.await?
-					.wait_for_in_block()
-					.await?;
+				tokio::time::timeout(
```
AlastairHolmes, line 759: Just one thing can you add a comment, mentioning the issue? Of course in the end I agree we should add timeouts everywhere. But just while this is a one off timeout.

AlastairHolmes: Are you sure it is the second await? Because if the first await on line 778 is the problem, then this timeout will not work.

AlastairHolmes: You could put the whole request is an async block and put the timeout around that whole thing?

AlastairHolmes (approved):

PR #4366, opened 2023-12-21
Improve "subscribe_prewitness_swaps", add cf_pool_price_v2, and some internal refactors

syan095:
# Pull Request

Closes: PRO-1035
Closes: PRO-1011

## Checklist

## Summary
All subscription now also returns a new BlockUpdate type which includes block number, block hash and the original data type.
Improved PoolPrice - added a new PoolPriceV2, which returns both the buy price and the sell price.
Use "base_asset" and "quote_asset" for new subscriptions and rpcs, instead of "from" and "to".

### DIFF
--- a/api/bin/chainflip-lp-api/Cargo.toml
+++ b/api/bin/chainflip-lp-api/Cargo.toml
@@ -38,7 +38,6 @@ frame-system = { git = 'https://github.com/chainflip-io/polkadot-sdk.git', tag =
 
 # Local
 chainflip-api = { path = "../../lib" }
-state-chain-runtime = { path = "../../../state-chain/runtime" }
 cf-utilities = { package = "utilities", path = "../../../utilities" }
 
 [build-dependencies]
--- a/api/bin/chainflip-lp-api/Cargo.toml
+++ b/api/bin/chainflip-lp-api/Cargo.toml
@@ -38,6 +38,7 @@ frame-system = { git = 'https://github.com/chainflip-io/polkadot-sdk.git', tag =
 
 # Local
 chainflip-api = { path = "../../lib" }
+state-chain-runtime = { path = "../../../state-chain/runtime" }
 cf-utilities = { package = "utilities", path = "../../../utilities" }
 
 [build-dependencies]
--- a/api/bin/chainflip-lp-api/src/main.rs
+++ b/api/bin/chainflip-lp-api/src/main.rs
@@ -34,6 +34,7 @@ use jsonrpsee::{
 use pallet_cf_pools::{AssetPair, AssetsMap, IncreaseOrDecrease, OrderId, RangeOrderSize};
 use rpc_types::{AssetBalance, OpenSwapChannels, OrderIdJson, RangeOrderSizeJson};
 use sp_core::U256;
+use state_chain_runtime::chainflip::BlockUpdate;
 use std::{
 	collections::{BTreeMap, HashMap, HashSet},
 	ops::Range,
@@ -194,11 +195,11 @@ pub trait Rpc {
 		executor_address: Option<EthereumAddress>,
 	) -> RpcResult<Hash>;
 
-	#[subscription(name = "subscribe_order_fills", item = OrderFills)]
+	#[subscription(name = "subscribe_order_fills", item = BlockUpdate<OrderFills>)]
 	fn subscribe_order_fills(&self);
 
 	#[method(name = "order_fills")]
-	async fn order_fills(&self, at: Option<Hash>) -> RpcResult<OrderFills>;
+	async fn order_fills(&self, at: Option<Hash>) -> RpcResult<BlockUpdate<OrderFills>>;
 }
 
 pub struct RpcServerImpl {
@@ -219,8 +220,6 @@ impl RpcServerImpl {
 
 #[derive(serde::Serialize, serde::Deserialize, Clone)]
 pub struct OrderFills {
-	block_hash: Hash,
-	block_number: BlockNumber,
 	fills: Vec<OrderFilled>,
 }
 
@@ -496,7 +495,7 @@ impl RpcServer for RpcServerImpl {
 		Ok(())
 	}
 
-	async fn order_fills(&self, at: Option<Hash>) -> RpcResult<OrderFills> {
+	async fn order_fills(&self, at: Option<Hash>) -> RpcResult<BlockUpdate<OrderFills>> {
 		let state_chain_client = &self.api.state_chain_client;
 
 		let block = if let Some(at) = at {
@@ -512,14 +511,14 @@ impl RpcServer for RpcServerImpl {
 async fn order_fills<StateChainClient>(
 	state_chain_client: Arc<StateChainClient>,
 	block: BlockInfo,
-) -> Result<OrderFills, jsonrpsee::core::Error>
+) -> Result<BlockUpdate<OrderFills>, jsonrpsee::core::Error>
 where
 	StateChainClient: StorageApi,
 {
-	Ok(OrderFills {
+	Ok(BlockUpdate::<OrderFills> {
 		block_hash: block.hash,
 		block_number: block.number,
-		fills: {
+		data: {
 			let (previous_pools, pools, events) = try_join!(
 				state_chain_client.storage_map::<pallet_cf_pools::Pools<
 					chainflip_api::primitives::state_chain_runtime::Runtime,
@@ -674,7 +673,7 @@ where
 				})
 				.collect::<Vec<_>>();
 
-			order_fills
+			OrderFills { fills: order_fills }
 		},
 	})
 }
--- a/api/bin/chainflip-lp-api/src/main.rs
+++ b/api/bin/chainflip-lp-api/src/main.rs
@@ -15,8 +15,8 @@ use chainflip_api::{
 		AccountRole, Asset, ForeignChain, Hash, RedemptionAmount,
 	},
 	settings::StateChain,
-	BlockInfo, ChainApi, EthereumAddress, OperatorApi, SignedExtrinsicApi, StateChainApi,
-	StorageApi, WaitFor,
+	BlockInfo, BlockUpdate, ChainApi, EthereumAddress, OperatorApi, SignedExtrinsicApi,
+	StateChainApi, StorageApi, WaitFor,
 };
 use clap::Parser;
 use custom_rpc::{
@@ -34,7 +34,6 @@ use jsonrpsee::{
 use pallet_cf_pools::{AssetPair, AssetsMap, IncreaseOrDecrease, OrderId, RangeOrderSize};
 use rpc_types::{AssetBalance, OpenSwapChannels, OrderIdJson, RangeOrderSizeJson};
 use sp_core::U256;
-use state_chain_runtime::chainflip::BlockUpdate;
 use std::{
 	collections::{BTreeMap, HashMap, HashSet},
 	ops::Range,
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -1,5 +1,5 @@
 use cf_amm::{
-	common::{Amount, Tick},
+	common::{Amount, Order, Tick},
 	range_orders::Liquidity,
 };
 use cf_chains::{
@@ -21,14 +21,15 @@ use jsonrpsee::{
 };
 use pallet_cf_governance::GovCallHash;
 use pallet_cf_pools::{
-	AskBidMap, AssetsMap, PoolInfo, PoolLiquidity, PoolPrice, UnidirectionalPoolDepth,
+	AskBidMap, AssetsMap, PoolInfo, PoolLiquidity, PoolPriceV1, UnidirectionalPoolDepth,
 };
 use sc_client_api::{BlockchainEvents, HeaderBackend};
 use serde::{Deserialize, Serialize};
-use sp_api::BlockT;
+use sp_api::{BlockT, HeaderT};
+use sp_core::U256;
 use sp_runtime::Permill;
 use state_chain_runtime::{
-	chainflip::Offence,
+	chainflip::{BlockUpdate, Offence},
 	constants::common::TX_FEE_MULTIPLIER,
 	runtime_apis::{
 		CustomRuntimeApi, DispatchErrorWithMessage, FailingWitnessValidators,
@@ -42,7 +43,7 @@ use std::{
 	sync::Arc,
 };
 
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 #[serde(untagged)]
 #[serde(
 	expecting = r#"Expected a valid asset specifier. Assets should be specified as upper-case strings, e.g. `"ETH"`, and can be optionally distinguished by chain, e.g. `{ chain: "Ethereum", asset: "ETH" }."#
@@ -292,6 +293,22 @@ pub struct RpcEnvironment {
 	pools: PoolsEnvironment,
 }
 
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct PoolPriceV2 {
+	pub base_asset: RpcAsset,
+	pub quote_asset: RpcAsset,
+	#[serde(flatten)]
+	pub price: pallet_cf_pools::PoolPriceV2,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct RpcPrewitnessedSwap {
+	pub base_asset: RpcAsset,
+	pub quote_asset: RpcAsset,
+	pub side: Order,
+	pub amounts: Vec<U256>,
+}
+
 #[rpc(server, client, namespace = "cf")]
 /// The custom RPC endpoints for the state chain node.
 pub trait CustomApi {
@@ -390,7 +407,14 @@ pub trait CustomApi {
 		from_asset: RpcAsset,
 		to_asset: RpcAsset,
 		at: Option<state_chain_runtime::Hash>,
-	) -> RpcResult<Option<PoolPrice>>;
+	) -> RpcResult<Option<PoolPriceV1>>;
+	#[method(name = "pool_price_v2")]
+	fn cf_pool_price_v2(
+		&self,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<PoolPriceV2>;
 	#[method(name = "swap_rate")]
 	fn cf_pool_swap_rate(
 		&self,
@@ -482,19 +506,26 @@ pub trait CustomApi {
 
 	#[method(name = "max_swap_amount")]
 	fn cf_max_swap_amount(&self, asset: RpcAsset) -> RpcResult<Option<AssetAmount>>;
-	#[subscription(name = "subscribe_pool_price", item = PoolPrice)]
+	#[subscription(name = "subscribe_pool_price", item = PoolPriceV1)]
 	fn cf_subscribe_pool_price(&self, from_asset: RpcAsset, to_asset: RpcAsset);
-
-	#[subscription(name = "subscribe_prewitness_swaps", item = Vec<AssetAmount>)]
-	fn cf_subscribe_prewitness_swaps(&self, from_asset: RpcAsset, to_asset: RpcAsset);
+	#[subscription(name = "subscribe_pool_price_v2", item = BlockUpdate<PoolPriceV2>)]
+	fn cf_subscribe_pool_price_v2(&self, base_asset: RpcAsset, quote_asset: RpcAsset);
+	#[subscription(name = "subscribe_prewitness_swaps", item = BlockUpdate<RpcPrewitnessedSwap>)]
+	fn cf_subscribe_prewitness_swaps(
+		&self,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+		side: Order,
+	);
 
 	#[method(name = "prewitness_swaps")]
 	fn cf_prewitness_swaps(
 		&self,
-		from_asset: RpcAsset,
-		to_asset: RpcAsset,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+		side: Order,
 		at: Option<state_chain_runtime::Hash>,
-	) -> RpcResult<Option<Vec<AssetAmount>>>;
+	) -> RpcResult<RpcPrewitnessedSwap>;
 
 	#[method(name = "supported_assets")]
 	fn cf_supported_assets(&self) -> RpcResult<HashMap<ForeignChain, Vec<Asset>>>;
@@ -554,7 +585,7 @@ fn map_dispatch_error(e: DispatchErrorWithMessage) -> jsonrpsee::core::Error {
 
 impl<C, B> CustomApiServer for CustomRpc<C, B>
 where
-	B: BlockT<Hash = state_chain_runtime::Hash>,
+	B: BlockT<Hash = state_chain_runtime::Hash, Header = state_chain_runtime::Header>,
 	C: sp_api::ProvideRuntimeApi<B>
 		+ Send
 		+ Sync
@@ -833,11 +864,32 @@ where
 		from_asset: RpcAsset,
 		to_asset: RpcAsset,
 		at: Option<state_chain_runtime::Hash>,
-	) -> RpcResult<Option<PoolPrice>> {
+	) -> RpcResult<Option<PoolPriceV1>> {
 		self.client
 			.runtime_api()
 			.cf_pool_price(self.unwrap_or_best(at), from_asset.try_into()?, to_asset.try_into()?)
 			.map_err(to_rpc_error)
+	}
+
+	fn cf_pool_price_v2(
+		&self,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<PoolPriceV2> {
+		let base = base_asset.try_into()?;
+		let quote = quote_asset.try_into()?;
+		let hash = self.unwrap_or_best(at);
+		Ok(PoolPriceV2 {
+			base_asset,
+			quote_asset,
+			price: self
+				.client
+				.runtime_api()
+				.cf_pool_price_v2(hash, base, quote)
+				.map_err(to_rpc_error)
+				.and_then(|result| result.map_err(map_dispatch_error))?,
+		})
 	}
 
 	fn cf_pool_swap_rate(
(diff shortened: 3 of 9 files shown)


syan095:
[test-subscription.zip](https://github.com/chainflip-io/chainflip-backend/files/13735333/test-subscription.zip)

Test program that tests the subscription and print returned data into the console

dandanlen (commented):
See my comments in the linked Linear issue.

AlastairHolmes (commented):
Can we use named parameters and include the number and the hash, for consistent. i.e.

block_hash:
block_number:
...

Thisis the direction I wanted to go.

AlastairHolmes:
Also ultimately we should do this for all the subscriptions too. Of course the other subscriptions (pool_price) we'll need to impl a "v2" (Probably should do the same for the prewitness sub as well) and keep the old version too.

AlastairHolmes:
> Also ultimately we should do this for all the subscriptions too. Of course the other subscriptions (pool_price) we'll need to impl a "v2" (Probably should do the same for the prewitness sub as well) and keep the old version too.

To do this you could add a serde(flatten) struct that includes the block_hash and block_number.

i.e. 

```rust
#[derive(...)]
struct BlockDetails<T> {
    block_hash: Hash,
    block_number: Number,
    #[serde(flatten)]
    t: T
}
```

Then we can impl a function (on top of the existing ) new_items_subscription, that adds this this `BlockDetails` wrapper.

syan095:
> This is the direction I wanted to go. See the subscribe_order_fills rpc.

@AlastairHolmes permission to also change `subscribe_order_fills` return type to this new type?

syan095:
Had to remove the serde::flatten, because it doesn't support u128 types.
Otherwise confirmed via the test program the subscriptions work.

syan095:
[test-subscription.zip](https://github.com/chainflip-io/chainflip-backend/files/13825369/test-subscription.zip)

Updated subscription test app.
Now support testing all 4 subscriptions: PoolsPrice, PoolsPriceV2, pre-witnessed swaps, filledOrders

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1093,7 +1129,20 @@ where
 	) -> Result<(), SubscriptionEmptyError> {
 		let from = from_asset.try_into().map_err(|_| SubscriptionEmptyError)?;
 		let to = to_asset.try_into().map_err(|_| SubscriptionEmptyError)?;
-		self.new_items_subscription(sink, move |api, hash| api.cf_prewitness_swaps(hash, from, to))
+		self.new_items_subscription(sink, move |api, hash, block_number| {
```
AlastairHolmes, line 1132: Can you not factor this mapping out into the new_items_subscription and new_update_subscription functions?

state-chain/runtime/src/chainflip.rs
```diff
@@ -626,3 +628,27 @@ pub fn calculate_account_apy(account_id: &AccountId) -> Option<u32> {
...
+/// Types used for subscription
+pub trait WithContent<Data> {
+	fn content(&self) -> Data;
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug, Encode, Decode)]
+pub struct SubscriptionUpdate<Data> {
+	pub block_hash: Hash,
+	pub block_number: BlockNumber,
+	pub data: Data,
```
AlastairHolmes, line 641: You should use serde(flatten) here, so the "data" doesn't show up in all the rpc responses. Also you need to create a OrderFills structure to use as Data instead of just a Vec<>. Again to preserve the response structure (For back compat).

state-chain/custom-rpc/src/lib.rs
syan095: This is very nice - we no longer need to pass blocknumber into here, and we no longer need the "withContent" trait since we have access to the inside content within the `new_items_subscription` function

state-chain/runtime/src/chainflip.rs
syan095: The OrderFilled is not contained in a struct. 
I run the test app to ensure the result is indeed flattened

syan095:
I've committed the latest version of subscription test app to our repository:
https://github.com/chainflip-io/subscription-testapp

state-chain/custom-rpc/src/lib.rs
```diff
@@ -280,6 +280,22 @@ pub struct RpcEnvironment {
 	pools: PoolsEnvironment,
 }
 
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct PoolPriceV2 {
+	pub from_asset: RpcAsset,
+	pub to_asset: RpcAsset,
+	pub price: PoolPrice,
```
AlastairHolmes, line 287: I'd suggest, flatten the PoolPrice into this struct (I dont mean serde(flatten)).

AlastairHolmes: Also adding the "V1" to the old types would be nice.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -467,8 +483,9 @@ pub trait CustomApi {
 	fn cf_max_swap_amount(&self, asset: RpcAsset) -> RpcResult<Option<AssetAmount>>;
 	#[subscription(name = "subscribe_pool_price", item = PoolPrice)]
 	fn cf_subscribe_pool_price(&self, from_asset: RpcAsset, to_asset: RpcAsset);
-
-	#[subscription(name = "subscribe_prewitness_swaps", item = Vec<AssetAmount>)]
+	#[subscription(name = "subscribe_pool_price_v2", item = SubscriptionUpdate<PoolPriceV2>)]
+	fn cf_subscribe_pool_price_v2(&self, from_asset: RpcAsset, to_asset: RpcAsset);
```
AlastairHolmes, line 487: Also for this new subscription can we use three arguments (base_asset, quote_asset, and side). We had problems with people not understanding the price for side is different.

AlastairHolmes:
Sorry I didn't see your previous notewrt u128 and serde(flatten). I think in general we don't want to use u128 in the json outputs/inputs anyway (We should stick to hex strings).

state-chain/custom-rpc/src/lib.rs
AlastairHolmes: Sorry, can we also introduce a cf_pool_price_v2, that matches the new subscription arguments and response structure.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -280,6 +280,22 @@ pub struct RpcEnvironment {
...
+pub struct PoolPriceV2 {
+	pub from_asset: RpcAsset,
+	pub to_asset: RpcAsset,
+	pub price: PoolPrice,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct RpcPrewitnessedSwap {
+	pub from_asset: RpcAsset,
+	pub to_asset: RpcAsset,
```
AlastairHolmes, line 293: Also I suggest using base, quote and side. Instead of having the order implicit based on which way around the assets are.

AlastairHolmes:
PRO-1011

state-chain/custom-rpc/src/lib.rs
syan095: Added cf_pool_price_v2 call and PoolPriceV2, which contains
```
buy_price: PoolPriceV1,
sell_price: PoolPriceV1,
```

state-chain/custom-rpc/src/lib.rs
syan095: Also renamed a bunch of "from"s to "base" and "to"s to "quote"

syan095:
> Sorry I didn't see your previous notewrt u128 and serde(flatten). I think in general we don't want to use u128 in the json outputs/inputs anyway (We should stick to hex strings).

Changed the amount type to `NumberOrHex`

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -121,10 +121,10 @@ decl_runtime_apis!(
 		fn cf_suspensions() -> Vec<(Offence, Vec<(u32, AccountId32)>)>;
 		fn cf_generate_gov_key_call_hash(call: Vec<u8>) -> GovCallHash;
 		fn cf_auction_state() -> AuctionState;
-		fn cf_pool_price(from: Asset, to: Asset) -> Option<PoolPrice>;
+		fn cf_pool_price(base: Asset, quote: Asset) -> Option<PoolPriceV1>;
```
AlastairHolmes, line 124:
```suggestion
		fn cf_pool_price(base_asset: Asset, quote_asset: Asset) -> Option<PoolPriceV1>;
```

Same below

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -121,10 +121,10 @@ decl_runtime_apis!(
...
 		fn cf_pool_simulate_swap(
```
AlastairHolmes, line 125: Here you need the "from" and "to", or you need to add a side (i.e. intro a v2 of the rpc)

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -164,7 +164,7 @@ decl_runtime_apis!(
 		fn cf_min_swap_amount(asset: Asset) -> AssetAmount;
 		fn cf_max_swap_amount(asset: Asset) -> Option<AssetAmount>;
 		fn cf_min_deposit_amount(asset: Asset) -> AssetAmount;
-		fn cf_prewitness_swaps(from: Asset, to: Asset) -> Option<Vec<AssetAmount>>;
```
AlastairHolmes, line 167: Again the from and to here are meaningful. As in the old rpcs they do implicitly describe the direction.

AlastairHolmes:
Maybe i'm missing something but it seems like you removed the new rpcs?

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -455,10 +454,16 @@ impl RpcServer for RpcServerImpl {
 				if let Err(option_error) = order_fills(state_chain_client.clone(), block)
 					.await
 					.map_err(Some)
-					.and_then(|order_fills| match sink.send(&order_fills) {
-						Ok(true) => Ok(()),
-						Ok(false) => Err(None),
-						Err(error) => Err(Some(jsonrpsee::core::Error::ParseError(error))),
+					.and_then(|order_fills| {
+						if !order_fills.data.fills.is_empty() {
```
AlastairHolmes, line 458: No we want to leave this, and not fill when there are no fills.

syan095: The reason I added this is because since we added blockhash+block number, the data that gets put into the data pipe is different every block. I've added this check here to prevent empty results from being returned every block. 
Is there a better way to prevent this spamming?

state-chain/runtime/src/runtime_apis.rs
syan095: After some thoughts, I think it is simpler to use (from, to) than (base, quote, side), because the first is more obvious and straight forward. 
What do you think ? whats your thought/preference on this?

api/bin/chainflip-lp-api/src/main.rs
AlastairHolmes: We want to return empty results, when the block has no fills, it is no spamming, the lp's need this information, and we cannot change the behaviour of the existing rpcs that are being used.

AlastairHolmes: Remember this subscription already had the block hash and number, I think you are misunderstanding something.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -369,15 +384,22 @@ pub trait CustomApi {
 	#[method(name = "pool_price")]
 	fn cf_pool_price(
 		&self,
-		from_asset: RpcAsset,
-		to_asset: RpcAsset,
+		from: RpcAsset,
```
AlastairHolmes, line 387: All the existing rpcs, need to be unchanged. The parameter names here are exposed as callers can provide json objects for the argument list i.e. { "from": "ETH"}. So changing the names of these will break the exising LP's code. Which we cannot do.

Whereas the new RPCs can have any naming, i.e. either "from" or "from_asset", because no one is using them.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -369,15 +384,22 @@ pub trait CustomApi {
...
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<PoolPriceV2>;
 	#[method(name = "swap_rate")]
 	fn cf_pool_swap_rate(
 		&self,
-		from_asset: RpcAsset,
-		to_asset: RpcAsset,
+		base_asset: RpcAsset,
```
AlastairHolmes, line 401: These need to not be change to preserve compatibility. And the "from" and "to" terminology implies a direction, that base and quote does not.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -791,29 +814,50 @@ where
 
 	fn cf_pool_price(
 		&self,
-		from_asset: RpcAsset,
-		to_asset: RpcAsset,
+		from: RpcAsset,
+		to: RpcAsset,
```
AlastairHolmes, line 818: Also can you revert the name changes of this function, for consistently with the trait definition.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -791,29 +814,50 @@ where
...
+	) -> RpcResult<PoolPriceV2> {
+		let base = base_asset.try_into()?;
+		let quote = quote_asset.try_into()?;
+		let hash = self.unwrap_or_best(at);
+		let sell_price = self
+			.client
+			.runtime_api()
+			.cf_pool_price(hash, quote, base)
+			.map_err(to_rpc_error)?;
 		self.client
```
AlastairHolmes, line 841: I think this would be cleaner, like this:

```
Ok(PoolPriceV2 {
    base_asset,
    quote_asset,
    sell_asset: self
			.client
			.runtime_api()
			.cf_pool_price(hash, quote, base)
			.map_err(to_rpc_error)?,
    buy_asset: self
			.client
			.runtime_api()
			.cf_pool_price(hash, base, quote)
			.map_err(to_rpc_error)?,
})
```

state-chain/custom-rpc/src/lib.rs
```diff
@@ -791,29 +814,50 @@ where
...
 			.map_err(to_rpc_error)
+			.map(|buy_price| PoolPriceV2 { base_asset, quote_asset, buy_price, sell_price })
 	}
 
 	fn cf_pool_swap_rate(
 		&self,
-		from_asset: RpcAsset,
-		to_asset: RpcAsset,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
```
AlastairHolmes, line 851: Here these should not be renamed.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1093,7 +1156,15 @@ where
...
+				res.map(|swaps| RpcPrewitnessedSwap {
+					from_asset,
+					to_asset,
+					amounts: swaps.into_iter().map(|s| s.into()).collect(),
+				})
+			})
+		})
 	}
 
 	fn cf_prewitness_swaps(
```
AlastairHolmes, line 1170: Also see how the return tpyes of the single requests and subscriptions for price match, I think we should probably do the same here, and make cf_prewitness_swaps return RpcPrewitnessedSwap.

Also as a note, the cf_subscribe_prewitness_swaps and cf_prewitness_swaps interfaces can be changed with worrying about backcompat as we know no one is using them. Unlikely all the other rpcs, so I'm fine to not introduce new v2 rpcs here for the prewitness rpcs.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1085,6 +1129,25 @@ where
...
 
+	fn cf_subscribe_pool_price_v2(
+		&self,
+		sink: SubscriptionSink,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+	) -> Result<(), SubscriptionEmptyError> {
+		let base = base_asset.try_into().map_err(|_| SubscriptionEmptyError)?;
+		let quote = quote_asset.try_into().map_err(|_| SubscriptionEmptyError)?;
+		self.new_update_subscription(sink, move |api, hash| {
```
AlastairHolmes, line 1140: Also can we use `new_items_subscription`, I think we will remove `new_update_subscription` when we can in the future, as we have complaints that it is unhelpful when the subs don't return every block.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1602,11 +1602,11 @@ impl<T: Config> Pallet<T> {
 		Self::try_mutate_pool(asset_pair, f)
 	}
 
-	pub fn current_price(from: Asset, to: Asset) -> Option<PoolPrice> {
-		let (asset_pair, order) = AssetPair::from_swap(from, to)?;
+	pub fn current_price(base: Asset, quote: Asset) -> Option<PoolPriceV1> {
```
AlastairHolmes, line 1605: These also should not be changed.

state-chain/runtime/src/lib.rs
```diff
@@ -1056,8 +1056,8 @@ impl_runtime_apis! {
 		///
 		/// Note: This function must only be called through RPC, because RPC has its own storage buffer
 		/// layer and would not affect on-chain storage.
-		fn cf_pool_simulate_swap(from: Asset, to:Asset, amount: AssetAmount) -> Result<SwapOutput, DispatchErrorWithMessage> {
-			LiquidityPools::swap_with_network_fee(from, to, amount).map_err(Into::into)
+		fn cf_pool_simulate_swap(base: Asset, quote:Asset, amount: AssetAmount) -> Result<SwapOutput, DispatchErrorWithMessage> {
```
AlastairHolmes, line 1059: Same here these parameter names should not be changed.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1171,6 +1244,11 @@ where
 		};
 
 		let mut previous = initial.clone();
+		let initial_update = SubscriptionUpdate {
```
AlastairHolmes, line 1247: I believe you can factor this out into a closure. I.e.

```
|info| async {
    f(&self.client.runtime_api(), info.best_hash).map(|update| {
        SubscriptionUpdate {
			block_hash: info.hash,
			block_number: info.number(),
			data: update,
		})
    })
}
```

state-chain/custom-rpc/src/lib.rs
AlastairHolmes: Actually we'll need to use a version "new_items_subscription", but one that includes an initial item like "new_update_subscription" does.

AlastairHolmes: So either add a new "new_items_subscription_with_initial" function or maybe add a boolean parameter to "new_items_subscription" that allows you to control if there is an initial item.

AlastairHolmes (commented):
Sorry that took awhile, I believe that is everything, and once these things are address we can merge.

state-chain/runtime/src/runtime_apis.rs
AlastairHolmes: Ok, I think we can leave this rpc with "from" and "to", but in the future add a new one with "base", "quote" and "side". The reason I'd like to move to suing "side" in the new rpcs, is because we had this exact problem with LPs not understanding the order of the two parameters mattered, and having a "side" arguments makes to very clear.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -280,6 +280,21 @@ pub struct RpcEnvironment {
...
+	pub quote_asset: RpcAsset,
+	pub buy_price: Option<PoolPriceV1>,
+	pub sell_price: Option<PoolPriceV1>,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct RpcPrewitnessedSwap {
+	pub from_asset: RpcAsset,
+	pub to_asset: RpcAsset,
+	pub amounts: Vec<NumberOrHex>,
```
AlastairHolmes, line 295: Is this the only place you added NumberOrHex?

AlastairHolmes: We should actually use `U256`, we are going to stop using NumberOrHex because it serializes strangely.

AlastairHolmes: Particularly depending on the size of the value, the serialization is different, which makes LPs lives pointlessly difficult.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -280,6 +280,21 @@ pub struct RpcEnvironment {
 	pools: PoolsEnvironment,
 }
 
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct PoolPriceV2 {
+	pub base_asset: RpcAsset,
+	pub quote_asset: RpcAsset,
+	pub buy_price: Option<PoolPriceV1>,
+	pub sell_price: Option<PoolPriceV1>,
```
niklasnatter, line 288: would it make sense to add the range order price (PRO-1050) here? @AlastairHolmes

api/bin/chainflip-lp-api/src/main.rs
syan095: I see

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1085,6 +1129,25 @@ where
...
+	fn cf_subscribe_pool_price_v2(
+		&self,
+		sink: SubscriptionSink,
+		base_asset: RpcAsset,
+		quote_asset: RpcAsset,
+	) -> Result<(), SubscriptionEmptyError> {
+		let base = base_asset.try_into().map_err(|_| SubscriptionEmptyError)?;
+		let quote = quote_asset.try_into().map_err(|_| SubscriptionEmptyError)?;
+		self.new_update_subscription(sink, move |api, hash| {
+			let sell_price = api.cf_pool_price(hash, quote, base)?;
```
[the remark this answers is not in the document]
syan095: I didn't add a pool_price_v2 to the Runtime API because the V2 only does what V1 does twice, I thought it was unnecessary to create a separate function for it. So here, I would do the same as the cf_pool_price_v2 and construct the PoolPriceV2 struct here. 
Does this look OK? I think the duplicate code is minimal

state-chain/custom-rpc/src/lib.rs
syan095: Oh I thought it was a bug to spam every block, but it looks like it is actually desired behaviuor

state-chain/custom-rpc/src/lib.rs
syan095: I think the initial value is obtained from client info, and subsequent info are obtained from incoming Block, these are not the same type so it would be very difficult to refactor them into shared code

state-chain/custom-rpc/src/lib.rs
syan095: Ok no problem. 
There's lots of other places that uses `NumberOrHex` to represent AssetAmounts. I'll change this to u256 here

syan095:
I unified the 2 subscription methods into one - with the option `always_update` - i.e. the user can choose to either return data every block, or return only when the inner data changes.
So now the new subscription pool-price-v2 and pre-witness will always return data every block.

kylezs:
@AlastairHolmes do you think there's still a point to leaving this open? @msgmaxim is working on a PR for the new JIT subscription, that reports swaps that are going to occurring in 2 blocks time. The idea would be, as Ben mentioned, that it's easier for them to use that then this one.

AlastairHolmes:
> @AlastairHolmes do you think there's still a point to leaving this open? @msgmaxim is working on a PR for the new JIT subscription, that reports swaps that are going to occurring in 2 blocks time. The idea would be, as Ben mentioned, that it's easier for them to use that then this one.

This should still be merged, just give a few days to get to it

api/bin/chainflip-lp-api/Cargo.toml
```diff
@@ -38,6 +38,7 @@ frame-system = { git = 'https://github.com/chainflip-io/polkadot-sdk.git', tag =
 
 # Local
 chainflip-api = { path = "../../lib" }
+state-chain-runtime = { path = "../../../state-chain/runtime" }
```
dandanlen, line 41: I don't think this is needed, looks like BlockUpdate is re-exported through chainflip-api. (Same goes for cf-primitives, not strictly necessary to import this here).

state-chain/runtime/src/chainflip.rs
```diff
@@ -638,3 +639,13 @@ pub fn calculate_account_apy(account_id: &AccountId) -> Option<u32> {
...
 }
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug, Encode, Decode)]
+pub struct BlockUpdate<Data> {
+	pub block_hash: Hash,
+	pub block_number: BlockNumber,
+	// NOTE: Flatten requires Data types to be struct or map
+	// Also flatten is incompatible with u128, so AssetAmounts needs to be String type.
+	#[serde(flatten)]
+	pub data: Data,
```
dandanlen, line 650: Do we really *need* to flatten here, given that it imposes these slightly weird conditions (AssetAmount is a string..)? 

Ie. could it not just be `update: <object>`?

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1245,12 +1252,18 @@ pub struct PoolOrderbook {
...
+pub struct PoolPriceV1 {
 	pub price: Price,
 	pub sqrt_price: SqrtPriceQ64F96,
 	pub tick: Tick,
 }
 
+#[derive(Serialize, Deserialize, Clone, Encode, Decode, TypeInfo, PartialEq, Eq, Debug)]
+pub struct PoolPriceV2 {
+	pub sell: Option<SqrtPriceQ64F96>,
+	pub buy: Option<SqrtPriceQ64F96>,
```
dandanlen, line 1264: Didn't one of the LPs have a preference for ticks? What's the reason for removing the additional information?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1229,69 +1340,43 @@ where
...
 		};
+		let stream = futures::stream::iter(std::iter::once(Ok(BlockUpdate {
+			block_hash: info.best_hash,
+			block_number: info.best_number,
+			data: initial.clone(),
+		})))
+		.chain(
+			self.client
+				.import_notification_stream()
+				.filter(|n| futures::future::ready(n.is_new_best))
```
dandanlen, line 1351: What is the meaning of `is_new_best`? Does this just mean that we ignore re-orgs?

dandanlen (approved):
Left a few comments but LGTM, i'm happy for this to be merged.

state-chain/pallets/cf-pools/src/lib.rs
AlastairHolmes: Had a very long discussion with them, they didn't understand a bunch of things. The conclusion was actually they only ever wanted the sqrt_price anyway, the others are possibly/likely misleading.

state-chain/custom-rpc/src/lib.rs
AlastairHolmes: Yes

AlastairHolmes: Sort of, it is more intended, so we skip to the latest block (if the node discovers a whole sequence of blocks at once)

state-chain/runtime/src/chainflip.rs
AlastairHolmes: Its really about, how I want the JSON to look, if we have that problem happy to just duplicate the structures in the future.

kylezs:
@msgmaxim might want to rebase your RPC PR with this merged

PR #4369, opened 2023-12-22
feat: add chaintracking metric

marcellorigotti:
# Pull Request

Closes: PRO-986

## Checklist

## Summary

Added a metric that reports the last reported block, this should help operators to quickly check if their engine is working correctly and reporting blocks from the external chains

### DIFF
--- a/engine/src/witness/common/chain_source.rs
+++ b/engine/src/witness/common/chain_source.rs
@@ -23,7 +23,7 @@ pub mod aliases {
 		}
 	}
 
-	define_trait_alias!(pub trait Index: core::fmt::Debug + Bounded + DeserializeOwned + Serialize + FullCodec + Step + PartialEq + Eq + PartialOrd + Ord + Clone + Copy + Send + Sync + Unpin + 'static);
+	define_trait_alias!(pub trait Index: core::fmt::Debug + Bounded + DeserializeOwned + Serialize + FullCodec + Step + PartialEq + Eq + PartialOrd + Ord + Clone + Copy + Send + Sync + Unpin + 'static + Into<u64>);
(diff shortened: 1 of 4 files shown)


engine/Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ["Chainflip team <https://github.com/chainflip-io>"]
 build = 'build.rs'
 edition = '2021'
 name = "chainflip-engine"
-version = '1.1.4'
```
AlastairHolmes, line 6: Is this what we are intending? @dandanlen @kylezs

AlastairHolmes: I thought we change all the version numbers together?

AlastairHolmes (approved):
Before we merge I'd like to be sure this is the intention wrt the version numbers. But everything is good. Thanks.

engine/Cargo.toml
marcellorigotti: I was trying out how to make it work! not sure what's the correct way here (first time trying out this new version bump thing).
Should I update all the versions of all our packages? lp-api, cli, etc.?

dandanlen: We do always bump all versions at the same time. However in this case i think you can bump them all to `1.2.0`. The next release is going to be a breaking one. There are already some PRs waiting to be merged that bump us to 1.2.0, so it will be in line with that.

dandanlen: In fact, since the version bump thing is not blocking, feel free to just remove the version bump - it will be added in another PR.

marcellorigotti: Ok! reverted back the version bump!

PR #4371, opened 2023-12-30
BitcoinFeeInfo: store single field, derive the rest

RGafiyatullin:
# Pull Request

Closes: PRO-1073

## Checklist

## Summary

As the value of `BitcoinFeeInfo` is stored, it is better to reduce its size.

As all the estimations are basically the function of `sats_per_kilo_byte` — only `sats_per_kilo_byte` is to be kept in this record.

As the estimations linearly depend on `sats_per_kilo_byte`, sorting by `sats_per_kilo_byte` instead of the estimations's values will not change the order (the median-value calculation in `decompose_recompose.rs` will not be affected).

### DIFF
--- a/state-chain/chains/src/btc.rs
+++ b/state-chain/chains/src/btc.rs
@@ -134,6 +134,8 @@ impl FeeEstimationApi<Bitcoin> for BitcoinTrackedData {
 	}
 }
 
+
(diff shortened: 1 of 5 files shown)


RGafiyatullin:
Closed in favour of #4372 (for CI-related reasons)

PR #4372, opened 2024-01-02
BitcoinFeeInfo: store single field, derive the rest (PRO-1073)

RGafiyatullin:
# Pull Request

Closes: PRO-1073

## Checklist

## Summary

As the value of `BitcoinFeeInfo` is stored, it is better to reduce its size.

As all the estimations are basically the function of `sats_per_kilo_byte` — only `sats_per_kilo_byte` is to be kept in this record.

As the estimations linearly depend on `sats_per_kilo_byte`, sorting by `sats_per_kilo_byte` instead of the estimations's values will not change the order (the median-value calculation in `decompose_recompose.rs` will not be affected).

## Migrations:

The `BitcoinFeeInfo` is persistently stored in two places:
- `cf-chain-tracking::CurrentChainState<T: Config, I>`;
- `cf-witnesser::ExtraCallData<T: Config>`.

### `cf-chain-tracking::CurrentChainState<T: Config, I>`

A single record with the last observed chain-state.
The old structure is replaced with the new one.

### `cf-witnesser::ExtraCallData<T: Config>`

The `ExtraCallData` is a map from `(Epoch, CallHash)` into `Vec<EncodedData>`.
This collection is filled by the data from runtime-calls of various types, the items of the collection have no hint of a Codec used to store the data.

The existing implementation would silently skip any records that can't be decoded.
I changed it to log a warning about the skipped records.

It is expected that the change of format will provoke such situations: this can get resolved when either the epoch or the call-hash changes.

**_So we can either require a new epoch to be started, or just wait for the next BTC-block to be available. Would that work?_**


state-chain/runtime/src/chainflip/decompose_recompose.rs
```diff
@@ -25,20 +21,8 @@ fn decode_many<T: Encode + Decode>(data: &mut [Vec<u8>]) -> Vec<T> {
 }
 
 fn select_median_btc_info(data: Vec<BitcoinFeeInfo>) -> Option<BitcoinFeeInfo> {
```
kylezs, line 23: I would say now we don't even need this function, and we can inline it - the tests for select_median_btc_info are basically just testing the `select_median` function now, so would delete them

state-chain/runtime/src/chainflip/decompose_recompose.rs
```diff
@@ -6,12 +6,8 @@ use pallet_cf_witnesser::WitnessDataExtraction;
 use sp_std::{mem, prelude::*};
 
 fn select_median<T: Ord + Copy>(mut data: Vec<T>) -> Option<T> {
-	if data.is_empty() {
-		return None
-	}
-
-	let len = data.len();
-	let median_index = (len - 1) / 2;
+	let len_sub_one = data.len().checked_sub(1)?;
```
kylezs, line 9: nit: would inline this too

state-chain/runtime/src/chainflip/decompose_recompose.rs
```diff
@@ -283,36 +267,14 @@ mod tests {
 		assert_eq!(select_median::<u16>(vec![]), None);
 	}
 
+	#[test]
 	// For BTC, we witness multiple values, and median should be
```
kylezs, line 271: the comment is now inaccurate - but yeah, as mentioned, I think is better to just delete the test and inline the function

state-chain/runtime/src/chainflip/decompose_recompose.rs
RGafiyatullin: I'd agree if the `Ord` would be defined for `BitcoinFeeInfo`.
Right now it can be defined as just a forward to `Ord::cmp(&self.sats_per_kilo_byte, &other.sats_per_kilo_byte)`.
But if some other data is added into `BitcoinFeeInfo`, such a definition of `Ord for BitcoinInfo` may become surprising.

state-chain/runtime/src/chainflip/decompose_recompose.rs
RGafiyatullin: I updated the test and the comment to it, so that it hopefully will make more sense.

state-chain/runtime/src/chainflip/decompose_recompose.rs
kylezs: fair

state-chain/pallets/cf-chain-tracking/src/migrations/v2.rs
```diff
@@ -0,0 +1,141 @@
...
+
+			fn from_v1(TrackedData { block_height, tracked_data }: TrackedData) -> Self {
+				log::info!("upgrading {} @{:?}", core::any::type_name::<Self>(), block_height);
+
+				fn undo(derived_fee: u64, size: u64) -> BtcAmount {
+					let a = derived_fee * BYTES_PER_KILOBYTE;
+					let q = a / size;
+					let r = a % size;
+
+					assert!(r == 0 || a == BtcAmount::MAX);
```
kylezs, line 117: Confused about the `a == BtcAmount::MAX` case, why does this matter? I assume you're saying "there might be a remainder if the multiplication saturated" ? - (which ofc requires use of saturating_mul above). 

I definitely like the use of asserts in general - but I think in this kind of code - i.e. runtime code, it's best to write defensive code that can't panic (which is what you've done by doing the `max()` chaining for example). So I'd probably err on the side of not including the assert - and using testing / reviews to check this - in particular because it's not clear that it would matter if this condition didn't hold, and could have very bad consequences if the assert is hit - i.e. a panic in the runtime, which is quite difficult to recover from.

state-chain/pallets/cf-chain-tracking/src/migrations/v2.rs
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		impl FromV1 for ChainState<cf_chains::Bitcoin> {
+			type OldType = TrackedData;
+
+			fn from_v1(TrackedData { block_height, tracked_data }: TrackedData) -> Self {
+				log::info!("upgrading {} @{:?}", core::any::type_name::<Self>(), block_height);
+
+				fn undo(derived_fee: u64, size: u64) -> BtcAmount {
+					let a = derived_fee * BYTES_PER_KILOBYTE;
```
kylezs, line 113:
```suggestion
					let a = derived_fee.saturating_mul(BYTES_PER_KILOBYTE);
```

state-chain/pallets/cf-chain-tracking/src/migrations/v2.rs
```diff
@@ -0,0 +1,141 @@
...
+					let q = a / size;
+					let r = a % size;
+
+					assert!(r == 0 || a == BtcAmount::MAX);
+
+					q
+				}
+
+				let via_out = undo(tracked_data.fee_per_input_utxo, INPUT_UTXO_SIZE_IN_BYTES);
+				let via_in = undo(tracked_data.fee_per_output_utxo, OUTPUT_UTXO_SIZE_IN_BYTES);
```
kylezs, line 123:
```suggestion
				let via_in = undo(tracked_data.fee_per_input_utxo, INPUT_UTXO_SIZE_IN_BYTES);
				let via_out = undo(tracked_data.fee_per_output_utxo, OUTPUT_UTXO_SIZE_IN_BYTES);
```

state-chain/pallets/cf-chain-tracking/src/migrations/v2.rs
```diff
@@ -0,0 +1,141 @@
...
+		use super::FromV1;
+		use crate::*;
+
+		// The following type-aliases and constants are defined here intentionally (as opposed to
+		// being imported). The types and values should correspond to the types and values that were
+		// in effect right before this migration.
+		pub type BtcBlockNumber = u64;
+		pub type BtcAmount = u64;
+
+		const BYTES_PER_KILOBYTE: BtcAmount = 1024;
```
kylezs, line 84: this is a general const, no need to dupe this

state-chain/runtime/src/chainflip/decompose_recompose.rs
```diff
@@ -20,25 +15,20 @@ fn select_median<T: Ord + Copy>(mut data: Vec<T>) -> Option<T> {
 fn decode_many<T: Encode + Decode>(data: &mut [Vec<u8>]) -> Vec<T> {
 	data.iter_mut()
 		.map(|entry| T::decode(&mut entry.as_slice()))
+		.inspect(|r|
+			if let Err(e) = r {
+				log::error!(
+					"Error decoding {}: {}.\nThis is expected to happen right after the runtime-upgrade, but should only be observed until the end of the epoch (PRO-1073)",
```
kylezs, line 21: What was the issue with migrating the ExtraCallData?

We know that ExtraCallData is always an old BitcoinFeeInfo in the Bitcoin instance of the pallet, and we know that we can decode it using decode_many and passing the old type in, then we can do the same `undo` as written in the other migration for each of these values, and then write it back ?

RGafiyatullin: `ExtraCallData` is a storage of the `witnesser` pallet.
The `witnesser` pallet does not seem to be separated in a way as `chain-traking` is.

I might be wrong though.

state-chain/pallets/cf-chain-tracking/src/migrations/v2.rs
RGafiyatullin: That would require extending the public interface of `cf-chains` without adding much value.

I will if you insist.

state-chain/chains/src/btc.rs
```diff
@@ -1388,4 +1385,12 @@ mod test {
 		assert!(!ConsolidationParameters::new(1, 1).are_valid());
 		assert!(!ConsolidationParameters::new(0, 10).are_valid());
 	}
+
+	#[test]
+	#[should_panic]
+	// XXX: @martin-chainflip 👋
```
RGafiyatullin, line 1391: @martin-chainflip 👋

AlastairHolmes: Also we should also be using saturating_add really when adding here: `self.btc_fee_info.min_fee_required_per_tx + self.btc_fee_info.fee_per_output_utx`

state-chain/pallets/cf-chain-tracking/src/migrations/v2.rs
```diff
@@ -0,0 +1,141 @@
...
+	fn on_runtime_upgrade() -> Weight {
+		// Runtime-check: only migrate the Bitcoin TrackedData
+		if T::TargetChain::NAME == cf_chains::Bitcoin::NAME {
+			// Compile-time: `impl v1::FromV1 for ChainState<Chain>`
+			//     should be defined for every `Chain` we use this migration with.
+			let translated_opt =
+				CurrentChainState::<T, I>::translate(|old| old.map(v1::FromV1::from_v1))
+					// XXX: is it okay to panic here? How to signalise an error otherwise?
+					.expect("failed to decode v1-storage");
+			if let Some(translated) = translated_opt {
```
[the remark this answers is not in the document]
RGafiyatullin: Indeed there is a call to `unhashed::put_raw/unhashed::kill` inside, but there is not a word about altering the storage in the doc :)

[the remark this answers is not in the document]
kylezs: yeah the docs certainly aren't the greatest in a lot of areas

state-chain/runtime/src/chainflip/decompose_recompose.rs
kylezs: Ah yes, ofc, you're right. hmm

kylezs (approved):
Nice work, wasn't the easiest first task 🙌

state-chain/chains/src/btc.rs
```diff
@@ -1388,9 +1385,14 @@ mod test {
...
-			BTC_FEE_MULTIPLIER.checked_mul_int(u64::MAX).expect("Martin says it'll be fine");
+	fn btc_fee_multiplier_arithmetics_negative() {
+		let _: u64 = BTC_FEE_MULTIPLIER
+			.checked_mul_int(u64::MAX)
+			.expect("fee is a u64, multiplier is 1.5, so this should never overflow");
+	}
+
+	#[test]
+	fn btc_fee_multiplier_arithmetics_positive() {
+		assert_eq!(u64::MAX, BTC_FEE_MULTIPLIER.saturating_mul_int(u64::MAX));
```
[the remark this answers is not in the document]
RGafiyatullin: It is the justification for changing the arithmetics with `BTC_FEE_MULTIPLIER` (Introduced with #4390): [My override here](020827b30)

state-chain/runtime/src/chainflip/decompose_recompose.rs
```diff
@@ -108,6 +75,39 @@ impl WitnessDataExtraction for RuntimeCall {
...
+
+fn decode_and_select<T, F>(data: &mut [Vec<u8>], mut select: F) -> Option<T>
+where
+	T: Decode,
+	F: FnMut(Vec<T>) -> Option<T>,
+{
+	match try_decode_all(data) {
+		Ok(entries) => select(entries),
+		Err(decode_err) => {
+			log::warn!("Error decoding {}: {}", core::any::type_name::<T>(), decode_err);
```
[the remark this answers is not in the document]
RGafiyatullin: According to [this](https://github.com/chainflip-io/chainflip-backend/discussions/1448) I'd keep it a warning: we failed to decode, that is misfortunate, but life goes on.

martin-chainflip (approved):

PR #4377, opened 2024-01-02
fix: pool_orders rpc filters empty orders (PRO-1039)

AlastairHolmes:


engine/multisig/Cargo.toml
```diff
@@ -19,7 +19,7 @@ futures = "0.3.14"
 futures-core = "0.3.14"
 futures-util = "0.3.14"
 hex = "0.4.3"
-itertools = "0.10"
+itertools = "0.11"
```
RGafiyatullin, line 22: That might be too petty, but switching the version of the rest of the `itertools` dependencies might shave some bytes off the runtime-size. :)

AlastairHolmes: Agreed

kylezs (approved):

PR #4380, opened 2024-01-03
chore: fix log message formatting

kylezs:
Annoying seeing the \n\t\t in some log messages like this:

```
{"timestamp":"2024-01-03T10:51:08.061388Z","level":"ERROR","fields":{"message":"Failed to connect to Polkadot node at https://polkadot.api.onfinality.io:443/rpc?apikey=2cc**** with error: Rpc error: RPC error: Networking or low-level protocol error: Server returned an error status code: 504. Please check your CFE\n\t\t\t\t\t\tconfiguration file. Retrying in 10s..."},"target":"chainflip_engine::dot::http_rpc"}
```

### DIFF
--- a/engine/src/btc/rpc.rs
+++ b/engine/src/btc/rpc.rs
@@ -86,15 +86,10 @@ impl BtcRpcClient {
 					match get_bitcoin_network(&client, &endpoint).await {
 						Ok(network) if network == expected_btc_network => break,
 						Ok(network) => {
-							error!(
-									"Connected to Bitcoin node but with incorrect network name `{network}`, expected `{expected_btc_network}` on endpoint {}. Please check your CFE
-									configuration file...",
-									endpoint.http_endpoint
-								);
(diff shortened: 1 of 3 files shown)


engine/src/btc/rpc.rs
```diff
@@ -86,15 +86,10 @@ impl BtcRpcClient {
 					match get_bitcoin_network(&client, &endpoint).await {
 						Ok(network) if network == expected_btc_network => break,
 						Ok(network) => {
-							error!(
```
AlastairHolmes, line 89: fyi you can use r#"..." to do this without having to put everything on one line. I believe, haven't used it myself before.

AlastairHolmes (approved):

engine/src/btc/rpc.rs
AlastairHolmes: They're called raw string literals.

kylezs: I think raw string literals is the opposite, it will always print the formatting, including any escapes i.e. it prints the raw message. Here I could use `\` to escape the newline in the message (not using raw string literals). I'll do that, a little nicer.

PR #4382, opened 2024-01-04
Feat: decouple CFE from SC events

msgmaxim:
# Pull Request

Closes PRO-1091

## Checklist

## Summary

After discussion with @dandanlen I decided to create a separate pallet for CFE events, so that all events are in the same place and we don't need to repeat the mechanism for cleaning up events etc (the alternative I first considered was having CFE read storage items from separate pallets).

All events are stored in the CfeEvents storage item, which can be queried by block number. In SC Observer, once we receive a new finalized header, we now query all events for that block. Events are stored for 20 blocks after which they are removed, which should give the CFE enough time to retrieve them.

All events are defined in the CfeEvent enum. I imagine we can create simple tests decoding/encoding each variant from/to some hardcoded values to ensure that we don't accidentally make incompatible changes (probably in a separate PR).

I created a few traits, some of which are generic over chain or crypto, so that our generic code (broadcaster, threshold signature etc) can use them while running slightly different code for each. I think for the most part I was able to avoid duplicating code where possible.

I'm pretty sure this doesn't require a runtime storage migration, but let me know if I'm wrong. I did test upgrading runtime from main to this branch and it worked as expected, initialising CfeEvents with the default value (empty container).


engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -292,306 +297,215 @@ where
...
-                                                        ),
-                                                        multisig::bitcoin::SigningPayload(payload),
-                                                    )
-                                                })
-                                                .collect::<Vec<_>>();
+                    for event in events {
+                    	match_event! {event, {
+                    		CfeEvent::EthThresholdSignatureRequest(req) => {
+                                handle_signing_request::<_, _, _, EthereumInstance>(
+                                    scope,
```
AlastairHolmes, line 313: formatting

state-chain/pallets/cf-cfe-event-emitter/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+	// TODO: consider taking directly taking ThresholdSignatureRequest as parameter (though this
+	// would require cf-traits importing ThresholdSignatureRequest, which is a non-primitive type)
+	fn signature_request(
+		ceremony_id: CeremonyId,
+		epoch_index: EpochIndex,
+		key: <EvmCrypto as ChainCrypto>::AggKey,
+		signatories: BTreeSet<T::ValidatorId>,
+		payload: <EvmCrypto as ChainCrypto>::Payload,
+	) {
+		add_event::<T>(CfeEvent::EthThresholdSignatureRequest(ThresholdSignatureRequest {
```
kylezs, line 137: It ought to be possible to remove a lot of the duplication in the SCO relating to these requests, and I think this might make it a little more difficult? I think that's worth keeping in mind anyway - that when we do this and put these type instances in their own enum variant - we may lose the benefit of the generic to some extent. But this is the kind of thing I think just requires some poking around to see it all fit together - might not be so simple given we probably do want to have a single unified enum of `CfeEvent` to encompass all CFE events. as we already do here.

Perhaps duplication can be removed later with a macro if we deem it necessary / nice.

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -709,6 +721,15 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
+				T::CfeEventEmitter::signature_request(
+					ceremony_id,
+					epoch,
+					key.clone(),
+					participants.clone(),
+					payload.clone(),
+				);
+
 				Event::<T, I>::ThresholdSignatureRequest {
```
kylezs, line 733: So the idea will be to leave these events here so that the upgrade is made a little simpler? Then once everyone's upgrading, we can remove the emission of these events?

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -292,306 +297,215 @@ where
...
+                            CfeEvent::DotKeygenRequest(req) => {
+                                handle_keygen_request::<_, _, _, PolkadotInstance>(
+                                    scope,
+                                    &dot_multisig_client,
+                                    state_chain_client.clone(),
+                                    req.ceremony_id,
+                                    req.epoch_index,
+                                    req.participants,
+                                ).await;
+                            }
```
kylezs, line 404: for clarity, I mean the duplication here (which I know existed before this PR) - would be nice if we can use the generics to dedup this somehow

state-chain/traits/src/lib.rs
```diff
@@ -446,6 +446,53 @@ where
...
+	fn key_handover_request(
+		_ceremony_id: CeremonyId,
+		_from_epoch: EpochIndex,
+		_to_epoch: EpochIndex,
+		_key_to_share: C::AggKey,
+		_sharing_participants: BTreeSet<T::ValidatorId>,
+		_receiving_participants: BTreeSet<T::ValidatorId>,
+		_new_key: C::AggKey,
+	) {
+		// Do nothing by default
```
kylezs, line 473: We could use `key_handover_is_required` which is on `ChainCrypto` here to have a stricter default

state-chain/pallets/cf-threshold-signature/src/lib.rs
msgmaxim: Yeah, I was going to ask whether we want to keep them or not. I prefer if we delete them, but would need to make sure that no one else is using these.

state-chain/traits/src/lib.rs
msgmaxim: Not sure if this can be done at compile time (using the current key_handover_is_required(), which is a "runtime" function), but I added an assert to check that key_handover_is_required is false in the default implementation.

state-chain/pallets/cf-cfe-event-emitter/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+	#[pallet::disable_frame_system_supertrait_check]
+	pub trait Config: Chainflip {
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn get_cfe_events)]
+	#[pallet::unbounded]
+	pub type CfeEvents<T: Config> =
+		StorageMap<_, Twox64Concat, BlockNumberFor<T>, Vec<CfeEvent<T>>>;
```
kylezs, line 58: Why do we need a block number here? Seems like an unnecessary addition to me. When you query for the events, you need to provide a block hash that those events were stored at anyway. Having the block number here is misleading since, I immediately think "I can query events by block number" but that's only the case if I also query within that block number's hash and that hash + 20... is just much easier if we query by hash and hash only.

state-chain/pallets/cf-cfe-event-emitter/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(block_number: BlockNumberFor<T>) -> frame_support::weights::Weight {
+			use frame_support::sp_runtime::Saturating;
+
+			CfeEvents::<T>::remove(block_number.saturating_sub(EVENT_LIFETIME.into()));
```
kylezs, line 65: Related to above, this won't be necessary as we'd just overwrite the existing the events in the storagevalue - rather than have a storage map

state-chain/pallets/cf-threshold-signature/src/lib.rs
kylezs: I think one strong reason to leave this at the moment is all the alerting infrastructure we have. I'm pretty sure some of the alerts use these events, we should convert them over before removing them. @marcellorigotti fyi

state-chain/traits/src/mocks/cfe_event_emitter_mock.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl<T: Chainflip> CfeEventEmitterT<T> for MockCfeEventEmitter {
+	fn peer_registered(
+		_account_id: <T as Chainflip>::ValidatorId,
+		_pubkey: cf_primitives::Ed25519PublicKey,
+		_port: u16,
+		_ip: cf_primitives::Ipv6Addr,
+	) {
+		// TODO: implement when needed for any test
```
kylezs, line 55: nit: unimplemented!() might make this easier to catch if/when someone comes to using this

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -49,45 +54,64 @@ thread_local! {
...
+				}
+			},
+			_ => {
+				// No other events used in these tests
+			},
 		}
 	}
 
-	fn process_event(event: RuntimeEvent, scenario: Scenario) {
+	fn process_event(event: RuntimeEvent) {
```
kylezs, line 110: I think now we can remove all the events from here - except the timedout case of course. Just to remove some clutter - it's clear which events are required by the CFE now

state-chain/cf-integration-tests/src/network.rs
```diff
@@ -643,8 +688,12 @@ impl Network {
 				.into_iter()
 				.map(|e| e.event)
 				.collect::<Vec<RuntimeEvent>>();
+
+			let cfe_events = state_chain_runtime::CfeEventEmitter::get_cfe_events(block_number)
+				.unwrap_or_default();
+
 			for engine in self.engines.values_mut() {
-				engine.handle_state_chain_events(&events);
+				engine.handle_state_chain_events(&events, &cfe_events);
```
kylezs, line 696: I think we should split these out, and have another processor for the runtime events, those not processed by the CFE. These integration tests have historically not been the easiest to follow, so anywhere we can make it a bit more logical, I think we should do it. Perhaps including this as a method on the network - which can then call it for each of the engines, for things like the key rotating.

This could wait for another PR though - as it may involve a bit of shifting around things in the tests.

state-chain/cf-integration-tests/src/network.rs
```diff
@@ -215,10 +218,12 @@ impl Engine {
...
-	fn handle_state_chain_events(&mut self, events: &[RuntimeEvent]) {
+	fn handle_state_chain_events(&mut self, events: &[RuntimeEvent], cfe_events: &[CfeEvent]) {
 		if self.live {
 			// Being a CurrentAuthority we would respond to certain events
 			if self.state() == ChainflipAccountState::CurrentAuthority {
+				// Note: these aren't events that the engine normally responds to, but
+				// we process them here due to the way integration tests are set up
 				on_events! {
 					events,
 					RuntimeEvent::Validator(
```
kylezs, line 229: This is actually outdated now that we optimistically rotate, the CFE should use the new key after the key activation begins.

state-chain/pallets/cf-threshold-signature/src/lib.rs
kylezs: I created an issue: https://linear.app/chainflip/issue/PRO-1096/move-alerts-that-depend-on-events-to-use-the-new-cfe-event-emitter

marcellorigotti: thanks! will talk to tom and have a look to be sure that we are not using these/to change these alerts

kylezs: @marcellorigotti checked and it seems we don't use any of these moved events in the alerts - so in that case I think we can remove these. But might be worth doing in a separate PR just in case it needs to be rolled back or something.

state-chain/pallets/cf-cfe-event-emitter/src/lib.rs
msgmaxim: > When you query for the events, you need to provide a block hash that those events were stored at anyway. 

This is only true if we immeidately clear the storage at the beginning of every new block (otherwise we'd also be getting events from previous blocks). Is that what you are suggesting? I wanted to give the CFE a bit of a buffer (EVENT_LIFETIME in this case) in case it falls behind and needs to catch up by reading older blocks. I'm worried that if we do what you are suggesting, then it would be possible for SC to clear events before CFE had a chance to read them (imagine a case where our node had a network hiccup and then received two blocks at once when recovered).

msgmaxim: Took me some time, but I see now what you are saying. Even if we overwrite the events and the CFE didn't get a chance to read them, they are still going to be available for some time when requested by the right block hash. Will simplify.

msgmaxim: Done. Note that I had to move `CfeEventEmitter` in Rutnime before any other pallet that may emit a cfe event so that we can delete old events in `on_initialize()` before any other event is added for the new block.

state-chain/traits/src/mocks/cfe_event_emitter_mock.rs
msgmaxim: This code does get executed though, so the tests would panic. It is not hard to add actual handling, but I'm not sure if it is worth it considering we don't test for these events and we are considering removing peer events in the future.

state-chain/cf-integration-tests/src/network.rs
msgmaxim: Sure, I'll save it for another PR.

state-chain/cf-integration-tests/src/network.rs
msgmaxim: I had a go at this, and it doesn't seem trivial to me. I tried using `AwaitingGovernanceActivation` instead, but then it is not emitted for Ethereum, so it would probably need a separate mechanism. I'm not even convinced that this would have been more accurate, since CFEs doesn't have a notion of "active" key. They just save every key they generate immediately and then it is up to the SC to decide when, if ever, to use them (so maybe having a map from pubkey to secret key would be more accurate). I think it is OK to leave it as is (and maybe make some improvements to integration tests in a separate PR).

kylezs (approved):
🙌  product team requests to update events can be accommodated much more easily now

state-chain/runtime/src/lib.rs
```diff
@@ -756,6 +770,7 @@ construct_runtime!(
 	{
 		System: frame_system,
 		Timestamp: pallet_timestamp,
+		CfeEventEmitter: pallet_cf_cfe_event_emitter,
```
kylezs, line 773: Discussion here: https://discord.com/channels/775961728608895008/1195210668819878030

kylezs (changes requested):

msgmaxim:
So just to summarise, in this PR we need a way to clear events at exact block boundary (so that if we request storage for a block hash, we only get events generated at that block). The current solution is/was to put the CFE events pallet before any pallet that might emit a CFE event and clear the events in on_initialize. However, as discussed on discord, this is prolematic as we don't want to change the indexes of the existing pallets. I see a few options as to how to proceed:

a) Go back to the original approach where we store events in a map so that events from different blocks can be queried separately (i.e. reverting https://github.com/chainflip-io/chainflip-backend/pull/4382/commits/9760930ee04affab2ea6f48f12b81a2d9fa53efd)

b) Clear events in the earliest pallet that currently emits CFE events, so that we can be sure that no new events are lost. @AlastairHolmes also suggested modifying the existing system pallet to do something similar (or even moving the storage for the CFE events there).

Any other suggestions? Kyle suggested that you might have some thoughts @dandanlen.

dandanlen:
We can customise the execution order of the hooks. 

In the runtime.rs we have: 

```rust
pub type Executive = frame_executive::Executive<
	Runtime,
	Block,
	frame_system::ChainContext<Runtime>,
	Runtime,
	AllPalletsWithSystem,
	PalletMigrations,
>;
```

We can replace `AllPalletsWithSystem` with our own custom tuple with the pallets in any order we please. This will then execute all the hooks in that order.

The only downsides I can think is that we need to remember to maintain this, ie. when we add a new pallet, the hooks need to be added to the custom tuple. This also applies to eg. integration tests: we need to replicate the same execution order instead of relying on the default. 

The benefit is that we don't need to worry about re-orderding pallets in the runtime declaration - as you mentioned @msgmaxim this will continue to cause issues whenever we want to add eg. a new chain. 

---

A third option would be to make the breaking change and re-order the pallets, but to use some pre-defined explicit ranges with some gaps for future additions, ie.:

```rust
construct_runtime!(
	pub struct Runtime
	{
		System: frame_system = 0,
		Timestamp: pallet_timestamp,
		CfeEventEmitter: pallet_cf_cfe_event_emitter,
		// 3: TBC
		// 4: TBC
		Environment: pallet_cf_environment = 5,
		Flip: pallet_cf_flip,
		Emissions: pallet_cf_emissions,
		// etc..
	}
);
```

This would buy us some time but will eventually also run into the same limitation (at some point we might run out of free indices between pallets).

state-chain/pallets/cf-cfe-event-emitter/src/benchmarking.rs
```diff
@@ -0,0 +1,21 @@
+#![cfg(feature = "runtime-benchmarks")]
+
+use super::*;
+use frame_benchmarking::benchmarks;
+
+benchmarks! {
+
+	clear_events {
```
dandanlen, line 8: I think this will break when we generate new benchmarks: it needs to be called `remove_events_for_block` or alternatively, the method in `WeightInfo` needs to be renamed to `clear_events`. (I actually prefer the latter given that you removed the index per-block).

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -832,6 +835,13 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				broadcast_id,
 			);
 
+			T::CfeEventEmitter::tx_broadcast_request(TxBroadcastRequest {
+				broadcast_id,
+				nominee: nominated_signer.clone(),
+				payload: broadcast_attempt.transaction_payload.clone(),
+			});
+
+			// TODO: consider removing this
```
dandanlen, line 844: Why would we not remove it? I guess polkaJS, debugging?

state-chain/traits/src/lib.rs
```diff
@@ -446,6 +447,31 @@ where
...
+	fn keygen_request(req: KeygenRequest<T::ValidatorId>);
+
+	fn signature_request(req: cfe_events::ThresholdSignatureRequest<T::ValidatorId, C>);
+
+	fn key_handover_request(_req: KeyHandoverRequest<T::ValidatorId, C>) {
+		assert!(!C::key_handover_is_required());
+	}
+}
+
+pub trait CfeEventEmitterT<T: Chainflip> {
```
dandanlen, line 460: Why not something like `CfePeerRegistration`? 

We don't need to assume that the underlying implementation emits events.

Similarly for the other traits. Could be `CfeMultisigRequest`, `CfeBroadcastRequest`.

state-chain/traits/src/mocks/cfe_event_emitter_mock.rs
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	fn peer_deregistered(
+		_account_id: <T as Chainflip>::ValidatorId,
+		_pubkey: cf_primitives::Ed25519PublicKey,
+	) {
+		// TODO: implement when needed for any test
```
dandanlen, line 62: Maybe use `todo!()` then anyone triggering this in a test will be obliged to implement?

dandanlen (commented):
FWIW I would prefer a less restrictive pallet name, for example cfe-interface - long term, we might use this pallet for things other than events, for example to group cfe-specific storage values, or to host cfe-specific extrinsics (for example the version number for compatibility checks, or submitting/storing the peer info, which is used only by the cfe, not by the runtime).

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: *edit - I saw there was another comment thread about this, think I agree with keeping it for now*

state-chain/traits/src/mocks/cfe_event_emitter_mock.rs
msgmaxim: Kyle brought this up before. This code does get called (so `todo!()` would result in a panic), it is just that we don't care about recording that event (and we have discussed removing peer update events in the future).

msgmaxim:
> The only downsides I can think is that we need to remember to maintain this

If we forget, does this mean the hooks for that pallet won't be executed? If that's the case, then I don't think it is a really big deal, since we will certaintly notice that something doesn't work.

> another downside is that the hooks are executed after runtime upgrades, meaning we would be unable to emit any cfe events during a runtime upgrade

Do you mean in migration code? Seems unlikely that there will be a need for that. Even if do we need something like that, I imagine it wouldn't be too difficult to come up with a workaround.

> We can replace AllPalletsWithSystem with our own custom tuple with the pallets in any order we please (same thing as we do with PalletMigrations). This will then execute all the hooks in that order.

Sould like a reasonable approach. Should I go ahead with this or do we want to think about this more? @dandanlen

dandanlen:
I think a custom tuple gives us the quickest win right now. Also, we'll need this anyway whenever we add a new chain. 

We will need to remain aware of the runtime upgrade weirdness: For example say we want to trigger a threshold signature or a broadcast during a runtime upgrade, the current approach would not allow this. 

I'll open a Linear issue to customize the executor/system pallet events reset mechanism.

state-chain/runtime/src/lib.rs
msgmaxim: Moved back to the end of the list (now using custom order of pallet execution instead)

state-chain/traits/src/lib.rs
msgmaxim: Renamed traits in https://github.com/chainflip-io/chainflip-backend/pull/4382/commits/d2fdc660b486df0f599e7739c66fa1b3cb2cd522

state-chain/pallets/cf-cfe-event-emitter/src/benchmarking.rs
msgmaxim: Updated in https://github.com/chainflip-io/chainflip-backend/pull/4382/commits/0b710a255652616188f3f8d6658db4ff4f0c338c

msgmaxim:
> I think a custom tuple gives us the quickest win right now. Also, we'll need this anyway whenever we add a new chain.

Made this change here: 
https://github.com/chainflip-io/chainflip-backend/pull/4382/commits/0b710a255652616188f3f8d6658db4ff4f0c338c. Tested on localnet, seems to work as expected. The order in the tuple is exactly what was used before with the cfe interface pallet moved up right after the Timestamp pallet.

dandanlen:
It looks like some integration tests are failing. Could be related to the change in ordering... either way, it looks like all the tests are failing at the same assertion, so should (hopefully!) be a simple fix. @syan095 might be able to help out here, he's quite familiar with the integration tests. (Also it looks like his recent PR conflicts with this one).

msgmaxim:
Fixed integration tests in https://github.com/chainflip-io/chainflip-backend/pull/4382/commits/71b2087565d6d59ee40097ab79db3191745fa444 and resolved the conflict with main @dandanlen.

dandanlen (approved):
@kylezs still needs to approve.

kylezs (approved):
🙌

PR #4383, opened 2024-01-04
chore: warn on attempt 0 in retrier

kylezs:
# Pull Request

Closes: PRO-1080

## Checklist

## Summary

We want to reduce some noise for monitoring people/systems. Given attempt 0 failures occur occasionally we want to make them warn instead of error - but only for requests that are `RetryLimit::NoLimit` - since requests that do have a limit, we want to know that they fail immediately, as they won't be retried for as long, or perhaps not as long. e.g. in the case of broadcasting, we want to know when we hit a broadcast error, as this could be indicative of lack of ETH in the account for example.

### DIFF
--- a/engine/src/retrier.rs
+++ b/engine/src/retrier.rs
@@ -406,7 +406,7 @@ where
 							let half_max = max_sleep_duration(initial_request_timeout, attempt) / 2;
 							let sleep_duration = half_max + rand::thread_rng().gen_range(Duration::default()..half_max);
 
-							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {}", sleep_duration);
+							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {:?}", sleep_duration);
 							if attempt == 0 && !matches!(retry_limit, RetryLimit::Limit(1)) {
 								tracing::warn!(error_message);
 							} else {
--- a/engine/src/retrier.rs
+++ b/engine/src/retrier.rs
@@ -407,7 +407,7 @@ where
 							let sleep_duration = half_max + rand::thread_rng().gen_range(Duration::default()..half_max);
 
 							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {}ms", sleep_duration.as_millis());
-							if attempt == 0 && matches!(retry_limit, RetryLimit::NoLimit) {
+							if attempt == 0 && !matches!(retry_limit, RetryLimit::Limit(1)) {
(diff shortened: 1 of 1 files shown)


engine/src/retrier.rs
```diff
@@ -405,7 +405,13 @@ where
 							// We avoid small delays by always having a time of at least half.
 							let half_max = max_sleep_duration(initial_request_timeout, attempt) / 2;
 							let sleep_duration = half_max + rand::thread_rng().gen_range(Duration::default()..half_max);
-							tracing::error!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {}ms", sleep_duration.as_millis());
+
+							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {}ms", sleep_duration.as_millis());
+							if attempt == 0 && matches!(retry_limit, RetryLimit::NoLimit) {
```
AlastairHolmes, line 410: Would this be better as, this is attempt 0 and attempt 0 is not the last attempt?

kylezs: true, think this is a nice improvement, pushed the change

engine/src/retrier.rs
```diff
@@ -405,7 +405,13 @@ where
 							// We avoid small delays by always having a time of at least half.
 							let half_max = max_sleep_duration(initial_request_timeout, attempt) / 2;
 							let sleep_duration = half_max + rand::thread_rng().gen_range(Duration::default()..half_max);
-							tracing::error!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {}ms", sleep_duration.as_millis());
+
+							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {}ms", sleep_duration.as_millis());
```
msgmaxim, line 409: Nit: `Duration`'s implementation of `Debug` is actually quite good, printing in human readable format, converting to seconds for long duration etc (e.g. see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f90cc8163855cbc791c87aa5fe83ef4e)). Should we prefer that to converting to milliseconds manually?

msgmaxim (approved):

engine/src/retrier.rs
kylezs: I think it makes it a bit easier to search the messages, so it doesn't change `ms` to `seconds` then to `minutes` etc. it'll just always be millis.

kylezs: Discussed with Alastair, and he said there were some real issues when trying to read the logs when debugging one of the issues. And upon reflection, I think ultimately while log searching is a valid concern, in this case, it'd be rare to be searching for particular "ms" values in the logs.

PR #4388, opened 2024-01-05
fix: bump subxt version to 0.32.1

kylezs:
# Pull Request

Closes: PRO-1000

## Checklist

## Summary

On bad blocks, which we did receive from getblock.io at some point - as in the issue, subxt would previously panic. Now, instead, they throw an error in 0.32.0+ version. So here I'm bumping the version to 0.32.1, since upgrading to 0.33.0 (latest at time of writing) requires a compiler version bump, which isn't necessary right now.

### DIFF
--- a/engine/src/dot/http_rpc.rs
+++ b/engine/src/dot/http_rpc.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use cf_chains::dot::{PolkadotHash, RuntimeVersion};
 use cf_primitives::PolkadotBlockNumber;
 use futures_core::Future;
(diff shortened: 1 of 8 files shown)


engine/src/dot/http_rpc.rs
```diff
@@ -72,22 +73,23 @@ impl RpcClientT for PolkadotHttpClient {
...
-	) -> RpcFuture<'a, RpcSubscription> {
+	) -> RawRpcFuture<'a, RawRpcSubscription> {
 		unimplemented!("HTTP Client does not support subscription");
 	}
 }
 
 #[derive(Clone)]
 pub struct DotHttpRpcClient {
 	online_client: OnlineClient<PolkadotConfig>,
+	rpc_methods: LegacyRpcMethods<PolkadotConfig>,
```
AlastairHolmes, line 84: Can you make an issue to move to the new rpcs?

kylezs: Created the issue: PRO-1094

AlastairHolmes (approved):

engine/src/dot/http_rpc.rs
AlastairHolmes: > afaik they're not fully stable yet - at least not within subxt, I assume subxt just called it "legacy" so they don't have to rename things again

Just for my benefit, what do you mean but not fully stable? Do you mean they said somewhere the interfaces may be changed? Or do you mean something else.

PR #4389, opened 2024-01-08
refactor: code cleaniness

AlastairHolmes:


engine/src/dot/http_rpc.rs
```diff
@@ -32,12 +32,14 @@ pub struct PolkadotHttpClient(HttpClient);
...
-		let mut headers = HeaderMap::new();
-		headers.insert(AUTHORIZATION, token.parse().unwrap());
-		let client = HttpClientBuilder::default().set_headers(headers).build(url)?;
-
-		Ok(Self(client))
+		Ok(Self(
+			HttpClientBuilder::default()
+				.set_headers(HeaderMap::from_iter([(
+					AUTHORIZATION,
+					const_eval!(HeaderValue, HeaderValue::from_static("Bearer TOKEN")),
```
kylezs, line 39: Do we even need this header actually? It is a little strange that the token is just TOKEN all the time 🤔

AlastairHolmes: Keine ahnung

kylezs (approved):

PR #4394, opened 2024-01-08
Bitcoin fee for broadcast was set below minimum resulting in broadcast aborted

martin-chainflip:
# Pull Request

Closes: PRO-1082

## Checklist

## Summary

To spend one of our deposit UTXOs, we need:
32 bytes for the TX ID
4 bytes for the output index
1 byte to indicate that this is a segwit input
4 bytes for the sequence number
= 41 bytes
in addition to that we need the witness data, which is:
1 byte for the element counter
65 bytes for the signature
1 length byte for the unlock script
x bytes for the salt
1 byte to drop the salt
33 bytes for the pubkey
1 bytes to check the signature
34 bytes for the internal pubkey
= 136 bytes + x

Initially, we assumed that the salt would be 1 byte long, but since it is varint encoded, it can grow up to 9 bytes.
In Bitcoin, each byte of witness data is counted as only 1/4 of a byte towards the size of the transaction, so assuming the largest possible salt value, we have:
utxo size = 41 + (136 + 9) / 4 = 77.25 bytes
since we may add multiple utxos together, the fractional parts may add up to another byte, so we are rounding up to be on the safe side and set the UTXO size to 78 bytes

### DIFF
--- a/state-chain/pallets/cf-environment/src/tests.rs
+++ b/state-chain/pallets/cf-environment/src/tests.rs
@@ -53,7 +53,7 @@ fn test_btc_utxo_selection() {
 		add_utxo_amount(dust_amount);
 
 		// select some utxos for a tx
-		const EXPECTED_CHANGE_AMOUNT: crate::BtcAmount = 24770;
+		const EXPECTED_CHANGE_AMOUNT: crate::BtcAmount = 24680;
 		assert_eq!(
 			Environment::select_and_take_bitcoin_utxos(UtxoSelectionType::Some {
 				output_amount: 12000,
@@ -70,7 +70,7 @@ fn test_btc_utxo_selection() {
 		assert_eq!(
 			Environment::select_and_take_bitcoin_utxos(UtxoSelectionType::SelectAllForRotation)
 				.unwrap(),
-			(vec![utxo(5000000), utxo(100000), utxo(EXPECTED_CHANGE_AMOUNT),], 5121970)
+			(vec![utxo(5000000), utxo(100000), utxo(EXPECTED_CHANGE_AMOUNT),], 5121790)
 		);
 
 		// add some more utxos to the list
@@ -88,7 +88,7 @@ fn test_btc_utxo_selection() {
 		assert_eq!(
 			Environment::select_and_take_bitcoin_utxos(UtxoSelectionType::SelectAllForRotation)
 				.unwrap(),
-			(vec![utxo(5000), utxo(15000),], 17950)
+			(vec![utxo(5000), utxo(15000),], 17890)
 		);
 	});
 }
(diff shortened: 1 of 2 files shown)


state-chain/chains/src/btc.rs
```diff
@@ -15,7 +15,7 @@ use bech32::{self, u5, FromBase32, ToBase32, Variant};
 pub use cf_primitives::chains::Bitcoin;
 use cf_primitives::{
 	chains::assets, NetworkEnvironment, DEFAULT_FEE_SATS_PER_KILO_BYTE, INPUT_UTXO_SIZE_IN_BYTES,
-	MINIMUM_BTC_TX_SIZE_IN_BYTES, OUTPUT_UTXO_SIZE_IN_BYTES,
+	MINIMUM_BTC_TX_SIZE_IN_BYTES, MINIMUM_FEE_SATS_PER_KILO_BYTE, OUTPUT_UTXO_SIZE_IN_BYTES,
```
kylezs, line 18: ik this was this way before in some places, but kilobyte is one word, so it should be `MINIMUM_FEE_SATS_PER_KILOBYTE`,

martin-chainflip:
changed the approach in this PR, i didn't like the previous idea

state-chain/primitives/src/lib.rs
```diff
@@ -58,7 +58,7 @@ pub const FLIPPERINOS_PER_FLIP: FlipBalance = 10u128.pow(FLIP_DECIMALS);
 pub const DEFAULT_FEE_SATS_PER_KILO_BYTE: u64 = 102400;
 
 // Approximate values calculated
-pub const INPUT_UTXO_SIZE_IN_BYTES: u64 = 75;
+pub const INPUT_UTXO_SIZE_IN_BYTES: u64 = 78;
```
kylezs, line 61: worth adding a comment explaining why it's this value

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -53,7 +53,7 @@ fn test_btc_utxo_selection() {
 		add_utxo_amount(dust_amount);
 
 		// select some utxos for a tx
-		const EXPECTED_CHANGE_AMOUNT: crate::BtcAmount = 24770;
+		const EXPECTED_CHANGE_AMOUNT: crate::BtcAmount = 24680;
```
kylezs, line 56: why is this the expected amount? could we include a comment here explaining this, as well as the other static values

state-chain/primitives/src/lib.rs
martin-chainflip: done

kylezs (approved):

PR #4396, opened 2024-01-09
chore: increase limit on max number of bitcoin payloads in a ceremony to theoretical maximum

kylezs:
# Pull Request

## Checklist

## Summary

I've increased the limit to the theoretical maximum number of signing payloads we could receive, rather than removing the limit. This is to protect against potential spam attacks. Created this PRO-992.

### DIFF
--- a/engine/multisig/src/crypto/bitcoin.rs
+++ b/engine/multisig/src/crypto/bitcoin.rs
@@ -70,7 +70,7 @@ impl ChainSigning for BtcSigning {
 
 	/// The window is smaller for bitcoin because its block time is a lot longer and it supports
 	/// multiple signing payloads
-	const CEREMONY_ID_WINDOW: u64 = 1500;
+	const CEREMONY_ID_WINDOW: u64 = 50;
 }
 
 impl CryptoScheme for BtcCryptoScheme {
--- a/engine/multisig/src/lib.rs
+++ b/engine/multisig/src/lib.rs
@@ -14,12 +14,9 @@ pub mod client;
 mod crypto;
 
 /// Maximum number of payloads in a single bitcoin signing ceremony
-// We choose 5000 because the maximum size of a Bitcoin block is 1MB.
-// The average size of a transaction is around 500 bytes, however, historically
-// this was 250 bytes. See this thread: https://bitcoin.stackexchange.com/q/31974/109777.
-// Given we only accept a single UTXO in each of these transactions, the maximum number of signing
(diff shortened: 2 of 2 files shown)


martin-chainflip (approved):

msgmaxim:
Just to summarise what I discussed with @kylezs yesterday:

- The limit's primary purpose is to mitigate a possible spam attack (rather than protecting the CFE from performing a ceremony that is likely to fail). We did, however, choose `MAX_BTC_SIGNING_PAYLOADS` as roughtly the maximum number of concurrent ceremonies we think the CFE can handle (though we haven't done thorough benchmarking for this).
- The spam attack is as follows: in the current design nodes have to be able to receive/store stage 1 messages from other nodes as some of them are almost guaranteed to arrive before we receive the corresponding request to sign from the SC (which would tell us the details of the ceremony, like the number of payloads). This means that without a limit, the nodes will have to accept messages even if they are for 1 million payloads or more, with the expectation that a request for 1 million payloads will arrive shortly. These messages are stored in memory and can lead to an OOM crash. What makes it more challenging is that it is technically possible (though somewhat uncommon) to receive such messages for more than a single ceremony (again, before we receive requests for any of them). We thus also limit the number of concurrent *unauthoriesd* (i.e. without a corresponding request from SC) ceremonies allowed to `CEREMONY_ID_WINDOW` in code, which we set (very generously) to 1500 for BTC.

The size of stage 1 message for BTC is: `66 * number_of_payloads + 8` bytes. One current authority can send us CEREMONY_ID_WINDOW * (66 * MAX_BTC_SIGNING_PAYLOADS + 8), which is about 100MB. This isn't much, but we wanted to leave room for future chains supporting multiple payloads and whatever else is happening on the node. We can easily increase `MAX_BTC_SIGNING_PAYLOADS`, especially if we reduce `CEREMONY_ID_WINDOW` (something like 10 would be more than enough given our usage), and then set `MAX_BTC_SIGNING_PAYLOADS` to whatever theoretically possible for BTC. I don't think we should remove the limit altogether though.

kylezs:
How do we feel about a value like 20,000?

msgmaxim:
> How do we feel about a value like 20,000?

I don't mind setting `MAX_BTC_SIGNING_PAYLOADS` to 20_000 and reducing `CEREMONY_ID_WINDOW` (only for BTC) to 50.

msgmaxim (approved):

PR #4401, opened 2024-01-10
feature/PRO-1037/include-fees-uncollected-positions

Janislav:
# Pull Request

Closes: PRO-1037

## Checklist

## Summary


bouncer/shared/lp_api_test.ts
```diff
@@ -47,7 +47,9 @@ async function provideLiquidityAndTestAssetBalances() {
   let ethBalance = 0;
   do {
     const balances = await lpApiRpc(`lp_asset_balances`, []);
-    ethBalance = balances.Eth;
+    ethBalance = parseInt(
```
Janislav, line 50: Like it seemed this test has never tested anything, ethBalance was undefined and undefined < a number is false :D. Thanks JavaScript :D.

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -297,14 +306,22 @@ impl RpcServer for RpcServerImpl {
...
 	async fn asset_balances(&self) -> RpcResult<BTreeMap<ForeignChain, Vec<AssetBalance>>> {
-		let mut balances = BTreeMap::<_, Vec<_>>::new();
-		for (asset, balance) in self.api.query_api().get_balances(None).await? {
-			balances
-				.entry(ForeignChain::from(asset))
-				.or_default()
-				.push(AssetBalance { asset, balance: balance.into() });
-		}
-		Ok(balances)
+		let result = self
```
AlastairHolmes, line 309: `result` is a bit ambiguous, maybe `asset_balances`

martin-chainflip (approved):

martin-chainflip:
Please don't merge until Alastair has approved.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -55,6 +55,12 @@ pub struct RuntimeApiAccountInfoV2 {
 	pub restricted_balances: BTreeMap<EthereumAddress, u128>,
 }
 
+#[derive(Encode, Decode, Eq, PartialEq, TypeInfo, Serialize, Deserialize)]
+pub struct ChainAsset {
```
AlastairHolmes, line 59: I assume you need this type, as RpcAsset isn't no-std compatible? I believe we could remove it in the future, and have added a note to PRO-1157 to do so.

state-chain/runtime/src/lib.rs
```diff
@@ -1019,6 +1019,12 @@ impl_runtime_apis! {
 				})
 				.collect()
 		}
+		fn cf_asset_balances(account_id: AccountId) -> Vec<(ChainAsset, AssetAmount)> {
+			LiquidityProvider::asset_balances(&account_id)
+				.into_iter()
+				.map(|asset_balance| (ChainAsset {chain: asset_balance.asset.into(), asset: asset_balance.asset}, asset_balance.balance))
```
AlastairHolmes, line 1025: Don't think this conversion is needed, yo seem to be going from ChainAsset to ChainAsset.

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -360,4 +364,16 @@ impl<T: Config> LpBalanceApi for Pallet<T> {
 		});
 		Ok(())
 	}
+
+	fn asset_balances(who: &Self::AccountId) -> Vec<AssetBalance> {
```
AlastairHolmes, line 368: I think it probably makes sense to delete this AssetBalance type, and just use a tuple of ChainAsset, and AssetBalance.

AlastairHolmes: Basically so it matches the interface of: cf_asset_balances

AlastairHolmes (approved):
Just this one change to remove the new `struct AssetBalance`, then happy to merge. Thank you.

AlastairHolmes:
You could instead keep the AssetBalances struct, but change the cf_asset_balances rpc to return it.

Currently cf_asset_balances would return something like this: `[{"chain":"Bitcoin","asset":"BTC"},34]`

If we keep the cf_asset_balances struct we could have the encoding look like this (via serde(flatten)):
{"chain":"Bitcoin","asset":"BTC", "amount": 34}

state-chain/pallets/cf-lp/src/lib.rs
Janislav: I think we can't because it's returned from the lp-api, and I'm pretty sure we don't want to have breaking changes, right?

state-chain/runtime/src/runtime_apis.rs
Janislav: In theory, we could move it to the runtime, but it's a type created for the RPC and I don't think it belongs there. Importing it from the RPC to the runtime is not possible because the dependencies are already set the other way around. But in any case we need a type that reflects the chain and the asset also for the custom RPC of the state-chain to the outer world. Internally, we can cast from Asset to Chain, but that's not possible if we return it from RPC. I think the cause of this type just shows that we have a general problem with our type system. The whole thing breaks at the moment with introduce USDC (for e.x) on any other chain.

state-chain/pallets/cf-lp/src/lib.rs
AlastairHolmes: There are two types both called AssetBalance, I mean the other one.

AlastairHolmes: But yes we don't want breaking changes.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -57,7 +57,9 @@ pub struct RuntimeApiAccountInfoV2 {
 
 #[derive(Encode, Decode, Eq, PartialEq, TypeInfo, Serialize, Deserialize)]
 pub struct ChainAsset {
+	#[serde(flatten)]
```
AlastairHolmes, line 60: I'm not sure this will do anything in this case, same with the asset below.

AlastairHolmes: I think this type isn't used now?

Janislav: Good catch 🕵️

AlastairHolmes (approved):

AlastairHolmes:
Good to merge

PR #4408, opened 2024-01-11
feat: bouncer command to check how many validator witnessed something

marcellorigotti:
# Pull Request

Closes: PRO-1087

## Checklist

## Summary

Added bouncer command to check the number of validator in the active set which witnessed something.
Currently it support 4 different arguments, DOT, BTC, ETH which will make the script wait for the next chaintracking extrinsic being reported for the specified chain and then after a couple of state chain blocks check how many validator witnessed it. (this could take a while for BTC since block production is supposed to be every 10m)

Otherwise HASH can be used followed by the hash to check, this will query the storage item with the given hash and immediately check how many validator reported it, in order to build the correct hash polkaJS can be used as explained in the INSTRUCTION at the beginning of the file.


marcellorigotti:
I had to disable the eslint check for restricted syntax otherwise we cannot use `for of`, which unfortunately cannot be replaced with `foreach` when doing async calls.

bouncer/commands/check_wintesses.ts
```diff
@@ -0,0 +1,226 @@
...
+// Go to PolkaJS > Developer > Extrinsic
+// from there, once you built the extrinsic you need the encoded call hash
+
+// For example: ./commands/check_witnesses.ts ETH
+// will wait for the next chainStateUpdate extrinsic for ethereum and after some blocks (2) it will check how many validator witnessed it
+
+import { blake2AsHex } from '@polkadot/util-crypto';
+import { runWithTimeout, sleep, getChainflipApi } from '../shared/utils';
+
+function hex2bin(hex0x: string) {
```
kylezs, line 19: You can just use `parseInt(someHexString, 16).toString(2)` ? Would have to strip out the 0x first still though

bouncer/commands/check_wintesses.ts
```diff
@@ -0,0 +1,226 @@
...
+          // parse the data and removed useless comas
+          finalData.new_chain_state.trackedData.baseFee = baseFee;
+          finalData.new_chain_state.blockHeight = blockHeight;
+          // create the extrinsic we need to witness (ETH chain tracking in this case)
+          const extrinsic = api.tx.ethereumChainTracking.updateChainState(
+            finalData.new_chain_state,
+          );
+          // obtain the hash of the extrinsic call
+          const blakeHash = blake2AsHex(extrinsic.method.toU8a(), 256);
+          witnessHash.add(blakeHash);
```
kylezs, line 129: Seems like this could be a function - repeated a number of times

bouncer/commands/check_wintesses.ts
```diff
@@ -0,0 +1,226 @@
...
+          }
+          console.log(
+            `${witnessNumber}/${validators?.length} witnessed ${elem} hash!\nThe extrinsic was in block ${currentBlockNumber}`,
+          );
+          failingValidators.forEach((element) => {
+            if (vanityNames[element] && vanityNames[element].substr(0, 2) === '0x') {
+              const vanity = vanityNames[element].substr(2);
+              const bytes = [];
+              for (let i = 0; i < vanity.length; i += 2) {
+                bytes.push(parseInt(vanity.substr(i, 2), 16));
```
kylezs, line 209: Looks like you actually use the parseInt version here already 🤔

bouncer/commands/check_wintesses.ts
marcellorigotti: This was problematic because of the max integer size in js. Which is way lower than the value represented by the hex, not sure if there is another direct way to convert hex to binary.

kylezs: I see - should at least leave a comment that this is the reason for this function. 

Also, just noticed the file has a typo, should be `check_witnesses`

dandanlen:
FWIW I think it would be easier / more accurate to add a runtime rpc that takes a call or a hash and looks up the witness votes / does the calculation natively in rust instead of trying to replicate everything in javascript. This would have the further advantage of being easy to use from the command line or polkaJS.

marcellorigotti:
EDIT:
- Addded cf_witness_count(hash) as a custom rpc
- Added CLI command count-witnesses <HASH>, which uses the custom rpc to retrieve how many validators are not witnessing something
- Modified the bouncer command to use the newly created custom rpc

state-chain/custom-rpc/src/lib.rs
```diff
@@ -487,6 +489,9 @@ pub trait CustomApi {
 		&self,
 		broadcast_id: BroadcastId,
 	) -> RpcResult<Option<<cf_chains::Ethereum as Chain>::Transaction>>;
+
+	#[method(name = "witness_count")]
+	fn cf_witness_count(&self, hash: String) -> RpcResult<Option<FailingWitnessValidators>>;
```
dandanlen, line 494: I think you can use `state_chain_runtime::Hash` as the type here and it will deserialize correctly. Avoids having to use `from_hex` to convert the string.

state-chain/runtime/src/lib.rs
```diff
@@ -1342,6 +1344,53 @@ impl_runtime_apis! {
...
+			match vote {
+				Some(bytes) => {
+					let (count, indices) = count_set_bits_and_indices(&bytes, validators.len());
+					let vanity_names: BTreeMap<AccountId, Vec<u8>> = pallet_cf_validator::VanityNames::<Runtime>::get();
+					result.number = count;
+					for elem in indices {
+						// we can safely unwrap since it is impossible to retrieve indices greater than the number of elements
+						let validator: &AccountId = validators.iter().nth(elem).unwrap();
+						// we can safely unwrap since it is impossible to not have a vanity record for an authority
+						let vanity: &Vec<u8> = vanity_names.get(validator).unwrap();
```
dandanlen, line 1386: AFAIK This isn't true - authorities are not required to have a vanity name.

state-chain/runtime/src/lib.rs
```diff
@@ -1342,6 +1344,53 @@ impl_runtime_apis! {
 				None
 			}
 		}
+
+		fn cf_witness_count(hash: pallet_cf_witnesser::CallHash) -> Option<FailingWitnessValidators> {
```
dandanlen, line 1348: I think it would be useful to give a full list of who did and didn't vote instead of just the missing votes. Basically a Vec of { id, vanity, vote }.

state-chain/runtime/src/lib.rs
```diff
@@ -1342,6 +1344,53 @@ impl_runtime_apis! {
...
+			// helper function to compute the number of bits set and their indices
+			fn count_set_bits_and_indices(bytes: &[u8], size: usize) -> (u32, Vec<usize>) {
```
dandanlen, line 1350: I think I didn't express very well what I meant with my other comment. 

The witnesser pallet already uses a crate `bitvec` that has utilities to index into a vector of bits. 

You could load the votes into a BitVec and then just iterate this to get a Vec of (authority, vote, vanity_name). Something like this:

```rust
let votes = BitVec::from_vec(Votes::<Runtime>::get(callhash)?);
let vote_count = votes.count_ones();
let missing_vote_count = votes.count_zeros();

let result: Vec<_> = votes.iter()
	// by_vals is needed to convert to true/false bool values.
	.by_vals()
	// authorities are stored in the same order as the votes
	.zip(authorities)
	.map(|(vote, account_id)| (
		account_id,
		vote,
		vanity_names[account_id],
	))
	.collect();
```

This can all be implemented as a public fn inside the witnesser pallet.

state-chain/runtime/src/lib.rs
marcellorigotti: You are right, I though even with no vanity name they still had an entry in the map (which turns out is not the case)

marcellorigotti:
A quick comment on this.
It would be nice if we can include this PR in the next release such that we can then use the bouncer command for mainnet/perseverance (otherwise it won't work since it uses a new runtime rpc)

state-chain/pallets/cf-witnesser/src/lib.rs
```diff
@@ -479,6 +479,25 @@ impl<T: Config> Pallet<T> {
...
 		CallHashExecuted::<T>::insert(witnessed_at_epoch, call_hash, ());
 	}
+
+	pub fn count_votes(callhash: CallHash) -> Option<Vec<(<T as Chainflip>::ValidatorId, bool)>> {
+		let current_epoch = T::EpochInfo::epoch_index();
+		let votes: BitVec<u8, Msb0> =
+			BitVec::from_vec(Votes::<T>::get(current_epoch, callhash).unwrap_or(vec![]));
+		if votes.is_empty() {
+			return None
+		}
```
dandanlen, line 489: nit: I think this is equivalent and more idiomatic:

```suggestion
			BitVec::from_vec(Votes::<T>::get(current_epoch, callhash)?);
```

A lot of these variables can also be inlined.

api/lib/src/queries.rs
```diff
@@ -220,6 +220,21 @@ impl QueryApi {
 		result.next_block_in = Some(compute_distance(index, current_relative_slot, validator_len));
 		Ok(result)
 	}
+
+	pub async fn check_witnesses(
+		&self,
+		_block_hash: Option<state_chain_runtime::Hash>,
```
dandanlen, line 226: Might be a good idea to allow querying historical blocks... This can be useful for debugging things in the past. It should be a relatively simple addition in the runtime rpc. LMK, I can do it on Monday.

api/bin/chainflip-cli/src/main.rs
```diff
@@ -285,6 +288,21 @@ async fn pre_update_check(api: QueryApi) -> Result<()> {
 	Ok(())
 }
 
+async fn count_witnesses(api: QueryApi, hash: state_chain_runtime::Hash) -> Result<()> {
+	let result = api.check_witnesses(None, hash).await?;
+	match result {
+		Some(value) => {
+			println!("Number of validator who failed to witness it: {}", value.failing_count);
+			println!("List of validator: {:?}", value.validators);
```
dandanlen, line 296: nit:

```suggestion
			println!("Number of authorities who failed to witness it: {}", value.failing_count);
			println!("List of witness votes:\n {:?}", value.validators);
```

dandanlen (commented):
Good work @marcellorigotti 

Just a final thing to add 😅 I think it would be useful to allow querying the witness votes at historical block hashes.

dandanlen (approved):

PR #4409, opened 2024-01-12
fix: restore correct restriction on redemption expiry (PRO-1072)

Janislav:
# Pull Request

Closes: PRO-1072

## Checklist

## Summary

If the redemption amount to a restricted address is bigger than the restricted amount **and** there are more free balances available than the restricted amount **and** the redemption is expiring, we use the total amount of the redemption and set it as the restricted amount. That in fact means that the address restriction is higher than before, which is wrong. To void that, we have to remember somewhere how high the restriction was to later be able to separate the restricted funds from the free one.

I was initially planning to extend the `PendingRedemptions` storage item to also take the restricted amount but due to the fact that this is not normal case and also will probably disappear one day completely after vesting is done (I guess) I decided to introduce a new storage item. This also better in terms of:

1) No risky and annoying migration needed
2) Less complex types
3) A restricted balance and redemption doesn't belong necessarily together conceptually (in my opinion)
4) Easier to understand


state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -529,11 +540,20 @@ pub mod pallet {
 
 			// If the address is still restricted, we update the restricted balances again.
 			if RestrictedAddresses::<T>::contains_key(address) {
+				let restricted_balance_for_address =
+					PendingRedemptionsRestrictedBalance::<T>::take(&account_id).unwrap_or_default();
+				// Ensure that the balance we set for the restricted address is not more than the
+				// balance we have recorded during redemption for the address.
+				let restricted_amount = if amount >= restricted_balance_for_address {
```
dandanlen, line 547: We don't expect this to ever be the case (and IMO adding this conditional clause suggests that it might legitimately occur). I would prefer a `debug_assert` here to ensure that the condition holds in our unit tests.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -406,6 +413,8 @@ pub mod pallet {
 
 			// If necessary, update account restrictions.
 			if let Some(restricted_balance) = restricted_balances.get_mut(&address) {
+				// We need to keep track of the restricted balance in case the redemption expires.
+				PendingRedemptionsRestrictedBalance::<T>::insert(&account_id, *restricted_balance);
```
dandanlen, line 417: This is incorrect. It should be `min(debit_amount, restricted_balance)`. 

Imagine the restricted balance is 100 and the user redeems 10 of this and it expires: The new restricted amount would become 190! 

We should add a test case for this.

dandanlen (commented):
Tbh. i think i'd prefer if we added the restricted amount to the existing `PendingRedemptions` storage. Makes it easier to be sure it's used consistently (related storage items are accessed together).

For example instead of a tuple we could store a struct:

```rust
struct PendingRedemptionInfo {
	pub total: FlipBalance,
	pub restricted: FlipBalance,
	pub redeem_address: EthereumAddress,
}
```

This is nice and explicit and the storage migration is very simple to write. It's true this functionality is likely to be phased out eventually, but we're talking 1 or 2 years... in the meantime it needs to be maintainable and resistant to regressions.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: No, we still take the amount in that case. You just check it here in advance and I in the expiry function.

dandanlen: Ok... I can see that now. This explains why I was confused by the if/else check below. 

I find this quite confusing: we are storing the entire 'old' restricted balance and then checking during expiry if that old balance was greater than the redemption amount. 

We could just do the check when we process the redemption, and store the amount of restricted funds actually debited. This also makes it easy to see (in polkaJS for example) how much was taken from the account, and how much of that was restricted.

ramizhasan111: agree with dan here. The storage should store the amount of restricted funds in the pending redemption. Easier to  reason and follow what is happening .

Janislav: Agree 👌

state-chain/pallets/cf-funding/src/tests.rs
```diff
@@ -596,14 +594,13 @@ fn redemption_expiry_removes_redemption() {
...
-		assert!(PendingRedemptionsRestrictedBalance::<Test>::get(&ALICE).is_none());
 
 		// Tax was paid, rest is returned.
 		assert_eq!(Flip::total_balance_of(&ALICE), TOTAL_FUNDS - REDEMPTION_TAX);
-		// Restricted funds are restricted again, minus redemption tax.
+		// Restricted funds are restricted again.
 		assert_eq!(
 			*RestrictedBalances::<Test>::get(&ALICE).get(&RESTRICTED_ADDRESS).unwrap(),
-			RESTRICTED_AMOUNT - REDEMPTION_TAX
+			RESTRICTED_AMOUNT
```
dandanlen, line 603: Actually, strictly speaking this was correct - the redemption tax *is* paid by the restricted funds. If we want to be totally correct, we should include this in our calcuation, ie line 464 of lib.rs:

```
restricted: min(total_restricted_balance, debit_amount)
```

should be something like:

```
restricted: min(total_restricted_balance, debit_amount).saturating_sub(redemption_tax)
```

Janislav: Why is this the case? The redemption tax should not affect the restricted amount in case of an expiry. In any case, the amount has to reset to the amount before the redemption attempt. Everything else seems a bit odd to me.

PR #4412, opened 2024-01-12
fix: await finalisation before starting broker

kylezs:
Fixes a few issues with the ugprade test - it's stil not currently working on main - still looking into that (I suspect there's a broken migration on main - but this should help).

### DIFF
--- a/.github/workflows/upgrade-test.yml
+++ b/.github/workflows/upgrade-test.yml
@@ -2,10 +2,6 @@ name: Test upgrade from latest release to main
 
 on:
   workflow_dispatch:
-  pull_request:
-    branches:
-      - main
-      - fix/broker-upgrade
 
 env:
   FORCE_COLOR: 1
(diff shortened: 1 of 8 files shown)


.github/workflows/upgrade-test.yml
```diff
@@ -182,4 +192,4 @@ jobs:
         with:
           args: |
             ❗️❗️❗️❗️ Sorry **${{ github.actor }}**, The Upgrade Test has not passed ❗️❗️❗️❗️
-            👾 Link to job: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+            👾 Link to job: <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}>
```
kylezs, line 195: won't show the link embed which takes up tons of screen real estate on discord

PR #4414, opened 2024-01-12
Solana

ramizhasan111:
# Pull Request

closes PRO-1353

## Checklist

## Summary

Current Status:

- Code for constructing txs, creating signing payload, signing (for tests), and serialization is working (some cleanup is pending).
- test cases include successfully tested txs. This includes both native txs (system program txs) and calls to our vault program.

Important things to note:

- The building blocks (types and methods) have been taken directly from the solana-sdk (https://docs.rs/solana-sdk/latest/solana_sdk/). Direct use of the solana sdk is not possible since it does not support the no-std environment.

- An important modification is the upgrade of ```bincode``` serializer that is used for serialization of solana txs. The version 1.3.3 used in solana sdk has dependency on std. Fortunately, the new version has an optional dependency on std and therefore can be used. The newer version of ```bincode``` works a little differently however. The code is updated such that it works exactly the same as the old version that the svm supports. (see comment in code)

- ```bincode``` uses serde to do custom serialization with a custom serialization implemented for short vecs that is included here.

- Program instructions can either use ```bincode``` or ```borsh``` to serialize. Most programs use borsh for se/deserialize instructions. The methods for constructing instructions with both ```bincode``` and ```borsh``` serialization are implemented, tx tested.


state-chain/chains/src/sol/program_instructions.rs
```diff
@@ -0,0 +1,241 @@
...
+#[derive(BorshDeserialize, BorshSerialize, Debug, Clone, PartialEq, Eq)]
+pub enum VaultProgram {
+	FetchSol { seed: Vec<u8>, bump: u8 },
+}
+
+impl VaultProgram {
+	pub fn get_instruction(self, accounts: Vec<AccountMeta>) -> Instruction {
+		let mut instruction =
+			Instruction::new_with_borsh(Pubkey::from_str(VAULT_PROGRAM).unwrap(), &self, accounts);
+		instruction.data.remove(0);
```
dandanlen, line 227: What is this remove(0) for?

state-chain/chains/src/sol.rs
```diff
@@ -0,0 +1,889 @@
...
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+struct CompiledKeyMeta {
+	is_signer: bool,
+	is_writable: bool,
+	is_invoked: bool,
+}
+
+fn position(keys: &[Pubkey], key: &Pubkey) -> u8 {
+	keys.iter().position(|k| k == key).unwrap() as u8
```
dandanlen, line 557: An unwrap 😱 
We should screen the code for these and make sure they are safe.

dandanlen (commented):
Nice work Ramiz. 

I think it would be good to move all of the Solana 'internals' (basically of the copy-pasted stuff) into its own crate or module, separate from our own code, just to make it even more obvious what code should *not* be touched by us.

state-chain/chains/src/sol/program_instructions.rs
ramizhasan111: This is basically because of custom instruction serialization that anchor (framework which we are using to write solana program) uses. Basically, raw solana program have indices of calls as function selectors but anchor uses the hash of the function name and so we remove the index byte and add the hash of the call as function selector.

ramizhasan111:
I have moved all the sol tx building blocks into its own module so that in future PR when we write code in ```sol.rs```, it doesnt mix with the building blocks code here which almost all is taken from solana sdk except some function call changes related to bincode mentioned above and in code comments. From my perspective it should be ready to merge. @dandanlen can you take a look and approve?

state-chain/chains/src/sol/sol_tx_building_blocks.rs
```diff
@@ -1084,7 +1088,7 @@ mod tests {
 			let transaction_bytes = serialized_tx.len();
 			println!(
 				"Full rotation - number of nonce accounts: {} - Bytes: {}",
-				i,
+				i + 1,
```
dandanlen, line 1091: In these tests, what are the println! statements for?
Should we be using assertions instead? If the prinlns are not being used I would rather we remove them, otherwise it just creates noise.

dandanlen (approved):
LGMT, just some test related nits.

state-chain/chains/src/sol/sol_tx_building_blocks.rs
albert-llimos: In general those prints are for development purposes and I haven't removed them because I'll most likely need them later on when testing and/or playing with some transaction parameters. But it makes sense to remove them and I'll just add them locally when needed.

PR #4420, opened 2024-01-16
Improved the retry queue storage

syan095:
# Pull Request

Closes: PRO-1089

## Checklist

## Summary

* Combined the normal retry queue with the delayed retry queue. 
* Retry queue now stores only the broadcast ID to minimize redundant storage 
* Fixed a bug where if no nominee can be selected initially for a broadcast, the broadcast data is not stored
* Added migration for Retry queue -> Delayed retry queue
* Renamed the Broadcast Attempt and other names to be more consistent (remove the use of "attempt" where appropriate)
* AwaitingBroadcast is now the only storage for broadcast data. All other storage uses broadcast ID or Set<BroadcastId>. This resolves the duplication problem as well.

Workflow summary:
On threshold-signing complete, Broadcast Data is generated and stored into AwaitingBroadcast with `None` set as the nominee.
On initial broadcast, a nominee is selected and stored in AwaitingBroadcast. If no nominee is selected, the broadcast is queued to be retried.
On failed broadcast, the broadcast Id is added to the retry queue.
On success, all broadcast data is removed from AwaitingBroadcast. Further success/fail/timeout are ignored.


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -102,21 +102,33 @@ pub mod pallet {
...
+		}
+	}
+
+	/// Broadcast data with a nominated broadcaster
 	#[derive(Clone, RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo)]
 	#[scale_info(skip_type_params(T, I))]
-	pub struct TransactionSigningAttempt<T: Config<I>, I: 'static> {
-		pub broadcast_attempt: BroadcastAttempt<T, I>,
-		pub nominee: T::ValidatorId,
+	pub struct BroadcastWithNominee<T: Config<I>, I: 'static> {
```
dandanlen, line 129: Does this really need to be a separate struct? Since nominee is optional anyway, can we not just flatten it into BroadcastData?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -371,8 +375,29 @@ pub mod pallet {
...
+
+				// Retry broadcast (allowed by broadcast barrier)
+				let next_block = block_number.saturating_add(One::one());
+				let id_limit = BroadcastBarriers::<T, I>::get()
+					.first()
+					.copied()
+					.unwrap_or(BroadcastId::max_value());
+				let mut remaining_retries = delayed_retries
+					.into_iter()
+					.filter(|broadcast_id| {
```
dandanlen, line 387: We can use `retain` here to avoid allocating a new set.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -501,20 +481,26 @@ pub mod pallet {
 					(broadcast_id, initiated_at),
 				);
 
-				let broadcast_attempt = BroadcastAttempt::<T, I> {
+				// Store the new broadcast data so it is ready to be broadcasted.
```
dandanlen, line 484: Nit: this comment feels unnecessary.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -501,20 +481,26 @@ pub mod pallet {
...
+				let broadcast_data = BroadcastData::<T, I> {
 					broadcast_id,
 					transaction_payload: T::TransactionBuilder::build_transaction(&signed_api_call),
 					threshold_signature_payload,
 					transaction_out_id,
 				};
+				// This write is required in-case if no nominee can be selected in the first try.
```
dandanlen, line 491: See my other comment - maybe we don't need this write if we flatten the nominee into the `BroadcastData` struct.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -835,33 +819,32 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-				DelayedBroadcastRetryQueue::<T, I>::append(retry, broadcast_attempt);
-				retry
-			},
-			None => {
-				BroadcastRetryQueue::<T, I>::append(broadcast_attempt);
-				current_block.saturating_add(One::one())
-			},
-		};
+		let attempt_count = Self::attempt_count(broadcast_id);
+		let _failed = FailedBroadcasters::<T, I>::iter_keys()
```
dandanlen, line 838: Looks like this was left here by accident?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -835,33 +819,32 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			.map(|key| {
+				(key, FailedBroadcasters::<T, I>::get(key).clone().into_iter().collect::<Vec<_>>())
+			})
+			.collect::<Vec<_>>();
+		// If no delay, retry in the next block.
+		let delay = T::RetryPolicy::next_attempt_delay(attempt_count).unwrap_or(One::one());
+		let retry_block = current_block.saturating_add(delay);
```
dandanlen, line 845: nit: delay can be inlined (actually so can attempt_count and current_block).

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -879,46 +862,57 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-			broadcast_id,
-			|failed_broadcasters: &mut BTreeSet<_>| {
-				if failed_broadcasters.insert(failed_broadcaster.clone()) {
-					Ok(failed_broadcasters.len())
+		if let Some(failed_broadcaster) = maybe_reporter {
+			if let Ok(attempt_count) =
+				FailedBroadcasters::<T, I>::try_mutate(broadcast_id, |failed_broadcasters| {
+					if broadcast_id == 11 {
+						let _a = failed_broadcaster.clone();
+					}
```
dandanlen, line 879: Some debugging code left in?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -879,46 +862,57 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		let broadcast = AwaitingBroadcast::<T, I>::get(broadcast_id)
 			.ok_or(Error::<T, I>::InvalidBroadcastId)?;
 
-		let failed_broadcaster = maybe_reporter.unwrap_or(signing_attempt.nominee);
+		if maybe_reporter.is_none() {
+			maybe_reporter = broadcast.nominee;
+		}
 
-		// NOTE: We don't use `append` here because we want to make sure not to insert duplicates
-		// and `append` doesn't guarantee that.
```
dandanlen, line 890: I think this comment is useful to avoid someone trying to optimise this in the future, by using StorageAppend::append.

dandanlen (commented):
I think we should also fix the fact that when a broadcast times out, we report the latest nominee, not the nominee whose broadcast timed out.

Example:

A is nominated at block 0, we set a timeout at block 10.
A reports 'unable to sign' at block 1, we schedule a retry at block 2.
B is nominated at block 2, timeout set at block 12.
B doesn't respond.
At block 10 we report B (early) and nominate C.
At block 12, B's timeout is triggered, we report C.


I think the solution is to add the nominee to the Timeouts, ie. Timeouts should be a `BTreeSet<(BroadcastId, Nominee)>`.

state-chain/pallets/cf-broadcast/src/lib.rs
syan095: Now that we use "set", it is absolutely fine to use append, so this comment is now redandunt.

syan095: the storage append takes in an individual item, whereas BTreeSet append takes a BTreeSet, so both can be used when appropriate.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -102,21 +102,33 @@ pub mod pallet {
...
-	pub struct BroadcastAttempt<T: Config<I>, I: 'static> {
+	pub struct BroadcastData<T: Config<I>, I: 'static> {
 		pub broadcast_id: BroadcastId,
 		pub transaction_payload: TransactionFor<T, I>,
 		pub threshold_signature_payload: PayloadFor<T, I>,
 		pub transaction_out_id: TransactionOutIdFor<T, I>,
 	}
 
-	/// The first step in the process - a transaction signing attempt.
+	impl<T: Config<I>, I: 'static> Clone for BroadcastData<T, I> {
```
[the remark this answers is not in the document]
syan095: Thanks that is great to know!

state-chain/pallets/cf-broadcast/src/lib.rs
syan095: Yeah Unifying these 2 structs makes everything simpler!

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: The issue with StorageAppend is that it causes `decode_len` to become inconsistent.

For example if you do this: 

```rust
// Use StorageAppend::append
FailedBroadcasters::<T>::append(Default::default());
FailedBroadcasters::<T>::append(Default::default());
let count = FailedBroadcasters::<T>::decode_len();
```

`count` will be 2.

But with this:

```rust
// Use BTreeSet::insert
FailedBroadcasters::<T>::mutate(|vals| vals.insert(Default::default()));
FailedBroadcasters::<T>::mutate(|vals| vals.insert(Default::default()));
let count = FailedBroadcasters::<T>::decode_len();
```

`count` will be 1.


Maybe the comment wasn't very clear, but the intention was that if someone notices that we use the second method (which is much less efficient), they might be tempted to change it to the first, but this would introduce a potential bug.

dandanlen (approved):
I made one more minor change, other than that LGTM.

PR #4431, opened 2024-01-19
chore: rename transfer to transfer_allow_death

dandanlen:
# Pull Request

I was looking through recent change to polkadot and noticed that balances::transfer was deprecated and will soon be removed. As it happens, we (correctly) already use transfer_allow_death, it just wasn't made explicit in the code. This PR changes that.


kylezs:
Is this waiting on anything?

PR #4432, opened 2024-01-19
fix: filter incompatible blocks

AlastairHolmes:
Note has been split into atomic/meaningful commits.


engine/src/state_chain_observer/client/mod.rs
```diff
@@ -141,7 +153,8 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static>
 		signing_key_file: &std::path::Path,
 		required_role: AccountRole,
 		wait_for_required_role: bool,
-		required_version_and_wait: Option<(SemVer, bool)>,
+		wait_for_version: bool,
```
AlastairHolmes, line 156: waitforequiredversion

kylezs (approved):

PR #4437, opened 2024-01-22
PRO-1056: Filter out zero amount transfers when building transaction to broadcast

RGafiyatullin:
# Pull Request

Closes: PRO-1056

## Checklist

## Summary

Subj.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -822,7 +822,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 											.unwrap_or(false)
 									},
 								),
-								FetchOrTransfer::Transfer { amount, .. } => !amount.is_zero(),
+								FetchOrTransfer::Transfer { .. } => true,
 							}
 					})
 					.collect()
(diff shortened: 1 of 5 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -822,7 +822,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 											.unwrap_or(false)
 									},
 								),
-								FetchOrTransfer::Transfer { .. } => true,
+								FetchOrTransfer::Transfer { amount, .. } => !amount.is_zero(),
```
dandanlen, line 825: Lower down in this method, we subtract some fees from this amount. This is where the check needs to happen (line 866).

dandanlen (commented):
I think the Linear issue is referring to the case where the amount in the transfer is less than the egress fee, rather than the case where the swap output is zero.

state-chain/pallets/cf-ingress-egress/src/lib.rs
RGafiyatullin: Oh...

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1179,6 +1183,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				(available_amount, Zero::zero())
 			});
 
+		// XXX: it is correct not to put the fee if the input amount is insufficient
+		// XXX: but the remaining amount in this case should be zero — so that no free egress would
+		// be performed
```
syan095, line 1188: I think it is correct to take the fee even if remaining is zero. The network needs to take the required fee (or as much as possible) in all cases to protect the network against DOS attacks.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -145,6 +145,75 @@ fn blacklisted_asset_will_not_egress_via_ccm() {
...
+		IngressEgress::on_finalize(1);
+
+		assert_has_event::<Test>(RuntimeEvent::IngressEgress(
+			crate::Event::BatchBroadcastRequested {
+				broadcast_id: 1,
+				egress_ids: vec![(ForeignChain::Ethereum, 1), (ForeignChain::Ethereum, 3)],
+			},
+		))
+		// XXX: Check the fee is collected correctly.
+		// XXX: Think how the transaction amount is reported (it wasn't collected as a fee, neither it was egressed).
```
syan095, line 213: Yeah should emit an event when the egress is cancelled.

dandanlen:
FYI @RGafiyatullin 

There are some other in-flight related tickets that are being worked on by @msgmaxim and might conflict with this.

Basically the three child issues of this one:
https://linear.app/chainflip/issue/PRO-1103/expose-ingressegress-fees-from-the-protocol

dandanlen:
I think this was covered by the below PRs, will close.

#4442 
#4458

PR #4440, opened 2024-01-22
fix: use custom subxt config + nextAccountIndex

kylezs:
# Pull Request

Closes: PRO-1126
Closes: PRO-1130

## Checklist

## Summary

Uses a custom config for our subxt client used with the SC. This allows us to successfully submit extrinsics to our chain with subxt - in particular the `cfe_version` extrinsic, which is the only one we use subxt for.

v0.32.1 uses the finalised nonce where it previously used the unfinalised nonce. Here we use the system_accountNextIndex call to also account for any mempool txs, ensuring we get the correct nonce.

### DIFF
--- a/engine/src/state_chain_observer/client/mod.rs
+++ b/engine/src/state_chain_observer/client/mod.rs
@@ -711,8 +711,10 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 			use crate::state_chain_observer::client::subxt_state_chain_config::StateChainConfig;
 			use subxt::tx::Signer;
 
+			let rpc_client = RpcClient::new(SubxtInterface(base_rpc_client.clone()));
+
 			let subxt_client = subxt::client::OnlineClient::<StateChainConfig>::from_rpc_client(
-				RpcClient::new(SubxtInterface(base_rpc_client.clone())),
+				rpc_client.clone(),
 			)
 			.await?;
 			let subxt_signer = {
@@ -762,9 +764,16 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 				// error (see https://linear.app/chainflip/issue/PRO-1064/new-cfe-version-gets-stuck-on-startup),
 				// so we use a timeout to ensure we can recover:
 				tokio::time::timeout(CFE_VERSION_SUBMIT_TIMEOUT, async {
+					let current_nonce = rpc_client
+						.request::<u64>(
+							"system_accountNextIndex",
+							subxt::rpc_params![&subxt_signer.account_id()],
+						)
+						.await?;
+
 					subxt_client
 						.tx()
-						.sign_and_submit_then_watch_default(
+						.create_signed_with_nonce(
 							&subxt::dynamic::tx(
 								"Validator",
 								"cfe_version",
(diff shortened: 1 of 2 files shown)


engine/src/state_chain_observer/client/subxt_state_chain_config.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub enum StateChainConfig {}
+
+impl Config for StateChainConfig {
+	type Hash = <state_chain_runtime::Runtime as frame_system::Config>::Hash;
+	type AccountId = subxt::utils::AccountId32;
+	type Address = subxt::utils::MultiAddress<Self::AccountId, ()>;
+	type Signature = subxt::utils::MultiSignature;
+	type Hasher = subxt::ext::sp_runtime::traits::BlakeTwo256;
+	type Header = subxt::ext::sp_runtime::generic::Header<u32, Self::Hasher>;
```
dandanlen, line 11: If we're creating our own StateChainConfig we might as well explicitly use the values from our Runtime? Ie. `type AccountId = <state_chain_runtime::Runtime as frame_system::Config>::AccountId;` etc. Does this not work?

kylezs: Unfortunately not

AlastairHolmes, line 11:
```suggestion
	// We cannot use our own Runtime's types for every associated type here, see comments below.
	type Hash = <state_chain_runtime::Runtime as frame_system::Config>::Hash;
	type AccountId = subxt::utils::AccountId32; // Requires EncodeAsType trait (which our AccountId doesn't)
	type Address = subxt::utils::MultiAddress<Self::AccountId, ()>; // Must be convertible from Self::AccountId
	type Signature = state_chain_runtime::Signature
	type Hasher = subxt::ext::sp_runtime::traits::BlakeTwo256; // Requires subxt's custom Hash trait
	type Header = subxt::ext::sp_runtime::generic::Header<u32, Self::Hasher>; // Requires subxt's custom Header trait
```

engine/src/state_chain_observer/client/mod.rs
```diff
@@ -779,7 +787,10 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 								)],
 							),
 							&subxt_signer,
-						)
+							current_nonce,
+							Default::default(),
```
AlastairHolmes, line 791: We probably should set a mortality here

AlastairHolmes: The SCC uses 128 blocks

engine/src/state_chain_observer/client/subxt_state_chain_config.rs
kylezs: Yeah, I didn't think it was much better having them inconsistent. But not a strong opinion, will make the change.

engine/src/state_chain_observer/client/mod.rs
kylezs: Will create an issue

kylezs: https://linear.app/chainflip/issue/PRO-1130/add-mortality-to-cfe-version-extrinsic-submission-on-cfe-startup

AlastairHolmes (approved):

PR #4441, opened 2024-01-22
WIP: run tests in release-mode

RGafiyatullin:


.github/workflows/_11_coverage.yml
```diff
@@ -28,9 +28,9 @@ jobs:
       - name: Setup sccache
         run: ./ci/scripts/install_sccache.sh $SCCACHE_VERSION
 
-      - name: cargo cf-test-ci --locked
+      - name: cargo cf-test-ci --locked --release
         run: |
-          cargo cf-test-ci --locked
```
dandanlen, line 33: We could add the `--release` arg to the `cf-` custom commands in the `config.toml` file.

I guess it makes sense to add this for cf-test-ci and cf-clippy-ci?

dandanlen (commented):
Thanks @RGafiyatullin , think this is a good 'quick win' for CI efficiency: it looks like it shaves ~15 minutes off in total!

I compared with [this CI run](https://github.com/chainflip-io/chainflip-backend/actions/runs/7613690804?pr=4438) which is the last PR that was merged before you branch this PR from main. 

As mentioned below I think it makes sense to add this to the config.toml instead of the CI files, and to use it for clippy too.

state-chain/chains/src/dot.rs
```diff
@@ -996,7 +996,7 @@ impl BenchmarkValueExtended for PolkadotChannelId {
 	}
 }
 
-#[cfg(debug_assertions)]
```
AlastairHolmes, line 999: I think just in the const's case, it should be `cfg(test)`.

AlastairHolmes: This then solves the problem, and tests can be run in release or debug.

state-chain/chains/src/mocks.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(debug_assertions)]
+#![cfg(feature = "std")]
```
AlastairHolmes, line 1: Seems that the mocks are inconsistently enabled. Sometimes they use `feature = "std"` and somethings `cfg(debug_assertions)`, probably worth making it consistent at the very least @dandanlen.

.github/workflows/_11_coverage.yml
RGafiyatullin: Now we can have a choice to run `cf-test-ci` or `cf-test-ci --release`, so I find it better.

If you'd insist on keeping the CI-files intact, I'd propose to also have a command `cf-test-ci-debug`. Which is a bit messy, in my opinion, but not tragically so.

state-chain/chains/src/dot.rs
RGafiyatullin: True

.github/workflows/_11_coverage.yml
RGafiyatullin: Okay. The `-ci` commands are for the CI.
CI runs in release-mode.

The `cf-test-ci` will have the `--release` baked into it.

kylezs:
This looks good to me, will let @AlastairHolmes approve if nothing else

AlastairHolmes (approved):

PR #4442, opened 2024-01-23
Feat: expose ingress and egress fees on events

msgmaxim:
# Pull Request

Closes: PRO-1119, PRO-1119

## Checklist

## Summary

- `withhold_transaction_fee` now also returns the deducted fee.
- Added `ingress_fee` to the `DepositReceived` event.
- Added `action` field to the `DepositReceived` event. This allows the frontend to explicitly associate an action with a deposit.
- Added `egress_fee` to the `EgressScheduled` event, which we either get from `withhold_transaction_fee` (for regular swaps) or set to `gas_budget` (for CCM swaps).
- As discussed, I moved `DepositBalances` to `schedule_egress`, which is now called on CCM swaps too.

Technically this needs a migration, since the meaning of "amount" in ScheduledEgressFetchOrTransfer has now changed to mean "after fee deducted", but because these items are cleared after a block is finalised (unless we disable egress), it shouldn't be necessary in practice. If we want, I could write a migration that would deduct egress fees from pending transfers in case they haven't been cleared for some reason.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -184,6 +184,16 @@ pub mod pallet {
 		},
 	}
 
+	/// Contains identifying information about the particular actions that have occurred for a
+	/// particular deposit.
+	#[derive(Clone, RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+	pub enum DepositAction<AccountId> {
+		// TODO: use a SwapId alias
+		Swap { swap_id: u64 },
+		LiquidityProvision { lp_account: AccountId },
+		CcmTransfer { principal_swap_id: Option<u64>, gas_swap_id: Option<u64> },
+	}
+
 	#[derive(
 		CloneNoBound,
 		DefaultNoBound,
@@ -401,6 +411,7 @@ pub mod pallet {
 			amount: TargetChainAmount<T, I>,
 			deposit_details: <T::TargetChain as Chain>::DepositDetails,
 			ingress_fee: TargetChainAmount<T, I>,
+			action: DepositAction<T::AccountId>,
 		},
 		AssetEgressStatusChanged {
 			asset: TargetChainAsset<T, I>,
@@ -1016,54 +1027,63 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				reason: DepositIgnoredReason::NotEnoughToPayFees,
 			});
 		} else {
-			match deposit_channel_details.action {
-				ChannelAction::LiquidityProvision { lp_account, .. } => T::LpBalance::add_deposit(
-					&lp_account,
-					asset.into(),
-					net_deposit_amount.into(),
-					ingress_fee.into(),
-				)?,
+			let deposit_action = match deposit_channel_details.action {
+				ChannelAction::LiquidityProvision { lp_account, .. } => {
+					T::LpBalance::add_deposit(
+						&lp_account,
+						asset.into(),
+						net_deposit_amount.into(),
+						ingress_fee.into(),
+					)?;
+
+					DepositAction::LiquidityProvision { lp_account }
+				},
 				ChannelAction::Swap {
 					destination_address,
 					destination_asset,
 					broker_id,
 					broker_commission_bps,
 					..
-				} => T::SwapDepositHandler::schedule_swap_from_channel(
-					deposit_address.clone().into(),
-					block_height.into(),
-					asset.into(),
-					destination_asset,
-					net_deposit_amount.into(),
-					destination_address,
-					broker_id,
-					broker_commission_bps,
-					ingress_fee.into(),
-					channel_id,
-				),
+				} => DepositAction::Swap {
+					swap_id: T::SwapDepositHandler::schedule_swap_from_channel(
+						deposit_address.clone().into(),
+						block_height.into(),
+						asset.into(),
+						destination_asset,
+						net_deposit_amount.into(),
+						destination_address,
+						broker_id,
+						broker_commission_bps,
+						ingress_fee.into(),
+						channel_id,
+					),
+				},
 				ChannelAction::CcmTransfer {
 					destination_asset,
 					destination_address,
 					channel_metadata,
 					..
-				} => T::CcmHandler::on_ccm_deposit(
-					asset.into(),
-					net_deposit_amount.into(),
-					destination_asset,
-					destination_address,
-					CcmDepositMetadata {
-						source_chain: asset.into(),
-						source_address: None,
-						channel_metadata,
-					},
-					SwapOrigin::DepositChannel {
-						deposit_address: T::AddressConverter::to_encoded_address(
-							deposit_address.clone().into(),
-						),
-						channel_id,
-						deposit_block_height: block_height.into(),
-					},
-				),
+				} => {
+					let (principal_swap_id, gas_swap_id) = T::CcmHandler::on_ccm_deposit(
+						asset.into(),
+						net_deposit_amount.into(),
+						destination_asset,
+						destination_address,
+						CcmDepositMetadata {
+							source_chain: asset.into(),
+							source_address: None,
+							channel_metadata,
+						},
+						SwapOrigin::DepositChannel {
+							deposit_address: T::AddressConverter::to_encoded_address(
+								deposit_address.clone().into(),
+							),
+							channel_id,
+							deposit_block_height: block_height.into(),
+						},
+					);
+					DepositAction::CcmTransfer { principal_swap_id, gas_swap_id }
+				},
 			};
 
 			Self::deposit_event(Event::DepositReceived {
@@ -1072,6 +1092,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				amount: deposit_amount,
 				deposit_details,
 				ingress_fee,
+				action: deposit_action,
 			});
 		}
 
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -196,6 +196,7 @@ pub mod pallet {
 		Swap { swap_id: SwapId },
 		LiquidityProvision { lp_account: AccountId },
 		CcmTransfer { principal_swap_id: Option<SwapId>, gas_swap_id: Option<SwapId> },
+		NoAction,
 	}
 
 	#[derive(
@@ -1068,7 +1069,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 					channel_metadata,
 					..
 				} => {
-					let (principal_swap_id, gas_swap_id) = T::CcmHandler::on_ccm_deposit(
+					if let Ok((principal_swap_id, gas_swap_id)) = T::CcmHandler::on_ccm_deposit(
 						asset.into(),
 						amount_after_fees.into(),
 						destination_asset,
@@ -1085,8 +1086,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 							channel_id,
 							deposit_block_height: block_height.into(),
 						},
-					);
-					DepositAction::CcmTransfer { principal_swap_id, gas_swap_id }
+					) {
+						DepositAction::CcmTransfer { principal_swap_id, gas_swap_id }
+					} else {
+						DepositAction::NoAction
+					}
 				},
 			};
 
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -165,6 +165,11 @@ pub mod pallet {
 		Egress,
 	}
 
+	pub struct AmountFeesWithheld<T: Config<I>, I: 'static> {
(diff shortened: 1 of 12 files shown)


RGafiyatullin:
Added `chainflip-io/web` to the reviewers, as the event format was changed.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1200,36 +1196,49 @@ impl<T: Config<I>, I: 'static> EgressApi<T::TargetChain> for Pallet<T, I> {
...
+						amount: amount_after_fees,
+						egress_id,
+					}
+				});
+				egress_fee
+			},
+		};
+
+		DepositBalances::<T, I>::mutate(asset, |tracker| {
+			tracker.register_transfer(amount);
```
RGafiyatullin, line 1233: Both `tracker.register_transfer` and `tracker.mark_as_fetched` used to be invoked in `fn do_egress_scheduled_fetch_transfer`. 

One of them is moved into `fn schedule_egress` as the original (gross?) amount is not saved after taking the fee.

Can the `tracker.mark_as_fetched` invocation be moved to the `fn schedule_egress` too?
IMO, it would make easier to reason about storage changes this way.

RGafiyatullin, line 1233: Both `tracker.register_transfer` and `tracker.mark_as_fetched` used to be invoked in `fn 
do_egress_scheduled_fetch_transfer`.

One of them is moved into `fn schedule_egress` as the original (gross?) amount is not saved after taking the fee.

Can the `tracker.mark_as_fetched` invocation be moved to the `fn schedule_egress` too?
IMO, it would make easier to reason about storage changes this way.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1184,7 +1180,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			fees.saturating_accrue(fee_estimate);
 		});
 
-		remaining_amount
```
dandanlen, line 1187: I'm not sure this is what we want. The fee estimate is in terms of the GAS_ASSET, ie. if someone sends USDC, it will be in terms of ETH. 

I think we want to return (remaining_amount, available_amount - remaining_amount).

Or maybe we want both 🤔 

(https://linear.app/chainflip/issue/PRO-1118/expose-fees-in-ingressegress-asset-amount)

msgmaxim: Updated to return `(remaining_amount, available_amount - remaining_amount)` for now.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -406,6 +409,7 @@ pub mod pallet {
 			asset: TargetChainAsset<T, I>,
 			amount: AssetAmount,
```
niklasnatter, line 410: will this still be the amount where fees have already been deducted?

niklasnatter (approved):
this will make our lives a lot easier - thanks!

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: For `mark_as_fetched`, I think the equivalent place would be `process_single_deposit` (where we append the fetch to `ScheduledEgressFetchOrTransfer`), but it is also where `deposits.register_deposit` is called (which adds the amount as "unfetched"). To be honest, I'm not 100% sure what the distinction between fetched/unfetched is (seems like all unfetched funds become "fetched" at the end of the block, so maybe the intention is to add a 1 block delay until we can use them?).

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: I haven't changed the behaviour here. Though from what I can tell (also having tested this on localnet now), the `amount` emitter here *includes* the egress fee.

niklasnatter: yes you are right, sorry. its the amount before the fee is deducted 👍

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1200,36 +1200,49 @@ impl<T: Config<I>, I: 'static> EgressApi<T::TargetChain> for Pallet<T, I> {
...
+		DepositBalances::<T, I>::mutate(asset, |tracker| {
+			tracker.register_transfer(amount);
+		});
 
 		Self::deposit_event(Event::<T, I>::EgressScheduled {
 			id: egress_id,
 			asset,
 			amount: amount.into(),
 			destination_address,
+			egress_fee,
```
kylezs, line 1245: Related to the CCM comment above, about how the ingress fee isn't required - this is also likely to cause confusion. In the case of a CCM, this "fee" means something different

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -587,6 +588,7 @@ pub mod pallet {
 				origin: swap_origin,
 				swap_type: SwapType::Swap(destination_address_internal),
 				broker_commission: None,
+				ingress_fee: None,
```
kylezs, line 591: This isn't true, we took a fee when the deposit was made, we just don't pass it through. We should have a test for this.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -272,6 +272,7 @@ pub mod pallet {
 			origin: SwapOrigin,
 			swap_type: SwapType,
 			broker_commission: Option<AssetAmount>,
+			ingress_fee: Option<AssetAmount>,
```
kylezs, line 275: A comment as to why this is optional would be helpful. It can be None when there's a contract initiated swap.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1802,6 +1817,7 @@ fn swap_with_custom_broker_fee(
 	to: Asset,
 	amount: AssetAmount,
 	broker_fee: BasisPoints,
+	ingress_fee: AssetAmount,
```
kylezs, line 1820: This is 0 in every test. We should add some tests where it's not 0.

kylezs: and as stated before - one of these tests should be a CcmTransfer scheduled from channel to ensure the ingress fee is set there

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1200,36 +1200,49 @@ impl<T: Config<I>, I: 'static> EgressApi<T::TargetChain> for Pallet<T, I> {
...
+					cf_parameters: channel_metadata.cf_parameters,
+					source_chain,
+					source_address,
+					gas_budget,
+				});
+				gas_budget
+			},
+			None => {
+				let (amount_after_fees, egress_fee) =
+					Self::withhold_transaction_fee(IngressOrEgress::Egress, asset, amount);
```
kylezs, line 1223: would perhaps make it easier to see the order is correct, and harder to make a mistake if we returned a struct instead of a tuple from `withhold_transaction_fee` so we get named fields... especially because the types are the same

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: Yeah I'm not sure what Dan intended by fetched/unfetched in DepositBalances here. But given it's not used anywhere else except tracking I don't think it matters a whole lot *when* things occur, so long as the amounts are correct. I think leaving the fetched where the tx is kicked off (where it is now).

kylezs: I actually think there's a bug here after looking at this a bit deeper now too. In `process_single_deposit` we add the `net_deposit_amount` to the DepositBalances. We pass through the `deposit_amount` to the fetch transaction. When we process that fetch in `do_egress_scheduled_fetch_transfer` we do:

```rust
FetchOrTransfer::<T::TargetChain>::Fetch {
					asset,
					deposit_address,
					deposit_fetch_id,
					amount,
				} => {
					fetch_params.push(FetchAssetParams {
						deposit_fetch_id: deposit_fetch_id.expect("Checked in extract_if"),
						asset,
					});
					addresses.push(deposit_address.clone());
					DepositBalances::<T, I>::mutate(asset, |tracker| {
						tracker.mark_as_fetched(amount);
					});
```

We've now fetched more than we had in unfetched.

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: On further reflection - it's not obvious what to do here - it's definitely inconsistent with the standard swap flow. But it's actually possible no event is emitted here at all... which is even more inconsistent?

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -117,6 +117,12 @@ pub mod pallet {
 			chain: ForeignChain,
 			address: ForeignChainAddress,
 		},
+		LiquidityDepositCredited {
```
niklasnatter, line 120: could we expose the information necessary to identify the channel of the deposit here?

msgmaxim:
> We've now fetched more than we had in unfetched.

I did notice that something was off with tracking. Would the correct behaviour be to add the full amount in `process_single_deposit`? @kylezs

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: I changed it so that the unfetched adds the deposit_amount rather than the amount_after_fees

state-chain/pallets/cf-lp/src/lib.rs
kylezs: chatted in dms, not necessary after adding the account id of the depositer to DepositReceived

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -396,6 +398,7 @@ pub mod pallet {
 			asset: TargetChainAsset<T, I>,
 			amount: TargetChainAmount<T, I>,
 			deposit_details: <T::TargetChain as Chain>::DepositDetails,
+			ingress_fee: TargetChainAmount<T, I>,
```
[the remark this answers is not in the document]
kylezs: I added a comment here to clarify that

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1005,10 +1025,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			deposit_channel_details.deposit_channel,
 		);
 		DepositBalances::<T, I>::mutate(asset, |deposits| {
-			deposits.register_deposit(net_deposit_amount)
+			deposits.register_deposit(deposit_amount)
```
kylezs, line 1028: note here this is now using the full amount. The full amount is fetched, and therefore the full amount will be used when we register as fetched.

niklasnatter (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -163,6 +163,11 @@ pub mod pallet {
 		Egress,
 	}
 
+	pub struct AmountFeesWithheld<T: Config<I>, I: 'static> {
```
msgmaxim, line 167: `AmountFeesWithheld` -> `AmountAndFeesWithheld`? Otherwise it is a bit misleading on the first read.

kylezs: I did this in the next PR targetting this: https://github.com/chainflip-io/chainflip-backend/pull/4458

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: I'm not sure this is correct. 

We *do* need to fetch the full (gross) deposit amount, but not all of that is actually deposited in the vault, since some of it is used for fees. 

I can see how the use of this isn't super clear, but the idea of the deposit balances is to track how much is available for swap egresses / lp withdrawals.

If we 'take' some of the deposit and put it in the withheld fees, we can't also put it in the deposits.

```
transfer_amount = deposit_amount + fees
^ 					^				^- Fees withheld at the witnessing step
|					|- amount actually credited to the vault and used in the swap/lp deposit 
|_ The total amount transferred by the user
```

I'll take a closer look at and may change this back.

kylezs: Would it make more sense to tie these two concepts a bit more closely together in that case?

dandanlen: Maybe. We could also improve the naming/documentation and distinguish more clearly between what the user sends vs what we actually count as the deposit.

But IMO this is out of scope for this PR, especially since deposit tracking is not used anywhere. I added it as a placeholder to iterate upon, because I wanted us to have *something*, but it needs improvement.

dandanlen (approved):
I made some minor changes and rebased against main. LGTM

PR #4443, opened 2024-01-23
Remove unused TX Version from transaction struct

martin-chainflip:
Block 5839f20446d7b9446e82c00117ee3699fa84154e970d57f09add60deef2eaa18 on BTC testnet contains a transaction with a version number 4294967295 reported by the BTC RPC. This violates the spec that the version should be a signed 32bit integer. This breaks our parser. Since we don't use the version anywhere we just remove it for now.

# Pull Request

Closes: PRO-1131

## Checklist

## Summary


acdibble (approved):
want my ingress/egress tracker back

engine/src/btc/rpc.rs
```diff
@@ -497,6 +496,16 @@ mod tests {
...
+		let jd = &mut serde_json::Deserializer::from_str(block_data);
+		let _result: Result<VerboseBlock, _> = serde_path_to_error::deserialize(jd);
```
kylezs, line 506: We'd want to unwrap this right? Else the test will always pass

engine/src/btc/rpc.rs
```diff
@@ -261,7 +261,6 @@ pub enum Difficulty {
 pub struct VerboseTransaction {
 	pub txid: Txid,
 	pub hash: Txid,
```
AlastairHolmes, line 263: I feel there should also be a comment here, stating why the Version is excluded, as we will forget.

kylezs (approved):

engine/src/btc/rpc.rs
RGafiyatullin: Keeping the field, but changing its type to something like this could be more self-explanatory:

```rust
#[derive(Clone, PartialEq, Debug, Deserialize, Serialize)]
#[serde(untagged)]
pub enum VerboseTransactionVersion {
    Valid(Version),
    Invalid(serde_json::Value),
}
```

```rust
#[test]
fn test_verbose_transaction_version() {
    for (input, expected_output) in [
        (json!(1), VerboseTransactionVersion::Valid(Version::from_consensus(1))),
        (json!(2), VerboseTransactionVersion::Valid(Version::from_consensus(2))),
        (json!(u32::MAX), VerboseTransactionVersion::Invalid(json!(u32::MAX))),
    ] {
        assert_eq!(serde_json::from_value::<VerboseTransactionVersion>(input).unwrap(), expected_output);
    }
}```

martin-chainflip: essentialy, any u32 value is allowed. Seems that the correct way to parse this would be to parse a u32 and then reinterpret it as i32

PR #4458, opened 2024-01-26
feat: expose all ignored egresses in ingress-egress pallet

kylezs:
Extends #4442 (that should be merged first)

# Pull Request

Closes: PRO-1148
Closes: PRO-1155

## Checklist

## Summary

Centralising the filtering of egresses so that we can catch all of them, and display the ignored egress consistently. 

Also: 
- EgressApi returns a Result containing id, egress amount and fee
- Swap and Lp Handlers emit the events
- Corresponding Ingress/Egress events removed and merged with swap/lp analogs
- Events show egress amount not swap output amount


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -372,6 +372,12 @@ pub mod pallet {
 	pub type MinimumDeposit<T: Config<I>, I: 'static = ()> =
 		StorageMap<_, Twox64Concat, TargetChainAsset<T, I>, TargetChainAmount<T, I>, ValueQuery>;
 
+	/// Defines the minimum amount for a single egress i.e. *not* of a batch, but the outputs of
+	/// each individual egress within that batch.
+	#[pallet::storage]
+	pub type MinimumEgress<T: Config<I>, I: 'static = ()> =
```
kylezs, line 378: This would be set to the 600 sats for Bitcon, and 0 for the rest - just getting initial feedback first, then will do this

acdibble: We could also read these values on the FE if they're added to the ingress/egress environment RPC

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -461,6 +467,12 @@ pub mod pallet {
 			deposit_details: <T::TargetChain as Chain>::DepositDetails,
 			reason: DepositIgnoredReason,
 		},
+		EgressIgnored {
+			egress_id: EgressId,
+			asset: TargetChainAsset<T, I>,
+			amount: TargetChainAmount<T, I>,
+			destination_address: TargetChainAccount<T, I>,
```
acdibble, line 474: should we add a `reason` enum? future-proof it?

kylezs: done

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: done

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -886,14 +905,23 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-				} => {
-					egress_ids.push(egress_id);
-					transfer_params.push(TransferAssetParams {
-						asset,
-						amount,
-						to: destination_address,
-					});
-				},
+				} =>
+					if amount > MinimumEgress::<T, I>::get(asset) {
```
RGafiyatullin, line 909: "Minimum Egress" sounds as if it stands for the smallest, but still an allowed amount: I'd put a `>=` here.

kylezs: Needs to be `>` because this will be 0 in some cases. If we make this >= 0 we will try egressing 0 amounts. I clarified in the comment on the storage - also made it consistent with the `BITCOIN_DUST_AMOUNT` that we use in the Bitcoin tx construction - we now use strictly greater everywhere, so it all ties nicely together (without any `-1`s all over the place).

kylezs: On second thought, I guess an alternative is to make the minimum 1 for the other chains 🤔 . That seems like a (very slightly) worse approach though

acdibble (approved):

RGafiyatullin (changes requested):
I'm not aboard with the storage type for `ingress-egress::MinimumEgress`

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -372,6 +386,14 @@ pub mod pallet {
 	pub type MinimumDeposit<T: Config<I>, I: 'static = ()> =
 		StorageMap<_, Twox64Concat, TargetChainAsset<T, I>, TargetChainAmount<T, I>, ValueQuery>;
 
+	/// Defines the minimum amount for a single egress i.e. *not* of a batch, but the outputs of
+	/// each individual egress within that batch. Note: The amount must be strictly greater than
+	/// this minimum.
+	#[pallet::storage]
+	#[pallet::getter(fn minimum_egress)]
+	pub type MinimumEgress<T: Config<I>, I: 'static = ()> =
+		StorageMap<_, Twox64Concat, TargetChainAsset<T, I>, TargetChainAmount<T, I>, ValueQuery>;
```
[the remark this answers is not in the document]
kylezs: Yeah, Ethereum, USDC, FLIP and ETH

state-chain/runtime/src/lib.rs
```diff
@@ -1191,6 +1191,26 @@ impl_runtime_apis! {
 			}
 		}
 
+		fn cf_min_egress_amount(asset: Asset) -> AssetAmount {
+			use pallet_cf_ingress_egress::MinimumEgress;
+			use cf_chains::assets::{eth, dot, btc};
+
+			match ForeignChain::from(asset) {
+				ForeignChain::Ethereum => MinimumEgress::<Runtime, EthereumInstance>::get(
```
RGafiyatullin, line 1199: A demo of redundancy of having [a map instead of a single-value](https://github.com/chainflip-io/chainflip-backend/pull/4458#discussion_r1469626108):

For each `$CHAIN` in all chains:
- get the storage for `$CHAIN`
- in the latter storage fetch the value for the Asset of `$CHAIN`

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -406,26 +402,21 @@ pub mod pallet {
 						});
 						// Handle swap completion logic.
 						match &swap.swap_type {
-							SwapType::Swap(destination_address) =>
-								if !egress_amount.is_zero() {
```
dandanlen, line 404: I guess this is the reason we use strictly greater? Just so we can avoid explicitly checking for zero?

dandanlen (commented):
I think it would be better to handle the zero egress case explicitly (as we do with ingress), rather than implicitly by requiring the minimum to be minimum-1.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1256,17 +1285,32 @@ impl<T: Config<I>, I: 'static> EgressApi<T::TargetChain> for Pallet<T, I> {
...
-					}
-				});
-				fees_withheld
+						asset,
+						amount,
+						destination_address,
+						reason: EgressIgnoredReason::BelowMinimumEgress,
+					});
+
+					return egress_id
```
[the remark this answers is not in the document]
kylezs: Yeah, so product can associate it

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: Correct

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -864,7 +893,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		let mut fetch_params = vec![];
 		let mut transfer_params = vec![];
-		let mut egress_ids = vec![];
+		let mut egress_ids_to_broadcast = vec![];
```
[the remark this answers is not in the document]
kylezs: Yeah this is from an earlier commit, where I was doing the ignoring here. Will revert this.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -400,26 +402,21 @@ pub mod pallet {
...
+									egress_amount,
+									destination_address.clone(),
+									None,
+								);
+
+								Self::deposit_event(Event::<T>::SwapEgressScheduled {
+									swap_id: swap.swap_id,
+									egress_id,
+									asset: swap.to,
+									amount: egress_amount,
```
dandanlen, line 417: I noticed this is the amount including fees - not what is actually received by the user. This should probably be fixed too.

kylezs: Would double check this with product first - there was some discussion about fees included or not included in these amounts

niklasnatter: as far as i see, we dont use the `amount` of the `SwapEgressScheduled` event at the moment, so changing it would be fine on our side

dandanlen: Saying we have scheduled an egress for X amount if the actual amount is (X - fee) seems wrong? I'd rather we expose the correct value. Less confusing.

state-chain/pallets/cf-emissions/src/lib.rs
```diff
@@ -171,18 +182,7 @@ pub mod pallet {
...
-							Asset::Flip,
-							flip_to_burn,
-							ForeignChainAddress::Eth(
-								T::EthEnvironment::state_chain_gateway_address(),
-							),
-							None,
-						);
-						T::Issuance::burn(flip_to_burn.into());
-					}
+					Self::burn_flip_network_fee();
```
dandanlen, line 185: I move this to function to make it easier to test independently of inflation (which is hard to predict).

state-chain/pallets/cf-emissions/src/lib.rs
```diff
@@ -308,6 +308,36 @@ impl<T: Config> Pallet<T> {
 			block_number.saturated_into(),
 		));
 	}
+
+	fn burn_flip_network_fee() {
+		let flip_to_burn = T::FlipToBurn::take_flip_to_burn();
```
kylezs, line 313: This is a bug and it's not caught in the test because the mock always returns the same value of `FLIP_TO_BURN`. If we take here and then error, we're going to remove the storage but not actually burn anything, and on the next attempt we'd be starting from 0, so we'd never burn anything

kylezs: We should ofc fix the test too

dandanlen: 👌  good catch

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -426,16 +446,6 @@ pub mod pallet {
 			asset: TargetChainAsset<T, I>,
 			disabled: bool,
 		},
-		EgressScheduled {
```
niklasnatter, line 429: we do use this event in the SDK at the moment. so we will need a bit of time to adjust our code when this is released to a testnet

kylezs: 👍 I think this will be released a bit slower than last time due to breaking changes like this. Will stay on testnet for a while before moving across.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -285,7 +287,9 @@ pub mod pallet {
 			source_asset: Asset,
 			deposit_amount: AssetAmount,
 			destination_asset: Asset,
+			#[deprecated(note = "Use swap_output instead")]
 			egress_amount: AssetAmount,
```
niklasnatter, line 291: if we rename this, would it make sense to rename the `deposit_amount` too? or is `deposit_amount` the appropriate term for the amount with the ingress fee deducted?

state-chain/node/src/chain_spec.rs
```diff
@@ -671,14 +673,17 @@ fn testnet_genesis(
 		bitcoin_ingress_egress: BitcoinIngressEgressConfig {
 			deposit_channel_lifetime: bitcoin_deposit_channel_lifetime.into(),
 			witness_safety_margin: Some(bitcoin_safety_margin),
+			min_egresses: vec![(btc::Asset::Btc, BITCOIN_DUST_LIMIT)],
 		},
 		ethereum_ingress_egress: EthereumIngressEgressConfig {
 			deposit_channel_lifetime: ethereum_deposit_channel_lifetime.into(),
 			witness_safety_margin: Some(ethereum_safety_margin),
+			min_egresses: Default::default(),
```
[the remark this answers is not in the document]
dandanlen: I changed it to an option - this way the default is 'not set' and we unwrap_or(One::one()).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -838,30 +850,25 @@ pub mod pallet {
...
-			if let Some(swap_id) = ccm_swap.principal_swap_id {
-				Self::deposit_event(Event::<T>::SwapEgressScheduled {
-					swap_id,
-					egress_id,
-					asset: ccm_swap.destination_asset,
-					amount: ccm_output_principal,
-				});
-			}
-			if let Some(swap_id) = ccm_swap.gas_swap_id {
-				Self::deposit_event(Event::<T>::SwapEgressScheduled {
```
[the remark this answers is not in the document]
dandanlen: I removed this because there's no actual egress scheduled for the gas, so it was wrong. Maybe we can extend the ccm event somehow, but that's out of scope here.

state-chain/runtime/src/lib.rs
```diff
@@ -1191,6 +1191,26 @@ impl_runtime_apis! {
...
+					eth::Asset::try_from(asset)
+						.expect("Conversion must succeed: ForeignChain checked in match clause.")
+				).unwrap_or(1),
```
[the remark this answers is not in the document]
kylezs: Could do both ?

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Yeah swap_input probably makes more sense here. 

I think terminology should be something like this:

`deposit_amount` (`ingress_amount` ?): the amount sent by the user
`swap_input`: the amount that is actually swapped (ie. deposit amount minus broker fee and gas fees)
`swap_output`: the result of the swap (net liquidity and network fee)
`egress_amount`: the amount sent to the user (swap output minus egress fee)

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1193,8 +1188,23 @@ fn basic_balance_tracking() {
...
+		// Trigger a non-swap transfer.
+		.then_execute_at_next_block(|_| {
+			assert_ok!(<IngressEgress as EgressApi<Ethereum>>::schedule_egress(
+				eth::Asset::Flip,
+				FLIP_EGRESS_AMOUNT,
+				Default::default(),
+				None
+			));
+		})
+		.check_deposit_balances(&[
```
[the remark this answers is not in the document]
dandanlen: I agree, could be more thorough, but this will require changing the ChainTracking mocks (since this is what 'delivers' the fee estimate). I'd rather leave this for another PR, this one has spun a little out of control already.

dandanlen:
I think I've addresses all of the above... should ok for re-review.

state-chain/pallets/cf-ingress-egress/src/tests.rs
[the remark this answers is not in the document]
dandanlen: (https://linear.app/chainflip/issue/PRO-1173/improve-chain-tracking-fee-estimation-mocks-and-tests)

state-chain/pallets/cf-emissions/src/tests.rs
```diff
@@ -175,7 +175,10 @@ fn burn_flip() {
 		let egresses = MockEgressHandler::<AnyChain>::get_scheduled_egresses();
 		assert!(egresses.len() == 1);
 		assert_eq!(egresses.first().expect("should exist").amount(), FLIP_TO_BURN);
-		assert_has_event::<Test>(crate::Event::NetworkFeeBurned { amount: FLIP_TO_BURN }.into());
+		assert_has_matching_event!(
+			Test,
+			RuntimeEvent::Emissions(crate::Event::NetworkFeeBurned { amount: FLIP_TO_BURN, .. }),
```
kylezs, line 183: worth checking that when called again right after we don't emit anything

state-chain/traits/src/mocks/egress_handler.rs
```diff
@@ -74,7 +74,7 @@ impl<C: Chain> EgressApi<C> for MockEgressHandler<C> {
 		destination_address: <C as Chain>::ChainAccount,
 		maybe_ccm_with_gas_budget: Option<(CcmDepositMetadata, <C as Chain>::ChainAmount)>,
 	) -> Result<ScheduledEgressDetails<C>, DispatchError> {
-		if amount.is_zero() {
+		if amount.is_zero() && maybe_ccm_with_gas_budget.is_none() {
```
kylezs, line 77: this is a good catch. we should add a test for this (could be another PR)

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -559,7 +560,7 @@ pub mod pallet {
 			let earned_fees = EarnedBrokerFees::<T>::take(account_id, asset);
 			ensure!(earned_fees != 0, Error::<T>::NoFundsAvailable);
 
-			let ScheduledEgressDetails { egress_id, egress_amount, .. } =
+			let ScheduledEgressDetails { egress_id, egress_amount, fee_taken } =
```
kylezs, line 563: nit: maybe `fees_withheld` for consistency with `AmountAndFeesWithheld`?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -260,7 +260,7 @@ pub struct IngressEgressEnvironment {
 	pub ingress_fees: HashMap<ForeignChain, HashMap<Asset, NumberOrHex>>,
 	pub egress_fees: HashMap<ForeignChain, HashMap<Asset, NumberOrHex>>,
 	pub witness_safety_margins: HashMap<ForeignChain, Option<u64>>,
-	pub minimum_egress_amounts: HashMap<ForeignChain, HashMap<Asset, NumberOrHex>>,
+	pub egress_dust_limits: HashMap<ForeignChain, HashMap<Asset, NumberOrHex>>,
```
kylezs, line 263: my thinking with minimum_egress_amounts was that it's more general and therefore future proof - while still quite clear what it is. not a strong opinion though

kylezs:
Looks good 👍  LGTM and a release branched!

state-chain/traits/src/mocks/egress_handler.rs
dandanlen: This is just the mock - behaviour in the runtime was already correct.

dandanlen (approved):
Approving with @kylezs consent (he opened the PR so can't approve himself)

PR #4464, opened 2024-01-30
perf: :zap: unlinked old migrations

Janislav:
# Pull Request

Closes: PRO-988

## Checklist

## Summary


state-chain/runtime/src/lib.rs
```diff
@@ -901,27 +901,12 @@ type PalletMigrations = (
...
-	// pallet_cf_threshold_signature::migrations::PalletMigration<Runtime, Instance3>,
 	pallet_cf_broadcast::migrations::PalletMigration<Runtime, Instance1>,
 	pallet_cf_broadcast::migrations::PalletMigration<Runtime, Instance2>,
 	pallet_cf_broadcast::migrations::PalletMigration<Runtime, Instance3>,
-	pallet_cf_swapping::migrations::PalletMigration<Runtime>,
-	pallet_cf_lp::migrations::PalletMigration<Runtime>,
-	pallet_cf_ingress_egress::migrations::PalletMigration<Runtime, Instance1>,
-	pallet_cf_ingress_egress::migrations::PalletMigration<Runtime, Instance2>,
-	pallet_cf_ingress_egress::migrations::PalletMigration<Runtime, Instance3>,
-	pallet_cf_pools::migrations::PalletMigration<Runtime>,
```
dandanlen, line 924: I think we should keep these here - we can unlink the individual upgrades but the 'framework' boilerplate can stay IMO.

dandanlen (commented):
For pallets, I'd rather we keep the migration module linked, and we can simply remove the individual old migrations, ie. if the migration module looks like this:

```rust
pub mod v2;
pub mod v3;

use cf_runtime_upgrade_utilities::VersionedMigration;

pub type PalletMigration<T, I> = (
	VersionedMigration<crate::Pallet<T, I>, v2::Migration<T, I>, 1, 2>,
	VersionedMigration<crate::Pallet<T, I>, v3::Migration<T, I>, 2, 3>,
);
```

And say v2 is no longer needed, we can change it to:

```rust
pub mod v3;

use cf_runtime_upgrade_utilities::VersionedMigration;

pub type PalletMigration<T, I> = (
	VersionedMigration<crate::Pallet<T, I>, v3::Migration<T, I>, 2, 3>,
);
```

If none are needed:

```rust

use cf_runtime_upgrade_utilities::VersionedMigration;

pub type PalletMigration = ();
```

IMO this is better than simply commenting out the whole module. We can continue to refer to it in the runtime, it will just be a no-op. This is fine. We should at least get a message in the logs saying "no migrations defined for pallet X".

And when someone wants to add a new migration, they just add it to the migrations module.

state-chain/runtime/src/lib.rs
Janislav: Well, but then we still compile it into the runtime and size stays the same. Then actually can just keep them. They are getting skipped anyway.

dandanlen: Maybe you misunderstood the suggestion - see the other longer comment.

I think we can make things as developer-friendly as possible by just removing the v1 v2 etc migrations and keeping the boilerplate 'shell'.

This way adding a new migration again is simply adding a new vX migration, and you don't need to remember to add something to the runtime, uncomment the migration module declaration etc. 

Nothing extra will be compiled into the runtime - if there is no migration the implementation is just `()` or similar.

dandanlen:
Looks like was duplicated in #4478

dandanlen:
I'll compare/merge the two PRs.

dandanlen:
Closing this in favour of the other PR, see https://github.com/chainflip-io/chainflip-backend/pull/4478#pullrequestreview-1859001619

PR #4465, opened 2024-01-30
chore: add basic check solana

albert-llimos:
## Checklist

## Summary

Add basic SOL transfer to the bouncer to make sure Solana is running properly and that we don't break it before we start adding the logic to the protocol.
For now, allow the SOL transfer to be skipped for anyone that doesn't have Solana installed yet to avoid issues. I will remove that once Solana is integrated.

### DIFF
--- a/bouncer/commands/send_sol.ts
+++ b/bouncer/commands/send_sol.ts
@@ -1,0 +1,27 @@
+#!/usr/bin/env -S pnpm tsx
+// INSTRUCTIONS
+//
+// This command takes two arguments.
+// It will fund the ethereum address provided as the first argument with the amount
+// provided in the second argument. The asset amount is interpreted in ETH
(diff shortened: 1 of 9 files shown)


bouncer/commands/send_sol.ts
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env -S pnpm tsx
+// INSTRUCTIONS
+//
+// This command takes two arguments.
+// It will fund the ethereum address provided as the first argument with the amount
```
martin-chainflip, line 5: This comment still references Ethereum addresses and ETH

albert-llimos: Missed that one. Fixed!

martin-chainflip (approved):

PR #4467, opened 2024-01-30
Feat/delay swaps after deposit

msgmaxim:
# Pull Request

Closes: PRO-1137

## Checklist

## Summary

- Added `execute_at` field to `Swap`.
- Added a migration that adds `execute_at` (set to the current block) to any pending swaps. Tested migration with `try-runtime`.
- Only swaps for which `execute_at <= current_block` are now executed, the others remain in the SwapQueue to be processed by future blocks. Note that the use of < ensures that swaps that missed their block for some reason (i.e. due to failure or safe mode) can still be processed in future blocks, and the order of swaps is preserved.
- Added `execute_at` to `SwapScheduled` event.
- Updated all existing tests, and added a new test.

I don't have a way to indicate to LPs that a swap is "re-scheduled". Have we decided how this should be done?


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1013,11 +1050,15 @@ pub mod pallet {
 				};
 
 			let gas_swap_id = if let Some(other_gas_asset) = other_gas_asset {
+				let execute_at =
```
kylezs, line 1053: we should probably ensure both ccm swaps are in the same block by using a variable used by both here

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -363,14 +375,21 @@ pub mod pallet {
...
 				return
 			}
 			// Wrap the entire swapping section as a transaction, any failed swap will rollback all
 			// storage changes.
 			if let Err(failed_swap) = with_storage_layer(|| -> Result<(), BatchExecutionError> {
-				let mut swaps = SwapQueue::<T>::take();
+				let all_swaps = SwapQueue::<T>::take();
+
+				let (mut swaps, remaining): (Vec<_>, Vec<_>) =
+					all_swaps.into_iter().partition(|s| s.execute_at <= n);
```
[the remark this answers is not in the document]
msgmaxim: As discussed, this is out of scope of this PR.

kylezs (commented):
👍 Looks good to me. We want to wait until release/1.2 is branched before merging this

dandanlen (approved):
Approving with @kylezs agreement.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -221,7 +227,7 @@ pub mod pallet {
 
 	/// Scheduled Swaps
 	#[pallet::storage]
-	pub(crate) type SwapQueue<T: Config> = StorageValue<_, Vec<Swap>, ValueQuery>;
+	pub(crate) type SwapQueue<T: Config> = StorageValue<_, Vec<Swap<T>>, ValueQuery>;
```
dandanlen, line 230: I think it might be better to use a Map<BlockNumber, Vec<Swap<T>> (+ Value<BlockNumber> for the last block executed)...

This way we always have to decode all swaps (including future swaps) and then filter them out, and write the remaining Vec back to storage. It's wasteful. Also the Vec can grow arbitrarily large (and repeats the block_number value for every swap, which is a waste). Also keeps the Swap struct unchanged - no need for the `T` type param and the migration is much simpler.

msgmaxim: > I think it might be better to use a Map<BlockNumber, Vec<Swap> (+ Value for the last block executed)...

I did consider this, but I thought it'd be better to keep things simple? With the map we'd need to be more careful when selecting which keys to query (because of failures). It is not very hard, but more room for error still.

> This way we always have to decode all swaps (including future swaps) and then filter them out, and write the remaining Vec back to storage. It's wasteful. 

I don't have a good intiution when it comes to SC runtime's requirements for performance, but we are only talking about a handful of swaps per block, I imagine (e.g. looks like TC does 30 swaps per minute on average)?

I'm happy to make the change if you think it is necessary, but to me it seems like an optimisation we could consider in the future?

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1372,7 +1436,7 @@ fn can_handle_swaps_with_zero_outputs() {
 			);
 
 			// Swaps are not egressed when output is 0.
-			assert_eq!(SwapQueue::<Test>::decode_len(), None);
+			assert_eq!(SwapQueue::<Test>::decode_len(), Some(0));
```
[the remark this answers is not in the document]
msgmaxim: Periously we were "taking" the value, e.g. the entire `Vec`, so it was getting replaced by `None`. In my code it is more natural to leave the empty Vec instead, and I think conceptually it makes more sense too.

state-chain/cf-integration-tests/src/swapping.rs
```diff
@@ -555,7 +566,7 @@ fn failed_swaps_are_rolled_back() {
 		System::reset_events();
 
 		// Usdc -> Flip swap will fail. All swaps are stalled.
-		Swapping::on_finalize(1);
+		Swapping::on_finalize(3);
```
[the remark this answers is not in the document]
msgmaxim: Updated to use SWAP_DELAY/execute_at in tests more consistently now. I noticed that most existing tests don't use test utils consistently in cases like these, so I didn't put much effort into improving tests that weren't super relevant to this PR. I did however use `execute_next_block` in the new tests that I added.

state-chain/pallets/cf-swapping/src/tests.rs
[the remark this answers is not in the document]
kylezs: Does it makes sense to use a ValueQuery here now instead then?

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: Yeah, I was thinking about this approach too, but I did consider it an optimisation, given we already do query for every swap. I do agree it's not nice having the block number duplicated in every swap, but in reality, how often are we getting two swaps in a 6 second window? So imo it's worth tracking this (creating an issue) as a future optimisation (if we ever needed to optimise it at all.

kylezs (approved):
Would wait for Dan's approval here too

kylezs:
I was a little confused by what was Dan suggested, but we chatted in person, and now I think I agree. Just writing this if it helps clarify. 

The per block method is slightly better due to the fact we can execute a blocks worth of swaps *atomically*. So we would only try to swap a blocks worth of swaps, and if one within that block fails we roll back and then wait until the next block. Rather than trying to do *all* swaps at once. Doing a block at a time, allows us to process that blocks worth of swaps before then trying to process the next block's swaps.  Doing all the swaps at the same time (as we do now in this PR), if we have low liquidity, makes it more difficult to recover since the total liquidity required increases and we can't process *anything*. Where as by doing it per block, the total liquidity to process *all* the swaps is still the same, but we're able to process a subset - which actually reduces the total liquidity required, because through swapping, LPs get the asset on the other side which they can then deploy into new orders.

msgmaxim:
Should be ready for re-review @dandanlen, @kylezs. I haven't got a chance to re-test benchmarks and the migration (the change there is relatively small), but if this PR is approved I will make sure to re-test them tomorrow before merging.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -383,102 +392,17 @@ pub mod pallet {
...
-							direction,
-							amount,
-						}),
-					BatchExecutionError::DispatchError { error } => {
-						log::error!("Failed to execute swap batch: {:?}", error);
-					},
-				}
+			// NOTE: we iterate manually because BlockNumberFor<T> does not implement Step:
+			while block_to_process <= current_block {
+				Self::process_swaps_for_block(block_to_process);
```
dandanlen, line 404: We want to exit the loop here if a block of swaps fails to execute.
Also we need to rememember to update `FirstBlockWithPendingSwaps`.

We can wrap `process_swaps_for_block` in `#[transactional]` to reduce boilerplate, and do something like this:

```rust
while block_to_process <= current_block {
	match Self::process_swaps_for_block(block_to_process) {
		Err(e) {
			// log error / deposit event
		} else {
			block_to_process += 1
		}
	}
}
FirstBlockWithPendingSwaps::put(block_to_process)
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -708,20 +724,10 @@ pub mod pallet {
 						debug_assert!(false, "Swap is not completed yet!");
 					}
 				}
+
+				FirstBlockWithPendingSwaps::<T>::set(block + 1u32.into());
```
dandanlen: nit: The block number in `FirstBlockWithPendingSwaps` doesn't necessarily contain any swaps so the name is a bit misleading. I would suggest storing a `LastBlockExecuted` or similar. Then we don't rely on having matching logic in the function and outside (ie. we increment the block here, and also in the loop where we increment `block_to_process`).

In fact we can write to storage outside the loop:

```rust
let mut block_to_execute = LastBlockExecuted::get();
while block_to_execute <= current_block - 2 {
	match process_swaps_for_block(block_to_execute) {
		Ok(()) => block_to_execute += 1,
		Err(_) => break;
	}
}
LastBlockExecuted::put(block_number);
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -644,90 +644,88 @@ pub mod pallet {
...
-				// Take NetworkFee for all swaps
-				for swap in swaps.iter_mut() {
-					debug_assert!(
-						swap.stable_amount.is_some(),
-						"All swaps should have Stable amount set here"
-					);
-					let stable_amount = swap.stable_amount.get_or_insert_with(Default::default);
-					*stable_amount = T::SwappingApi::take_network_fee(*stable_amount);
-				}
+			let mut swaps = SwapQueue::<T>::take(block);
```
dandanlen, line 650: I think it makes sense to return Ok(()) early here if there are no swaps in the batch.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: > I would suggest storing a LastBlockExecuted or similar.

That was my first thought too, but then I realised that it would be awkward around block 0, since I would need to use None or something like that to indicate that block 0 hasn't been processed. Of course, in practise we could just initialize it to 0 anyway, but I wanted to be more precise and went for a different name instead.

> The block number in FirstBlockWithPendingSwaps doesn't necessarily contain any swaps so the name is a bit misleading.

How about `FirstUnprocessedBlock`?

> In fact we can write to storage outside the loop

My intention was to write block number as part of the "transaction". I know `LastBlockExecuted::put(block_number)` probably can't fail, but conceptually it belongs there. I don't think it is much of an optimisation either since usually we would be processing 1 block only.

msgmaxim: Update: I renamed `FirstBlockWithPendingSwaps` -> `FirstUnprocessedBlock`, and moved `FirstUnprocessedBlock::<T>::set(block_to_process);` outside of `process_swaps_for_block` (and outside of the loop). Let me know if you still feel strongly about calling the `LastBlockExecuted` naming after my arguments against it.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Updated

msgmaxim:
> I haven't got a chance to re-test benchmarks and the migration (the change there is relatively small), but if this PR is approved I will make sure to re-test them tomorrow before merging.

Update: now have tested benchmarks and the migration. Turns out I needed to re-enable migration for cf-swapping in runtime/lib.rs, but now that's done as well.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -605,11 +546,14 @@ pub mod pallet {
 				Self::validate_destination_address(&destination_address, to)?;
 			let swap_origin = SwapOrigin::Vault { tx_hash };
 
+			let execute_at = frame_system::Pallet::<T>::block_number() + SWAP_DELAY_BLOCKS.into();
```
kylezs, line 549: We duplicate this in 3 places, why not just assign this inside `schedule_swap_internal` and then pass it out so it can be used in the events?

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1934,3 +1985,61 @@ fn broker_bps_is_limited() {
 		);
 	});
 }
+
+#[test]
+fn swaps_are_executed_according_to_execute_at_field() {
```
kylezs, line 1990: We should add a test for failure too. 
That a) the state is rolled back and no state changes have occurred
and b) that success on the next block they go through, as well as the subsequent block's swaps being attempted

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1872,6 +1917,7 @@ fn swap_scheduled_event_witnessed(
 	deposit_amount: AssetAmount,
 	destination_asset: Asset,
 	broker_commission: AssetAmount,
+	execute_at: u64,
```
kylezs, line 1920: not part of your changes by the name of this function is a little strange, it has nothing to do with witnessing 🤔 . I think just `assert_swap_scheduled_event` like we do for other things is clearer

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -216,71 +221,80 @@ fn cannot_swap_with_incorrect_destination_address_type() {
...
+				None,
+				0
+			));
+			// 2. Schedule the swap -> SwapScheduled
+			<Pallet<Test> as SwapDepositHandler>::schedule_swap_from_channel(
+				ForeignChainAddress::Eth(Default::default()),
+				Default::default(),
+				Asset::Eth,
+				Asset::Usdc,
+				500,
```
kylezs, line 249: can use a const here instead of magic number

kylezs (commented):
Nice, think just some tidy up and a little more testing

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -192,7 +194,7 @@ pub mod pallet {
 
 	#[pallet::config]
 	#[pallet::disable_frame_system_supertrait_check]
-	pub trait Config: Chainflip {
+	pub trait Config: Chainflip + core::fmt::Debug {
```
dandanlen, line 197: nit: we shouldn't need this constraint.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -383,103 +392,35 @@ pub mod pallet {
...
-							},
-						};
-					} else {
-						debug_assert!(false, "Swap is not completed yet!");
+			let mut block_to_process = FirstUnprocessedBlock::<T>::get();
+
+			// NOTE: we iterate manually because BlockNumberFor<T> does not implement Step:
+			while block_to_process <= current_block {
+				if let Err(failed_swap) = Self::process_swaps_for_block(block_to_process) {
+					match failed_swap {
```
dandanlen, line 405: nit: this could be single `match` instead of nesting a match inside `if let Err(..)`

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -383,103 +392,35 @@ pub mod pallet {
...
+						BatchExecutionError::SwapLegFailed { asset, direction, amount } =>
+							Self::deposit_event(Event::<T>::BatchSwapFailed {
+								asset,
+								direction,
+								amount,
+							}),
+						BatchExecutionError::DispatchError { error } => {
+							log::error!("Failed to execute swap batch: {:?}", error);
```
dandanlen, line 413: Not sure why we weren't emitting the BatchSwapFailed event here, we probably should.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Naming is fine 👌

state-chain/pallets/cf-swapping/src/benchmarking.rs
```diff
@@ -117,10 +119,12 @@ mod benchmarks {
...
 		assert_eq!(
-			SwapQueue::<T>::get(),
+			SwapQueue::<T>::get(
+				<frame_system::Pallet<T>>::block_number() + SWAP_DELAY_BLOCKS.into()
+			),
 			vec![
-				Swap::new(1, Asset::Usdc, Asset::Eth, 1_000 - 1, SwapType::CcmPrincipal(1)),
-				Swap::new(2, Asset::Usdc, Asset::Eth, 1, SwapType::CcmGas(1))
+				Swap::new(1, Asset::Usdc, Asset::Eth, 1_000 - 1, SwapType::CcmPrincipal(1),),
+				Swap::new(2, Asset::Usdc, Asset::Eth, 1, SwapType::CcmGas(1),)
```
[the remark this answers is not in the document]
msgmaxim: Not sure what you are saying here. They do have different ids here: 1 and 2.

state-chain/pallets/cf-swapping/src/migrations/schedule_swaps.rs
```diff
@@ -0,0 +1,78 @@
...
+				swap_id: s.swap_id,
+				from: s.from,
+				to: s.to,
+				amount: s.amount,
+				swap_type: s.swap_type,
+				stable_amount: s.stable_amount,
+				final_output: s.final_output,
+				fee_taken: s.fee_taken,
+			})
+			.collect();
```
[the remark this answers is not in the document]
msgmaxim: Yeah, this was from when I had a differnet type. Updated now.

state-chain/pallets/cf-swapping/src/tests.rs
msgmaxim: Renamed to `assert_swap_scheduled_event_emitted`.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: I had a quick look and it doesn't seem like a trivial change. I'm not sure if we will always be able to associate data like asset/direction/amount with a generic `DispatchError`. (I think some of these errors come from `with_transaction` and don't have much to do with specific swaps.) Are you suggesting adding another `BatchSwapFailed`  event not coinaining this data or changing the existing event?

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: I have made the change as you suggested, but I'm not sure if it is a big improvement considering that now I have to duplicate `break` (and remember to add it in case we add other error types).

state-chain/pallets/cf-swapping/src/tests.rs
msgmaxim: The ability to process swaps for blocks other than the current was already covered by safe mode tests, but I now added `swaps_get_retried_on_next_block_after_failure` to test for swaping errors explicitly.

msgmaxim:
I have either addressed or responded to all comments @kylezs @dandanlen. Let me know what you think.

state-chain/pallets/cf-swapping/src/mock.rs
```diff
@@ -67,7 +69,23 @@ parameter_types! {
...
+}
+
 pub struct MockSwappingApi;
+
+impl MockSwappingApi {
+	pub fn set_swaps_should_fail(should_fail: bool) {
+		SWAPS_SHOULD_FAIL.with(|cell| cell.set(should_fail));
+	}
+
+	fn get_swaps_should_fail() -> bool {
```
kylezs, line 84: nit: `swaps_should_fail`

kylezs (approved):
👍 LGTM

state-chain/pallets/cf-swapping/src/benchmarking.rs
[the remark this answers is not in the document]
kylezs: inside the SwapType

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: I didn't realise we couldn't emit the event in the other error case...

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: I think maybe an event with just the block number of the failed swap would be useful. But not critical here, we can add this another time if we deem it necessary.

dandanlen (approved):
LGTM thanks!

state-chain/pallets/cf-swapping/src/benchmarking.rs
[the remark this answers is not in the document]
msgmaxim: But that's CcmId, which I understand should be the same for gas and principal, right?

PR #4476, opened 2024-02-01
Chore/benchmarking improvements

syan095:
# Pull Request

Closes: PRO-1151

## Checklist

## Summary

Fixed an incomplete benchmarking.
Remove  `let _ = `s in benchmarking code to ensure the intended data is returned
Removed 2 TODOs after investigation:
1. Governance::on_initialize(): This benchmarks the partitioning of expired proposals. Partition splits the vec into 2 separate vecs, whether the item is added to the left or right does not matter. We also do not want to add expired proposals as the handling of expired proposals is benchmarked separately. Having any proposal expire during this call will double count the weight.

2. Witnesser:: witness_at_epoch(): The correct code path is used here. We do not want to trigger the actual proposed call during benchmarking, as the weight of the call itself is added separately. Benchmarking is now done with 3 node chain spec so during actual Runtime benchmarking the proposed call should not be dispatched. This is working as intended.

### DIFF
--- a/state-chain/pallets/cf-broadcast/Cargo.toml
+++ b/state-chain/pallets/cf-broadcast/Cargo.toml
@@ -43,7 +43,7 @@ sp-io = { git = "https://github.com/chainflip-io/polkadot-sdk.git", tag = "chain
 cf-test-utilities = { path = '../../test-utilities' }
 
 [features]
-default = ['std']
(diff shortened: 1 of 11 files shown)


dandanlen (approved):
Thanks Roy.

PR #4477, opened 2024-02-01
feat: CFE interface version check

j4m1ef0rd:
# Pull Request

Closes: PRO-1128

## Checklist

## Summary

- Added a storage value for a CFE Events version number.
	- Added it to the genesis config so we will not read 0 on block 0 with localnets.
- Changed the SCO to read this new version number each block so it can decode the events using the correct struct.
- Added a function to the storage api that reads a storage value without decoding it.

### DIFF
--- a/engine/src/state_chain_observer/sc_observer/mod.rs
+++ b/engine/src/state_chain_observer/sc_observer/mod.rs
@@ -2,15 +2,16 @@ mod crypto_compat;
 #[cfg(test)]
 mod tests;
 
-use anyhow::{anyhow, Context};
+use anyhow::{anyhow, bail, Context};
 use cf_chains::{
 	btc::{self, PreviousOrCurrent},
 	Chain,
 };
 use cf_primitives::{BlockNumber, CeremonyId, EpochIndex};
+use codec::Decode;
 use crypto_compat::CryptoCompat;
 use futures::{FutureExt, StreamExt};
-use pallet_cf_cfe_interface::{ThresholdSignatureRequest, TxBroadcastRequest};
+use pallet_cf_cfe_interface::{ThresholdSignatureRequest, TxBroadcastRequest, CFE_EVENTS_VERSION};
 
 type CfeEvent = pallet_cf_cfe_interface::CfeEvent<Runtime>;
 
@@ -48,7 +49,7 @@ use multisig::{
 };
 use utilities::task_scope::{task_scope, Scope};
 
-use super::client::chain_api::ChainApi;
+use super::client::{chain_api::ChainApi, BlockInfo};
 
 async fn handle_keygen_request<'a, StateChainClient, MultisigClient, C, I>(
 	scope: &Scope<'a, anyhow::Error>,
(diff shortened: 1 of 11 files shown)


engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -525,6 +516,39 @@ where
...
+	// First we check the version number of the CFE events
+	let cfe_events_version = state_chain_client
+		.storage_value::<pallet_cf_cfe_interface::CfeEventsVersion<Runtime>>(block_info.hash)
+		.await
+		.expect("Failed to get CFE events version");
+
+	// For now we only have one version of CFE events, so we panic if the version number is not a
+	// match.
+	if cfe_events_version != CFE_EVENTS_VERSION {
+		panic!("Cannot process CFE events at block {}. Incompatible CFE events version: {cfe_events_version}. Using version: {CFE_EVENTS_VERSION}", block_info.number);
```
kylezs, line 533:
```suggestion
		panic!("Cannot process CFE events at block {}. Incompatible CFE events version. State Chain returned: {cfe_events_version}. CFE expects: {CFE_EVENTS_VERSION}", block_info.number);
```

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -525,6 +516,39 @@ where
...
+/// Get the CFE events from the SC at the given block with version checking
+pub async fn get_cfe_events(
+	state_chain_client: Arc<impl StorageApi>,
+	block_info: BlockInfo,
+) -> Result<Vec<CfeEvent>, anyhow::Error> {
+	// First we check the version number of the CFE events
+	let cfe_events_version = state_chain_client
+		.storage_value::<pallet_cf_cfe_interface::CfeEventsVersion<Runtime>>(block_info.hash)
+		.await
+		.expect("Failed to get CFE events version");
```
AlastairHolmes, line 528: I think we should move away from using expects really, and given the code didn't previously, can we return the error?

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -525,6 +516,39 @@ where
...
+	// For now we only have one version of CFE events, so we panic if the version number is not a
+	// match.
+	if cfe_events_version != CFE_EVENTS_VERSION {
+		panic!("Cannot process CFE events at block {}. Incompatible CFE events version: {cfe_events_version}. Using version: {CFE_EVENTS_VERSION}", block_info.number);
+	}
+
+	// Getting the events without decoding yet, so in the future we can decode them with the correct
+	// struct for the detected version.
+	match state_chain_client
+		.storage_value_raw::<pallet_cf_cfe_interface::CfeEvents<Runtime>>(block_info.hash)
```
kylezs, line 539: Think this is a bug in the same way we had the Polkadot bug. We should be getting the version at the *previous* block? 

I.e. events that are version 2, change to version 3 in block 20. The blocks *in* block 20 are version 2 but the blocks in block 21 are now version 3. And querying returns the value at the *end* of the block, if we queried for block 20, we'd get version 3, but the events within that block are version 2 events.

Would be nice if we could write a test for this.

engine/src/state_chain_observer/sc_observer/mod.rs
AlastairHolmes, line 533: same here, doesnt it make sense to return an error.

state-chain/pallets/cf-cfe-interface/src/lib.rs
```diff
@@ -11,17 +11,16 @@ use cf_chains::{
...
-	traits::StorageVersion,
-};
+use frame_support::traits::StorageVersion;
 use frame_system::pallet_prelude::*;
 pub use pallet::*;
 use sp_std::vec::Vec;
 
 pub const PALLET_VERSION: StorageVersion = StorageVersion::new(0);
 
+// If any breaking change is made the CFE events, increment this version.
```
kylezs, line 21:
```suggestion
// If any change is made to the CFE events, increment this version.
```

All changes to the CFE events are breaking to the CFE - possibly include that in the comment too

state-chain/pallets/cf-cfe-interface/src/lib.rs
```diff
@@ -37,9 +36,29 @@ pub type TxBroadcastRequest<T, C> =
 pub mod pallet {
 
 	use cf_traits::Chainflip;
+	use frame_support::pallet_prelude::*;
 
 	use super::*;
 
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T: Config> {
+		pub cfe_events_version: u32,
```
kylezs, line 45: I don't think this needs to be a parameter, we can just set it to CFE_EVENTS_VERSION

state-chain/pallets/cf-cfe-interface/src/lib.rs
```diff
@@ -55,9 +74,14 @@ pub mod pallet {
 	#[pallet::unbounded]
 	pub type CfeEvents<T: Config> = StorageValue<_, Vec<CfeEvent<T>>, ValueQuery>;
 
+	#[pallet::storage]
+	pub type CfeEventsVersion<T: Config> = StorageValue<_, u32, ValueQuery>;
+
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(_block_number: BlockNumberFor<T>) -> frame_support::weights::Weight {
+			CfeEventsVersion::<T>::put(CFE_EVENTS_VERSION);
```
kylezs, line 83: What's the intention here? This isn't required. Currently what this is doing is setting the CfeEventsVersion to the same thing every block

engine/src/state_chain_observer/sc_observer/mod.rs
```diff
@@ -525,6 +516,39 @@ where
...
+		.await
+	{
+		Ok(Some(events_data)) => <Vec<CfeEvent>>::decode(&mut &events_data[..])
+			.map_err(anyhow::Error::msg)
+			.context(format!("Failed to decode events at block {}", block_info.number)),
+		Ok(None) => Ok(vec![]),
+		Err(error) => {
+			bail!("Failed to read events at block {}. {error}", block_info.number);
```
kylezs, line 547: nit: when not returning early, we use Err(anyhow::anyhow!())
```suggestion
			bail!("Failed to read events at block {}. {error}", block_info.number);
```

AlastairHolmes:
Would it be better to instead add an rpc to the node for getting the events (where it has a parameter for the events "version"), and the node can translate the events into whatever previous version the engine needs. Same as how Substrate handles metadata versions. This avoids needing new engine binaries.


Also I feel an even better approach would be to simply add new CFE events (instead of changing existing events), and then deprecate the old ones (And remove them much later). Not only would this approach avoid needing new engine binaries when the cfe events are changed, but also the need to write a translation from the new to old events (or vice versa).

kylezs:
I sent out a calendar invite to discuss the CFE upgrades to makes sure we're clear on the direction we want to go.

state-chain/pallets/cf-cfe-interface/src/lib.rs
msgmaxim: What do you have in mind instead? If you are thinking of using the const value directly, I don't think that would work as it wouldnt' have the correct behaviour around runtime upgrades (historical data wouldn't be accurate). I mentioned that in the linear issue [(PRO-1128)](https://linear.app/chainflip/issue/PRO-1128/store-cfe-events-version-inside-cfe-interface-pallet).

msgmaxim:
> Also I feel an even better approach would be to simply add new CFE events (instead of changing existing events), and then deprecate the old ones

The advantage of using an explicit version is that we can keep the event enum tidy as we won't be accumulating a whole bunch of legacy events just to keep decoding consistent with past versions. Imagine changing signging requests where we'd need to deprectate one event for each chain. This would also let us keep related events grouped (e.g. signature requests) together the way they are now. In general, I see having a version as giving us more flexibility in the future, and the cost of having it is fairly minimal.

Even if we have the version, we can still add new events without changing the version if we really want. So we'll have the flexibility to decide which approach suits better for any given change to the events.

msgmaxim:
> Would it be better to instead add an rpc to the node for getting the events (where it has a parameter for the events "version"), and the node can translate the events into whatever previous version the engine needs.

This may be a reasonable solution too, but still not convinced how much not needing new engine binaries buys us (considering that we'd want to be able handle these events most likely). The solution we have here is also reasonable in my view, so I don't know if it is worth throwing it away. Maybe something we could discuss during the upcoming meeting.

engine/src/state_chain_observer/sc_observer/mod.rs
msgmaxim: I'm not sure if this is true. Both the version and the events come from storage, and storage migration happen at the beginning of the block, I believe? Worth checking of course, but I would say a manual test with a runtime upgrade would suffice?

state-chain/pallets/cf-cfe-interface/src/lib.rs
msgmaxim: In theory, you could remove some events from the end of the enum, and it wouldn't be a breaking change.

msgmaxim (approved):
LGTM, but might be worth doing the test that Kyle mentioned.

state-chain/pallets/cf-cfe-interface/src/lib.rs
kylezs: You would update it as part of a runtime upgrade and not every block

kylezs:
Let's not merge this until the meeting.

j4m1ef0rd:
After a team meeting we have decided that this PR is not the correct approach. We are more in favour of using an RPC to get the events with some fuzzy logic to decode them, possibly using JSON. But this is not a priority right now, because the 2 dll's upgrade solution will avoid compatibility issues for now. Closing this PR

PR #4478, opened 2024-02-02
chore: remove storage migrations already applied to the Mainnet

syan095:
# Pull Request

Closes: PRO-1115
Closes: PRO-1162
Closes: PRO-988

## Checklist

## Summary

Removed storage migrations that are already completed on the mainnet.
Removed v1 from AMM module


dandanlen:
This looks like a (partial?) duplicate of #4464 

Looks like the Linear issues were duplicated too... sorry.

dandanlen (commented):
I'm going to close the other PR in favour of this one since I think this is closer what we want.

I would like us to be consistent in how we deal with this (removing/deprecating migrations). Here are some concrete suggestions:

1. We delete old migrations. I can see why it might sometimes be useful to keep these in the file tree, but we can access them at any time anyway using git. (for example in vs code just git -> compare with branch release/1.x).
2. We add a migration 'template' to the runtime upgrade utils and import this when there are no migrations. -> more consistent
3. We avoid commenting out migrations. -> fewer diffs / surprises 
4. The template migration should be `unimplemented!()` -> returns an error if we accidentally use it
5. The template migration should be added like this: `VersionedMigration<Pallet<T>, template::Migration, $current_pallet_version, ${current_pallet_version + 1}>` -> placeholder for the next migration that needs to be added


@Janislav / @syan095 please agree amongst yourselves which of you wants to complete this.

syan095:
Are these changes what you had in mind @dandanlen ?

dandanlen (approved):
Thanks @syan095 this is great.

PR #4491, opened 2024-02-06
refactor: Remove RpcAsset (PRO-1187)

AlastairHolmes:
This changes the serde impls of any::Asset, to match those of the existing RpcAsset, and then replaces every usage of RpcAsset with Asset. Also in the places we previously used the serde impl of Asset, I've added a OldAsset type to provide matching behaviour (so we don't break the interfaces). The existing serde impl on Asset, only allowed "Eth"/"Usdc" etc, which is not what we want so we will need to replace those uses of OldAsset, with v2 rpcs etc at some point, which I will create a ticket to do.

This also introduces the AssetMap types, that provide compile time safe "asset->item" maps.

In a few of the CLI commands we allow parameters like source_asset, and source_chain. I removed this option, as I feel strongly using separate arguments to allow users to explicitly specify the chain is not the right approach. If we care about this functionality, I'd suggest we use something allow something like: "Ethereum:USDC" (Just for the CLI). (I can change this back, but I feel the current design doesn't work nicely with the direction we intend to go with always specifying the chain).

I'm going to replace `Asset` with `ForeignChainAndAsset` (of course renaming it to `Asset`), but will do so later. As this PR makes sense on its own (And doing so will take abit more time then I'd like to spend atm).

**Note this is split into atomic commits:**
- First I change the serde impl
- Then I remove rpc asset type, and replace it with the new Asset type
- Then introduce the AssetMap type, replacing the existing HashMap<ForeignChain, HashMap<Asset, T>> we use.
- Then find all the locations we were using the Old Asset's serde impl (which we probably shouldn't have been doing), and replace those cases with the OldAsset type (with has matching behaviour to the old Asset type).

**Finally the only breaking change in this PR, is the change to the CLI mentioned above, all the other interfaces are unchanged, but I had to change one of the snapshots because in the unit tests previous in the HashMap<ForeignChain, HashMap<Asset, T>> we didn't include all the assets and chains, but my new AssetMaps always have all assets/chains, but in all the real interfaces this is true always anyway**.


AlastairHolmes:
Also I added the option to do "{"asset": "ETH"}", which I think makes sense while we allow the chain to be implicit. But happy to change back.

AlastairHolmes:
Just need to fix this: `try_from_iter`, which is why the unit tests aren't passing, which I'll do today.

state-chain/primitives/src/chains/assets.rs
```diff
@@ -302,16 +368,16 @@ macro_rules! assets {
 assets!(pub enum Asset {
 	// 0 is reserved for particular cross chain messaging scenarios where we want to pass
 	// through a message without making a swap.
-	(eth, Ethereum) => {
-		(Eth, "eth") = 1u32 (GAS_ASSET),
-		(Flip, "flip") = 2u32,
-		(Usdc, "usdc") = 3u32,
+	(eth, Ethereum, "Ethereum") => {
```
dandanlen, line 379: This could also just be stringify!($chain) ?

dandanlen (commented):
Very nice. 

Wen ChainMap<T> 😆 ?

Would be good to talk about how this extends to new chains (ie. arbitrum: ArbEth vs (Arb, Eth) etc.

dandanlen (approved):

state-chain/primitives/src/chains/assets.rs
AlastairHolmes: Yeah I tried, in this case it is impossible because stringify!(...) is an expression (Kind of), but the serde attribute needs a string literal. I think it is about the ordering of macro eval.

state-chain/primitives/src/chains/assets.rs
```diff
@@ -13,182 +13,307 @@
...
+				PartialEq,
+				Eq,
+				Encode,
+				Decode,
+				TypeInfo,
+				MaxEncodedLen,
+				Hash,
+				PartialOrd,
+				Ord,
+				EnumIter,
```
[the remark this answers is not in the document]
AlastairHolmes: I just tried, its actually kind of annoying to do (Without making the iterator object an Array, and thereby making it take more space (u32 verse n*u32)), so I'm going to leave. If that's ok.

PR #4496, opened 2024-02-07
feat: store prewitnessed deposits with id

j4m1ef0rd:
# Pull Request

Closes: PRO-1141

## Checklist

## Summary

- id counter for prewitnessed deposits
- new extrinsic for prewitnessing and executing the call. used only for prewinessing deposits.
	- new origin for executing a prewitness call
- Change the processing of witnesses to check the origin and either add it as a prewitness or process the witness as before.
- Storing the prewitnessed deposits in a double map
- Cleanup the prewitnessed deposits in on_idle and when witnessed
- Changed some test code to now use the `process_deposits` extrinsic anymore due to issues with the origin. Instead they just call the `process_deposit_witnesses` function directly.


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -43,6 +43,13 @@ pub use pallet::*;
 use sp_runtime::traits::UniqueSaturatedInto;
 use sp_std::{vec, vec::Vec};
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub struct BoostableDeposit<C: Chain> {
```
kylezs, line 47: When we have the extrinsic, if someone references this, how can we know what action to take? We need to store the deposit address here so that we can look up `DepositChannelLookup` to find the action. Also probably just useful for product anyway.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -809,6 +830,44 @@ pub mod pallet {
...
+			for DepositWitness { ref deposit_address, asset, amount, .. } in deposit_witnesses {
+				let id = BoostableDepositIdCounter::<T, I>::mutate(|id| -> u64 {
+					*id = id.saturating_add(1);
+					*id
+				});
+
+				let deposit_channel_details = DepositChannelLookup::<T, I>::get(deposit_address)
+					.ok_or(Error::<T, I>::InvalidDepositAddress)?;
+
+				BoostableDeposits::<T, I>::insert(
```
kylezs, line 860: We're currently just inserting, but we also need to remove to make sure this storage doesn't expand forever. There are a few cases where it's removed:
1. It's executed (not part of this PR)
2. It's finalised (not part of this PR)
3. Nothing happens. This is possible if we get a pre-witnessed deposit that then never reaches finalisation (it's re-org'd out). In this case we want to clean up when the channel expires. We should do this as part of this PR. 

So, we just have to work out how to avoid a linear search, which will require a little bit of storage shifting. 

What we could do is this:
BoostableDeposits stored as a storage double map `Map<Address<Map<BoostableDepositId, Deposit>`. 

When we create the deposit, we also add to another storage item `BoostId -> DepositAddress`. Note, that the BoostId would still increment in the same way, on *every* deposit i.e. not just per channel.

Now, when a channel expires, we can just remove the `address` key. 

When we call `boost` extrinsic we can still use just `boost_id`, but in the code we would do a `id -> address` lookup, so then we can lookup the deposit. using `address`, then `boost_id`. 

Tagging @dandanlen here too for thoughts on above^

dandanlen: Sounds good. To avoid the extra lookup, which would (I think?) require an extra map to store the correspondance between id and address, we could require boosters to provide the channel id or address with the deposit id. 

Ie. the booster would provide (effectively) a tuple of (channel_id, deposit_id) to identify the BoostableDeposit.

kylezs: Yeah avoided wanted to avoid that for simplicity for boosting - but perhaps it's an over-optimisation for now.

kylezs: Actually, I think (channel_id, deposit_id) is the only way to do it, because we have to clean up the ` BoostId -> DepositAddress` mapping too otherwise, which we would have to do with *another* storage item.

@GabrielBuragev tl;dr for you. Proposed change is that the `boost()` extrinsic would take `deposit_address, boost_id` rather than just `boost_id` as before. I don't think this should have an impact on you though?

GabrielBuragev: Yep, no impact on our end. Thank you for tagging me here!

engine/src/witness/btc/btc_deposits.rs
```diff
@@ -28,6 +28,7 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 	pub fn btc_deposits<ProcessCall, ProcessingFut>(
 		self,
 		process_call: ProcessCall,
+		is_prewitness: bool,
```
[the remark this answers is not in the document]
msgmaxim: I would say `is_prewitness` works better because to me `should_prewitness = true` implies that this would both pre-witness and post-witness, which is not the case.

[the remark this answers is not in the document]
kylezs: hmm maybe, I more thought `should` because it implies an action rather than a state. will leave it up to Jamie, idm either way

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: I would suggest (channel_id, deposit_id) because BoostableDeposit doesn't contain the address. Otherwise the booster will have to look up the address.

kylezs: > BoostableDeposit doesn't contain the address

I suggested above that it should include the address which would solve this

engine/src/witness/btc/btc_deposits.rs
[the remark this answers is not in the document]
j4m1ef0rd: I would lean towards Maxim's thinking here. I prefer `is_prewitness`

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -530,35 +557,70 @@ pub mod pallet {
...
+
+					used_weight = used_weight.saturating_add(read_weight);
+
+					if let Some(details) = DepositChannelLookup::<T, I>::take(address) {
+						let item_count = BoostableDeposits::<T, I>::iter_prefix(
 							details.deposit_channel.channel_id,
-							DepositChannel { state, ..details.deposit_channel },
+						)
+						.count() as u32;
+						let _removed_deposits = BoostableDeposits::<T, I>::clear_prefix(
```
kylezs, line 582: https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/child/struct.MultiRemovalResults.html

Looks like we can just use loops or something instead of doing the count above

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -530,35 +557,70 @@ pub mod pallet {
...
+							None,
 						);
+
+						let cleanup_weight = T::WeightInfo::remove_boostable_deposits(item_count)
+							.saturating_add(
+								frame_support::weights::constants::RocksDbWeight::get()
+									.reads_writes(0, 2),
+							);
+
+						if !remaining_weight.saturating_sub(used_weight).all_gte(cleanup_weight) {
```
kylezs, line 594: this is kinda dangerous no? We've taken a channel from the storage. If this happens, we won't recycle the address, and nor will we on the next iteration.

engine/src/witness/btc/btc_deposits.rs
AlastairHolmes, line 31: Why cannot whatever the logic instead here just go into the process_call?

kylezs: Good point, this is a rememnant of a previous approach, we should do it this way

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: yeah i was not finished refactoring the code. I want to do only reads to calculate the exact weight before starting the actual cleanup.

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: `loops` would tell us the weight after it is done, but we need to know beforehand so we can make sure we have enough weight. So I get the item count and use that to calculate the weight required.

engine/src/witness/btc/btc_deposits.rs
j4m1ef0rd: It is difficult to do while keeping the `ProcessCall` generic. Maybe i'm not understanding. Can you dm me with details if you have time.

state-chain/pallets/cf-ingress-egress/src/weights.rs
```diff
@@ -2,25 +2,25 @@
 //! Autogenerated weights for pallet_cf_ingress_egress
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2023-11-03, STEPS: `20`, REPEAT: `10`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2024-02-09, STEPS: `20`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! WORST CASE MAP SIZE: `1000000`
-//! HOSTNAME: `ip-172-31-9-222`, CPU: `Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz`
```
j4m1ef0rd, line 7: Do the benchmarks get ran by the CI at some point and pushed? Am I supposed to leave this file in the PR?

state-chain/TROUBLESHOOTING.md
```diff
@@ -77,7 +77,7 @@ Make sure to add `my-pallet/try-runtime` in the runtime's Cargo.toml, otherwise
 ### Compile the node with benchmark features enabled
 
 ```bash
-cargo cf-build-with-benchmarks
+cargo cf-build
```
j4m1ef0rd, line 80: From what i understand, the `benchmark.sh` script uses the release build, so this cmd is not the right one. I can't see a cargo alias for building release with benchmarks. What cmd do ppl use for this case?

kylezs: There's no need for you to run the specific benchmarks anyway, just get something that compiles. The real benchmarks will be run on reference hardware within a CI job

j4m1ef0rd: > There's no need for you to run the specific benchmarks anyway, 

But I need to use a new weight function in `on_idle`, It will not build until the weight file is updated, so i need to run the benchmark manually so i can continue work.

> the real benchmarks will be run on reference hardware within a CI job

When does this happen? Am I supposed to have the updated weight file in my PR?

I am still unsure about the procedure around the weight files. This troubleshooting guide does not explain much, and it is out of date, the commands don't exist.

kylezs: No need to update the weights file in the PR, it happens as part of a release.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -530,35 +557,71 @@ pub mod pallet {
...
 						);
+						DepositChannelLookup::<T, I>::remove(address);
+
+						used_weight = used_weight
+							.saturating_add(T::WeightInfo::remove_boostable_deposits(item_count))
+							.saturating_add(write_weight);
 					}
+
+					DepositChannelRecycleBlocks::<T, I>::mutate(|recycle_queue| {
+						recycle_queue.remove(0);
```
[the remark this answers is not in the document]
kylezs: > could we not just assume that there is 1 boostable deposit on average for every channel

I think this would be a reasonable heuristic

engine/src/witness/btc/btc_deposits.rs
```diff
@@ -65,15 +66,27 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
-						pallet_cf_ingress_egress::Call::<_, BitcoinInstance>::process_deposits {
-							deposit_witnesses,
-							block_height: header.index,
-						}
-						.into(),
-						epoch.index,
-					)
-					.await;
+					if is_prewitness {
+						process_call(
```
AlastairHolmes, line 70: Can you factor out the process_call here, its basically impossible to see the difference etween these two branches otherwise.

kylezs:
Dm'd this to Jamie, but thought I'd post here for visibility too:
I was chatting with @AlastairHolmes about this, and he suggested a nice solution to this if statement in the CFE problem we were trying to avoid, and I agree it's better.
Basically, the CFE code could remain more or less unchanged, and so it would still look like process_call(process_deposits) inside the btc deposits adapter, as it was before.
In the same way the code was before, where we wrap the calls in the prewitness extrinsics, we can create another origin within the witnesser pallet.
Currently we have:
```rust
    pub enum RawOrigin {
        HistoricalActiveEpochWitnessThreshold,
        CurrentEpochWitnessThreshold,
    }
```

Then we can add PrewitnessThreshold or something. What this does is call the extrinsic it wraps, with the PrewitnessThreshold origin - we do something like this already, see dispatch_call in the witness lib.rs . Then inside the process_deposits extrinisc it can do a check if origin is prewitness else if origin is witness etc. 

This then keeps the nice CFE structure and makes it clear that there's one set of data we witness (the process_deposits extrinsic). And the rest of the SC is pretty much the same. The two paths can be pulled into separate functions for easier testing and benchmarking and what not.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1474,3 +1461,35 @@ fn consolidation_tx_gets_broadcasted_on_finalize() {
...
+		let deposit_witnesses = vec![deposit_detail.clone()];
+
+		// Submit the prewitnessed deposit
+		assert_ok!(IngressEgress::add_boostable_deposits(deposit_witnesses.clone(),));
+
+		// Check that the deposit is stored in the storage
+		let deposit_id = BoostableDepositIdCounter::<Test>::get();
+		let channel_id = ChannelIdCounter::<Test>::get();
+		assert_eq!(
+			BoostableDeposits::<Test>::get(channel_id, deposit_id),
```
kylezs, line 1491: We can check the Prewitness event is emitted here too

state-chain/pallets/cf-witnesser/src/lib.rs
```diff
@@ -495,7 +497,32 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Emits an event to notify that this call has been witnessed. Then, it dispatches the call
+		/// using the prewitness threshold origin.
+		#[pallet::call_index(3)]
+		#[pallet::weight(call.get_dispatch_info().weight)] // TODO JAMIE: weight
+		pub fn prewitness_and_execute(
+			origin: OriginFor<T>,
+			call: Box<<T as Config>::RuntimeCall>,
```
kylezs, line 510: initially thought there might have been an issue here, but there's not - however still think it would be a good idea to add a test for this case (could be in a separate PR). The case being: 

A call is `prewitness_and_execute`d successfully. And then it's witnessed successfully. 

The reason for testing this explicitly is to ensure that we don't, for example, store the `CallHash` as executed after the prewitness goes through, and then we can't execute the main witness call.

state-chain/pallets/cf-ingress-egress/src/benchmarking.rs
```diff
@@ -163,6 +163,74 @@ mod benchmarks {
...
+		}
+
+		#[block]
+		{
+			let _old_deposits = BoostableDeposits::<T, I>::clear_prefix(0, u32::MAX, None);
+		}
+	}
+
+	// #[benchmark]
+	// fn prewitness_deposits(n: Linear<1, 255>) {
```
kylezs, line 187: to be removed

kylezs (commented):
Looks good, let me know when it's cleaned up for a final look over

state-chain/pallets/cf-ingress-egress/src/tests.rs
j4m1ef0rd: The `Prewitnessed` event is not emitted by the `add_boostable_deposits` function. It is emitted by the `prewitness_and_execute`, so we can't test for it here.

state-chain/pallets/cf-witnesser/src/lib.rs
```diff
@@ -495,7 +497,32 @@ pub mod pallet {
 			call: Box<<T as Config>::RuntimeCall>,
 		) -> DispatchResult {
 			T::EnsureWitnessed::ensure_origin(origin)?;
-			Self::deposit_event(Event::<T>::Prewitnessed { call: *call });
+			Self::deposit_event(Event::<T>::Prewitnessed { call: *call.clone() });
```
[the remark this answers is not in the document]
j4m1ef0rd: Don't we need it for prewitnessing ccm deposits? How else can `RuntimeEvent::Witnesser(pallet_cf_witnesser::Event::Prewitnessed { RuntimeCall::Swapping(pallet_cf_swapping::Call::ccm_deposit {..}}` exist.
See line 1440 in runtime/src/lib.rs

state-chain/pallets/cf-witnesser/src/lib.rs
```diff
@@ -489,7 +491,7 @@ pub mod pallet {
 		/// Simply emits an event to notify that this call has been witnessed. Implicitly signals
 		/// that we expect the same call to be witnessed at a later block.
 		#[pallet::call_index(2)]
-		#[pallet::weight(call.get_dispatch_info().weight)]
```
j4m1ef0rd, line 492: This extrinsic does not execute the call, so i think this weight was wrong?

state-chain/pallets/cf-witnesser/src/lib.rs
j4m1ef0rd: I thought the test was easy, but i ran into an issue with the origin. Yeah might leave it for another PR. probs a few more tests we can do.

state-chain/pallets/cf-ingress-egress/src/tests.rs
kylezs: yeah you're right 👍

state-chain/pallets/cf-witnesser/src/lib.rs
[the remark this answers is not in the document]
kylezs: That's true, yeah let's keep it in that case

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -692,36 +738,19 @@ pub mod pallet {
 
 		/// Called when funds have been deposited into the given address.
 		///
-		/// Requires `EnsureWitnessed` origin.
+		/// Requires `EnsurePrewitnessed` or `EnsureWitnessed` origin.
 		#[pallet::call_index(2)]
 		#[pallet::weight(T::WeightInfo::process_single_deposit().saturating_mul(deposit_witnesses.len() as u64))]
```
kylezs, line 743: We should be able to do benchmark each function individually (`add_boostable_deposits` and `process_deposit_witnesses` ) and then add the condition of the origin inside this weight. e.g. you can do something like:
```rust
#[pallet::weight(if T::EnsurePrewitnessed::ensure_origin(origin.clone()).is_ok() {
			T::WeightInfo::process_boostable_deposits(deposit_witnesses.len() as u32)
		} else {
			T::WeightInfo::process_deposits_witnesses(deposit_witnesses.len() as u32)
		})]
```

kylezs: (could still use the `.saturating_mul` method too - main point is just that you can use the condition and benchmark each of the functions separately)

kylezs: RIP, seems like a limitation of substrate, will create an issue

kylezs: Think best we can do then is just use the more expensive direction here

kylezs: Issue already exists... and is 3.5 years old :/ https://github.com/paritytech/polkadot-sdk/issues/355

j4m1ef0rd: Dang, I guess that the `process_deposit_witnesses` is the more expensive one, so ill just leave it as that for now.

state-chain/pallets/cf-ingress-egress/src/benchmarking.rs
```diff
@@ -145,6 +159,26 @@ mod benchmarks {
...
+					asset: BenchmarkValue::benchmark_value(),
+					amount: BenchmarkValue::benchmark_value(),
+					deposit_address: BenchmarkValue::benchmark_value(),
+				},
+			);
+		}
+
+		#[block]
+		{
+			let _old_deposits = PrewitnessedDeposits::<T, I>::clear_prefix(0, u32::MAX, None);
```
kylezs, line 178: Better to pull the benchmarked part into a function rather than effectively duping this here and where it's actually called in the running code

kylezs: and then we can add a simple check that they were removed too

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -590,11 +621,27 @@ pub mod pallet {
 							details.deposit_channel.channel_id,
 							DepositChannel { state, ..details.deposit_channel },
 						);
+						used_weight = used_weight.saturating_add(
+							frame_support::weights::constants::RocksDbWeight::get()
+								.reads_writes(1, 0),
```
kylezs, line 626: other way around? (0, 1) as we inserted, which is a write

kylezs:
Be good if @msgmaxim can review this too - as he'll be building off it, make sure the assumptions are aligned

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -566,12 +590,15 @@ pub mod pallet {
 	impl<T: Config<I>, I: 'static> Hooks<BlockNumberFor<T>> for Pallet<T, I> {
 		/// Recycle addresses if we can
 		fn on_idle(_n: BlockNumberFor<T>, remaining_weight: Weight) -> Weight {
-			let read_write_weight =
-				frame_support::weights::constants::RocksDbWeight::get().reads_writes(1, 1);
+			let mut used_weight = Weight::zero();
+
+			let cleanup_weight = frame_support::weights::constants::RocksDbWeight::get()
```
msgmaxim, line 595: Should this be `recycle_weigth_per_block`? Also, maybe we could comments where `reads_writes(2, 2)` comes from?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -582,6 +609,10 @@ pub mod pallet {
 					T::ChainTracking::get_block_height(),
 				)
 			});
+			used_weight = used_weight.saturating_add(
```
msgmaxim, line 612: Does reading `n` elements really counts as `n` reads? I'm sure, but I would imagine this is more to do with how many lookups by key we do (1 in this case) rather than the size of the data read. (If we do need to count elements, should this be `can_recycle` elements written, or the number of elements left?)

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1035,6 +1065,56 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		deposit_witnesses: Vec<DepositWitness<T::TargetChain>>,
+	) -> DispatchResult {
+		for DepositWitness { deposit_address, asset, amount, .. } in deposit_witnesses {
+			let id = PrewitnessedDepositIdCounter::<T, I>::mutate(|id| -> u64 {
+				*id = id.saturating_add(1);
+				*id
+			});
+
+			let deposit_channel_details =
+				DepositChannelLookup::<T, I>::get(deposit_address.clone())
```
msgmaxim, line 1106: You can use a reference here.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -590,11 +621,27 @@ pub mod pallet {
 							details.deposit_channel.channel_id,
 							DepositChannel { state, ..details.deposit_channel },
 						);
+						used_weight = used_weight.saturating_add(
+							frame_support::weights::constants::RocksDbWeight::get()
+								.reads_writes(1, 0),
+						);
 					}
+					// Cleanup the deposit data
```
msgmaxim, line 629: Does having items here indicate that a deposit was pre-witnessed, but never fully witnessed? I wonder if we want to log this somehow.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1576,3 +1566,113 @@ fn can_update_multiple_items_at_once() {
...
+		let deposit_witnesses = vec![deposit_detail.clone()];
+		assert_ok!(IngressEgress::add_prewitnessed_deposits(deposit_witnesses.clone(),));
+
+		// Check that the deposit is stored in the storage
+		let prewitnessed_deposit_id = PrewitnessedDepositIdCounter::<Test>::get();
+		let channel_id = ChannelIdCounter::<Test>::get();
+		assert_eq!(
+			PrewitnessedDeposits::<Test>::get(channel_id, prewitnessed_deposit_id),
+			Some(PrewitnessedDeposit { asset: ASSET, amount: DEPOSIT_AMOUNT, deposit_address })
+		);
```
msgmaxim, line 1629: Is the test until here exactly copies `handle_prewitness_deposit`? If so, maybe the first test is redundant? Alternatively, maybe you could extract some of the common code into a helper function?

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1576,3 +1566,113 @@ fn can_update_multiple_items_at_once() {
...
+
+		// Fast forward the block height to the recycle block of the created deposit channel
+		let expiry_block = IngressEgress::expiry_and_recycle_block_height().2;
+		BlockHeightProvider::<MockEthereum>::set_block_height(expiry_block);
+
+		// Run the cleanup
+		IngressEgress::on_idle(1, Weight::MAX);
```
msgmaxim, line 1636: Looks like `1` isn't used in `on_idle`, but it is meant to be block height, which you are setting to `expiry_block`, so I think it would be more correct to use `expiry_block` here too.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1576,3 +1566,113 @@ fn can_update_multiple_items_at_once() {
...
+
+		// Submit the prewitnessed deposit twice
+		let deposit_detail: DepositWitness<Ethereum> = DepositWitness::<Ethereum> {
+			deposit_address,
+			asset: ASSET,
+			amount: DEPOSIT_AMOUNT,
+			deposit_details: (),
+		};
+		let deposit_witnesses = vec![deposit_detail.clone()];
+		assert_ok!(IngressEgress::add_prewitnessed_deposits(deposit_witnesses.clone()));
```
msgmaxim, line 1662: We could cover more logic if we use deposit with different amounts and check that the correct one gets removed.

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: This `used_weight` addition is not just for the above mutate, it is for the for loop below, thats why it uses `n` read/writes. Ill add a comment.

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: Yes it does. I think it might only matter if it was boosted.

state-chain/pallets/cf-ingress-egress/src/tests.rs
j4m1ef0rd: `expiry_block` is the eth block, but the `1` is the SC block. So it does not need to be the same value. I will set it to default to show that it does not matter for the test.

msgmaxim (approved):

PR #4497, opened 2024-02-07
feat: expose ingress egress env fees in asset amounts

msgmaxim:
# Pull Request

Closes: PRO-1118

## Checklist

## Summary

- Added `estimate_swap_input_for_desired_output` to `AssetConverter`, which similar to `convert_asset_to_approximate_output` uses `swap_with_network_fee` wrapped into `TransactionOutcome::Rollback` to simulate a swap.
- As discussed, because we don't know input amount, we simply swap from output to input to get an approximation of what the input should have been.
- Because the (simulated) swap can fail, I chose to return Option<Amount> instead of Amount. The `IngressEgressEnvironment` has been changed accordingly. Note that Some(Amount) seems to be encoded the same way as Amount, however for the None case `null` is now also a possible value.
- Note that I am reusing the existing fields inside `IngressEgressEnvironment`, even though they should be interpreted differently now. Let me know if this is a problem, and maybe we could create new fields to make the transition more smooth for the product team.


dandanlen (approved):

dandanlen:
@niklasnatter can you please check if this is ok for you (fees are now optional if we can't convert the asset)?

dandanlen:
I think we should add a `None` output to the test cases btw.

msgmaxim:
> I think we should add a `None` output to the test cases btw.

Added to `test_environment_serialization` test.

niklasnatter (approved):
looks good from my side 🙂

dandanlen (approved):

PR #4503, opened 2024-02-08
fix: continuous adapter (PRO-684)

AlastairHolmes:
Closes: PRO-684

I finally worked out what problem this ticket was actually about. Its not really a significant bug (Explained in the issue).

Of course I feel the loop_select! macro is probably needs some work, as its clear from this and some of the other uses it's design doesn't work nicely in these complex situations, but it is still better than the tokio::select! (As what I do here, isn't possible *simply* with that macro).

We should review this together, and there is zero urgency to merge this.

### DIFF
--- a/engine/src/witness/common/chunked_chain_source/chunked_by_vault/continuous.rs
+++ b/engine/src/witness/common/chunked_chain_source/chunked_by_vault/continuous.rs
@@ -122,6 +122,7 @@ where
 
 									break Some((header, (chain_stream, chain_client, epoch, unprocessed_indices, inprogress_indices, processed_indices)))
 								},
+								// Allows the stream to exit while waiting for blocks, if the epoch becomes historic
 								if let true = epoch.historic_signal.clone().wait().map(|(_, historic_at, _)| is_epoch_complete(&processed_indices, historic_at)) => {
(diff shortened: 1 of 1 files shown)


engine/src/witness/common/chunked_chain_source/chunked_by_vault/continuous.rs
```diff
@@ -113,12 +122,9 @@ where
 
 									break Some((header, (chain_stream, chain_client, epoch, unprocessed_indices, inprogress_indices, processed_indices)))
 								},
-								if epoch.historic_signal.get().is_some() && processed_indices.is_superset(&{
-									let mut bitmap = RleBitmap::new(true);
-									bitmap.set_range(..epoch.info.1, false);
-									bitmap.set_range(epoch.historic_signal.get().unwrap().1.., false);
-									bitmap
-								}) => break None,
+								if let true = epoch.historic_signal.clone().wait().map(|(_, historic_at, _)| is_epoch_complete(&processed_indices, historic_at)) => {
```
kylezs, line 125: as discussed, just including a comment here explaining why we do this / what it's solving

kylezs (approved):

PR #4504, opened 2024-02-08
feat: bump substrate deps to polkadot-sdk 1.6

dandanlen:
# Pull Request

Closes: PRO-1129

## Checklist

## Summary

- Updates our substrate dependency to the 1.6 polkadot release. 
- Bump the rust compiler version.
- Fix some new clippy lints.
- Cargo-updates our dependencies.
- Resolve some cargo-audit issues.
- Update cryptography libs to latest version now that they are compatible with substrate. (@msgmaxim please take a look: [6aa9e2a](https://github.com/chainflip-io/chainflip-backend/pull/4504/commits/6aa9e2a933a287f63c599c1f72d173b3870f9e64))

Most notable substrate changes since 1.1:
- Genesis configs are now serializable as JSON.
- BTreeSet no longer supports StorageDecodeLength (now uses the wonderfully named `StorageDecodeNonDedupLength`).
- There's a `derive_impl` macro that makes it easier to 'inherit' runtime implementations from a base config (used throughout the mocks).
- The Imbalances trait had a new method added.
- All accounts need to have a non-zero reference counter. This includes governance accounts (see [059ccea](https://github.com/chainflip-io/chainflip-backend/pull/4504/commits/059ccea4dd3b59a74d08b903c35fe579bfd4d222))

The only thing I consciously ignored was this change: https://github.com/paritytech/polkadot-sdk/pull/2390

I doesn't really apply to us - we don't have any lints configured at the workspace level. I opened a Linear issue to collect our scattered linting config in a workspace-level file but I don't think it's high priority. (PRO-1117)


dandanlen:
Converting back to draft - looks like I need to update something in CI.

msgmaxim:
> Update cryptography libs to latest version now that they are compatible with substrate. (@msgmaxim please take a look: https://github.com/chainflip-io/chainflip-backend/commit/d90263667e9a7b07d3e8dac45fe8dd65f839bc97)

This part looks good to me 👍

bouncer/tests/broker_fee_collection_test.ts
```diff
@@ -189,14 +189,18 @@ async function testBrokerFees(asset: Asset, seed?: string): Promise<void> {
   const balanceBeforeWithdrawalBigInt = BigInt(
     amountToFineAmount(balanceBeforeWithdrawal, assetDecimals[asset]),
   );
-  const minExpectedBalanceAfterWithdrawal =
-    balanceBeforeWithdrawalBigInt + earnedBrokerFeesAfter - maxWithdrawalFee[asset];
   const detectWithdrawalGasFee =
     balanceBeforeWithdrawalBigInt + earnedBrokerFeesAfter - balanceAfterWithdrawalBigInt;
+  if (['FLIP', 'DOT', 'USDC'].includes(asset.toString())) {
```
AlastairHolmes, line 194: ?

AlastairHolmes: Not clear to me how this related to the polkadot version

engine/multisig/src/client/common.rs
```diff
@@ -34,6 +34,7 @@ pub type PublicKey<C> = <<C as ChainSigning>::CryptoScheme as CryptoScheme>::Pub
 pub type SigningPayload<C> = <<C as ChainSigning>::CryptoScheme as CryptoScheme>::SigningPayload;
 pub type Signature<C> = <<C as ChainSigning>::CryptoScheme as CryptoScheme>::Signature;
 
+#[allow(clippy::manual_non_exhaustive)]
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
```
AlastairHolmes, line 38: Instead of think it would better to actually use the attribute, as the lint suggests: https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute

engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
```diff
@@ -566,6 +566,7 @@ impl<'a, 'env, BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static>
 								matching_request.until_in_block_sender
 							{
 								let _result = until_in_block_sender.send(
+									#[allow(clippy::useless_asref)]
 									result.as_ref().map(Clone::clone).map_err(
```
AlastairHolmes, line 570: I assume you tried removing the as_ref()?

martin-chainflip (approved):

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -913,7 +913,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 	pub fn attempt_count(broadcast_id: BroadcastId) -> AttemptCount {
 		// NOTE: decode_len is correct here only as long as we *don't* use `append` to insert items.
```
AlastairHolmes, line 915: Guessing this comment is out of date?

bouncer/tests/broker_fee_collection_test.ts
dandanlen: It's a typo, should be flip, eth, usdc... I added this to debug a bouncer issue, I'll fix the typo. 

The bouncer issue is that the fee broker fee estimation can return different values depending on which order they run in.

state-chain/pallets/cf-governance/src/migrations.rs
```diff
@@ -1,5 +1,9 @@
 use crate::Pallet;
+use cf_runtime_upgrade_utilities::{migration_template, VersionedMigration};
 
-use cf_runtime_upgrade_utilities::{migration_template::Migration, VersionedMigration};
+pub mod v2;
 
-pub type PalletMigration<T> = VersionedMigration<Pallet<T>, Migration<T>, 0, 1>;
+pub type PalletMigration<T> = (
+	VersionedMigration<Pallet<T>, v2::Migration<T>, 1, 2>,
+	VersionedMigration<Pallet<T>, migration_template::Migration<T>, 2, 3>,
```
AlastairHolmes, line 8: Seems wrong?

AlastairHolmes: `VersionedMigration<Pallet<T>, migration_template::Migration<T>, 2, 3>,`

engine/multisig/src/client/common.rs
dandanlen: I don't mind changing it but my thinking was that the purpose of the attribute is not relevant to what is being done here, so it felt better to keep the comment but ignore the lint.

The comment here says it's so that you can't use the struct literal to construct the value. 

The attribute is related to public apis and the assumption that the exposed values won't change.

state-chain/runtime-utilities/macros/Cargo.toml
```diff
@@ -6,8 +6,11 @@ edition = '2021'
...
+[lints]
+workspace = true
+
 [dependencies]
 # proc-macro-crate = '0.1.5'
 # proc-macro-error = '1.0.4'
 quote = '1.0'
-syn = '1.0'
\ No newline at end of file
+syn = '1.0'
```
AlastairHolmes, line 16: Having an extra empty line at the end of files in nice, because it makes diffs nicer, when you add new lines.

engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
dandanlen: Yep.

dandanlen: I guess I can add a comment.

state-chain/pallets/cf-governance/src/migrations.rs
dandanlen: It's intentional. The migration is required to increment the reference counter for governance accounts. If you mean the presence of the migration_template, this is just a convention, makes it more convenient to add a new migration.

utilities/src/with_std/spmc.rs
```diff
@@ -61,6 +61,7 @@ impl<T> Sender<T> {
 	}
 }
 
+#[allow(dead_code)]
```
AlastairHolmes, line 64: ?

engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
AlastairHolmes: no need

engine/multisig/src/client/common.rs
AlastairHolmes: I feel it does what we want, but I'm happy for you to leave it.

AlastairHolmes (approved):

state-chain/pallets/cf-governance/src/migrations.rs
AlastairHolmes: Like the Pallet version is 2 in the lib.rs, but we have what I assume is a blank migration to version 3?

utilities/src/with_std/spmc.rs
dandanlen: I meant to ask you about this - there is a `tokio::sync::watch::Receiver<()>` that is isn't explicitly used, so clippy flags it as dead code.

But I think it needs to remain there because the other end of the channel waits for it to be dropped (there's a doc comment a few lines above).

state-chain/runtime-utilities/macros/Cargo.toml
dandanlen: Not actually sure how this was added 😅 
Presumably some new cargo fmt feature?

state-chain/pallets/cf-governance/src/migrations.rs
dandanlen: Yes - actually it's an `unimplemented!()` migration that won't be applied (because the pallet version is 2). And if you accidentally do use it, the upgrade checks will panic and tell you off. 

It means that when you delete an old migration, you can leave the template as a placeholder. Makes the diff nicer and makes it easier to add the 'next' migration.

utilities/src/with_std/spmc.rs
AlastairHolmes: Ill try a look after the meeting

PR #4507, opened 2024-02-12
feat: Punish nodes that do not witness in time

syan095:
# Pull Request

Closes: PRO-1004

## Checklist

## Summary

When witnessed calls are dispatched, a deadline is set for remaining authorities to witness. 
Once the deadline is up, any nodes that failed to witness the all are reported to be punished


kylezs:
assigned myself as reviewer just to check there's no impact to how the CFE is witnessing

dandanlen (approved):
Good work @syan095 

I made some changes to the implementation to avoid re-reading storage and building the vote count lookup for every loop iteration. 

If you're happy with this, please merge.

dandanlen:
Btw. we will have to remember to set a penalty for this when we actually roll it out.

kylezs:
Looks like nothing touching CFE side 👍

kylezs:
> Btw. we will have to remember to set a penalty for this when we actually roll it out.

Does this mean we won't see any penalties on localnet and the like?

PR #4512, opened 2024-02-12
feat: charge a fee for opening swap deposit addresses

dandanlen:
# Pull Request

Closes: PRO-1134

## Checklist

## Summary

Charges a fee for opening swap channels. 

Currently the fee is zero, can be set using the new governance extrinsic `update_pallet_config`. 

Bonus feature: it's now possible to do multiple updates at once (eg. update all the swap limits at once).


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -363,6 +384,9 @@ pub mod pallet {
 			amount: AssetAmount,
 			reason: DispatchError,
 		},
+		ChannelOpeningFeeSet {
+			fee: T::Amount,
```
kylezs, line 388: here we call it fee, but in the config we call it Flipperinos, could make it consistent

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2128,3 +2142,65 @@ fn deposit_address_ready_event_contain_correct_boost_fee_value() {
...
+				PalletConfigUpdate::ChannelOpeningFee { flipperinos: 100 },
+				PalletConfigUpdate::MaximumSwapAmount { asset: Asset::Btc, amount: Some(100) },
+				PalletConfigUpdate::MaximumSwapAmount { asset: Asset::Dot, amount: Some(200) },
+			]
+			.try_into()
+			.unwrap()
+		));
+		assert_eq!(ChannelOpeningFee::<Test>::get(), 100);
+		assert_eq!(MaximumSwapAmount::<Test>::get(Asset::Btc), Some(100));
+		assert_eq!(MaximumSwapAmount::<Test>::get(Asset::Dot), Some(200));
```
kylezs, line 2204: we should assert they're different to this before doing the update - else the function could be a noop and this test still pass

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -267,6 +283,10 @@ pub mod pallet {
 	#[pallet::getter(fn maximum_swap_amount)]
 	pub type MaximumSwapAmount<T: Config> = StorageMap<_, Twox64Concat, Asset, AssetAmount>;
 
+	/// The fixed fee charged for opening a channel, in Flipperinos.
+	#[pallet::storage]
+	pub type ChannelOpeningFee<T: Config> = StorageValue<_, T::Amount, ValueQuery>;
```
kylezs, line 288: rather than require another round of governance actions (which we already have to do for the upgrade), we could have a migration to set this to some const

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: I figured when you're setting it in the governance extrinsic, it's useful to know that it needs to be set in flipperinos. If you see an event with fee: 1,000,000,000,000, it's already obvious that it's not measured in flip.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: We could but I didn't want to unilaterally decide on a value - might require some discussion.

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: would be the first time exposing `flipperinos`, currently we only refer to it internally - not necessarily against it, but might not be the clearest to external readers

dandanlen: I've changed it to fee - the doc comment mentions the denomination.

dandanlen:
Kyle mentioned that it might make sense to apply the fee to all channels (swap *and* lp), and i agree. So i moved it into the ingress/egress pallet. 

I also moved the fee to the rpcs instead of returning it in the apis.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -505,6 +533,12 @@ pub mod pallet {
 		UtxoConsolidation {
 			broadcast_id: BroadcastId,
 		},
+		ChannelOpeningFeePaid {
```
kylezs, line 536: We should include who opened the channel (and therefore who paid the fee). Perhaps it makes sense to make this just a `ChannelOpened` event and include who opened it, the asset, and the action?

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1474,3 +1488,67 @@ fn consolidation_tx_gets_broadcasted_on_finalize() {
...
+	new_test_ext().execute_with(|| {
+		const CHANNEL_REQUESTER: u64 = 789;
+		const FEE: u128 = 100;
+		MockFundingInfo::<Test>::credit_funds(&CHANNEL_REQUESTER, FEE);
+		assert_eq!(MockFundingInfo::<Test>::total_balance_of(&CHANNEL_REQUESTER), FEE);
+		assert_ok!(IngressEgress::update_pallet_config(
+			OriginTrait::root(),
+			vec![PalletConfigUpdate::ChannelOpeningFee { fee: FEE }].try_into().unwrap()
+		));
+		assert_ok!(IngressEgress::open_channel(
```
kylezs, line 1503: we can also check the event is emitted

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2128,3 +2140,22 @@ fn deposit_address_ready_event_contain_correct_boost_fee_value() {
 		);
 	});
 }
+
+#[test]
+fn can_update_multiple_items_at_once() {
```
kylezs, line 2145: nit: we're only updating one at once here now

dandanlen: It's updating the max swap amounts for btc and dot - that's two?

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: I wanted to avoid duplicating too much info... this is always emitted in the context of a swap or lp channel, and both of these emit their own events. What's the benefit of adding the extra info here?

kylezs (approved):

kylezs:
@acdibble some environment RPC snapshots changing here

dandanlen:
Sorry @acdibble I thought you would be auto-tagged for the review. The change is an additional item `channel_opening_fees` in the ingress egress environment return value.

PR #4516, opened 2024-02-13
fix: downgrade upload/download artifact

tomjohnburton:


tomjohnburton:
https://github.com/chainflip-io/chainflip-backend/actions/runs/7876151080/job/21509355965#step:10:40

tomjohnburton:
No we want the old behaviour which still exists in version 3
The new behaviour requires many more changes
I raised a ticket here
https://linear.app/chainflip/issue/PLA-1057/review-google-workspace-groups

PR #4525, opened 2024-02-14
Feat: Scheduled Swaps Subscription

msgmaxim:
# Pull Request

Closes: PRO-1140

## Checklist

## Summary

Took me a bit longer than it should have... I initially thought we should need a stream of finalized stream, and I wrote some code to deal with that, only to realise that with a finalized stream a 2 block swap delay wouldn't be long enough. So here I'm exposing swaps from unfinalised blocks using existing `new_subscription`.

I also ran into an issue after merging recent changes in main. Took me some time to realise this, but depending on what we use as T in `new_subscription`, we could get a silent failure when `pipe_from_try_stream` was failing to encode `BlockUpdate` internally (due to `serde(flatten)`), and this wasn't closing the subscription either. I now added error handling around `pipe_from_try_stream` to make it easier detect failures like this in the future.

Example usage:

```
wscat -c ws://localhost:9944
> {"id":1, "jsonrpc":"2.0", "method": "cf_subscribe_scheduled_swaps", "params": ["FLIP", "BTC"]}
```

This would return a notification on every block, even when the list of swaps is empty:

```
{"jsonrpc":"2.0","method":"cf_subscribe_scheduled_swaps","params":{"subscription":"SgUMJq00PQ7tetsx","result":{"block_hash":"0xdb82a90439be5826f8dece231d35ee8b23c9baaf98839aa345fc188bdee64423","block_number":58,"swaps":[{"swap_id":1,"from":"BTC","to":"FLIP","amount":4949884,"swap_type":{"Swap":{"Eth":"0x41ad2bc63a2059f9b623533d87fe99887d794847"}},"stable_amount":null,"final_output":null,"fee_taken":false,"execute_at":60}]}}}
< {"jsonrpc":"2.0","method":"cf_subscribe_scheduled_swaps","params":{"subscription":"SgUMJq00PQ7tetsx","result":{"block_hash":"0x4f70862ca1544f5add9c61598a0e1875e9ca3dd9637e243f0cfb68cefdb7db3a","block_number":59,"swaps":[{"swap_id":1,"from":"BTC","to":"FLIP","amount":4949884,"swap_type":{"Swap":{"Eth":"0x41ad2bc63a2059f9b623533d87fe99887d794847"}},"stable_amount":null,"final_output":null,"fee_taken":false,"execute_at":60}]}}}
< {"jsonrpc":"2.0","method":"cf_subscribe_scheduled_swaps","params":{"subscription":"SgUMJq00PQ7tetsx","result":{"block_hash":"0xc5737b08cf7133c1fdb88f7932f1b8c67945e216c9e5101988323ad1e627f2af","block_number":60,"swaps":[]}}}
```

Note how the same swap is returned twice (i.e. in two notifications) in this case. This is becasue we return all swaps that are "pending", and it takes 2 blocks untils swaps get executed.


state-chain/runtime/src/lib.rs
```diff
@@ -1408,6 +1408,25 @@ impl_runtime_apis! {
 			all_prewitnessed_swaps
 		}
 
+		fn cf_scheduled_swaps(from: Asset, to: Asset) -> Vec<(Swap, BlockNumber)> {
```
kylezs, line 1411: given we don't actually care about direction here, perhaps naming these just `asset1` and `asset2` would make this a bit clearer

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1212,6 +1223,30 @@ where
 		)
 	}
 
+	fn cf_subscribe_scheduled_swaps(
```
kylezs, line 1226: doc comment would be helpful here, in particular explaining that the asset order doesn't matter here, it'll get swaps in both directions for that pair

state-chain/runtime/src/lib.rs
```diff
@@ -1408,6 +1408,25 @@ impl_runtime_apis! {
...
+			let mut swaps = vec![];
+
+			while first_block <= last_block {
+				let swaps_selected_from_block = Swapping::swap_queue(first_block).into_iter()
+					.filter(|swap| { (swap.from == from && swap.to == to) || (swap.from == to && swap.to == from)})
+					.map(|swap| (swap, first_block));
+
+				swaps.extend(swaps_selected_from_block);
+				first_block += 1u32;
+			}
```
kylezs, line 1425: Can remove the muts and increment: 
```rust
(first_block..=last_block).into_iter().map(|block| {
				Swapping::swap_queue(block).into_iter()
					.filter(|swap| { (swap.from == from && swap.to == to) || (swap.from == to && swap.to == from)})
					.map(move |swap| (swap, block))
			}).flatten().collect()
```

state-chain/runtime/src/lib.rs
```diff
@@ -1408,6 +1408,25 @@ impl_runtime_apis! {
 			all_prewitnessed_swaps
 		}
 
+		fn cf_scheduled_swaps(from: Asset, to: Asset) -> Vec<(Swap, BlockNumber)> {
+
+			let mut first_block = Swapping::first_unprocessed_block();
+			let last_block = System::block_number() + pallet_cf_swapping::SWAP_DELAY_BLOCKS;
```
kylezs, line 1414: perhaps a debug_assert that first_block < last_block

state-chain/custom-rpc/src/lib.rs
```diff
@@ -518,6 +526,9 @@ pub trait CustomApi {
 		side: Order,
 	);
 
+	#[subscription(name = "subscribe_scheduled_swaps", item = Vec<AssetAmount>)]
```
kylezs, line 529: Shouldn't the items here be `Vec<ScheduledSwap>`? else we can't know the direction

kylezs:
Once this is included we should add docs to the docs repo, I added a sub-issue on linear so we don't forget: PRO-1203

kylezs:
Rest looks fine to me. Will let @AlastairHolmes approve as he's been working on these RPCs a lot recently

state-chain/runtime/src/lib.rs
AlastairHolmes, line 1411: These should be consistent with the other interfaces, and therefore called base_asset and quote_asset.

AlastairHolmes: And the ordering should be enforced to match the pool.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -40,12 +40,14 @@ const BASIS_POINTS_PER_MILLION: u32 = 100;
...
 pub enum SwapType {
 	Swap(ForeignChainAddress),
 	CcmPrincipal(SwapId),
 	CcmGas(SwapId),
 }
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
 pub struct Swap {
 	pub swap_id: SwapId,
 	pub from: Asset,
```
AlastairHolmes, line 53: The returned response should use "base_asset", "quote_asset", and "side: Order"

state-chain/runtime/src/lib.rs
AlastairHolmes: Right, I see a problem. I believe what we want (and what was discussed in discord before) from this sub is per pool scheduled swaps, otherwise the sub isn't very useful. Where as my (somewhat brief) reading of this, is this implementation is a sub of end to end scheduled swaps. The problem with that is LPs would have to do not insignificant amount of work to work out what the per pool scheduled swaps are.

kylezs: ok yeah, let's be consistent with the other ones, as per Alastair's comment

AlastairHolmes:
Ignore my comments, I've realised there's been some miss-communication about the intended behaviour. I will write on the ticket (Shortly, not immediately (Lunch)).

state-chain/runtime/src/lib.rs
msgmaxim: Indeed. I had assumed the block number was generic here, and didn't implement Step.

state-chain/custom-rpc/src/lib.rs
msgmaxim: Yeah. I copy pasted from somewhere. Not sure what this actually affects, since the stream does correctly return `ScheduledSwap` items in practice.

kylezs (commented):
Looks good to me, will let @AlastairHolmes hit approve

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -57,6 +59,16 @@ pub struct Swap {
 	pub fee_taken: bool,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct SwapLegInfo {
+	pub swap_id: SwapId,
+	pub from: Asset,
```
AlastairHolmes, line 66: Instead of `from` and `to`, we should use `base_asset`, `quote_asset`, and `side: Order` (This type is being changed to be called Side fyi, by Jan).

I'd like to do this for consistency with the new rpcs.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -451,6 +465,12 @@ pub trait CustomApi {
 	#[subscription(name = "subscribe_prewitness_swaps", item = BlockUpdate<RpcPrewitnessedSwap>)]
 	fn cf_subscribe_prewitness_swaps(&self, base_asset: Asset, quote_asset: Asset, side: Order);
 
+	// Subscribe to a stream that on every block produces a list of all scheduled/pending
+	// swaps in the base_asset/quote_asset pool, including any "implicit" half-swaps (as a
+	// part of a swap involving two pools)
+	#[subscription(name = "subscribe_scheduled_swaps", item = SwapResponse)]
+	fn cf_subscribe_scheduled_swaps(&self, base_asset: Asset, quote_asset: Asset);
```
AlastairHolmes, line 472: I suggest we add a cf_scheduled_swaps(base_asset, quote_asset, at) rpc too

AlastairHolmes: Was a request of LPs, helps them now what as happened in the past

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1121,6 +1141,28 @@ where
...
+		base_asset: Asset,
+		quote_asset: Asset,
+	) -> Result<(), SubscriptionEmptyError> {
+		self.new_subscription(
+			false, /* only_on_changes */
+			true,  /* end_on_error */
+			sink,
+			move |api, hash| {
+				let swaps = api
+					.cf_scheduled_swaps(hash, base_asset, quote_asset)?
```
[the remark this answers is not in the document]
AlastairHolmes: Also then you could just delete `SwapLegInfo` and inline it into The ScheduledSwap structure, i.e. avoiding the `serde(flatten)`.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -57,6 +59,16 @@ pub struct Swap {
...
+	pub to: Asset,
+	pub amount: AssetAmount,
+	pub swap_type: SwapType,
```
AlastairHolmes, line 69: Had a discussion with Kyle and Dan, we agreed that we don't want to expose the SwapType.

Also this means you don't need the serde derives on the SwapType enum.

state-chain/custom-rpc/src/lib.rs
AlastairHolmes, line 472: This rpc call doesn't seem to have any checks on quote_asset? It should atleast fail, if it is not USDC, ideally I'd suggest just checking if the pool exists in `cf_scheduled_swaps`, and returning an error if it doesn't (In the pools pallet there is an error for this)

state-chain/runtime/src/lib.rs
```diff
@@ -1423,6 +1423,23 @@ impl_runtime_apis! {
...
 		}
 
+		fn cf_scheduled_swaps(base_asset: Asset, _quote_asset: Asset) -> Vec<(SwapLegInfo, BlockNumber)> {
+
+			let first_block = Swapping::first_unprocessed_block();
+			let last_block = System::block_number() + pallet_cf_swapping::SWAP_DELAY_BLOCKS;
+
+			debug_assert!(first_block < last_block);
+
+			(first_block..=last_block).flat_map(|block| {
```
AlastairHolmes, line 1433: You could just read the entire SwapQueue map into memory (Instead of one element at a time), then you can make this code much simpler.

state-chain/runtime/src/lib.rs
```diff
@@ -1423,6 +1423,23 @@ impl_runtime_apis! {
...
+				let swaps_for_block = Swapping::swap_queue(block);
+
+				let swaps: Vec<_> = swaps_for_block.iter().filter(|swap| swap.from == base_asset || swap.to == base_asset).cloned().collect();
+
+				Swapping::get_scheduled_swap_legs(swaps, base_asset).unwrap().into_iter().map(move |swap| (swap, block))
+			}).collect()
+
```
AlastairHolmes, line 1440:
```suggestion
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -241,11 +253,13 @@ pub mod pallet {
 
 	/// Scheduled Swaps
 	#[pallet::storage]
+	#[pallet::getter(fn swap_queue)]
 	pub(crate) type SwapQueue<T: Config> =
 		StorageMap<_, Twox64Concat, BlockNumberFor<T>, Vec<Swap>, ValueQuery>;
 
 	/// The first block for which swaps haven't yet been processed
 	#[pallet::storage]
+	#[pallet::getter(fn first_unprocessed_block)]
```
AlastairHolmes, line 262: I'm not sure there is much point in using these getters really, particularly as you cannot update through a similarly named setter?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -658,17 +672,47 @@ pub mod pallet {
...
+
+			Ok(swaps
+				.into_iter()
+				.filter_map(|swap| {
+					if swap.from == base_asset {
+						Some(SwapLegInfo {
+							swap_id: swap.swap_id,
+							from: swap.from,
+							// All swaps from `base_asset` have to go through Usdc:
+							to: Asset::Usdc,
```
AlastairHolmes, line 691: Can you use the STABLE_ASSET constant? Same below.

AlastairHolmes (commented):
Once all these are addressed we can merge.

state-chain/custom-rpc/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: I changed `cf_scheduled_swaps` to return `Vec<ScheduledSwap>` now, and simplified the code a bit.

However, I'm not sure merging the two structs is obviously justified. The reason I have separate structs is bacause `get_scheduled_swap_legs` doesn't deal with block numbers, so we'd need to pass the block number through that function just so it can construct `ScheduledSwap` on our behalf, which doesn't seem particularly nice. I would aslo need to add a generic parameter (for the block number) to `ScheduledSwap` to make it work. I don't think having this extra struct is much worse than the alternative. I can make this change if you have a strong preference though.

state-chain/custom-rpc/src/lib.rs
msgmaxim: I reused `cf_pool_info` for this (creating a separate method returning true/false didn't seem worth it)

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: I would say it is cleaner than writing `pallet_cf_swapping::FirstUnprocessedBlock::<Runtime>::get()`. I also don't think every getter needs a setter (in fact, it is a common incapsulation pattern to expose a value through a getter only so that it can only be modified internally).

AlastairHolmes: "(in fact, it is a common incapsulation pattern to expose a value through a getter only so that it can only be modified internally)", yeah but that isn't hwat you are doing here.

"IMO it is a bit cleaner than writing pallet_cf_swapping::FirstUnprocessedBlock::<Runtime>::get()" I disagree as naming the storage item you are reading is inherently clearer than using some other name for it (When that other name isn't consistently used elsewhere), and I believe there are basically zero places we use these getters substrate generates elsewhere in the codebase.

AlastairHolmes: If we do what I suggest inregards to the code that uses this getter, we wouldn't even need to read this storage item anyway.

state-chain/custom-rpc/src/lib.rs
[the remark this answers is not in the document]
AlastairHolmes: You should just inline get_scheduled_swap_legs, the fact that it causes you to need two struct's here (that aren't used separate other than in this one function), imo is clear indication the split isn't a good idea.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -40,12 +40,14 @@ const BASIS_POINTS_PER_MILLION: u32 = 100;
...
 
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
 pub enum SwapType {
 	Swap(ForeignChainAddress),
 	CcmPrincipal(SwapId),
 	CcmGas(SwapId),
 }
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
```
AlastairHolmes, line 50: Sorry forgot to mention this. This serde derive is not used, and so can be removed.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1121,6 +1141,28 @@ where
 		)
 	}
 
+	fn cf_subscribe_scheduled_swaps(
+		&self,
+		sink: SubscriptionSink,
+		base_asset: Asset,
+		quote_asset: Asset,
+	) -> Result<(), SubscriptionEmptyError> {
```
AlastairHolmes, line 1149: I've realised there is something else we need to deal with, swaps can fail will limited liquidity, so the simulation of the first leg can final. I think what we should do is if the simulation fails, don't output the second leg as scheduled, but do output the first leg as being scheduled.

kylezs: In this case if the liquidity gets added for the first leg, since the LPs can see that it requires liquidity, then it'll be swapped, but then the no one has accounted for the second leg? that seems bad? Would it not be better to use some kind approximate price for the first leg, to the estimate the second leg input - and perhaps making it clear through the API that this second amount is an estimate or something

state-chain/custom-rpc/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: The advantage of having a separate `get_scheduled_swap_legs` is that we can write unit tests against it. It is possible to re-write tests with it inlined, but that wouldn't be as nice. I don't think having a separate struct for what the RPC call returns is such a bad thing, particularly since we wrap the internal type (with `serde(flatten)`), and so we don't even have any duplication.

state-chain/custom-rpc/src/lib.rs
msgmaxim: added

msgmaxim:
Addressed/responded to all ouf your comments now @AlastairHolmes. Let me know is we can merge this (and whether your other comment re failing due to limited liquidity can be addressed separately).

AlastairHolmes:
There is also another bug, the subscription can list swaps as being scheduled in past blocks, we can just bound the scheduled block to after the current block. Then I think this can be merged. This could happen when swaps fail, as swaps aren't rescheduled so their scheduled block doesn't change.

state-chain/runtime/src/lib.rs
```diff
@@ -1427,9 +1427,15 @@ impl_runtime_apis! {
 
 		fn cf_scheduled_swaps(base_asset: Asset, _quote_asset: Asset) -> Vec<ScheduledSwap> {
 
+			let current_block = System::block_number();
+
 			pallet_cf_swapping::SwapQueue::<Runtime>::iter().flat_map(|(block, swaps_for_block)| {
+
+				// In case `block` has already passed, the swaps will be re-tried at the next block:
+				let execute_at = if block > current_block { block } else { current_block.saturating_add(1) };
```
AlastairHolmes, line 1435: I think this will should be `block >= current_block`, because this code is run at the end of the block N, but I assume current_block will still have the value N. But those swaps that remain in the SwapQueue at N, will be executed at N + 1.

AlastairHolmes, line 1435: I think this would be a little easier to read as 
```suggestion
				let execute_at = core::cmp::max(block, current_block.saturating_add(1));
```

AlastairHolmes (approved):

PR #4534, opened 2024-02-14
fix: Wait for ThresholdSignature success before switching to NewKeysActivated

marcellorigotti:
# Pull Request

Closes: PRO-1062

## Checklist

## Summary

- Modified rotation to wait for ThresholdSignature success before switching to NewKeysActivated
- Modified tests to match the new behaviour

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-broadcast/src/lib.rs
+++ b/state-chain/pallets/cf-broadcast/src/lib.rs
@@ -664,7 +664,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		maybe_failed_callback_generator: impl FnOnce(
 			BroadcastId,
 		) -> Option<<T as Config<I>>::BroadcastCallable>,
-	) -> (BroadcastId, u32) {
+	) -> (BroadcastId, ThresholdSignatureRequestId) {
 		let broadcast_id = Self::next_broadcast_id();
 
 		PendingBroadcasts::<T, I>::append(broadcast_id);
@@ -921,7 +921,9 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
 	type ApiCall = T::ApiCall;
 	type Callback = <T as Config<I>>::BroadcastCallable;
 
-	fn threshold_sign_and_broadcast(api_call: Self::ApiCall) -> (BroadcastId, u32) {
+	fn threshold_sign_and_broadcast(
+		api_call: Self::ApiCall,
+	) -> (BroadcastId, ThresholdSignatureRequestId) {
 		Self::threshold_sign_and_broadcast(api_call, None, |_| None)
 	}
 
@@ -948,7 +950,9 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
 		Self::clean_up_broadcast_storage(broadcast_id);
 	}
 
-	fn threshold_sign_and_broadcast_rotation_tx(api_call: Self::ApiCall) -> (BroadcastId, u32) {
+	fn threshold_sign_and_broadcast_rotation_tx(
+		api_call: Self::ApiCall,
+	) -> (BroadcastId, ThresholdSignatureRequestId) {
 		let (broadcast_id, request_id) =
 			<Self as Broadcaster<_>>::threshold_sign_and_broadcast(api_call);
 
--- a/state-chain/pallets/cf-broadcast/src/lib.rs
+++ b/state-chain/pallets/cf-broadcast/src/lib.rs
@@ -664,7 +664,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		maybe_failed_callback_generator: impl FnOnce(
 			BroadcastId,
 		) -> Option<<T as Config<I>>::BroadcastCallable>,
-	) -> BroadcastId {
+	) -> (BroadcastId, u32) {
 		let broadcast_id = Self::next_broadcast_id();
 
 		PendingBroadcasts::<T, I>::append(broadcast_id);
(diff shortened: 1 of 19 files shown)


kylezs:
If you haven't already, I'd run a 3 node localnet and try to break this - running through some rotations, swaps through rotations etc. bouncer only does one rotation which isn't enough to catch a lot of the more likely small issues here

state-chain/traits/src/lib.rs
```diff
@@ -192,7 +192,12 @@ pub trait VaultActivator<C: ChainCrypto> {
 
 	/// Activate key/s on particular chain/s. For example, setting the new key
 	/// on the contract for a smart contract chain.
-	fn activate(new_key: C::AggKey, maybe_old_key: Option<C::AggKey>);
+	fn activate(
+		new_key: C::AggKey,
+		maybe_old_key: Option<C::AggKey>,
+	) -> Option<ThresholdSignatureRequestId>;
+
+	fn activate_key();
```
dandanlen, line 200: Doc comment would be nice.

dandanlen: Maybe rename `activate` -> `start_key_activation`.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -678,7 +678,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		let _threshold_signature_id = Self::threshold_sign(api_call, broadcast_id, true);
 
-		broadcast_id
+		(broadcast_id, _threshold_signature_id)
```
dandanlen, line 681: nit: remove the leading `_` or simply inline the call.

dandanlen (commented):
LGTM apart from some minor comments. 

I think it makes sense to await the approval of @ramizhasan111 before merging.

state-chain/traits/src/lib.rs
marcellorigotti: I added something, not sure how specific I have to be.
Is it ok like this?

state-chain/pallets/cf-threshold-signature/src/mock.rs
```diff
@@ -193,7 +196,12 @@ impl pallet_cf_threshold_signature::Config<Instance1> for Test {
 pub struct MockVaultActivator;
 impl VaultActivator<MockEthereumChainCrypto> for MockVaultActivator {
 	type ValidatorId = <Test as Chainflip>::ValidatorId;
-	fn activate(_new_key: MockAggKey, _maybe_old_key: Option<MockAggKey>) {}
+	fn start_key_activation(
+		_new_key: MockAggKey,
+		_maybe_old_key: Option<MockAggKey>,
+	) -> Option<ThresholdSignatureRequestId> {
+		None
```
kylezs, line 203: This suggests we don't have any unit tests that test the new code path? Since the new code path takes the path when this returns Some(). We should probably add a couple. I ran every code path in my brain to confirm this was the case. 

(I added a panic here and unit tests all pass)
```
						if Signature::<T, I>::get(tss_request_id) == AsyncResult::Void {
							panic!("check");
							T::VaultActivator::activate_key();
						};
```

marcellorigotti: we have an integration test for it, I just tried adding the panic and it correctly fails the test.

I am referring to: integration-tests/src/authorities/authority_rotates_with_the_correct_sequence

kylezs: My thinking is that having at least one unit test to isolate this particular flow means that when code changes and something like an integration test (which is harder to ensure is correct) is changed, nothing is missed, and it's caught in the tests

state-chain/pallets/cf-threshold-signature/src/key_rotator.rs
```diff
@@ -147,14 +147,25 @@ impl<T: Config<I>, I: 'static> KeyRotator for Pallet<T, I> {
 						),
 					},
 				KeyRotationStatusVariant::AwaitingActivation => {
-					let new_public_key = match PendingKeyRotation::<T, I>::get() {
-						Some(KeyRotationStatus::AwaitingActivation { new_public_key }) =>
+					let (new_public_key, request_id) = match PendingKeyRotation::<T, I>::get() {
```
ramizhasan111, line 150: to keep the naming consistent with elsewhere maybe we should rename request_id to maybe_request_id and rename tss_request_id to request_id below.

ramizhasan111 (approved):

state-chain/pallets/cf-threshold-signature/src/key_rotator.rs
```diff
@@ -147,14 +147,25 @@ impl<T: Config<I>, I: 'static> KeyRotator for Pallet<T, I> {
...
+						}) => (new_public_key, request_id),
 						_ => unreachable!(
 							"Unreachable because we are in the branch for the AwaitingActivation variant."
 						),
 					};
 
+					if let Some(tss_request_id) = request_id {
+						// After a tss ceremony completes, it is consumed and Void is left
+						// behind At this point we are sure the ceremony existed and we
+						// completed it we can activate the key
```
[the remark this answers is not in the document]
ramizhasan111: // After a tss ceremony completes, it is consumed and Void is left
						// behind At this point we are sure the ceremony existed and we
						// completed it we can activate the key
```suggestion
						// After the ceremony completes, the signature is consumed and Void is left
						// behind. At this point we are sure the ceremony existed and succeeded.
						//  We can activate the key
```

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -141,6 +141,7 @@ pub enum KeyRotationStatus<T: Config<I>, I: 'static = ()> {
 	},
 	/// We are waiting for the key to be updated on the contract, and witnessed by the network.
 	AwaitingActivation {
+		request_id: Option<RequestId>,
```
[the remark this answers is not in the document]
marcellorigotti: As discussed IRL, a migration is not necessary in this case, since we won't perform an upgrade during a rotation

ramizhasan111:
Please dont merge this PR yet. It will have conflicts with the Arbitrum PR. I will merge this after merging the arbitrum PR.

kylezs:
Discussed with @ramizhasan111 - merging this, as we're gonna keep the arb one open until we include some changes for the next release - including this one :)

PR #4535, opened 2024-02-15
feat: more information added to AllBatchError. (PRO-1171)

syan095:
# Pull Request

Closes: PRO-1171

## Checklist

## Summary

When AllBatch build fails, an event is emitted that contains the error, this way both the pallet and the error details is visible.


state-chain/chains/src/btc/api.rs
```diff
@@ -89,7 +89,7 @@ where
 				.ok_or(AllBatchError::NotRequired)?,
 			number_of_outputs: (btc_outputs.len() + 1) as u64, // +1 for the change output
 		})
-		.ok_or(AllBatchError::Other)?;
+		.ok_or(AllBatchError::InvalidUtxo)?;
```
dandanlen, line 92: I don't think this is an accurate name for this error.

I would call it UtxoSelectionFailed or UnableToSelectUtxos.

state-chain/chains/src/lib.rs
```diff
@@ -341,9 +341,24 @@ pub trait RegisterRedemption: ApiCall<<Ethereum as Chain>::ChainCrypto> {
 	fn amount(&self) -> u128;
 }
 
-#[derive(Debug)]
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo)]
 pub enum AllBatchError {
+	// Empty transaction - the call is not required.
```
dandanlen, line 346: Prefer to use `///` then the doc string will be included in the metadata.

state-chain/chains/src/lib.rs
```diff
@@ -341,9 +341,24 @@ pub trait RegisterRedemption: ApiCall<<Ethereum as Chain>::ChainCrypto> {
...
+	VaultAccountNotSet,
+
+	// The Aggregate key lookup failed
+	AggKeyNotSet,
+
+	// Utxo lookup failed.
+	InvalidUtxo,
+
+	// Other errors.
 	Other,
```
dandanlen, line 362: Is other still used, can it be removed?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -942,9 +945,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 			},
 			Err(AllBatchError::NotRequired) => TransactionOutcome::Commit(Ok(())),
-			Err(AllBatchError::Other) => TransactionOutcome::Rollback(Err(DispatchError::Other(
-				"AllBatch ApiCall creation failed, rolled back storage",
-			))),
+			Err(error) => {
+				Self::deposit_event(Event::<T, I>::FailedToBuildAllBatchCall { error });
+				TransactionOutcome::Rollback(Err(DispatchError::Other(
+					"AllBatch ApiCall creation failed, rolled back storage",
+				)))
```
dandanlen, line 952: This event will not be visible because of the storage rollback.

We will have to return the error from `do_egress_scheduled_fetch_transfer` along with the transaction outcome and then emit it after the storage transaction rollback.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1474,3 +1474,28 @@ fn consolidation_tx_gets_broadcasted_on_finalize() {
...
+
+		// Expect the egress to fail
+		matches!(
+			IngressEgress::do_egress_scheduled_fetch_transfer(),
+			sp_runtime::TransactionOutcome::Rollback(..)
+		);
+
+		System::assert_last_event(RuntimeEvent::IngressEgress(
+			crate::Event::<Test>::FailedToBuildAllBatchCall {
+				error: cf_chains::AllBatchError::Other,
```
dandanlen, line 1497: I think this test passes because calling `do_egress_scheduled_fetch_transfer` outside of a transactional wrapper *doesn't* roll back the storage. 

A better approach might be to refactor `do_egress_scheduled_fetch_transfer` so that it returns a Result and simply decorate the fn with `#[transactional]`.

dandanlen (changes requested):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -983,12 +990,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-				TransactionOutcome::Commit(Ok(()))
+				Ok(())
 			},
-			Err(AllBatchError::NotRequired) => TransactionOutcome::Commit(Ok(())),
-			Err(AllBatchError::Other) => TransactionOutcome::Rollback(Err(DispatchError::Other(
-				"AllBatch ApiCall creation failed, rolled back storage",
-			))),
+			Err(AllBatchError::NotRequired) => Ok(()),
+			Err(error) => Err(DispatchError::Other(
+				format!("Failed to build AllBatch call. Error: {:?}", error).leak(),
```
kylezs, line 997: could we also include the chain information here - don't think that'd be included?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -567,8 +567,11 @@ pub mod pallet {
 		/// Take all scheduled Egress and send them out
 		fn on_finalize(_n: BlockNumberFor<T>) {
 			// Send all fetch/transfer requests as a batch. Revert storage if failed.
-			if let Err(e) = with_transaction(|| Self::do_egress_scheduled_fetch_transfer()) {
+			if let Err(e) = Self::do_egress_scheduled_fetch_transfer() {
 				log::error!("Ingress-Egress failed to send BatchAll. Error: {e:?}");
+				Self::deposit_event(Event::<T, I>::FailedToBuildAllBatchCall {
+					error: format!("{:?}", e),
```
dandanlen, line 607: I'm surprised that we have access to `format!` and `String` in the runtime...?

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: We emit it as an event, so we will know which pallet instance it came from. 

I would prefer not to use `leak` here... better to implement Into<DispatchError> for AllBatchError or explicitly match on the result.

dandanlen (commented):
I'll fix these and then merge. Thanks @syan095

PR #4540, opened 2024-02-16
fix: set_range_order now returns an error if no position is created

syan095:
# Pull Request

Closes: PRO-1063

## Checklist

## Summary

In set_range_order, if the input asset (size) is > 0 but the resultant liquidity change is 0, an error is returned.


syan095:
So this's been hanging here for half a year already. I've tagged Alastair for review but he just ignored it. Maybe we can get this merged in?

dandanlen:
I can review it.

dandanlen (approved):
LGTM - I just changed a variable name and added a comment.

The previous name `liquidity_change` suggested that it was the net change in liquidity, but it is actually the resultant liquidity after the change is applied.

PR #4547, opened 2024-02-19
feat: Relative Slippage Limits (PRO-1207)

AlastairHolmes:
Note: I noticed the benchmarks haven't been run for a while in the pools pallet, and so the schedule_limit_order_update didn't have upto date weights. So that is fixed here too.


dandanlen:
Looks fine to me - some tests would be nice...

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -987,13 +1002,49 @@ impl<T: Config> SwappingApi for Pallet<T> {
...
+						order,
+						input_amount,
+						output_amount,
+					),
+				);
+				let bounded_swap_tick = if tick_after < tick_before {
+					core::cmp::min(core::cmp::max(tick_after, swap_tick), tick_before)
+				} else {
+					core::cmp::min(core::cmp::max(tick_before, swap_tick), tick_after)
+				};
```
dandanlen, line 1034: Are these checks just there to be defensive? How could the swap tick not be bounded by the start and end ticks?

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -962,6 +964,19 @@ pub mod pallet {
 				_ => Err(Error::<T>::UnsupportedCall)?,
 			}
 		}
+
+		/// Sets the maximum percentage increase (in number of ticks) of the price of the brought
+		/// asset during a swap.
```
dandanlen, line 969: nit: I think this merits a better explanation. If it's measured in ticks, it's not really a percentage. It's also not necessarily an increase. Also, I think the price could actually move by more than the maximum expressed by this number (because of the whole max/min thing).

state-chain/pallets/cf-pools/src/lib.rs
AlastairHolmes: Rounding, can cause that. Particularly with small values, i.e. 1 outputing 0, would slip the price infinitely, causing these checks to always fail, this is a defence again that.

This is not just a problem with directly around zero, like 10 outputing 9, slips the price 10%, but real slippage may only be 1%, but we only output 9 becuase of rounding.

state-chain/pallets/cf-pools/src/lib.rs
AlastairHolmes: Ticks have a direct relationship to a percentage (1 to 1 mapping). It is a percentage measured in ticks.

Not sure what you mean by it is not always an increase?

I'll add clarification regarding the fuzzy nature of the maximum.

dandanlen: I thought ticks were multiplicative? They can be mapped to a percentage, but they aren't a representation of a percentage (not like pips/bips for example). 

Re. increase - I meant the price can also decrease but I see you mention it's in terms of the 'bought asset'.

dandanlen (approved):

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -962,6 +964,21 @@ pub mod pallet {
 				_ => Err(Error::<T>::UnsupportedCall)?,
 			}
 		}
+
+		/// Sets the allowed percentage increase (in number of ticks) of the price of the brought
```
martin-chainflip, line 968: should be "bought"

PR #4548, opened 2024-02-20
feat: add Range order pool price to the pool_price_v2 rpc call

syan095:
# Pull Request

Closes: PRO-1050

## Checklist


dandanlen (commented):
Looks fine to me. @AlastairHolmes ?

state-chain/amm/src/lib.rs
```diff
@@ -57,6 +57,14 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
 		})
 	}
 
+	/// Returns the Range Order sub-pool's current price.
+	/// SwapDirection is ignored as the price are the same for both directions.
+	pub fn current_range_order_pool_price(&mut self) -> Option<(Price, SqrtPriceQ64F96, Tick)> {
```
AlastairHolmes, line 62: So we want to just use the value in range_orders::PoolState::current_sqrt_price (which is not Option), because this is for calculating range order values, not for swaps.

syan095: Done!

PR #4549, opened 2024-02-20
chore: arbitrum updates

albert-llimos:
Closes: PRO-1225
Closes: PRO-778

## Checklist

## Summary

Changes and/or updates for Arbitrum. Explanations added as comments to this PR.
Do not take the changes of this PR as-is, we probably want to do some refactoring as explained in the comments. But the Arbitrum specific values should be updated as described here.


engine/src/witness/arb.rs
```diff
@@ -49,26 +49,32 @@ where
 		+ 'static,
 	ProcessingFut: Future<Output = ()> + Send + 'static,
 {
-	let key_manager_address = state_chain_client
```
albert-llimos, line 52: I know this is not a valid change. I am having it here to highlight that we need to have a starting flow that works for the bouncer. I know @ramizhasan111 already discussed this with Alastair, please elaborate on what is the solution.

state-chain/chains/src/evm.rs
```diff
@@ -24,12 +24,21 @@ use sp_std::{convert::TryFrom, str, vec};
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct EvmCrypto;
 
+// TODO: Since they are different we can probably move this to its own
+// respective files (arb.rs and eth.rs)
```
albert-llimos, line 28: As explained in the comments, ARB & ETH have different gas consumptions. I assume this should be refactored in their own files.

state-chain/chains/src/evm.rs
```diff
@@ -24,12 +24,21 @@ use sp_std::{convert::TryFrom, str, vec};
...
-	pub const GAS_COST_PER_TRANSFER_TOKEN: u128 = 40_000;
+	pub const ETH_BASE_COST_PER_BATCH: u128 = 50_000;
+	pub const ETH_GAS_COST_PER_FETCH: u128 = 30_000;
+	pub const ETH_GAS_COST_PER_TRANSFER_NATIVE: u128 = 20_000;
+	pub const ETH_GAS_COST_PER_TRANSFER_TOKEN: u128 = 40_000;
+
+	// TODO: We might have to be able to update these on the fly. They are 
+	// variable with the L1 calldata prize and L1 calldata size. L1 calldata
+	// price will change in 4844 and L1 calldata size will change as Arb improves.
+	pub const ARB_BASE_COST_PER_BATCH: u128 = 5_200_000;
```
albert-llimos, line 38: Could we have this values so they can be updated on a runtime upgrade? Governance is a bit unnecessary, but we want some way to decrease this amounts when it makes sense.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -456,7 +456,8 @@ pub mod pallet {
 			if channel_metadata.is_some() {
 				// Currently only Ethereum supports CCM.
 				ensure!(
-					ForeignChain::Ethereum == destination_asset.into(),
+					ForeignChain::Ethereum == destination_asset.into() ||
+						ForeignChain::Arbitrum == destination_asset.into(),
```
albert-llimos, line 460: This should probably be refactored into a function/type that contains chains that support CCM, since in the future there will be more chains with CCM (e.g. Solana).

state-chain/runtime/src/chainflip.rs
```diff
@@ -164,7 +164,11 @@ impl EvmPriorityFee<Ethereum> for EthTransactionBuilder {
 		Some(U256::from(tracked_data.priority_fee))
 	}
 }
-impl EvmPriorityFee<Arbitrum> for ArbTransactionBuilder {}
+impl EvmPriorityFee<Arbitrum> for ArbTransactionBuilder {
+	fn get_priority_fee(tracked_data: &<Arbitrum as Chain>::TrackedData) -> Option<U256> {
+		Some(U256::from(0))
```
albert-llimos, line 169: Setting this to zero instead of None because otherwise the CFE ends up estimating and setting a priority fee as part of the `estimate_gas` call`. We want it to be zero.

engine/src/witness/arb.rs
albert-llimos: tracked here: https://linear.app/chainflip/issue/PRO-1224/arbitrum-contracts-should-be-injected-on-runtime-upgrade

state-chain/chains/src/evm.rs
dandanlen: The can already be changed with a runtime upgrade. And we recently introduced a per-chain multplier for gas adjustments. 

We just need to make sure that these values are correct relative to each other.

state-chain/chains/src/evm.rs
dandanlen: I agree separate modules would be better.

state-chain/runtime/src/chainflip.rs
dandanlen: Maybe add a comment to this effect so nobody feels tempted to change it.

state-chain/chains/src/evm.rs
albert-llimos: Cool, didn't know that. Then we're good :+1:

engine/src/witness/arb.rs
ramizhasan111: We have the functionality to insert the contract addresses on runtime upgrade for live networks. For localnets we can have the contract addresses inserted on genesis via chainspec.

ramizhasan111: That way the engine will read the correct addresses when starting up

albert-llimos: Yes agree. This is outdated and already removed. (just writing it here in case Dan sees these comments and gets confused).

albert-llimos:
I just made the changes myself as it seemed easy to do. @ramizhasan111 @dandanlen this is now fully ready for review and merge.

ramizhasan111 (approved):
LGTM

state-chain/chains/src/arb.rs
```diff
@@ -82,19 +82,26 @@ impl ArbitrumTrackedData {
 	}
 }
 
+pub mod fees {
+	pub const ARB_BASE_COST_PER_BATCH: u128 = 5_200_000;
+	pub const ARB_GAS_COST_PER_FETCH: u128 = 1_700_000;
+	pub const ARB_GAS_COST_PER_TRANSFER_NATIVE: u128 = 1_500_000;
+	pub const ARB_GAS_COST_PER_TRANSFER_TOKEN: u128 = 1_800_000;
+}
```
dandanlen, line 90: naming nit: since the fees are already namespaced in `arb::fees`, the `ARB_` prefix seems redundant. Same for eth::fees and ETH_.

state-chain/runtime/src/chainflip.rs
```diff
@@ -147,13 +147,13 @@ impl cf_traits::WaivedFees for WaivedFees {
 /// We are willing to pay at most 2x the base fee. This is approximately the theoretical
 /// limit of the rate of increase of the base fee over 6 blocks (12.5% per block).
 const ETHEREUM_BASE_FEE_MULTIPLIER: FixedU64 = FixedU64::from_rational(2, 1);
-/// TODO!!:We are willing to pay at most 2x the base fee. This is approximately the theoretical
-/// limit of the rate of increase of the base fee over 6 blocks (12.5% per block).
-const ARBITRUM_BASE_FEE_MULTIPLIER: FixedU64 = FixedU64::from_rational(2, 1);
-// We arbitrarily set the MAX_GAS_LIMIT we are willing broadcast to 10M.
+/// For Arbitrum we could theoretically set the multiplier to 1 but we add
+/// a smaller multiplier for future proofing.
+const ARBITRUM_BASE_FEE_MULTIPLIER: FixedU64 = FixedU64::from_rational(3, 2);
```
dandanlen, line 152: 3/2 is greater than 1 is the comment wrong or the value?

PR #4552, opened 2024-02-20
feat: add boost lp account to bouncer and fund it on setup_swaps

marcellorigotti:
# Pull Request

Closes: PRO-1188

## Checklist

## Summary

- Added a genesis account called LP_BOOST, which will be used to run the bouncer tests for boost (in localnet)
- Fund the LP_BOOST account when setting up swaps


marcellorigotti:
There is one thing left to do: 
How do we retrieve the private key of this newly created account?
I would like to add it to `localnet/keys` like we did for LP_1/LP_2/BROKER_1/BROKER_2

bouncer/shared/provide_liquidity.ts
```diff
@@ -15,13 +15,18 @@ import {
 } from '../shared/utils';
 import { send } from '../shared/send';
 
-export async function provideLiquidity(ccy: Asset, amount: number, waitForFinalization = false) {
+export async function provideLiquidity(
+  ccy: Asset,
+  amount: number,
+  waitForFinalization = false,
+  lpKey = '',
```
msgmaxim, line 22: Nit: you can use `lpKey?: string`, which will automatically have the default value of `undefined`, and then later you can do:
```ts
  const lpUri = lpKey ?? process.env.LP_URI || '//LP_1';
```
It is a more ideomatic way of using "optional" values in JS.

martin-chainflip (approved):

dandanlen (approved):

PR #4557, opened 2024-02-22
refactor/PRO-1159/remove-order-caches

Janislav:
# Pull Request

Closes: PRO-1159

## Checklist

## Summary

- Refactored the AMM in away that we can access orders by LP by given LP
- Removed all order caches as well as the usage of them and replaced with an equivalent
- Added necessary migrations to transform the state of the old AMM
- Tested the migrations against a local net build from the current main and 1.3 branch
- Added/Removed tests were needed

#### Non-Breaking changes


Janislav:
As discussed, converting this to draft until we have time to back to it.

dandanlen:
Having read the issue description, it seems like it would have been a worthwhile change - what was the blocker here? Just finding the time to review?

PR #4559, opened 2024-02-22
feat: limit the number of utxos selected for egress

syan095:
# Pull Request

Closes: PRO-1102

## Checklist

## Summary

Use consolidation size (100 by default) to limit the number of utxos that can be selected for egress.
Modified the algorithm for utxo selection


state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -453,6 +453,7 @@ impl<T: Config> Pallet<T> {
 						output_amount +
 							number_of_outputs * fee_per_output_utxo +
 							min_fee_required_per_tx,
+						Self::consolidation_parameters().consolidation_size,
```
msgmaxim, line 456: I wonder if this limit to too small. The number of utxos for consolidation is not as important as it is for egress. We were a little converative with consolidation since the only downside of setting a small limit is us paying more in fees. For egress, however, if we can't select utxos due to the limit, this would actually hurt user experience. We could probably be a little less conservative for egress and try, say 200, utxos at once and at least give a singing ceremony a chance to succeed (which it very well might).

There *is* a concern that we might be making ceremonies with 100+ payloads more likely (due to always selecting smallest utxos first), but in practice I doubt this will happen very often (a counter-argument to this could be that by allowing more payloads/utxos we won't need as many of them in the long term since we consolidate more efficiently).

dandanlen (commented):
I feel like there definitely must be a way of expressing this algorithm more simply and more efficiently. It's kind of hard to understand: we return early in places, we exit with None even though 'this should never happen', there are lots of loops and conditionals checking different vec with different things in them and mutating vectors. The control flow is hard to follow. None of this is anyone's fault btw., it has just evolved to this. But I think we can do better. 

I think if we start with a vector / iterator of cumulative net utxo value, we should be able to iterate/search that vector to find the optimal utxos.

With a vec of cumulative amount, total contained in the utxos between two indices i and j is `cumulative_amount[j] - cumulative_amount[i]`.

So we could just start i being the index of the first viable utxo, and then search the indices until we get a range of values that work. The we just splice the vector between those indices.

This way only need to allocate single intermediate vec (with the cumulative values). Not putting back values etc.

And we avoid mutating the vec in the failure case.


Thoughts?

msgmaxim:
> Thoughts?

Does this mean we will effecitvely be swapping the smallest utxo with the first larger utxo that's not in our list (as opposed to the largest overall)? I thought about this too, but wasn't sure if we were willing to change the semantics of the algorithm. Instead of allocating a vec with cumulative sum, could we not just keep track of the amount in the currently selected range and modifying it by `utxos[j+1] - utxos[i]` as we shift "right" (similar to how it is done currently, except that the selection is a contiguous range, so we don't need to allocate anything)?

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -1,54 +1,114 @@
...
+	// 4. If total amount is not reached, keep swapping smallest elected with the largest
+	// unselected.
+	// Since the largest N element is > target, worse case scenario is to select largest N utxos.
+	while cumulative_amount < amount_to_be_spent &&
+		!available_utxos.is_empty() &&
+		!selected_utxos.is_empty()
+	{
+		// selected is sorted small >>> large
+		// available is sorted large >>> small
+		let small_utxo = selected_utxos.remove(0);
```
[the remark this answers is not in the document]
syan095: This is refactored out

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -100,99 +216,155 @@ fn test_utxo_selection() {
...
+		let fee_info = BitcoinFeeInfo { sats_per_kilobyte: 1_000_000 };
+		test_case(&mock_utxos(), &fee_info, 1, Some(3), None);
+
+		// lowest 5 are skipped due to high fee.
+		let fee_info = BitcoinFeeInfo { sats_per_kilobyte: 4_000 };
+		test_case(
+			&mock_utxos(),
+			&fee_info,
+			1_500,
+			Some(3),
```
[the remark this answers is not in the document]
syan095: Now the "take extra one" only happens if the selection limit is not reached. (before we always add one, even if we already at the limit, resulting in 1 more utxo than the limit allows)

state-chain/pallets/cf-environment/src/lib.rs
syan095: I was trying to avoid changing storage, but I think it is unavoidable.
I now renamed the consolidation_parameter to a more generic utxo_parameter, so the selection_limit can be configured separately to be more aggressive.

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -1,54 +1,114 @@
...
+///
+/// In the worst case scenarios, only the largest N utxos are selected. Checks in step 2 guarantees
+/// enough value are selected.
+///
+/// 5. We then take 1 more utxo as safety measure (in the case that the fees are more expensive than
+/// expected).
+/// The skipped utxos are appended back to the `available_utxos` for future use.
+///
+/// Note that on failure (when `None` is returned) this may still modify `available_utxos`. It is
+/// expected that the user will roll back storage on failed cases.
```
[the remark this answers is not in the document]
syan095: Yeah this is a much cleaner algorithm. I refactored the code to implement this. It's much cleaner

syan095:
I've refactored the ConsolidationParameter to a more generic UtxoParameter and added an extra field that controls utxo selection limit. 
I needed to add a storage migration for the renaming of storage.
It's very simple one - to insert the "INITIAL_UTXO_PARAMTER" const into storage. Do we need to manually remove the old storage value stored in "ConsolidationParameter::<T>"?

kylezs:
I haven't looked at this in a lot of detail, but my expectation was that this would require some changes to the rotation and/or UTXO consolidation in the case we have more UTXOs than necessary and we can't send them all across in one tx.

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -1,68 +1,126 @@
...
+		.extract_if(|utxo| utxo.amount <= fee_info.fee_for_utxo(utxo))
+		.collect::<Vec<_>>();
+
+	if available_utxos.is_empty() || selection_limit == 0usize {
+		return Err(UtxoSelectionError::NoUtxoAvailableAfterFeeReduction)
 	}
 
+	// 2. Find the optimal `first` and `last` index, such that
+	// `sum(utxo[first ..= last]) >= target_amount`
 	available_utxos.sort_by_key(|utxo| sp_std::cmp::Reverse(utxo.amount));
```
dandanlen, line 74: If we don't use `Reverse` then the values will be in ascending order and the algorithm should be easier to write (since we won't have the confusing situation where the 'first' index is greater than the 'last').

dandanlen: Btw. we should sort by the utxo's *net* value, not by utxo amount.

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -1,68 +1,126 @@
...
 
-	let mut selected_utxos: Vec<Utxo> = vec![];
-	let mut skipped_utxos: Vec<Utxo> = vec![];
-
+	let mut first = available_utxos.len();
+	let mut last = first - 1;
```
dandanlen, line 77: Having last and first with last < first is quite confusing...

state-chain/chains/src/btc.rs
```diff
@@ -217,21 +217,29 @@ pub struct EpochStartData {
 }
 
 #[derive(Encode, Decode, Default, PartialEq, Copy, Clone, TypeInfo, RuntimeDebug)]
-pub struct ConsolidationParameters {
+pub struct UtxoParameters {
```
dandanlen, line 220: `UtxoSelectionParameters`?

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -1,68 +1,126 @@
...
+	// If no selection limit is given, selecting all utxos is allowed.
+	let selection_limit = match maybe_limit {
+		Some(limit) => limit as usize,
+		None => available_utxos.len(),
+	};
+
+	// 1. Filter out utxos whose fees are too high.
+	let mut skipped_utxos = available_utxos
+		.extract_if(|utxo| utxo.amount <= fee_info.fee_for_utxo(utxo))
+		.collect::<Vec<_>>();
```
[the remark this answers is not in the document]
syan095: great idea!

dandanlen:
For visibility - discussed with @syan095 and the case where fail to select utxos is already hanndled gracefully in the sense that we would just retry on the next block. The only way this can currently happen is the (highly unlikely) case where even the 100 highest-valued utxos are insufficient to pay for an output.

kylezs:
> we would just retry on the next block.

Not quite sure i follow, could you expand on this?

syan095:
Also we need a reasonable number to set as UtxoSelectionLimit. currently `INITIAL_UTXO_SELECTION_PARAMETERS` is set to 250, this might be too aggressive. We should come up with a reasonable value.

syan095:
> > we would just retry on the next block.
> 
> Not quite sure i follow, could you expand on this?
> 
> The scenario I have in mind is that we want to rotate the vaults and we have > max number of utxos to sign over. So the activation tx would fail over to the next block?

For vault rotation we take all utxos that have positive net value (UtxoSelectionType::SelectAllForRotation), it's a different code path from the utxo selection. 
I think If we are worried that we have too many utxos and should wait for consolidation, we should handle this in a separate ticket

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -1,122 +1,108 @@
...
-	{
-		available_utxos.append(&mut skipped_utxos);
-		return None
-	}
+	while last < available_utxos.len() {
+		cumulative_amount += available_utxos[last].net_value(fee_info);
 
-	let mut selected_utxos: Vec<Utxo> = vec![];
+		if last - first >= selection_limit {
+			// Move the `last` pointer forward by one to keep selection size within limit.
```
msgmaxim, line 79: Do you mean "first"?

syan095: Corrected!

dandanlen (commented):
I took the liberty of moving some things around and tidying up some comments. The utxo parameters didn't really need to be in the main btc module. Also some of the comments were stale so I just removed them. The code is clear enough on its own.

I would also like us to think about the values before merging this:

```rust
const INITIAL_UTXO_CONSOLIDATION_PARAMETERS: utxo_selection::ConsolidationParameters =
	ConsolidationParameters { consolidation_threshold: 200, consolidation_size: 100 };

const INITIAL_UTXO_SELECTION_PARAMETERS: utxo_selection::UtxoSelectionParameters =
	UtxoSelectionParameters { selection_limit: 250 };
```

First of all, having a selection limit that is much higher than the consolidation limit *and* the consolidation threshold makes no sense. So the selection limit should be lower IMO. 

But also, I feel like the consolidation_size and selection_limit are related and should possibly just be the same. Why would we consolidate less than the limit? If want to consolidate, surely we should consolidate as much as possible? 

So maybe we don't need a separate `UtxoSelectionParameters` and we can just use the consolidation params? 

Any opinions? (@msgmaxim @syan095 )

dandanlen:
One more comment - not really related to this PR, but I feel like the tests could be clearer. It's very hard when reading to be able to tell that we've covered all possible scenarios... they could do with a refactor to be a bit clearer and more structured. This can wait for another time but I feel like it's worth doing.

msgmaxim:
> Why would we consolidate less than the limit? If want to consolidate, surely we should consolidate as much as possible?

I talked about it here: https://github.com/chainflip-io/chainflip-backend/pull/4559#discussion_r1500146828

My thinking is that we want consolidations to always succeed, because it is possible that they will be quite frequent (if everyone starts selling BTC, for example). We don't know how many simultaneous ceremonies we can do exactly, but for consolidation we can be conservative and go with a smaller number. For egress, however, we have a target amount that we must reach, and so to give a large egress a chance to succeed, we may want to be more aggressive and allow ceremonies with a slightly larger set of utxos than we normally alllow for consolidation.

For example, given current parameters, say the vault has 150 utxos. The next consolidation will only happen when we reach 200 utxos, so it make take a long time. It is possible though, that at some point we we need to satisfy a large egress, and this can only be done with 105 utxos. If we can somewhat confidently do 100 utxos, surely we can also almost always do 105 (even if it takes a few attempts), but if we use the same limit for egress, we won't even try that and the egress would be stuck.

I'm less convinced about the current 250 limit though. I think it might not make sense for it to be larger than `consolidation_threshold` (it is unlikely that we will ever have more than `consolidation_threshold` utxos in total, and even if we do, we can wait a few blocks for consolidation to happen). Making it equal to `consolidation_threshold` is probably reasonable. However, maybe `selection_limit` shouldn't be that much higher than `consolidation_size` as we don't want ceremonies to fail too often (we probably won't know until we try). I do think it makes sense to have it as a separate paramter so we can adjust it separately, even if we decide to make it equal to `consolidation_size` initially.

I'm also realising that due to how we select utxos (starting from smallest), it is possible that we end up selecting more than `consolidation_size` utxos more often than striclty necessary (because we could have selected fewer if we were froced to include some larger utxos). I don't know if this is a concern in practice or not.

syan095:
Yeah I agree that it doesn't make sense for the selection limit to be > consolidation limit. We expect the utxos to fluctuate between 100 - 200. I think it would be safe to set it to 200 to start off with and adjust later. 

The only case we will fail to select enough value for utxo while we DO have enough, is in the edge case where we suddenly got a influx of utxos in 1 block (so the total exceeds 200) - in which case we will fail to egress, and the consolidation will kick in. In the next block, we would have enough utxos to egress successfully.

dandanlen:
I can see an argument (albeit a very speculative one) for making the selection_limit larger than the consolidation_size, but larger than the consolidation_threshold seems pointless to me... we will basically never reach it. 

Also the current algorithm ensures that we fall back to the larger utxos, so we can only fail if the N *largest* utxos are insufficient to build the transaction. I find this highly unlikely. I would rather we have fewer configuration options unless there's a good reason to introduce more.

syan095:
> I would rather we have fewer configuration options unless there's a good reason to introduce more.

We can set the limit to the Consolidation Threshold (200). This saves extra storage, extrinsic and migration.

dandanlen:
Yeah I think it makes sense to consolidate both into one.

state-chain/chains/src/btc/utxo_selection.rs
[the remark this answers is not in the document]
dandanlen: Having discussed this a bit with Maxim, i don't think it's necessary as long as the selection limit is comparable to the consolidation limit.

syan095:
> Yeah I think it makes sense to consolidate both into one.

Reverted the extra selection parameter.
We now use the Consoludation_threshold (200) for the selection limit.

dandanlen (approved):

PR #4560, opened 2024-02-23
feat(custom-rpc): add broker info [WEB-925]

acdibble:
# Pull Request

Closes: WEB-925

## Checklist

## Summary

#### Non-Breaking changes

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -33,7 +33,7 @@ use state_chain_runtime::{
 	chainflip::{BlockUpdate, Offence},
 	constants::common::TX_FEE_MULTIPLIER,
 	runtime_apis::{
-		CustomRuntimeApi, DispatchErrorWithMessage, FailingWitnessValidators,
+		BrokerInfo, CustomRuntimeApi, DispatchErrorWithMessage, FailingWitnessValidators,
 		LiquidityProviderInfo, RuntimeApiAccountInfoV2,
 	},
 	NetworkFee,
@@ -60,6 +60,7 @@ pub enum RpcAccountInfo {
 	},
 	Broker {
 		flip_balance: NumberOrHex,
+		earned_fees: any::AssetMap<NumberOrHex>,
 	},
 	LiquidityProvider {
 		balances: any::AssetMap<NumberOrHex>,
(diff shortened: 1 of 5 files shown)


state-chain/runtime/src/runtime_apis.rs
```diff
@@ -125,7 +130,7 @@ decl_runtime_apis!(
 		fn cf_flip_supply() -> (u128, u128);
 		fn cf_accounts() -> Vec<(AccountId32, VanityName)>;
 		fn cf_account_flip_balance(account_id: &AccountId32) -> u128;
-		fn cf_account_info_v2(account_id: &AccountId32) -> RuntimeApiAccountInfoV2;
+		fn cf_validator_info(account_id: &AccountId32) -> RuntimeApiAccountInfoV2;
```
acdibble, line 133: it's a breaking change but it's a good time to do it with 1.3 coming soon and this old one being deprecated and only used for getting validator info

state-chain/custom-rpc/src/lib.rs
```diff
@@ -312,12 +322,6 @@ pub trait CustomApi {
 		account_id: state_chain_runtime::AccountId,
 		at: Option<state_chain_runtime::Hash>,
 	) -> RpcResult<RpcAccountInfo>;
-	#[method(name = "account_info_v2")]
-	fn cf_account_info_v2(
-		&self,
-		account_id: state_chain_runtime::AccountId,
-		at: Option<state_chain_runtime::Hash>,
-	) -> RpcResult<RpcAccountInfoV2>;
```
acdibble, line 320: it's a breaking change but it's a good time to do it with 1.3 coming soon and this old one being deprecated and only used for getting validator info

state-chain/runtime/src/lib.rs
```diff
@@ -1070,7 +1070,7 @@ impl_runtime_apis! {
 		fn cf_account_flip_balance(account_id: &AccountId) -> u128 {
 			pallet_cf_flip::Account::<Runtime>::get(account_id).total()
 		}
-		fn cf_account_info_v2(account_id: &AccountId) -> RuntimeApiAccountInfoV2 {
+		fn cf_validator_info(account_id: &AccountId) -> RuntimeApiAccountInfoV2 {
```
dandanlen, line 1073: If `RuntimeApiAccountInfoV2` is now only used for validators, we could rename it as `ValidatorInfo` to be consistent with other naming. 

As a next step (I'll open issue) we could factor out duplicate information and have a struct like `AccountInfo<Role>` where the role defines the additional information for each role. Something like this:

```rust
struct AccountInfo<RoleInfo> {
	balance: NumberOrHex,
	#[serde(flatten)]
	role_info: RoleInfo,
}

enum RpcAccountInfo {
	Unregistered(AccountInfo<()>),
	Lp(AccountInfo<LpAccountInfo>),
	//..
}
```

It should be possible to do this without chaning the encoding, and it will make things easier to maintain.

dandanlen (approved):
Tagging @AlastairHolmes for visibility.

state-chain/runtime/src/lib.rs
acdibble: will do, sounds good!

AlastairHolmes:
We we sure that rpc isn't being used, did we ask?

acdibble:
> We sure that rpc isn't being used, did we ask?

No, but all the information is available in `account_info`, I'll ask around

acdibble:
I've reverted the breaking change because I don't think anyone is using this because it's not documented and we haven't encouraged anyone to use it, but they might be and I don't feel like figuring it out. It's more important to get the additions in now.

PR #4561, opened 2024-02-23
feat: Expose tx_hash on BroadcastSuccess event

marcellorigotti:
# Pull Request

Closes: PRO-1195

## Checklist

## Summary

Updated transaction_succeeded and BroadcastSuccess Event to include the tx_ref which represent the transaction hash (or an equivalent way of identifying a transaction depending on the chain)

#### Breaking changes

This PR is breaking, both an extrinsic used in the engine and an event changed

### DIFF
--- a/api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
+++ b/api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
@@ -540,7 +540,7 @@ mod tests {
 						max_priority_fee_per_gas: None,
 						contract: H160::from([0; 20]),
 						gas_limit: None,
-						tx_ref: Default::default()
+						tx_ref: Default::default(),
 					},
 				},
 			),
--- a/api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
+++ b/api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
@@ -540,6 +540,7 @@ mod tests {
 						max_priority_fee_per_gas: None,
 						contract: H160::from([0; 20]),
 						gas_limit: None,
+						tx_ref: Default::default()
 					},
 				},
 			),
--- a/api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
+++ b/api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
@@ -540,7 +540,7 @@ mod tests {
 						max_priority_fee_per_gas: None,
 						contract: H160::from([0; 20]),
 						gas_limit: None,
-						tx_ref: Default::default(),
+						tx_hash: Default::default(),
 					},
 				},
 			),
--- a/engine/src/witness/btc.rs
+++ b/engine/src/witness/btc.rs
@@ -5,12 +5,14 @@ pub mod btc_source;
 use std::sync::Arc;
 
 use bitcoin::BlockHash;
-use cf_chains::btc::{self, deposit_address::DepositAddress, BlockNumber, CHANGE_ADDRESS_SALT, BitcoinTransactionMetadata};
+use cf_chains::btc::{
+	self, deposit_address::DepositAddress, BitcoinTransactionMetadata, BlockNumber,
+	CHANGE_ADDRESS_SALT,
+};
 use cf_primitives::{EpochIndex, NetworkEnvironment};
 use futures_core::Future;
 use secp256k1::hashes::Hash;
-use utilities::task_scope::Scope;
-use utilities::ArrayCollect;
+use utilities::{task_scope::Scope, ArrayCollect};
 
 use crate::{
 	btc::{
@@ -57,7 +59,9 @@ pub async fn process_egress<ProcessCall, ProcessingFut, ExtraInfo, ExtraHistoric
 					signer_id: DepositAddress::new(epoch.info.0.current, CHANGE_ADDRESS_SALT)
 						.script_pubkey(),
 					tx_fee: tx.fee.unwrap_or_default().to_sat(),
-					tx_metadata: BitcoinTransactionMetadata { tx_ref: tx_hash.iter().rev().map(|elem| elem.to_owned()).collect_array() },
+					tx_metadata: BitcoinTransactionMetadata {
+						tx_ref: tx_hash.iter().rev().map(|elem| elem.to_owned()).collect_array(),
+					},
 				},
 			),
 			epoch.index,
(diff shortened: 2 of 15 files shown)


api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -540,6 +540,7 @@ mod tests {
 						max_priority_fee_per_gas: None,
 						contract: H160::from([0; 20]),
 						gas_limit: None,
+						tx_ref: Default::default(),
```
kylezs, line 543: i think the naming is a little bit the other way around in this PR. On these chain specific types, we should use the specific name they refer to. e.g. in this case it should be `tx_hash`, and in polkadot it should be `tx_id`. And then on the generic type i.e. the `BroadcastSuccess` event, the field should be `tx_ref` since it's more general.

engine/src/witness/dot.rs
```diff
@@ -157,7 +157,12 @@ pub async fn process_egress<ProcessCall, ProcessingFut>(
 								tx_out_id: signature,
 								signer_id: epoch.info.0,
 								tx_fee,
-								tx_metadata: (),
+								tx_metadata: PolkadotTransactionMetadata {
+									tx_ref: PolkadotTransactionSignature {
```
kylezs, line 161:
```suggestion
									tx_ref: PolkadotTransactionId {
```

state-chain/chains/src/any.rs
```diff
@@ -24,6 +24,7 @@ impl Chain for AnyChain {
 	type TransactionMetadata = ();
 	type ReplayProtectionParams = ();
 	type ReplayProtection = ();
+	type TransactionHashItem = ();
```
kylezs, line 27: Here it makes more sense to use `TransactionRef` or `TransactionId` imo

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -316,6 +319,7 @@ pub mod pallet {
 		BroadcastSuccess {
 			broadcast_id: BroadcastId,
 			transaction_out_id: TransactionOutIdFor<T, I>,
+			transaction_hash: TransactionHashFor<T, I>,
```
kylezs, line 322:
```suggestion
			transaction_ref: TransactionHashFor<T, I>,
```

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -996,6 +997,7 @@ fn aborted_broadcasts_can_still_succeed() {
 				crate::Event::<Test, Instance1>::BroadcastSuccess {
 					broadcast_id,
 					transaction_out_id,
+					transaction_hash: Default::default(),
```
kylezs, line 1000: it's not clear why this is Default::default() - should be some association we can make in the test?

state-chain/chains/src/lib.rs
```diff
@@ -254,6 +258,13 @@ where
 	}
 }
 
+pub trait TransactionMetadataHash<C: Chain> {
+	fn get_transaction_hash(&self) -> C::TransactionHashItem;
```
kylezs, line 262:
```suggestion
	fn get_transaction_ref(&self) -> C::TransactionHashItem;
```

state-chain/chains/src/lib.rs
```diff
@@ -254,6 +258,13 @@ where
 	}
 }
 
+pub trait TransactionMetadataHash<C: Chain> {
```
kylezs, line 261:
```suggestion
pub trait GetTransactionRef<C: Chain> {
```

state-chain/chains/src/dot.rs
```diff
@@ -314,6 +314,38 @@ impl FeeEstimationApi<Polkadot> for PolkadotTrackedData {
...
+	fn verify_metadata(&self, _expected_metadata: &Self) -> bool {
+		true
+	}
+}
+
+#[derive(
+	Encode, Decode, TypeInfo, Clone, RuntimeDebug, Default, PartialEq, Eq, Serialize, Deserialize,
+)]
+pub struct PolkadotTransactionSignature {
+	pub block_hash: H256,
```
kylezs, line 345: we should use the `PolkadotBlockNumber` here instead. As it's only 32 bits, compared to the 32 byte hash - but still uniquely identifies the transaction.

state-chain/chains/src/dot.rs
```diff
@@ -314,6 +314,38 @@ impl FeeEstimationApi<Polkadot> for PolkadotTrackedData {
...
+	Encode, Decode, TypeInfo, Clone, RuntimeDebug, Default, PartialEq, Eq, Serialize, Deserialize,
+)]
+pub struct PolkadotTransactionMetadata {
+	pub tx_ref: PolkadotTransactionSignature,
+}
+
+impl<C: Chain<TransactionHashItem = PolkadotTransactionSignature>> TransactionMetadataHash<C>
+	for PolkadotTransactionMetadata
+{
+	fn get_transaction_hash(&self) -> <C as Chain>::TransactionHashItem {
```
kylezs, line 327: can just combine this trait with the other one?

kylezs (commented):
Mostly naming related - seems like @dandanlen is chatting to you about this too, so will pause the review here :)

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
marcellorigotti: Yeah sorry about the naming, I wasn't really sure what to use, followed your suggestions, makes everything clearer!

engine/src/witness/btc.rs
```diff
@@ -56,7 +59,9 @@ pub async fn process_egress<ProcessCall, ProcessingFut, ExtraInfo, ExtraHistoric
 					signer_id: DepositAddress::new(epoch.info.0.current, CHANGE_ADDRESS_SALT)
 						.script_pubkey(),
 					tx_fee: tx.fee.unwrap_or_default().to_sat(),
-					tx_metadata: (),
+					tx_metadata: BitcoinTransactionMetadata {
+						tx_hash: tx_hash.iter().rev().map(|elem| elem.to_owned()).collect_array(),
```
kylezs, line 63: Probably better to have a type to do this - then we can write tests too, to see that the bytes display the correct way

kylezs (approved):
Nice stuff!

kylezs:
Get at least a review from product here too

niklasnatter (approved):
looks good from my side 🙂 would be very cool if we could include it in the ingress-egress-tracker next!

kylezs:
> looks good from my side 🙂 would be very cool if we could include it in the ingress-egress-tracker next!

It should be included for free there - but I'm not sure how you guys are getting and writing the data now with the redis addition

marcellorigotti:
I'll close this PR and start working on the ingress-egress-tracker on a new one

PR #4563, opened 2024-02-23
refactor: pass out CFE incompatibility exit information to main

kylezs:
## Checklist

 I've tested every case of error returning from the client - and that the downcast works as expected. alles gut.

## Summary

A step towards single-binary CFE upgrades. This will allow the reason for shutting down to bubble up to the runner. 

Also allows for passing a starting block number to the SCC to ensure it won't miss blocks on the handover to the new version in an upcoming PR.

Non breaking but no point including it in the release.

### DIFF
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -1,4 +1,3 @@
-use crate::state_chain_observer::client::CreateStateChainClientError;
 use anyhow::Context;
 use cf_chains::dot::PolkadotHash;
 use cf_primitives::AccountRole;
@@ -13,7 +12,7 @@ use chainflip_engine::{
 		self,
 		client::{
 			chain_api::ChainApi, extrinsic_api::signed::SignedExtrinsicApi,
-			storage_api::StorageApi, STATE_CHAIN_CONNECTION,
+			storage_api::StorageApi, CreateStateChainClientError, STATE_CHAIN_CONNECTION,
 		},
 	},
 	witness,
@@ -227,32 +226,9 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 
 	if let Err(e) = &root_result {
 		let sc_error = e.downcast_ref::<CreateStateChainClientError>();
-		match sc_error {
-			Some(CreateStateChainClientError::NoLongerCompatible {
-				cfe_version,
-				cfe_version_required,
-				at_block,
-			}) => {
-				tracing::info!(
-					"Chainflip Engine is no longer compatible with the current state chain. \
-					Engine version: {}, required version: {}, first incompatible block: {}",
-					cfe_version,
-					cfe_version_required,
-					at_block
-				);
-			},
-			Some(CreateStateChainClientError::NotYetCompatible {
-				cfe_version,
-				cfe_version_required,
-			}) => {
-				tracing::info!(
-					"Chainflip Engine is not yet compatible with the current state chain. \
-					Engine version: {}, required version: {}",
-					cfe_version,
-					cfe_version_required
-				);
-			},
-			_ => {},
+		if let Some(CreateStateChainClientError::CompatibilityError(block_compatibility)) = sc_error
+		{
+			tracing::info!("Block compatibility on shutdown: {:?}", block_compatibility);
 		}
 	}
 	root_result
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -1,3 +1,4 @@
+use crate::state_chain_observer::client::CreateBlockStreamError;
 use anyhow::Context;
 use cf_chains::dot::PolkadotHash;
 use cf_primitives::AccountRole;
@@ -45,7 +46,7 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn run_main(settings: Settings) -> anyhow::Result<()> {
-	task_scope(|scope| {
+	let root_result = task_scope(|scope| {
 		async move {
 			let mut start_logger_server_fn =
 				Some(utilities::logging::init_json_logger(settings.logging.clone()).await);
@@ -220,5 +221,39 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 		}
 		.boxed()
 	})
-	.await
+	.await;
+
+	if let Err(e) = root_result {
+		let sc_error = e.downcast_ref::<CreateBlockStreamError>();
+		match sc_error {
+			Some(CreateBlockStreamError::NoLongerCompatible {
+				cfe_version,
+				cfe_version_required,
+				first_incompatible_block,
+			}) => {
+				tracing::info!(
+					"Chainflip Engine is no longer compatible with the current state chain. \
+					Engine version: {}, required version: {}, first incompatible block: {}",
+					cfe_version,
+					cfe_version_required,
+					first_incompatible_block
+				);
+			},
+			Some(CreateBlockStreamError::NotYetCompatible {
+				cfe_version,
+				cfe_version_required,
+			}) => {
+				tracing::info!(
+					"Chainflip Engine is not yet compatible with the current state chain. \
+					Engine version: {}, required version: {}",
+					cfe_version,
+					cfe_version_required
+				);
+			},
+			_ => {
+				tracing::error!("Error in main: {:?}", e);
+			},
+		}
+	}
+	Ok(())
 }
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -62,6 +62,8 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 					true,
 					true,
 					true,
+					// TODO: Pass this in from the CFE runner: PRO-935
+					None,
 				)
 				.await?;
 
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -1,4 +1,4 @@
-use crate::state_chain_observer::client::CreateBlockStreamError;
+use crate::state_chain_observer::client::CreateStateChainClientError;
 use anyhow::Context;
 use cf_chains::dot::PolkadotHash;
 use cf_primitives::AccountRole;
@@ -226,9 +226,9 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 	.await;
 
 	if let Err(e) = &root_result {
-		let sc_error = e.downcast_ref::<CreateBlockStreamError>();
+		let sc_error = e.downcast_ref::<CreateStateChainClientError>();
 		match sc_error {
-			Some(CreateBlockStreamError::NoLongerCompatible {
+			Some(CreateStateChainClientError::NoLongerCompatible {
 				cfe_version,
 				cfe_version_required,
 				first_incompatible_block,
@@ -241,7 +241,7 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 					first_incompatible_block
 				);
 			},
-			Some(CreateBlockStreamError::NotYetCompatible {
+			Some(CreateStateChainClientError::NotYetCompatible {
 				cfe_version,
 				cfe_version_required,
 			}) => {
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -231,14 +231,14 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 			Some(CreateStateChainClientError::NoLongerCompatible {
 				cfe_version,
 				cfe_version_required,
-				first_incompatible_block,
+				at_block,
 			}) => {
 				tracing::info!(
 					"Chainflip Engine is no longer compatible with the current state chain. \
 					Engine version: {}, required version: {}, first incompatible block: {}",
 					cfe_version,
 					cfe_version_required,
-					first_incompatible_block
+					at_block
 				);
 			},
 			Some(CreateStateChainClientError::NotYetCompatible {
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -252,9 +252,7 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 					cfe_version_required
 				);
 			},
-			_ => {
-				tracing::error!("Error in main: {:?}", e);
-			},
+			_ => {},
 		}
 	}
 	root_result
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -225,7 +225,7 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 	})
 	.await;
 
-	if let Err(e) = root_result {
+	if let Err(e) = &root_result {
 		let sc_error = e.downcast_ref::<CreateBlockStreamError>();
 		match sc_error {
 			Some(CreateBlockStreamError::NoLongerCompatible {
@@ -257,5 +257,5 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 			},
 		}
 	}
-	Ok(())
+	root_result
 }
--- a/engine/src/main.rs
+++ b/engine/src/main.rs
@@ -45,7 +45,7 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn run_main(settings: Settings) -> anyhow::Result<()> {
-	let root_result = task_scope(|scope| {
+	task_scope(|scope| {
 		async move {
 			let mut start_logger_server_fn =
 				Some(utilities::logging::init_json_logger(settings.logging.clone()).await);
@@ -222,14 +222,12 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 		}
 		.boxed()
 	})
-	.await;
-
-	if let Err(e) = &root_result {
-		let sc_error = e.downcast_ref::<CreateStateChainClientError>();
-		if let Some(CreateStateChainClientError::CompatibilityError(block_compatibility)) = sc_error
+	.await
+	.inspect_err(|e| {
+		if let Some(CreateStateChainClientError::CompatibilityError(block_compatibility)) =
+			e.downcast_ref::<CreateStateChainClientError>()
 		{
 			tracing::info!("Block compatibility on shutdown: {:?}", block_compatibility);
 		}
-	}
-	root_result
+	})
 }
--- a/engine/src/state_chain_observer/client.rs
+++ b/engine/src/state_chain_observer/client.rs
@@ -590,7 +590,7 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
 						_ => {
 							// No need to fetch any earlier blocks, we just have to move the stream
 							// to the correct starting point.
-							let skip_to = start_from.unwrap_or(finalized_header.number + 1);
+							let skip_to = start_from.unwrap_or(finalized_header.number);
 							for block_number in finalized_block_stream.cache().number + 1..skip_to {
 								assert_eq!(
 									finalized_block_stream.next().await.unwrap()?.number,
--- a/engine/src/state_chain_observer/client.rs
+++ b/engine/src/state_chain_observer/client.rs
@@ -106,8 +106,6 @@ pub enum CreateStateChainClientError {
 		cfe_version_required: SemVer,
 		first_incompatible_block: state_chain_runtime::BlockNumber,
 	},
-	#[error(transparent)]
-	Other(#[from] anyhow::Error),
 }
 
 pub struct StateChainClient<
--- a/engine/src/state_chain_observer/client.rs
+++ b/engine/src/state_chain_observer/client.rs
@@ -287,33 +287,33 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
 				base_rpc_client.check_block_compatibility(*block_stream.cache()).await?;
 
 			match block_compatibility.compatibility {
-				CfeCompatibility::NoLongerCompatible =>
-					if error_on_incompatible_block {
-						return Err(CreateStateChainClientError::CompatibilityError(
-							block_compatibility,
-						)
-						.into());
-					} else {
-						block_stream
-					},
-				CfeCompatibility::NotYetCompatible => {
+				CfeCompatibility::Compatible => block_stream,
+				CfeCompatibility::NoLongerCompatible if error_on_incompatible_block => {
+					return Err(CreateStateChainClientError::CompatibilityError(
+						block_compatibility,
+					)
+					.into());
+				},
+				// Either:
+				// - We're no longer compatible but we don't want to exit on error
+				// - We're not yet compatible and we're waiting for the required version
+				_ => {
 					// TODO: After the whole "single binary CFE upgrade" is complete, we should be
 					// able to remove `wait_for_required_version` as we'll never wait. We'll always
 					// return out the error and let the runner start the other version.
 					if wait_for_required_version {
-						let incompatible_blocks =
-							block_stream
-								.by_ref()
-								.try_take_while(|block_info| {
-									let base_rpc_client = base_rpc_client.clone();
-									let block_info = *block_info;
-									async move {
-										{
-											let block_compatibility = base_rpc_client
-												.check_block_compatibility(block_info)
-												.await?;
+						let incompatible_blocks = block_stream
+							.by_ref()
+							.try_take_while(|block_info| {
+								let base_rpc_client = base_rpc_client.clone();
+								let block_info = *block_info;
+								async move {
+									{
+										let block_compatibility = base_rpc_client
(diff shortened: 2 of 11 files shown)


engine/src/main.rs
```diff
@@ -220,5 +223,39 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
...
+					cfe_version,
+					cfe_version_required
+				);
+			},
+			_ => {
+				tracing::error!("Error in main: {:?}", e);
+			},
+		}
+	}
+	Ok(())
```
dandanlen, line 260: Do we not want to return `root_result` here?

state-chain/primitives/src/lib.rs
```diff
@@ -234,7 +234,31 @@ pub struct SemVer {
...
+
 impl SemVer {
+	pub fn compatibility_with_runtime(&self, version_runtime_requires: SemVer) -> CfeCompatibility {
+		if self.is_compatible_with(version_runtime_requires) {
+			CfeCompatibility::Compatible
+		} else if self < &version_runtime_requires {
+			CfeCompatibility::NotYetCompatible
+		} else {
+			CfeCompatibility::NoLongerCompatible
+		}
```
dandanlen, line 259: I think these might be the wrong way round - if the cfe is running a version lower than the runtime, then it will never be compatible (the runtime never downgrades).

engine/src/main.rs
```diff
@@ -220,5 +223,39 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
...
+				cfe_version,
+				cfe_version_required,
+				first_incompatible_block,
+			}) => {
+				tracing::info!(
+					"Chainflip Engine is no longer compatible with the current state chain. \
+					Engine version: {}, required version: {}, first incompatible block: {}",
+					cfe_version,
+					cfe_version_required,
+					first_incompatible_block
```
dandanlen, line 241: FWIW I think thiserror defines display methods so I don't think you need to deconstruct the errors here.

kylezs: It's setting up for a future PR where this will convert the error to something readable by the runner

engine/src/state_chain_observer/client.rs
```diff
@@ -94,6 +96,20 @@ impl DefaultRpcClient {
 	}
 }
 
+#[derive(Error, Debug)]
+pub enum CreateBlockStreamError {
```
AlastairHolmes, line 100:
```suggestion
pub enum CreateStateChainClientError {
```

state-chain/primitives/src/lib.rs
```diff
@@ -234,7 +234,30 @@ pub struct SemVer {
...
 	pub patch: u8,
 }
+
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum CfeCompatibility {
+	/// The version is currently compatible with the target.
+	Compatible,
+
+	/// The version is not yet compatible with the target. Should wait for the new version.
+	NotYetCompatible,
```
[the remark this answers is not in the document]
AlastairHolmes: And youll need to take BlockInfo instead of just the hash.

engine/src/state_chain_observer/client.rs
```diff
@@ -276,55 +297,125 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+					dyn TryCachedStream<
+							Error = anyhow::Error,
+							Ok = BlockInfo,
+							Item = Result<BlockInfo, anyhow::Error>,
+						> + Send,
+				>,
+			> = if let Some(start_from) = start_from {
+				let base_rpc_client = base_rpc_client.clone();
+
+				let mut start_from_headers = Box::pin(
```
AlastairHolmes, line 312: Because we know of node caching problems, I feel we probably want to not fail if the node has deleted some of the blocks between start_from and latest_block, because we could get into a bad situation where engines cannot start.

AlastairHolmes: Or atleast we need some way to allow the engine / single binary to try without the start_from if it fails.

engine/src/state_chain_observer/client.rs
```diff
@@ -292,18 +297,64 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
 
 		let mut processed_stream = {
 			let latest_block: BlockInfo = block_stream.next().await.unwrap()?;
-			let mut block_stream =
-				process_stream_fn(Box::pin(block_stream.make_try_cached(latest_block))).await?;
+
```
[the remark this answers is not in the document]
AlastairHolmes: Actually I think its ok, because we don't return the errors for the unfinalized stream i believe.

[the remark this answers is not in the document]
AlastairHolmes: ?

engine/src/state_chain_observer/client.rs
```diff
@@ -94,6 +96,20 @@ impl DefaultRpcClient {
...
+	#[error("The runtime version '{cfe_version}' is not yet compatible with the current release '{cfe_version_required}'")]
+	NotYetCompatible { cfe_version: SemVer, cfe_version_required: SemVer },
+	#[error("The runtime version '{cfe_version}' is no longer compatible with the current release '{cfe_version_required}' at block {first_incompatible_block}")]
+	NoLongerCompatible {
+		cfe_version: SemVer,
+		cfe_version_required: SemVer,
+		first_incompatible_block: state_chain_runtime::BlockNumber,
+	},
+	#[error(transparent)]
+	Other(#[from] anyhow::Error),
```
AlastairHolmes, line 110: I thought we weren't going this direction. I think you can just remove this variant.

state-chain/primitives/src/lib.rs
```diff
@@ -234,7 +234,30 @@ pub struct SemVer {
...
+
+	/// The version of the engine is no longer compatible with the runtime. Should switch to the new version.
+	NoLongerCompatible,
+}
+
 impl SemVer {
+	pub fn compatibility_with_runtime(&self, version_runtime_requires: SemVer) -> CfeCompatibility {
```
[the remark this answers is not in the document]
kylezs: Can remove the `is_compatible` easily - makes sense to leave the `is_more_recent_than` though - as that is different to using one of the variants in this function.

engine/src/state_chain_observer/client.rs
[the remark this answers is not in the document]
kylezs: My thinking was that you could just pass None to hear for the unfinalized stream... which I wasn't doing. But I think by doing it that way then it works sensibly - i.e. we don't have a start_from block for unfinailsed at all

engine/src/state_chain_observer/client.rs
```diff
@@ -279,52 +295,75 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+											.check_block_compatibility(block_info.hash)
+											.await?;
+
+									match cfe_compatibility {
+										CfeCompatibility::Compatible => false,
+										// We want the stream to continue as before
+										CfeCompatibility::NotYetCompatible => {
+											info!(
+												"{} WAITING for a compatible release version.",
+												incompatible_block_message(
```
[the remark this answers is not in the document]
kylezs: I forgot to inline this - I don't think it makes sense now to log it in the other cases, since we're just exiting, and the error will be logged at the top level

state-chain/primitives/src/lib.rs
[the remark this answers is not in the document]
kylezs: > You could probably also merge change CreateBlockStreamError in be a sub-enum of a Incompatible variant of CfeCompatibility (i.e. have two variants instead of three).

Maybe I'm not fully understanding the suggestion, but I'm not convinced of this, we'd be leaking the CreateStateChainClientError into the SC? e.g. where we do `NodeCFEVersion::<T>::get(validator_id).compatibility_with_runtime(version) ==
						CfeCompatibility::Compatible` in the validator pallet. 

Also having it as a tuple is nice because the version doesn't have to be unpacked from every variant separately

engine/src/state_chain_observer/client.rs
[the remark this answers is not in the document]
kylezs: I've updated it to pass None for the unfinalised

engine/src/state_chain_observer/client.rs
kylezs: This will only be the case if there is a `start_from`. Basically the start_from would only ever be provided on the handover. So if the engine exited and was restarted it'd be started directly from the new version, and therefore not have a `start_from` and so it would just start as it does now.

engine/src/state_chain_observer/client.rs
```diff
@@ -265,66 +284,137 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
-			block_stream
+			let (cfe_compatibility, latest_version_runtime_requires) =
+				base_rpc_client.check_block_compatibility(first_block.hash).await?;
+
+			match cfe_compatibility {
+				CfeCompatibility::NoLongerCompatible => {
+					return Err(CreateStateChainClientError::NoLongerCompatible {
+						cfe_version: *CFE_VERSION,
+						cfe_version_required: latest_version_runtime_requires,
+						first_incompatible_block: first_block.number,
```
AlastairHolmes, line 346: This is not really the `first_imcompatible_block`, I think maybe renaming to `at_block`, or something is more accurate. Because the version could have been changes many blocks ago.

kylezs: fair

engine/src/state_chain_observer/client.rs
```diff
@@ -209,6 +227,7 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
 		base_rpc_client: Arc<BaseRpcClient>,
 		wait_for_required_version: bool,
 		error_on_incompatible_block: bool,
+		start_from: Option<u32>,
```
AlastairHolmes, line 230: state_chain_runtime::BlockNumber

engine/src/state_chain_observer/client.rs
```diff
@@ -265,66 +284,137 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
-						)
-					},
-				)?;
-			}
+				Box::pin(block_stream.make_try_cached(latest_block))
+			};
+
+			let first_block = *block_stream.cache();
+
+			let mut block_stream = process_stream_fn(block_stream).await?;
```
AlastairHolmes, line 336: The process_stream_fn skips forward in the stream, so kind makes the start_from option do nothing. We'll need to think about how make these cases interact. Initially it seems to make sense to move this start from logic into the process_stream_fn only of the finalized case.

AlastairHolmes: Also the stream is still sparse here, is you may miss the first block of compatibility or incompatibly (Not sure if that matters)

state-chain/primitives/src/lib.rs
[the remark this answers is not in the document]
kylezs: I see what you're saying, I'll have a look

engine/src/state_chain_observer/client.rs
```diff
@@ -265,66 +284,137 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+							Error = anyhow::Error,
+							Ok = BlockInfo,
+							Item = Result<BlockInfo, anyhow::Error>,
+						> + Send,
+				>,
+			> = if let Some(start_from) = start_from {
+				let base_rpc_client = base_rpc_client.clone();
+
+				let mut start_from_headers = Box::pin(
+					futures::stream::iter(start_from..latest_block.number)
```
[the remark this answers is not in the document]
AlastairHolmes: Look below, you unwrap on the stream

engine/src/state_chain_observer/client.rs
AlastairHolmes: Also you check the first_block here for compatiblity, but then the stream is moved forward to a possibly uncompatible block, that is not checked. Which I think means the StateChainClient::new could pass out a stream with a incompatible cached block.

engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
```diff
@@ -414,8 +414,15 @@ impl<'a, 'env, BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static>
 			}
 			self.block_cache.push_back((
 				block_hash,
-				if self.base_rpc_client.check_block_compatibility(block_hash).await?.0 ==
-					CfeCompatibility::Compatible
+				if self
+					.base_rpc_client
+					.check_block_compatibility(BlockInfo {
+						parent_hash: block.block.header.parent_hash,
```
AlastairHolmes, line 420: There is an into impl from header to BlockInfo

engine/src/state_chain_observer/client.rs
kylezs: > Also you check the first_block here for compatiblity, but then the stream is moved forward to a possibly uncompatible block, that is not checked. Which I think means the StateChainClient::new could pass out a stream with a incompatible cached block.

I guess taking the first block *after* the stream processing will protect against this

AlastairHolmes: yeah

kylezs: > The process_stream_fn skips forward in the stream, so kind makes the start_from option do nothing. We'll need to think about how make these cases interact. Initially it seems to make sense to move this start from logic into the process_stream_fn only of the finalized case.

Yeah I considered this - but thought there might be a case for having it in the unfinalised too - but since we're not using it for unfinalised anyway, may make sense to move it in there

engine/src/main.rs
```diff
@@ -220,5 +222,14 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
 		}
 		.boxed()
 	})
-	.await
+	.await;
+
+	if let Err(e) = &root_result {
```
AlastairHolmes, line 227: You can inspect_err now I think

engine/src/main.rs
```diff
@@ -220,5 +222,14 @@ async fn run_main(settings: Settings) -> anyhow::Result<()> {
...
+		let sc_error = e.downcast_ref::<CreateStateChainClientError>();
```
AlastairHolmes, line 228: Can't you inline this

engine/src/state_chain_observer/client.rs
```diff
@@ -491,23 +524,77 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+										latest_block_from_cache,
+									))))
+									.chain(finalized_block_stream)
+									.make_try_cached(start_block),
+							)
+						},
+						_ => {
+							// No need to fetch any earlier blocks, we just have to move the stream
+							// to the correct starting point.
+							let skip_to = start_from.unwrap_or(finalized_header.number + 1);
```
kylezs, line 587: @AlastairHolmes this mimics the previous code, having the +1 effectively becomes the inclusive range. However, it's not clear to me that we actually want this. Skipping the inclusive range means that the next block that we pull on the stream will be the block *after* the header number - but if the header from the RPC call is the latest block, why would we want to do this, we're just delaying 6 seconds potentially?

engine/src/state_chain_observer/client.rs
```diff
@@ -265,66 +277,80 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+					// return out the error and let the runner start the other version.
+					if wait_for_required_version {
+						let incompatible_blocks =
+							block_stream
+								.by_ref()
+								.try_take_while(|block_info| {
+									let base_rpc_client = base_rpc_client.clone();
+									let block_info = *block_info;
+									async move {
+										Ok({
```
AlastairHolmes, line 308: Also I'd suggest removing the Ok( wrapping here, so you can avoid the return below.

engine/src/state_chain_observer/client.rs
```diff
@@ -265,66 +277,80 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
-					},
-				)?;
+						.into());
+					}
+				},
+				CfeCompatibility::Compatible => block_stream,
 			}
-
-			block_stream
 		};
```
AlastairHolmes, line 355: I'd like to add an assert here, that the cached block is compatible.

AlastairHolmes: in the processed_stream

engine/src/state_chain_observer/client.rs
```diff
@@ -265,66 +277,80 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
-					|latest_block_version| {
-						anyhow::anyhow!(
-							"{}",
-							incompatible_block_message(latest_block_version, latest_block)
+			let first_block = *block_stream.cache();
+
+			let block_compatibility =
+				base_rpc_client.check_block_compatibility(first_block).await?;
+
+			match block_compatibility.compatibility {
```
AlastairHolmes, line 291: Sorry I believe there is another problem, if the engine starts between the upgrade, i.e. the finalized block is the old version, and the unfinalized is the new version. The old engine will creation of the unfinalized stream will error here. And incorrectly state the start_from block.

AlastairHolmes: I think the unfinalized case needs to wiat like to didn't before, instead of erroring.

AlastairHolmes: You could use the exist bool `error_on_incompatible_block`

AlastairHolmes: Seems like a simple fix

AlastairHolmes: Probably makes sense for both the NotYet, and NoLonger cases?

kylezs: What do you mean? We don't have a start_from for the unfinalised?

engine/src/state_chain_observer/client.rs
AlastairHolmes: Pretty sure the old doesn't do that. This was something I thought about alot.

engine/src/state_chain_observer/client.rs
```diff
@@ -491,23 +524,77 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
-						if finalized_block_stream.cache().number < finalised_header.number {
-							let blocks_to_skip =
-								finalized_block_stream.cache().number + 1..=finalised_header.number;
-							for block_number in blocks_to_skip {
+					let finalized_hash = base_rpc_client.latest_finalized_block_hash().await?;
+					let finalized_header = base_rpc_client.block_header(finalized_hash).await?;
+
+					let base_rpc_client = base_rpc_client.clone();
+					let mut finalized_block_stream: Pin<
+						Box<
```
AlastairHolmes, line 538: If you write the branches in the other order, you will not need to put that explicit type here. You'll need to use an if instead of match, I think because I don't think you can put a wildcard as the first condition.

AlastairHolmes: "I think because I do not think"

- René Descartes

engine/src/state_chain_observer/client.rs
```diff
@@ -491,23 +524,77 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+							dyn TryCachedStream<
+									Ok = BlockInfo,
+									Error = anyhow::Error,
+									Item = Result<BlockInfo, anyhow::Error>,
+								> + Send,
+						>,
+					> = match start_from {
+						Some(start_from) if start_from < finalized_block_stream.cache().number => {
+							// save this to avoid fetching the same block again
+							let latest_block_from_cache = *finalized_block_stream.cache();
```
AlastairHolmes, line 548: This can be moved down, close to where it is used

AlastairHolmes: Maybe put it in a block?

engine/src/state_chain_observer/client.rs
```diff
@@ -491,23 +524,77 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+							);
+
+							let start_block = start_from_headers
+								.next()
+								.await
+								.expect("start_from is less than block stream cache number")?;
+
+							Box::pin(
+								start_from_headers
+									.chain(futures::stream::once(futures::future::ready(Ok(
```
AlastairHolmes, line 583: I think there are some other places we do this, could factor this idea out, into a function on the cached stream that removes the cache and adds it on the front of the stream.

engine/src/state_chain_observer/client.rs
AlastairHolmes: Just to be clear, I think we should remove the +1, as that is how the old code works. Which having looking now, does appear to be the case.

engine/src/state_chain_observer/client.rs
```diff
@@ -491,23 +524,77 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
...
+						Some(start_from) if start_from < finalized_block_stream.cache().number => {
+							// save this to avoid fetching the same block again
+							let latest_block_from_cache = *finalized_block_stream.cache();
+
+							// we can make this range exclusive because we've saved the cached block
+							// above, which we chain later.
+							let earlier_blocks_to_fetch =
+								start_from..finalized_block_stream.cache().number;
+
+							let base_rpc_client = base_rpc_client.clone();
```
AlastairHolmes, line 555: My only real concern is this piece of code, seems quite messy, but nothing comes to mind to fix it immediately.

AlastairHolmes: So should be good to go

engine/src/state_chain_observer/client.rs
```diff
@@ -355,6 +355,14 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
 		let (mut block_sender, block_receiver) = spmc::channel::<BlockInfo>(BLOCK_CAPACITY);
 
 		let latest_block = *processed_stream.cache();
+
+		{
+			let block_compatibility =
```
AlastairHolmes, line 360: Cannot you inline this?

engine/src/state_chain_observer/client.rs
kylezs: Using an `if` then means I have to have two else's that return the same error which is a little ugly, that's why I did it this way

engine/src/state_chain_observer/client.rs
kylezs: The cases are all slightly different it seems, don't think it's super obvious which way to factor it out at this point

engine/src/state_chain_observer/client.rs
```diff
@@ -288,9 +288,14 @@ impl<BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static, SignedExtr
 
 			match block_compatibility.compatibility {
 				CfeCompatibility::NoLongerCompatible =>
-					return Err(
-						CreateStateChainClientError::CompatibilityError(block_compatibility).into()
-					),
+					if error_on_incompatible_block {
```
AlastairHolmes, line 291: This case should still error, or should faill thorugh the to wait_for_compatibly case otherwise we will return a incpmpatible block from the stream i think.

kylezs: Ah yep you're right, we do that check in between, now it'd panic, thought it'd fall through to be caught on the next part 👍 will fix

AlastairHolmes: Ok, its kind of hard, in this case (in the unfinalized stream case), if we get here, what we want to wait for the finalized stream to fail. We may need to remove the cache the from stream, so the unfinalized stream doesn't need an initial block.

AlastairHolmes: So that we don't need to check compatibility to create the unfinalized stream in the first place.

AlastairHolmes: We'd just filter the incompatible blocks. As we already do, below.

AlastairHolmes (approved):

PR #4564, opened 2024-02-26
fix: more lenient max deposit fee

kylezs:
Fixes bouncer on main, depending on the order of the txs, the fees can increase making these very tight bounds be exceeded - just widening the bounds a bit more.


kylezs:
It's not just testing these particular amounts, it's testing the whole end-to-flow, so I think it's still valuable

PR #4576, opened 2024-02-27
fix: disable try-state checks

dandanlen:
Disabling these checks since they throw a lot of false positives. 

We can re-enable when the below issue is resolved:

https://github.com/paritytech/polkadot-sdk/issues/2560

### DIFF
--- a/bouncer/shared/try_runtime_upgrade.ts
+++ b/bouncer/shared/try_runtime_upgrade.ts
@@ -13,6 +13,7 @@ function tryRuntimeCommand(runtimePath: string, blockParam: string, networkUrl: 
   const stderrFile = path.join(os.tmpdir(), `cmd-stderr-${Date.now()}`);
   try {
     execSync(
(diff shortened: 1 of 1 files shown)


bouncer/shared/try_runtime_upgrade.ts
```diff
@@ -13,7 +13,7 @@ function tryRuntimeCommand(runtimePath: string, blockParam: string, networkUrl:
   const stderrFile = path.join(os.tmpdir(), `cmd-stderr-${Date.now()}`);
   try {
     execSync(
-      `try-runtime --runtime ${runtimePath} on-runtime-upgrade --disable-spec-version-check --disable-idempotency-checks --checks all ${blockParam} --uri ${networkUrl} 2> ${stderrFile}`,
+      `try-runtime --runtime ${runtimePath} on-runtime-upgrade --disable-spec-version-check --disable-idempotency-checks --checks pre-and-post ${blockParam} --uri ${networkUrl} 2> ${stderrFile}`,
```
kylezs, line 16: could we drop a TODO with a link to the issue here too

kylezs (approved):

PR #4581, opened 2024-02-28
feat: expose command for broker fee withdrawal

marcellorigotti:
# Pull Request

Closes: PRO-1219

## Checklist

## Summary

Expose command for broker fee withdrawal

#### Non-Breaking changes


api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -108,6 +115,22 @@ impl RpcServer for RpcServerImpl {
...
+	) -> RpcResult<String> {
+		Ok(self
+			.api
+			.broker_api()
+			.withdraw_fee_asset(
+				asset,
+				clean_foreign_chain_address(asset.into(), &destination_address)?,
+			)
+			.await
+			.map(|tx_hash| format!("{tx_hash:#x}"))?)
```
dandanlen, line 132: As discussed, better to return the hash here and rely on Serialize rather than converting to a string explicitly.

api/lib/src/lib.rs
```diff
@@ -356,6 +356,22 @@ pub trait BrokerApi: SignedExtrinsicApi {
...
+		destination_address: EncodedAddress,
+	) -> Result<H256> {
+		let (tx_hash, ..) = self
+			.submit_signed_extrinsic(RuntimeCall::from(pallet_cf_swapping::Call::withdraw {
+				asset,
+				destination_address,
+			}))
+			.await
+			.until_in_block()
+			.await
```
dandanlen, line 371: At this point we can also get the egress details, we should return these too.

api/lib/src/lib.rs
```diff
@@ -356,6 +361,42 @@ pub trait BrokerApi: SignedExtrinsicApi {
...
+			.context("Request to withdraw broker fee for ${asset} failed.")?;
+
+		if let Some(state_chain_runtime::RuntimeEvent::Swapping(
+			pallet_cf_swapping::Event::WithdrawalRequested {
+				egress_amount: _,
+				egress_fee: _,
+				destination_address: _,
```
dandanlen, line 383: Why not return all of the information? All of this is interesting to the broker.

api/bin/chainflip-cli/src/main.rs
```diff
@@ -75,6 +75,23 @@ async fn run_cli() -> Result<()> {
...
+								params.asset.into(),
+								&params.destination_address,
+							)?,
+						)
+						.await?;
+					println!(
+						"Withdrawal request successfull submitted. Tx hash: {:#x}",
+						withdraw_details.tx_hash
+					);
+					println!("Egress id: {:?}", withdraw_details.egress_id);
```
dandanlen, line 93:
```suggestion
					println!(
						"Withdrawal request successfully submitted: {}",
						withdraw_details
					);
```

(and implement Display for WithdrawalDetails.

api/lib/src/lib.rs
```diff
@@ -356,6 +361,42 @@ pub trait BrokerApi: SignedExtrinsicApi {
...
+				state_chain_runtime::RuntimeEvent::Swapping(
+					pallet_cf_swapping::Event::WithdrawalRequested { .. }
+				)
+			)
+		}) {
+			Ok(WithdrawFeesDetail { tx_hash, egress_id: *egress_id })
+		} else {
+			bail!("No WithdrawalRequested event was found");
+		}
+		// Ok(tx_hash)
```
dandanlen, line 398: nit: stale comment

marcellorigotti:
I left the impl of Deserialize for EncodedAddress as unimplemented, is that ok? 
I needed to implement it to satisfy the trait bunds but we never really use it if I am correct

state-chain/chains/src/address.rs
```diff
@@ -68,6 +68,28 @@ pub enum EncodedAddress {
...
+		}
+	}
+}
+
+impl<'de> Deserialize<'de> for EncodedAddress {
+	fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
+	where
+		D: serde::Deserializer<'de>,
+	{
+		unimplemented!()
```
dandanlen, line 89: I would rather we didn't have this, seems a bit dangerous. 

The implementation is just `clean_foreign_chain_address` from the api crate, but refactoring and moving all this is beyond the scope of this PR. 

I'll make the change and then we can merge.

marcellorigotti: thanks

dandanlen (approved):

dandanlen:
Thanks @marcellorigotti

PR #4584, opened 2024-02-29
feat: use swapping queue when swapping network fee for burn

j4m1ef0rd:
# Pull Request

Closes: PRO-1214

## Checklist

## Summary

- Created a `SwapQueueApi` so the pools pallet can add the swap to the queue
- Created a new swap type that is only used for this network fee swap.
- Moved the FlipToBurn storage item to the swapping pallet.
	- Added migration for this and tested it against perseverance.


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1224,3 +1234,16 @@ pub mod pallet {
...
+impl<T: Config> SwapQueueApi for Pallet<T> {
+	type BlockNumber = BlockNumberFor<T>;
+
+	fn schedule_swap(
+		from: Asset,
+		to: Asset,
+		amount: AssetAmount,
+		swap_type: SwapType,
+	) -> (u64, Self::BlockNumber) {
+		Self::schedule_swap_internal(from, to, amount, swap_type)
```
kylezs, line 1247: I don't think we want to do the MaximumSwapAmount checks for a network fee swap

kylezs: if we add it to the SwapQueue and use `Self::swap_into_stable_taking_network_fee(&mut swaps)?;` which is what's happening now, then we'll be taking a network fee for the network fee swap. which we don't want either. 
We should add a test for this case - after the network swap goes through then we should have nothing left to swap in the Network fee storage

kylezs: I'm wondering if it makes sense to have a second queue here. What matters isn't that it's in the `SwapQueue` specifically - but that it is queued by the delay and it shows in the subscription. I think this is probably the cleaner solution than trying to retrofit this into the existing flow which takes network fees etc. Also means we get the ability to emit a separate event for the network swap (if we wanted to). 

CC: @dandanlen ^

dandanlen: I don't think we really care that much if the network fee swap also pays a fee. Adding another queue raises more questions than it answers (do bundle it with other swaps? why/why not? what about delays, slippage limits etc...) 

Much simpler to treat it like a normal swap - notify the LPs, bundle it with everything else, recover like everything else.

kylezs: Yeah I guess it's not a big deal

dandanlen: If we really care about this, we can address it in the swap bundling / unbundling logic. Fee Swaps are always USDC->FLIP, so they happen on the second leg. We could just add the network fee swap to the bundle after we take the network fee. 

This would be less invasive than adding a whole new queue.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1008,7 +1018,7 @@ pub mod pallet {
 
 		/// Schedule the swap, assuming all checks already passed. Return swap_id along
 		/// with the block at which the swap is scheduled to be executed.
-		fn schedule_swap_internal(
+		pub fn schedule_swap_internal(
```
msgmaxim, line 1021: It is not very "internal" if you make it public 😄 Also, looks like the `pub` is not actually needed?

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Is it not as simple as just checking swap_type in `swap_into_stable_taking_network_fee` to see if the fee should be taken? Something like:

```rust
fn swap_into_stable_taking_network_fee(
	swaps: &mut Vec<Swap>,
) -> Result<(), BatchExecutionError> {
	Self::do_group_and_swap(swaps, SwapLeg::ToStable)?;

	// Take NetworkFee for all swaps
	for swap in swaps.iter_mut() {
                 // ***** NEW CODE BEGINS *****
		if swap.swap_type == SwapType::NetworkFee {
			// Don't take network fee for network fee swaps
			continue;
		}
                 // ***** NEW CODE END *****
		debug_assert!(
			swap.stable_amount.is_some(),
		);
		let stable_amount = swap.stable_amount.get_or_insert_with(Default::default);
		*stable_amount = T::SwappingApi::take_network_fee(*stable_amount);
	}

	Ok(())
}
```

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: I think it can be inlined into the trait function

state-chain/chains/src/lib.rs
```diff
@@ -536,3 +536,11 @@ impl RetryPolicy for DefaultRetryPolicy {
...
 	}
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub enum SwapType {
+	Swap(ForeignChainAddress),
+	CcmPrincipal(SwapId),
+	CcmGas(SwapId),
+	NetworkFee,
+}
```
dandanlen, line 546: I don' think this should go in the Chains crate. 

Either we consider it part of the SwapQueueApi interface and define it there. Or we consider it a primitive and add it to primitives. But it has nothing to do with chains.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1051,6 +1054,12 @@ impl<T: Config> SwappingApi for Pallet<T> {
 			Ok(output_amount)
 		})
 	}
+
+	fn add_flip_to_burn(amount: AssetAmount) {
+		FlipToBurn::<T>::mutate(|total| {
+			total.saturating_accrue(amount);
+		});
+	}
```
dandanlen, line 1062: It might be easier/cleaner to just move the FLipToBurn storage into the Swapping pallet rather than extend the SwappingApi with this arbitrary `add_flip_to_burn` method.

state-chain/chains/src/lib.rs
j4m1ef0rd: was not sure where to put it. It can't go in primitives because it uses `ForeignChainAddress` from chains. So ill put it next to the `SwapQueueApi` for now.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -814,6 +819,22 @@ pub mod pallet {
...
+								});
+							} else {
+								debug_assert!(
+									false,
+									"NetworkFee burning should not be in asset: {:?}",
+									swap.to
+								);
+								log::error!(
+									"NetworkFee burning is not supported for asset: {:?}",
+									swap.to
```
dandanlen, line 835: We have a utility, log_or_panic, that does this.

dandanlen (commented):
This looks fine to me, will leave final approval to @kylezs / @msgmaxim 

FWIW (and this is out of scope of this PR) I think we should consider moving all of the network fee stuff into the swapping pallet. 

We now seem to have two redundant interfaces:
- `SwapQueueApi`: only used for network fee swaps?
- `SwappingApi::take_network_fee`: used to calculate the network fee.

If we make it the responsibility of the swapping pallet to trigger the fee swaps, we can remove both of the above: the swapping pallet can calculate the USDC fee, store it, and then schedule a swap (or even schedule an immediate swap every time we process a batch). 

It's not quite trivial because we have `swap_with_network_fee` which is used in some places for price estimation, and this would need a bit of a refactor. But IMO is the right direction in the longer term. Taking the network fee is something we do for swaps, it should be in the swapping pallet. It's not really a concern of the pools. (@AlastairHolmes maybe you have an opinion here)

msgmaxim (approved):

state-chain/runtime/src/lib.rs
```diff
@@ -164,7 +164,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("chainflip-node"),
 	impl_name: create_runtime_str!("chainflip-node"),
 	authoring_version: 1,
-	spec_version: 130,
+	spec_version: 131,
```
kylezs, line 166: why is the spec_version being bumped?

kylezs (approved):

PR #4594, opened 2024-03-01
feat: Introduce tx fee multiplier storage item

marcellorigotti:
# Pull Request

Closes: PRO-1077

## Checklist

## Summary

In this PR we introduced a new storage item containing the fee_multiplier, this value is going to replace the const multiplier we previously defined.
Added a new governance extrinsic to update such storage item.

The new storage item is introduced in the ingress-egress pallet, and a migration is added to set the initial value for the bitcoinIngressEgress pallet to 1,5 (same as the previously used constant), the value for Ethereum/Polkadot IngressEgress pallet is set to 1, currently these multipliers are not used.


state-chain/chains/src/dot.rs
```diff
@@ -298,6 +298,7 @@ impl FeeEstimationApi<Polkadot> for PolkadotTrackedData {
 	fn estimate_ingress_fee(
 		&self,
 		_asset: <Polkadot as Chain>::ChainAsset,
+		_fee_multiplier: FixedU128,
```
dandanlen, line 301: If we add this we should use it and then set the multiplier to 1 in the config. Otherwise changing it has not effect.

state-chain/chains/src/eth.rs
```diff
@@ -91,6 +93,7 @@ impl FeeEstimationApi<Ethereum> for EthereumTrackedData {
 	fn estimate_ingress_fee(
 		&self,
 		asset: <Ethereum as Chain>::ChainAsset,
+		_fee_multiplier: FixedU128,
```
dandanlen, line 96: Same as above.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -898,6 +903,22 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Update the fee multiplier with the provided value
+		///
+		/// Requires Governance.
+		#[pallet::call_index(7)]
+		#[pallet::weight(T::WeightInfo::ccm_broadcast_failed())]
+		pub fn update_fee_multiplier(
```
dandanlen, line 912: We can add this to the PalletConfigUpdate enum and handle it there instead of defining a new extrinsic.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -898,6 +903,22 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			new_fee_multiplier: FixedU128,
+		) -> DispatchResult {
+			T::EnsureGovernance::ensure_origin(origin)?;
+
+			FeeMultiplier::<T, I>::put(new_fee_multiplier);
+
```
dandanlen, line 919: We should emit an event (i think we should add an event `PalletConfigUpdated(PalletConfigUpdate)`)

dandanlen (commented):
Some comments - but in reading the PR, I think the multiplier belongs in the chain tracking pallet... I have an idea how we can do this.

marcellorigotti:
PR should be ready now, I created a new extrinsic since we don't have the update_pallet_config for chain-tracking pallet, lmk if that's fine 👍

dandanlen:
Looking good, i just want to make some minor changes and will rebase against main 👌

dandanlen (approved):

PR #4602, opened 2024-03-05
feat: boost extrinsic

msgmaxim:
# Pull Request

Closes: PRO-1145

## Checklist

## Summary

- New `boost_deposit` extrinsic, which takes `deposit_address` and `deposit_id`
- Only LP accounts can boost
- `deposit_address` and `deposit_id` must match an existing `PrewitnessedDeposit`
- Booster fee is deducted first from `deposit_amount` (the pre-witnessed amount), then we also deduct ingress fee
- Booster's account is debited for `deposit_amount - booster_fee` (aborting on insufficient funds)
- Note that ingress fee is computed and decucted at the point of boosting rather than when the deposit is fully witnessed
- Channel action is immediately executed with the amount minus both fees, and `DepositBoosted` event is emitted (this contains the same info as old `DepositReceived` event plus the new `booster_id` field)

- New storage item: `BoostRecords` stores for a given channel: booster's account id, boosted amount and asset (the latter is not strictly necessary, but I added it so we can do a sanity check)
- Once the channel is boosted, it cannot be boosted again until the deposit (the amount must match exactly) is fully witnessed or the channel is recycled (booster loses money in this case). This is ensured by checking `BoostRecords`.
- When a deposit is witnessed and If there is no boost, the behaviour is the same as before (note however that we renamed `DepositReceived -> DepositFinalised`)
- When a deposit is witnessed, and there is a boost, but the amounts don't match, the boost will be ignored (see `multiple_deposits` test)
- When a deposit is witnessed and there is a matching boost, the booster gets credited the entire amount (no ingress is payed at this point); `DepositFinalised` is also emitted with `BoosterCredited` action
- I factored out `perform_channel_action` so it can be called from two places

Note that I added new fields to `PrewitnessedDeposit`. Technically we would need a migration, but I was hoping that the code with this storage item hasn't been deployed, and we can skip the migration?


msgmaxim:
Turning this to a draft since we decided to change the scope of this PR to include passive boosting (and remove boost extrinsic)

msgmaxim:
Will make a new PR when ready

PR #4611, opened 2024-03-06
fix: swap subscription amounts as hex

msgmaxim:
# Pull Request

## Checklist

## Summary

Upading amounts to be in hex as discussed in https://github.com/chainflip-io/chainflip-docs/pull/203#discussion_r1507172631. I chose to remove nesting from `ScheduledSwap` and move it out of runtime (`NumberOrHex` seems like an RPC-only concept). Tested on localnet, this is what the output looks like now:

```
{"jsonrpc":"2.0","method":"cf_subscribe_scheduled_swaps","params":{"subscription":"XSY3MSys5BEB6nOx","result":{"block_hash":"0x27c3e08998aeda7b41071d90d2d9162a48d543c109223df66a7c6b0854c04714","block_number":242,"swaps":[{"swap_id":2,"base_asset":{"chain":"Ethereum","asset":"FLIP"},"quote_asset":{"chain":"Ethereum","asset":"USDC"},"side":"sell","amount":"0x1ad54e6809d4cf6e69","execute_at":243}]}}}
```

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -9,8 +9,8 @@ use cf_chains::{
 };
 use cf_primitives::{
 	chains::assets::any::{self, OldAsset},
-	AccountRole, Asset, AssetAmount, BroadcastId, ForeignChain, NetworkEnvironment, SemVer,
-	SwapOutput,
+	AccountRole, Asset, AssetAmount, BlockNumber, BroadcastId, ForeignChain, NetworkEnvironment,
+	SemVer, SwapId, SwapOutput,
 };
 use cf_utilities::rpc::NumberOrHex;
 use core::ops::Range;
@@ -22,6 +22,7 @@ use jsonrpsee::{
 };
 use pallet_cf_governance::GovCallHash;
 use pallet_cf_pools::{AskBidMap, PoolInfo, PoolLiquidity, PoolPriceV1, UnidirectionalPoolDepth};
+use pallet_cf_swapping::SwapLegInfo;
 use sc_client_api::{BlockchainEvents, HeaderBackend};
 use serde::{Deserialize, Serialize};
 use sp_api::ApiError;
@@ -35,7 +36,7 @@ use state_chain_runtime::{
 	constants::common::TX_FEE_MULTIPLIER,
 	runtime_apis::{
 		BrokerInfo, CustomRuntimeApi, DispatchErrorWithMessage, FailingWitnessValidators,
-		LiquidityProviderInfo, ScheduledSwap, ValidatorInfo,
+		LiquidityProviderInfo, ValidatorInfo,
 	},
 	NetworkFee,
 };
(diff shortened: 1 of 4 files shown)


state-chain/custom-rpc/src/lib.rs
```diff
@@ -45,6 +46,38 @@ use std::{
...
 };
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ScheduledSwap {
+	pub swap_id: SwapId,
+	pub base_asset: Asset,
+	pub quote_asset: Asset,
+	pub side: Side,
+	pub amount: NumberOrHex,
+	#[serde(skip_serializing_if = "Option::is_none")]
```
dandanlen, line 56: Is this not the default behaviour? (skipping of None values)

state-chain/custom-rpc/src/lib.rs
```diff
@@ -45,6 +46,38 @@ use std::{
 	sync::Arc,
 };
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ScheduledSwap {
```
dandanlen, line 50: Rather than repeating this, couldn't this type be defined as:

```
pub struct ScheduledSwap {
	#[serde(flatten)]
	swap_leg: SwapLegInfo,
	execute_at: BlockNumber,
}
```

(and then derive de/serialize traits on SwapLegInfo with appropriate annotations).

Do we do it this way just so we can use NumberOrHex? 


*edit: it looks like this how it was done previously* 😅

dandanlen: Can we use serde_with?

```rust
#[serde(with = "hex::serde")]
amount: AssetAmount
```

If we're happy to use NumberOrHex we also have a utility:
```
#[serde(with = "utilities::serde::number_or_hex")]
```

state-chain/custom-rpc/src/lib.rs
```diff
@@ -45,6 +46,38 @@ use std::{
...
+	pub swap_id: SwapId,
+	pub base_asset: Asset,
+	pub quote_asset: Asset,
+	pub side: Side,
+	pub amount: NumberOrHex,
```
AlastairHolmes, line 55: Can you use U256 instead (same below), we're trying to remove NumberOrHex, it doesn't behave how you think that it does. The serialization changes depending on the size of the number. But U256 is always hex, but otherwise this is good to go imo.

AlastairHolmes (approved):
Note my comment, once that is done happy to merge.

state-chain/custom-rpc/src/lib.rs
AlastairHolmes: This is want the other rpcs do (use U256)

state-chain/custom-rpc/src/lib.rs
dandanlen: We could create a `#[serde(with = "utilities::serde::u256")]`... If you feel like it. Otherwise I don't mind duplicating with explicit conversion to U256.

state-chain/custom-rpc/src/lib.rs
msgmaxim: > it doesn't behave how you think that it does

I did check that `u128` always encodes as hex before using this.

> This is what the other rpcs do (use U256)

I see plenty of rpcs using `NumberOrHex` for amounts. Hence, I wasn't sure if we wanted to use `NumberOrHex` or `U256`. Happy to change to `U256`.

state-chain/custom-rpc/src/lib.rs
msgmaxim: I experimented with adding `utilities::serde::u256`, but looks like it doesn't compose nicely with `Option`. I'm sure it is possible to make it work, but didn't seem worth spending more time on it, so I will keep the separate struct for now.

state-chain/custom-rpc/src/lib.rs
msgmaxim: This is what the output looks like if I don't skip explicitly:
```
{"source_asset":null,"source_amount":null, ...}]
```
I imagine the product can still consume this, but last I heard we were going to skip the fields altogether.

msgmaxim:
Happy to merge this @dandanlen ?

state-chain/custom-rpc/src/lib.rs
dandanlen: Yes, I think skipping is preferred, I'm just surprised we have to explicitly annotate this.

PR #4612, opened 2024-03-06
chore: migration housekeeping

dandanlen:
Adds a housekeeping migration to clean up state when we upgrade the network.

### DIFF
--- a/state-chain/runtime/src/migrations/housekeeping.rs
+++ b/state-chain/runtime/src/migrations/housekeeping.rs
@@ -1,0 +1,36 @@
+use crate::{EthereumInstance, PolkadotInstance, Runtime};
+use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
+
+pub struct Migration;
+
+impl OnRuntimeUpgrade for Migration {
+	fn on_runtime_upgrade() -> Weight {
+		use cf_runtime_upgrade_utilities::genesis_hashes;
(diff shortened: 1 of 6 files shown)


state-chain/runtime/src/migrations/housekeeping.rs
```diff
@@ -0,0 +1,36 @@
...
+				pallet_cf_broadcast::AbortedBroadcasts::<Runtime, EthereumInstance>::mutate(
+					|all| all.retain(|id| ![964, 1093].contains(id)),
+				);
+				pallet_cf_broadcast::AbortedBroadcasts::<Runtime, PolkadotInstance>::mutate(
+					|all| all.retain(|id| ![168, 169, 170].contains(id)),
+				);
+				// Corrupted storage from a previous runtime upgrade.
+				frame_support::storage::unhashed::kill(
+					// Raw storage key retrieved from try-runtime error message.
+					&hex_literal::hex!("09f888937e67e4859e4ee6a943cc9e08347279749a2449c0b500c3a2462071c57f7df985518de74827010000")[..],
```
kylezs, line 22: do we know what this is/was?

kylezs (approved):

state-chain/runtime/src/migrations/housekeeping.rs
dandanlen: Yes it was an entry in bitcoinBroadcaster.thresholdSignatureData. It's old, pre-dates the last migration.

PR #4613, opened 2024-03-06
refactor: return error out of print_start_and_end

kylezs:
Hide whitespace for review.

## Checklist

## Summary

Required to pass out error out to the top to the runner for single CFE upgrade.


kylezs:
Actually I'll still need to make changes to this, but require changes in the main PR, so this will make more sense if I include it in the other one.

PR #4614, opened 2024-03-06
chore: allow OldAsset to support unambiguously encoding Arb USDC and Eth USDC, while maintaining backcompat

AlastairHolmes:
OldAsset will encode assets labels as "legacy_encoding" with only their name i.e. "ETH". But allows decoding "ETH" or "{"chain": "Ethereum", "asset": "ETH"}" or "{"asset": "ETH"}".

Asset's encoding remains the same.


dandanlen (approved):
Awesome.

PR #4622, opened 2024-03-07
feat: move broker_fee_collection_test inside all_concurrent_test

marcellorigotti:
# Pull Request

Closes: PRO-1175

## Checklist

## Summary

Move broker_fee_collection test inside all_concurrent test (using a different broker to avoid interference with other tests run simultaneously)


bouncer/tests/broker_fee_collection_test.ts
j4m1ef0rd, line 1: You said "move" but you actually just copied the code. Can we not just use the same functions from the "shared" file instead of duplicating the entire file?

bouncer/tests/broker_fee_collection_test.ts
```diff
@@ -29,11 +31,24 @@ const swapAssetAmount = {
   [Assets.USDC]: 1000,
 };
 const commissionBps = 1000; // 10%
+const cfNodeEndpoint = process.env.CF_NODE_ENDPOINT ?? 'ws://127.0.0.1:9944';
 
-const chainflip = await getChainflipApi();
+const chainflip = await ApiPromise.create({
```
j4m1ef0rd, line 35: Why can we not just use `getChainflipApi` here anymore?

bouncer/tests/broker_fee_collection_test.ts
marcellorigotti: My bad I forgot to change the broker_fee_collection_test to point to the code in the shared folder!

bouncer/tests/broker_fee_collection_test.ts
marcellorigotti: We need those types definition added to the API. 
I just added it to getChainflipApi so we can directly use that.

bouncer/shared/broker_fee_collection.ts
```diff
@@ -0,0 +1,248 @@
...
+    (event) =>
+      event.data.destinationAddress[destinationChain]?.toLowerCase() ===
+      observeDestinationAddress.toLowerCase(),
+  );
+  console.log(`Running swap ${asset} -> ${swapAsset}`);
+
+  const rawDepositForSwapAmount = swapAssetAmount[asset].toString();
+
+  // we need to manually create the swap channel and observe the relative event
+  // because we want to use a separate broker to not interfere with other tests
```
j4m1ef0rd, line 78: Could be worth changing `performSwap` to have an optional argument for the broker URL, then we can avoid this duplication.

j4m1ef0rd (approved):

bouncer/shared/utils.ts
```diff
@@ -221,6 +221,15 @@ function getCachedSubstrateApi(defaultEndpoint: string) {
     api = await ApiPromise.create({
       provider: new WsProvider(endpoint),
       noInitWarn: true,
+      types: {
+        EncodedAddress: {
+          _enum: {
+            Eth: '[u8; 20]',
+            Dot: '[u8; 32]',
+            Btc: '[u8; 34]',
```
dandanlen, line 229: I don't think we can guarantee that bitcoin addresses are always 34 bytes (and in fact the encoding is a Vec<u8> which is probably not compatible with [u8; 34] encoding) - is there another way to express this? For example `Vec<u8>` or `String` or something equivalent?

dandanlen (approved):

PR #4626, opened 2024-03-11
feat: only sign eth tx's once

j4m1ef0rd:
# Pull Request

Closes: PRO-893

## Checklist

## Summary

- Split the signing and send of the eth transaction
- Set the signing part of the broadcast to not retry
- Changed the retrier error logs to show more info including the root cause.

TODO: Tests.

### DIFF
--- a/engine/src/eth/retry_rpc.rs
+++ b/engine/src/eth/retry_rpc.rs
@@ -261,13 +261,16 @@ impl<Rpc: EthRpcApi> EthersRetryRpcApi for EthRetryRpcClient<Rpc> {
 
 #[async_trait::async_trait]
 impl<Rpc: EthSigningRpcApi> EthersRetrySigningRpcApi for EthRetryRpcClient<Rpc> {
-	/// Estimates gas and then sends the transaction to the network.
+	/// Estimates gas, signs and then sends the transaction to the network.
 	async fn broadcast_transaction(
 		&self,
 		tx: cf_chains::evm::Transaction,
 	) -> anyhow::Result<TxHash> {
 		let log = RequestLog::new("broadcast_transaction".to_string(), Some(format!("{tx:?}")));
-		self.rpc_retry_client
+
+		// Estimate gas
+		let transaction_request = self
+			.rpc_retry_client
 			.request_with_limit(
 				Box::pin(move |client| {
 					let tx = tx.clone();
@@ -306,9 +309,43 @@ impl<Rpc: EthSigningRpcApi> EthersRetrySigningRpcApi for EthRetryRpcClient<Rpc> 
 								estimated_gas.saturating_mul(U256::from(4u64)) / 3u64
 							},
 						});
+						Ok(transaction_request)
+					})
+				}),
+				log.clone(),
+				MAX_BROADCAST_RETRIES,
+			)
+			.await?;
 
+		// Sign the transaction
+		let tx = self
+			.rpc_retry_client
+			.request_with_limit(
+				Box::pin(move |client| {
+					let transaction_request = transaction_request.clone();
+					#[allow(clippy::redundant_async_block)]
+					Box::pin(async move {
 						client
-							.send_transaction(transaction_request)
+							.sign_transaction(transaction_request)
+							.await
+							.context("Failed to Sign ETH transaction")
+					})
+				}),
+				log.clone(),
+				// Only sign the transaction once
+				0,
+			)
+			.await?;
+
+		// Send the transaction
+		self.rpc_retry_client
+			.request_with_limit(
+				Box::pin(move |client| {
+					let tx = tx.clone();
+					#[allow(clippy::redundant_async_block)]
+					Box::pin(async move {
+						client
+							.send_raw_transaction(tx)
 							.await
 							.context("Failed to send ETH transaction")
 					})
--- a/engine/src/eth/retry_rpc.rs
+++ b/engine/src/eth/retry_rpc.rs
@@ -35,6 +35,8 @@ const ETHERS_RPC_TIMEOUT: Duration = Duration::from_millis(4 * 1000);
 const MAX_CONCURRENT_SUBMISSIONS: u32 = 100;
 
 const MAX_BROADCAST_RETRIES: Attempt = 2;
+// Only sign the transaction once
+const MAX_SIGNING_RETRIES: Attempt = 1;
 
 impl<Rpc: EthRpcApi> EthRetryRpcClient<Rpc> {
 	fn from_inner_clients<ClientFut: Future<Output = Rpc> + Send + 'static>(
@@ -332,8 +334,7 @@ impl<Rpc: EthSigningRpcApi> EthersRetrySigningRpcApi for EthRetryRpcClient<Rpc> 
 					})
 				}),
 				log.clone(),
-				// Only sign the transaction once
-				0,
+				MAX_SIGNING_RETRIES,
 			)
 			.await?;
 
--- a/engine/src/eth/rpc.rs
+++ b/engine/src/eth/rpc.rs
@@ -208,7 +208,11 @@ pub trait EthRpcApi: Send + Sync + Clone + 'static {
 pub trait EthSigningRpcApi: EthRpcApi {
 	fn address(&self) -> H160;
 
-	async fn send_transaction(&self, tx: Eip1559TransactionRequest) -> Result<TxHash>;
+	// Sign the transaction request and return the RLP encoded signed transaction bytes
+	async fn sign_transaction(&self, tx: Eip1559TransactionRequest) -> Result<Bytes>;
+
+	// Send an already signed RLP encoded transaction
+	async fn send_raw_transaction(&self, tx: Bytes) -> Result<TxHash>;
 }
 
 #[async_trait::async_trait]
@@ -260,17 +264,27 @@ impl EthSigningRpcApi for EthRpcSigningClient {
 		self.signer.address()
 	}
 
-	async fn send_transaction(&self, mut tx: Eip1559TransactionRequest) -> Result<TxHash> {
+	async fn sign_transaction(&self, mut tx: Eip1559TransactionRequest) -> Result<Bytes> {
 		tx.nonce = Some(self.get_next_nonce().await?);
+		let tx = TypedTransaction::Eip1559(tx);
 
-		let res = self.signer.send_transaction(tx, None).await;
-		if res.is_err() {
-			// Reset the nonce just in case (it will be re-requested during next broadcast)
-			tracing::warn!("Resetting eth broadcaster nonce due to error");
-			*self.nonce_info.lock().await = None;
+		self.signer
+			.signer()
+			.sign_transaction(&tx)
+			.await
+			.map_err(anyhow::Error::msg)
+			.map(|signature| tx.rlp_signed(&signature))
+	}
+
+	async fn send_raw_transaction(&self, tx: Bytes) -> Result<TxHash> {
+		match self.signer.inner().send_raw_transaction(tx).await {
+			Ok(pending_tx) => Ok(pending_tx.tx_hash()),
+			Err(e) => {
+				// Reset the nonce just in case (it will be re-requested during next broadcast)
+				*self.nonce_info.lock().await = None;
+				Err(anyhow::Error::new(e))
+			},
 		}
-
-		Ok(res?.tx_hash())
 	}
 }
 
--- a/engine/src/eth/rpc.rs
+++ b/engine/src/eth/rpc.rs
@@ -266,7 +266,9 @@ impl EthSigningRpcApi for EthRpcSigningClient {
 
 	async fn sign_transaction(&self, mut tx: Eip1559TransactionRequest) -> Result<Bytes> {
 		tx.nonce = Some(self.get_next_nonce().await?);
-		let tx = TypedTransaction::Eip1559(tx);
+		let mut tx = TypedTransaction::Eip1559(tx);
+
+		self.signer.fill_transaction(&mut tx, None).await.map_err(anyhow::Error::msg)?;
 
 		self.signer
 			.signer()
@@ -282,7 +284,7 @@ impl EthSigningRpcApi for EthRpcSigningClient {
 			Err(e) => {
 				// Reset the nonce just in case (it will be re-requested during next broadcast)
 				*self.nonce_info.lock().await = None;
-				Err(anyhow::Error::new(e))
+				Err(e.into())
 			},
 		}
 	}
--- a/engine/src/retrier.rs
+++ b/engine/src/retrier.rs
@@ -402,14 +402,18 @@ where
 
 							let attempt_number = attempt.saturating_add(1);
 							match retry_limit {
+								// Log it as an error on any attempt after the first or if there's only one attempt.
 								RetryLimit::Limit(max_attempts) if attempt_number >= max_attempts => {
-									tracing::error!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, Reached maximum attempts `{attempt_number}/{max_attempts}`: {e:#}");
+									tracing::error!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, final attempt `{attempt_number}/{max_attempts}`: {e:#}");
+								},
+								RetryLimit::Limit(max_attempts) if attempt > 0 =>{
+									tracing::error!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt_number}/{max_attempts}`: {e:#}. Delaying for {sleep_duration:?}");
 								},
 								RetryLimit::Limit(max_attempts) => {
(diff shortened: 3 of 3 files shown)


engine/src/eth/rpc.rs
```diff
@@ -260,17 +264,27 @@ impl EthSigningRpcApi for EthRpcSigningClient {
...
+	}
 
-		Ok(res?.tx_hash())
+	async fn send_raw_transaction(&self, tx: Bytes) -> Result<TxHash> {
+		match self.signer.inner().send_raw_transaction(tx).await {
+			Ok(pending_tx) => Ok(pending_tx.tx_hash()),
+			Err(e) => {
+				// Reset the nonce just in case (it will be re-requested during next broadcast)
+				*self.nonce_info.lock().await = None;
+				Err(anyhow::Error::new(e))
```
msgmaxim, line 285: A nice thing about `?` is that it calls `into()` for you. You are no longer using `?`, but you can still call `into()` manually here: `Err(e.into())`.

engine/src/retrier.rs
```diff
@@ -400,11 +400,17 @@ where
...
 							let sleep_duration = half_max + rand::thread_rng().gen_range(Duration::default()..half_max);
 
-							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {:?}", sleep_duration);
-							if attempt == 0 && !matches!(retry_limit, RetryLimit::Limit(1)) {
-								tracing::warn!(error_message);
-							} else {
-								tracing::error!(error_message);
+							let attempt_number = attempt.saturating_add(1);
+							match retry_limit {
+								RetryLimit::Limit(max_attempts) if attempt_number >= max_attempts => {
```
msgmaxim, line 405: In the past we would also log error (rather than a warn) in the specific case of `Limit = 1`. Looks like you code changes that. I'm assuming it is not intentional?

j4m1ef0rd: It was intentional. The old logic did not make sense to me, why a limit of 1? It was not clear why or when it would log as warn and when it would log as error. So now it is clear, only log error on the last try. Makes more sense.

msgmaxim: As discussed, the error is still printed in case `Limit = 1` as it would hit the first branch.

j4m1ef0rd: @kylezs Maybe you can clear up the intent behind when we want this to be logged as error/warn?
1) Log error on last try.
2) Log error on last try and any try after first.

engine/src/eth/rpc.rs
```diff
@@ -260,17 +264,27 @@ impl EthSigningRpcApi for EthRpcSigningClient {
...
 		tx.nonce = Some(self.get_next_nonce().await?);
+		let tx = TypedTransaction::Eip1559(tx);
 
-		let res = self.signer.send_transaction(tx, None).await;
-		if res.is_err() {
-			// Reset the nonce just in case (it will be re-requested during next broadcast)
-			tracing::warn!("Resetting eth broadcaster nonce due to error");
-			*self.nonce_info.lock().await = None;
-		}
+		self.signer
```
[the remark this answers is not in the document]
j4m1ef0rd: I wanted to use `SignerMiddleware::sign_transaction`, but it is private. 
I think you are right about `fill_transaction`. On closer inspection, i think it might be the only place we set the nonce. Thought it was just for the chain id and address.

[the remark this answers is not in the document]
j4m1ef0rd: Pretty sure we can skip `fill_transaction` because we set the chain id, nonce, and address ourselves, but it will not hurt to add it in. Might help in the future.

engine/src/retrier.rs
```diff
@@ -400,11 +400,17 @@ where
 							let half_max = max_sleep_duration(initial_request_timeout, attempt) / 2;
 							let sleep_duration = half_max + rand::thread_rng().gen_range(Duration::default()..half_max);
 
-							let error_message = format!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, attempt `{attempt}`: {e}. Delaying for {:?}", sleep_duration);
-							if attempt == 0 && !matches!(retry_limit, RetryLimit::Limit(1)) {
-								tracing::warn!(error_message);
-							} else {
-								tracing::error!(error_message);
+							let attempt_number = attempt.saturating_add(1);
```
kylezs, line 403: why are we incrementing an attempt number here? If it failed on attempt, 0, why do we log it as a failure for attempt 1? The attempt should only be incremented in one place, makes the code easier to read and reason about.

engine/src/retrier.rs
```diff
@@ -400,11 +400,17 @@ where
...
+							match retry_limit {
+								RetryLimit::Limit(max_attempts) if attempt_number >= max_attempts => {
+									tracing::error!("Retrier {name}: Error for request `{request_log}` with id `{request_id}`, Reached maximum attempts `{attempt_number}/{max_attempts}`: {e:#}");
```
kylezs, line 406: We don't need to log that the maximum attempt reached here - it's logged below when we actually remove the request.

engine/src/retrier.rs
kylezs: The real solution to this is this: https://linear.app/chainflip/issue/PRO-1206/timeouts-per-rpc-retrier-query-type so that we can just have them all be errors but adjust the times based on which we expect to take longer / shorter times.

> 1.  Log error on last try.
> 2. Log error on last try and any try after first.

By 2. do you mean just "any try after first" which would include the last try? This is how the code works at the moment, and is the intention. So I think we want to just leave this as is. Feel free to leave a comment there though, as I can see how that might not be clear.

engine/src/eth/retry_rpc.rs
```diff
@@ -306,9 +309,43 @@ impl<Rpc: EthSigningRpcApi> EthersRetrySigningRpcApi for EthRetryRpcClient<Rpc>
...
+					Box::pin(async move {
+						client
+							.sign_transaction(transaction_request)
+							.await
+							.context("Failed to Sign ETH transaction")
+					})
+				}),
+				log.clone(),
+				// Only sign the transaction once
+				0,
```
kylezs, line 336: should be 1 right? - technically don't think it matters as it'll short circuit at the same place - but 1 would be more inline with the `MAX_BROADCAST_RETRIES`

kylezs: also maybe a const rather than magic number

engine/src/eth/rpc.rs
kylezs, line 1: @albert-llimos if you could have a quick peak at this file's changes that'd be handy

engine/src/eth/rpc.rs
```diff
@@ -260,17 +264,29 @@ impl EthSigningRpcApi for EthRpcSigningClient {
...
+			.map_err(anyhow::Error::msg)
+			.map(|signature| tx.rlp_signed(&signature))
+	}
+
+	async fn send_raw_transaction(&self, tx: Bytes) -> Result<TxHash> {
+		match self.signer.inner().send_raw_transaction(tx).await {
+			Ok(pending_tx) => Ok(pending_tx.tx_hash()),
+			Err(e) => {
+				// Reset the nonce just in case (it will be re-requested during next broadcast)
+				*self.nonce_info.lock().await = None;
```
kylezs, line 286: Hmm I think this may be the culprit of the `replacement transaction underpriced`. The nonce is already signed over by this point, so it's strange to try resetting it at the broadcast level, resetting the nonce if there's a signing error would make more sense to me @albert-llimos

engine/src/eth/retry_rpc.rs
albert-llimos: I assume that what you mean that it will only retry if it has failed to sign. If we are sure that there is no way it can retry signing unless it has failed before, then I'm also fine with using `MAX_BROADCAST_RETRIES`. However, doing it only once is the safest way to make sure it's not double signed and this is unrelated to the broadcast retries, so I think leaving it as is is fine.

engine/src/eth/rpc.rs
albert-llimos: While I was about to agree with your statment, I actually don't think so after putting in some thought.
Signing won't fail even if the nonce is already used, it will be the broadcast that will fail in that case. Ideally we should only reset the nonce if the transaction broadcast fails due to the nonce having been used before(`replacement transaction underpriced`) and that is from a previous transaction that is still in the mempool. In the case of being the second broadcast retrial of the same tx, then it's intended. Given that we can't really know which scenario it is, it feels like the safest and correct behavior is to just reset the nonce for the next signing.

engine/src/eth/rpc.rs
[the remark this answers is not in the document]
albert-llimos: I think using `fill_transaction` seems right to me.

engine/src/eth/rpc.rs
albert-llimos: Also, there is another error, something along the lines of `nonce already used` that might be the result of the broadcast step that should also trigger a nonce reset. So the current approach makes sense to me even if it feels very counterintuitive.

engine/src/eth/retry_rpc.rs
kylezs: In this case, it's unrelated to the number of tries, it'd still be 1 try here is the point of my comment - more just how we represent that in this particular case.

engine/src/eth/rpc.rs
kylezs: Any ideas why we would see the `replacement transaction underpriced` in that case? doesn't happen consistently but it has

albert-llimos: For background , that happens when another transaction is already in the mempool with that nonce. So it's not that we are reusing a nonce used in a mined transaction.
I can't tell the exact reason of why that would happen, I would expect that to be some edge case where we query the latest nonce, which includes pending tx's, and for some reason a pending transaction is not included. However, as long as the error is not `nonce already used` I think it's alright. That would mean that we are using a nonce from a mined transaction.

albert-llimos: @kylezs @j4m1ef0rd do we have any engine log that has a `replacement transaction underpriced message`? It would be useful to know if a previous transaction has failed to broadcast or what exactly is the behavior around it.

kylezs: https://github.com/chainflip-io/chainflip-backend/actions/runs/8244260007/job/22546404358

engine/src/retrier.rs
j4m1ef0rd: > do you mean just "any try after first" which would include the last try

No, If there is no retries, then the last try is also the first, so it would not be an error. I think it's clearer to say last try and any try after first.

engine/src/retrier.rs
j4m1ef0rd: It is much clearer when it says "attempt 2/2" meaning 2 of 2. If we log it from 0, it would say "attempt 1" and you will not know if that is the first or second attempt, at a glance "attempt 1" would seem like the first attempt when in fact it was the second.

engine/src/retrier.rs
j4m1ef0rd: But it was displaying "Delaying for ..." which was deceptive, because it was the last attempt.

engine/src/eth/rpc.rs
kylezs: I just re-read what you had above Albert. You're right re the nonce error only occurring on broadcast, so doesn't make sense to move it. The direction I was thinking though was, because we signed and broadcasted in an atomic step before, it meant that if you signed and it failed to broadcast you'd reset the nonce by signing again. However, now, because we don't reset it on broadcast, it'll just broadcast the same thing again and fail again.

engine/src/retrier.rs
kylezs: As I said, we don't want to increment the attempt number in two places - this could be dangerous down the line, and is a little confusing to read in the code. As for making the log messages clearer, we can still do this, perhaps by making it clear that the attempts are 0 indexed... I'm not sure we necessary need to show this "2/2" format. As long as we show the values that are actually used (and not incremented) then it's going to be relatively easy to work out what's happening.

engine/src/retrier.rs
kylezs: Ok, but the code still works like before this change anyway? The current change just splits 2 conditions into 3.

kylezs: If it's the first attempt and there are many retries (with limit or without limit) then warn.
Otherwise log error for every attempt. (This condition includes:  If it's the first attempt and there's only one try then it will error.)

engine/src/retrier.rs
kylezs: But it still does get delayed here?

kylezs:
```rust
			// Delay the request before the next retry.
							retry_delays.push(Box::pin(
								async move {
									tokio::time::sleep(sleep_duration).await;
									// pass in primary or secondary so we know which client to use.
									(request_id, request_log, attempt, retry_limit, primary_or_secondary)
								}
							));
```

kylezs: Basically there's no need to dupe the conditions between the two branches, that was the reason it was implemented this way - keep it in one place where possible

engine/src/eth/rpc.rs
albert-llimos: > now, because we don't reset it on broadcast, it'll just broadcast the same thing again and fail again.
Yes. That makes sense if the purpose of retrying the broadcast is to handle cases where there is a temporary disconnection or some RPC issue. If it's a nonce issue that doesn't matter indeed.
So as you say, resigning if there is a nonce related error makes sense to me. We need to ensure we cover all the error cases where it can be nonce related. Do you think we can rely on that?
This would be particularly important in 1-node networks to prevent broadcast aborted when there is some nonce issue but it should be very irrelevant in live networks as some other node will be able to broadcast it anyway. We still have to implement the right approach ofc.

Besides that, I struggle to imagine what scenario could cause those `replacement transaction underpriced message`. I have added some debug prints and tried reproducing it locally and I can't reproduce it so far.

kylezs: >  We need to ensure we cover all the error cases where it can be nonce related. Do you think we can rely on that?

Yeah we couldn't really ensure this I don't think, nodes could change their response at any time. 

Question then is - is it better to resign every time (as in before this PR) or do as is currently the case in this PR?

kylezs: Yeah... bit of an annoying one. Really don't wanna introduce flakiness in localnets/CI, but can see why we wouldn't want to sign twice. There ought to be a reason we submitted with the wrong nonce anyway - so probably solving the fact that we didn't update the nonce correctly, but signing once, is the solution

albert-llimos: I think I agree with that. I will look more into it and see if I can figure out the reason for that.

kylezs: Have you had a chance to look into it @albert-llimos ?

msgmaxim: Is this just a the problem due to signing and broadcasting being scheduled separartely in the retrier? (I haven't looked how that works in detail, but that seems plausible @kylezs?) It is a little unfortunate that signing a tx has be done as a "request" even thought it is done completely locally...

albert-llimos: @kylezs I did try many times but I haven't been able to reproduce it, not even one single time. Looking at the code I can't really figure out why that would happen.
I've just been able to come up with one potential scenario that could explain this issue, related to what Maxim mentioned. There is a nonce timeout that forces a re-fetching of the nonce, which effectively looks at transcations mined and in the mempool. 
1. Transaction A is signed with nonce 5
2. Transaction A is requested for broadcast but not broadcasted yet
3. Transaction B is tried to be signed. It hits the timeout and fetches the latest nonce. Since Transaction A is not broadcasted yet, it gets nonce 5
4. Transaction B is now signed with nonce 5.
5. Either Transaction A or B will be rejected by the blockchain client with `replacement transaction underpriced`.

Another possibility, is that a broadcast request has failed, and we therefore reset the nonce, but for some reason the blockchain client has gotten the transaction anyway. Then we are trying to broadcast it again, which will return the mentioned error and I suspect it might end up with a BroadcastAborted.

You guys tell me if any of these two explanations make any sense to you.

kylezs: > It is a little unfortunate that signing a tx has be done as a "request" even thought it is done completely locally...

I guess it doesn't have to be this way. We could put a synchronous signer alongside the retriers for these local requests if we wanted. We can do this. Makes it conceptually clearer for sure.

kylezs: > Is this just a the problem due to signing and broadcasting being scheduled separately in the retrier?

It could be, especially given that when they were scheduled together we didn't see this issue. But I think I'm willing to say this is exceedingly rare. And get this PR in (after improving the above, having the signer be available for signing without going through the retrier (which is intended to be for RPC requests primarily.

kylezs: Yeah they seem reasonable @albert-llimos, as above comment, I think they are quite unlikely and that only signing once is a slight improvement, as we will tend to use less gas on failed txs.

j4m1ef0rd:
Stale PR. Going to try a different approach with more of a refactor.

PR #4634, opened 2024-03-12
feat: allow for single binary CFE upgrades

kylezs:
Picked changes for 1.3: https://github.com/chainflip-io/chainflip-backend/pull/4782

# Pull Request

Closes: PRO-935

## Checklist

- [x] I am confident that the code works - will continue testing with some testnet operators
  - [ ] Update the release notion document - to do after merge

## Summary

Allows us to do an upgrade by releasing a single CFE binary with a `sudo apt update + upgrade`.

An overview of some of the components can be found in the `engine-runner-bin` README file.


engine/src/lib.rs
```diff
@@ -21,3 +21,300 @@ pub mod witness;
...
+use futures::FutureExt;
+use libc::c_char;
+use std::{
+	sync::{atomic::AtomicBool, Arc},
+	time::Duration,
+};
+use utilities::{cached_stream::CachedStream, metrics, task_scope::task_scope};
+
+// The `cfe_entrypoint` macro adds the required C parameters to the function signature
+#[cfe_entrypoint]
```
kylezs, line 59: Most of this file is just moved to lib, as main is no longer required. The main change is using this macro to generate the entrypoint.

engine-runner/src/main.rs
```diff
@@ -0,0 +1,5 @@
+use engine_proc_macros::engine_runner;
+
+// Specify the version upgrading from and then the version being upgraded to.
+// e.g. engine_runner!("1.3.0", "1.4.0");
+engine_runner!("1.3.0", "1.4.0");
```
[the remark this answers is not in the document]
kylezs: https://linear.app/chainflip/issue/PRO-1252/inject-current-engine-version-into-the-engine-runner

engine/src/lib.rs
```diff
@@ -21,3 +21,299 @@ pub mod witness;
...
+			let has_completed_initialising = Arc::new(AtomicBool::new(false));
+
+			let (state_chain_stream, unfinalised_state_chain_stream, state_chain_client) =
+				state_chain_observer::client::StateChainClient::connect_with_account(
+					scope,
+					&settings.state_chain.ws_endpoint,
+					&settings.state_chain.signing_key_file,
+					AccountRole::Validator,
+					true,
+					// TODO: remove wait for required version entirely
```
kylezs, line 130: I will do this in another PR. Because we're not running the engines side by side, we should be able to remove this now. (note this TODO: wasn't introduced here, the intention was to remove it once this PR is in)

kylezs: We don't need `wait_for_required_version` any more because we should never wait. The SCC should always fail out so that the runner can then start the required version.

engine-runner-bin/README.md
```diff
@@ -0,0 +1,35 @@
+# Chainflip Engine (Runner)
+
+This README describes how the Chainflip Engine (CFE) actually runs and how the components interact to come together into a single runnable binary.
+
+The Chainflip Engine as it is run by a validator, consists of 3 build artefacts:
+
+1. The `enginer-runner` binary.
```
dandanlen, line 7: typo here

engine-runner-bin/README.md
```diff
@@ -0,0 +1,35 @@
...
+2. A shared library of the old [engine-dylib](./../engine-dylib/). This is fetched from the previous release build when a release is built. Or directly from the repo ([`old-engine-dylib](./../old-engine-dylib/)) for localnets
```
dandanlen, line 8: How does this fetch happen? Does cargo access github or somthing?

engine-upgrade-utils/src/lib.rs
```diff
@@ -142,5 +151,11 @@ fn test_c_str_array_with_args() {
 	let args = vec!["arg1".to_string(), "arg2".to_string()];
 
 	let c_args: CStrArray = args.clone().try_into().unwrap();
+	// check the Clone/drop implementations
+	{
+		let c_args_2: CStrArray = c_args.clone();
+		drop(c_args_2);
```
dandanlen, line 157: Does this actually check anything? Don't we have to check memory or something?

engine-runner-bin/build.rs
```diff
@@ -0,0 +1,68 @@
...
+
+	let build_dir_str = build_dir.to_str().unwrap(); // target/debug or target/release
+
+	println!("cargo:rustc-link-search=native={old_version_str}");
+	println!("cargo:rustc-link-search=native={build_dir_str}");
+
+	let old_version_suffix = OLD_VERSION.replace('.', "_");
+	let new_version_suffix = NEW_VERSION.replace('.', "_");
+
+	println!("cargo:rustc-link-lib=dylib=chainflip_engine_v{}", old_version_suffix);
```
dandanlen, line 31: We could define a constant for the lib in the utils crate: `const OLD_LIB_NAME: &str = concat!("chainflip_engine_v", OLD_VERSION)`.

Then we don't have to repeat the the replace(..) boilerplate all over the place.

state-chain/pallets/cf-chain-tracking/src/migrations/set_fee_multiplier.rs
dandanlen, line 1: Does this belong here? Bad  merge?

engine-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+pub fn cfe_entrypoint(_attrs: TokenStream, item: TokenStream) -> TokenStream {
+	// Parse the input function
+	let input_fn = parse_macro_input!(item as ItemFn);
+
+	// Get the version from your Cargo.toml file
+	let version = env!("CARGO_PKG_VERSION").replace('.', "_");
+
+	// Construct the new function name
+	let new_fn_name =
+		syn::Ident::new(&format!("cfe_entrypoint_v{}", version), input_fn.sig.ident.span());
```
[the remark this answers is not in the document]
kylezs: this is deliberate, it's just another place to introduce inconsistency if we actually use the specified name.

engine/src/lib.rs
```diff
@@ -70,7 +70,7 @@ pub fn settings_and_run_main(
 		},
 	};
 
-	match tokio::runtime::Builder::new_current_thread()
+	match tokio::runtime::Builder::new_multi_thread()
```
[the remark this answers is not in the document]
kylezs: It was multi thread before, but we just didn't use runner in this way

engine-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+		// Define the entrypoints into each version of the engine
+		#[link(name = #old_dylib_name)]
+		extern "C" {
+			fn #old_version_fn_ident(c_args: CStrArray, start_from: u32) -> engine_upgrade_utils::ExitStatus;
+		}
+
+		#[link(name = #new_dylib_name)]
+		extern "C" {
+			fn #new_version_fn_ident(c_args: CStrArray, start_from: u32) -> engine_upgrade_utils::ExitStatus;
+		}
```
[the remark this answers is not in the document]
kylezs: If I'm understanding the suggestion correctly, the names will conflict and it won't link

engine-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+	let [(old_version_fn_ident, old_dylib_name, old_version), (new_version_fn_ident, new_dylib_name, new_version)] =
+		[OLD_VERSION, NEW_VERSION]
+			.map(|i| i.replace('.', "_"))
+			.map(|version| {
+				(
+					syn::Ident::new(&format!("cfe_entrypoint_v{}", version), Span::call_site()),
+					version,
+				)
+			})
+			.map(|(ident, version)| (ident, format!("chainflip_engine_v{}", version), version));
```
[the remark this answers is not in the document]
kylezs: where exactly? I know there was on some earlier commits, is it resolved now?

engine-runner-bin/Cargo.toml
```diff
@@ -27,8 +27,8 @@ assets = [
     # The old version gets put into target/release by the package github actions workflow.
     # It downloads the correct version from the releases page.
     [
-        "target/release/libchainflip_engine_v1_3_0.so",
-        "usr/lib/chainflip-engine/libchainflip_engine_v1_3_0.so",
+        "target/release/libchainflip_engine_v1_3_3.so",
+        "usr/lib/chainflip-engine/libchainflip_engine_v1_3_3.so",
```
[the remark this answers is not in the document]
kylezs: Well, I didn't want to name the dylib changes I'm doing on 1.3 1.3.2 which is already released

engine-runner-bin/README.md
kylezs: Currently it's manually copy pasted - we can probably improve this, but for something rarely done, can be done later

engine-upgrade-utils/src/lib.rs
kylezs: Yes, before this commit's change, this panics, it was doing a double free

state-chain/pallets/cf-chain-tracking/src/migrations/set_fee_multiplier.rs
kylezs: Thought I select do not include... will remove this.

engine-runner-bin/Cargo.toml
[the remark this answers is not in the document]
kylezs: Ah, yeah, that's correct. Being stricter here means we have more freedom with modifying the runner, don't have to worry about accidentally linking to the wrong libs

engine-proc-macros/src/lib.rs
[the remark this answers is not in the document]
dandanlen: But we're relying on using `cfe_entrypoint` literals in all the right places (see other comments re. repetition) - if any of these changes it all falls apart. Not erasing the annotated function name seems less surprising than replacing it.

engine-proc-macros/src/lib.rs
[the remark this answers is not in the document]
dandanlen: `chainflip_engine_v` is repeated/interpolated in a bunch of places. So is `cfe_entrypoint` (not clear why we sometimes use engine and sometimes cfe). I think we could reduce a lot of boilerplate etc with some well-placed constants and/or conventions.

engine-runner-bin/Cargo.toml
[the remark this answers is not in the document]
dandanlen: I'm not sure I understand the benefit but ok. My gut feeling is that more flexibility would be better (and closer to expectations: we're using dynamic libs in a very un-dynamic way).

engine-runner-bin/README.md
dandanlen: In that case I don't understand what 'fetched from the previous release build' means in this context. Is the plan that there will be a CI job that somehow updates this on github when we make a release?

engine-proc-macros/src/lib.rs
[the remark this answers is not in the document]
kylezs: 👍  Updated to use a const for these names

engine-proc-macros/src/lib.rs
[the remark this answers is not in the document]
kylezs: uses a const now

engine-runner-bin/build.rs
kylezs: Yeah not the biggest fan of this repetition either. Unfortunately .replace() is not a const fn, and I think it makes more sense to have the version specified as the standard "1.3.1" than "1_3_1".

engine-upgrade-utils/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+		Self { c_args: std::ptr::null_mut(), n_args: 0 }
+	}
+}
+
+fn malloc_size<T: Sized>(number_of_ts: usize) -> *mut c_void {
+	unsafe { malloc(size_of::<T>() * number_of_ts) }
+}
+
+impl TryFrom<Vec<String>> for CStrArray {
+	type Error = anyhow::Error;
```
kylezs, line 85: Need to check this

dandanlen: (I tried it on a localnet, seems to work?)

kylezs: > Need to check this

using anyhow error is fine here as it's only used in the runner (not across dylib boundaries).

> Given that this has caused a few headaches - do we really need it? We only use it to pass in the command line args. Why can the engine lib not pull the args from the environment?

As discussed, to do this we'd need to have a "catch all" settings on each. This isn't ideal because someone can do something `--arp.ws_endpoint` and typo, in order to override their settings file for example, it won't error, but it won't use the specified endpoint because there's a typo. Better to have strict enforcement, and use the explicit filter command `args_compatible_with_old` to do the conversion between the versions.

AlastairHolmes (approved):
Reviewed in person

kylezs:
Will fix CI and do some final tests tomorrow before merging.

PR #4621, opened 2024-03-07
fix: correct usdc address

tomjohnburton:

PR #4635, opened 2024-03-12
feat: update sisyphos and perseverance chainspecs (#4621)

martin-chainflip:


state-chain/node/chainspecs/perseverance.chainspec.raw.json
dandanlen, line 1: Is this intentional? (merging the chainspecs)

dandanlen: Never mind - I think this is the right thing to do.

dandanlen (approved):

PR #4638, opened 2024-03-13
feat: RPC for returning scale-encoded System events

syan095:
# Pull Request
Closes: PRO-1185

## Checklist

## Summary:
Created a RPC call that returns System events (In EventRecord) scale encoded.
Changed the original get_events rpc to return event records instead of just the Raw events

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -1373,6 +1373,7 @@ where
 			.client
 			.runtime_api()
 			.cf_get_system_events(self.unwrap_or_best(at))
+			.map_err(to_rpc_error)?
 			.into_iter()
 			.map(|event| hex::encode(event.encode()))
 			.collect::<Vec<_>>())
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -13,6 +13,7 @@ use cf_primitives::{
 	SemVer, SwapId, SwapOutput,
 };
 use cf_utilities::rpc::NumberOrHex;
+use codec::Encode;
 use core::ops::Range;
 use jsonrpsee::{
 	core::{error::SubscriptionClosed, RpcResult},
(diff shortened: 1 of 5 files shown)


syan095:
The returned data is Vec<frame_system::Event<Runtime>> in scale-encoded bytes format (hex encoded).
{"jsonrpc":"2.0","result":["5c00c255053000020000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c000000824bf665798c0000"],"id":1}%  

It does not include extra information in the EventRecord

dandanlen:
Pretty sure we need the `EventRecord` here too. This is what allows us to correlate the event (for example ExtrinsicSuccess) with the extrinsic that triggered it.

syan095:
> Pretty sure we need the `EventRecord` here too. This is what allows us to correlate the event (for example ExtrinsicSuccess) with the extrinsic that triggered it.

Done. 
I took the chance to change the get_events rpc call to return EventRecord as well.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1347,16 +1354,32 @@ where
...
 			.into_iter()
-			.map(|event| event_decoder.decode_data(event))
+			.map(|event_record| event_decoder.decode_data(event_record.encode()))
+			.collect::<Vec<_>>())
+	}
+
+	fn cf_get_system_events_encoded(
+		&self,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<Vec<String>> {
```
dandanlen, line 1375: Instead of returning a string we should use `sp_core::Bytes` for example, and rely on the serializer to do the hex encoding.

dandanlen (commented):
Just one comment 👌

dandanlen (approved):

PR #4644, opened 2024-03-14
fix: remove unused cli command line options

dandanlen:
# Pull Request

I removed some old options from the cli. They are unused, I think left over from a previous implementation where we sent redeem transactions on the user's behalf.


dandanlen:
@j4m1ef0rd can you give this a quick look-over? Thanks.

j4m1ef0rd (approved):

PR #4655, opened 2024-03-18
chore: update systemd config to match 1.4 version 🚀

ahasna:
This prepares for 1.4 release so we don't forget to update this.


tomjohnburton (approved):

kylezs:
There was no need to do this, there won't be an engine 1.4, will be single binary upgrade

PR #4656, opened 2024-03-18
fix(chainflip-broker-api): replace u128 with U256

acdibble:
# Pull Request

## Checklist

## Summary

For consistency, this should return `NumberOrHex` instead of `u128`. `u128` is too large to be safely represented in JSON, but in this case there shouldn't be any issues because the fee shouldn't exceed 2^53 - 1 (at least immediately).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1687,6 +1687,7 @@ dependencies = [
  "hex",
  "jsonrpsee 0.16.3",
  "serde",
+ "sp-core 21.0.0 (git+https://github.com/chainflip-io/polkadot-sdk.git?tag=chainflip-substrate-1.6+1)",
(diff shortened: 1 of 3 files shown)


marcellorigotti (approved):

AlastairHolmes (changes requested):
Please use U128 or U256. We are trying to move away from NumberOrHex.

AlastairHolmes:
Also I believe this is a breaking change right?

acdibble:
Actually I looked into it and the u128 will get serialized to a string.

acdibble:
> Also I believe this is a breaking change right?

It won't be if it's included in 1.3 release

AlastairHolmes (approved):

PR #4660, opened 2024-03-19
chore(localnet): append timestamp to log files 🪵

ahasna:


kylezs (approved):
Await Tom's review too I'd say

tomjohnburton (approved):

PR #4669, opened 2024-03-20
fix: dry run gets fresh runtime version (PRO-1249)

AlastairHolmes:
We had the problem where dry-run wouldn't re-obtain the runtime version when it is out of date, so here I changed the dry-run to always re-obtain the runtime version.


AlastairHolmes:
Going to make a slight tweat to move the runtime_version call to be in parallel with both other rpcs.

kylezs (approved):

PR #4680, opened 2024-03-22
feat: continuous transfer of old utxos

syan095:
# Pull Request

Closes: PRO-1244

## Checklist

## Summary

Refactored the way utxos are transferred from the previous vault to the current vault.
Instead of transferring all utxos during vault rotation, the transfer is combined with Consolidation. 

Utxos from the previous vault are sent to the current vault, independently or together with consolidation utxos.
Utxos that aren't from the current or previous vault are discarded.

When transferring utxos, only take a limited number per transaction. the rest
is returned to storage to be transferred in the next block.

Added unit test and integration test.


state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -230,6 +246,81 @@ pub mod pallet {
...
+
+				// Utxos from the previous vault are rotated into the new vault.
+				if !to_send.is_empty() {
+					// Send out the Utxos into the new vault.
+					if let Some((post_fee_utxos, change_amount)) =
+						Self::calculate_utxos_and_change(to_send)
+					{
+						Self::deposit_event(Event::<T>::UtxoTransferred {
+							broadcast_id: T::BitcoinBroadcaster::threshold_sign_and_broadcast(
+								T::BitcoinCall::transfer_utxo(
```
kylezs, line 308:
```suggestion
								T::BitcoinCall::transfer_utxos(
```

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -230,6 +246,81 @@ pub mod pallet {
...
+					Self::deposit_event(Event::<T>::StaleUtxoDiscarded { utxos: stale });
+				}
+
+				// Utxos from the previous vault are rotated into the new vault.
+				if !to_send.is_empty() {
+					// Send out the Utxos into the new vault.
+					if let Some((post_fee_utxos, change_amount)) =
+						Self::calculate_utxos_and_change(to_send)
+					{
+						Self::deposit_event(Event::<T>::UtxoTransferred {
```
kylezs, line 306:
```suggestion
						Self::deposit_event(Event::<T>::UtxosTransferred {
```

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -230,6 +246,81 @@ pub mod pallet {
...
+		/// Some Utxo from before the previous epochs are discarded from storage.
+		StaleUtxoDiscarded { utxos: Vec<Utxo> },
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		/// Sweep the current utxo storage and:
+		/// 1. Transfer a limited number of utxos from the previous vault into the current vault.
+		/// 2. Any utxos from earlier than the previous vault are discarded.
+		fn on_finalize(_block_number: BlockNumberFor<T>) {
```
kylezs, line 260: Is there any reason we'd leave the current consolidation logic in the ingress egress pallet? There are two different scenarios:
1. Consolidate because we have a lot of UTXOs (sending from vault A to Vault A)
2. Rotate move over all the utxos to the new vault (sending from Vault A to Vault B).

However, they are very closely related, and keeping these two "automatic" utxo sending code paths in the same place would make this a bit easier to follow imo.

state-chain/chains/src/btc/api.rs
```diff
@@ -17,11 +17,41 @@ pub enum BitcoinApi<Environment: 'static> {
...
+
+impl<E> TransferUtxoCall for BitcoinApi<E> {
+	fn transfer_utxo(
+		old_key: [u8; 32],
+		new_key: [u8; 32],
+		utxos: Vec<Utxo>,
+		change_amount: BtcAmount,
+	) -> Self {
+		// We will use the bitcoin address derived with the salt of 0 as the vault address.
+		let new_vault_change_script =
```
kylezs, line 39: Also see here, this is very similar to the consolidation call - it seems like these can just be merged into one call. Both calls are already key agnostic. Sending from old to new. This would help making these two btc automatic send code paths into one.

kylezs:
Also just added a small commit - I'm not sure what was going to happen with that part of the code, but so it's not forgotten, we weren't adding any egress ids to the BatchBroadcastRequested for consolidation, which makes the event contents already covered by the specific consolidation event

dandanlen (commented):
I agree with Kyle - it feels like we're duplicating a lot of stuff. 

I feel like we could just change the logic for `select_utxos_for_consolidation`. 

Currently the logic is something like this:

```
if spendable_utxos.len() > consolidation_limit {
	// consolidate spendable_utxos
} else {
	// do nothing
}
```

It could instead be:

```
let old_utxos = spendable_utxos.filter(|utxo| utxo.key == old_key); // need to pass in the old key somehow
if spendable_utxos.len() > consolidation_limit {
	// consolidate spendable_utxos (which includes any old ones)
} else if !old_utxos.is_empty() {
	// consolidate old utxos only
} else {
	// do nothing
}
```

And then we don't need to add any additional logic or hooks ro anything like that.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -230,6 +246,81 @@ pub mod pallet {
...
+						Default::default()
+					};
+
+					(to_send, stale)
+				});
+
+				// Stale utxos are removed from storage and discarded
+				if !stale.is_empty() {
+					Self::deposit_event(Event::<T>::StaleUtxoDiscarded { utxos: stale });
+				}
```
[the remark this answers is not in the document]
syan095: Nice. yeah using drain is a cleaner way of doing it.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -568,6 +633,26 @@ impl<T: Config> Pallet<T> {
 				}),
 		}
 	}
+
+	fn calculate_utxos_and_change(spendable_utxos: Vec<Utxo>) -> Option<(Vec<Utxo>, BtcAmount)> {
+		let bitcoin_fee_info = T::BitcoinFeeInfo::bitcoin_fee_info();
```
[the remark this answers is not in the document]
syan095: Done

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -497,46 +506,53 @@ impl<T: Config> Pallet<T> {
...
+						.drain(
+							..sp_std::cmp::min(
+								old_utxos.len(),
+								consolidation_parameter.consolidation_size as usize,
+							),
+						)
+						.collect::<Vec<_>>();
+
+					// If consolidation is required, add into the same batch
+					if available_utxos.len() > consolidation_parameter.consolidation_size as usize {
```
kylezs, line 543: above, we limit the number of utxos for selection. But now here we're potentially adding more to that set, so we can go over the limit. We could add a test for this case too. To make sure that that the number of utxos selected doesn't exceed the consolidation size. We can also add a debug_assert after this `if` that the `to_send.len() < consolidation_size`

kylezs: So the logic in this case could be:
1. Rotate
2. Send to new vault
3. Full, wait until next block
4. Consolidate

which is fine. The primary thing is that we don't try to sign too many payloads at once

kylezs: also not really sure what the intention of this `if` statement is? 
It's currently saying "if we have more utxos associated with the current key than maximum number we are allowed to consolidate then run this code" but that doesn't seem relevant? If anything at all, it should be the `params.consolidation_threshold`? In which case this is already checked in `select_utxos_for_consolidation` and so isn't required here.

kylezs: Also, I think there's a small bug below too:
```rust
select_utxos_from_pool(
						available_utxos,
						&bitcoin_fee_info,
						output_amount +
							number_of_outputs * fee_per_output_utxo +
							min_fee_required_per_tx,
						Some(Self::consolidation_parameters().consolidation_threshold),
```
It should also be consolidation_size here too? Here's what it says the meanings of the parameters are.

```rust
pub struct ConsolidationParameters {
	/// Consolidate when total UTXO count reaches this threshold
	pub consolidation_threshold: u32,
	/// Consolidate this many UTXOs
	pub consolidation_size: u32,
}
```

dandanlen: Why not just do what I suggested and put all this inside the implementation of `select_utxos_for_consolidation`, and then inside there, first try the consolidation, and then if no consolidation is required, check for old utxos? It seems like we're handling a bunch of edge cases that we know will basically never happen. 

If we know we will do *either* 'consolidation above threshold' *or* 'consolidation of old utxos', then this is a lot easier to think about.

This way we would also make it easier to check whether the utxos are even worth rotating or not (net_value is currently not taken into account so we could rotate utxos that effectively unspendable).

syan095: I think I didn't fully understand your design before. 
Now the logic is part of the consolidation, I also think this is cleaner. Please take a look and see if this is what you had in mind.

syan095: > Also, I think there's a small bug below too:

WIth this, I was thinking about the case where we need more than consolidation_size number of utxos for the desired amount. Consider the edge case (which in practice should never happen)
e.g. threshold: 200, size: 100. If we use the "size" as the limit here:
we currently have 101 utxos, and we need all of them for the desired amount. We will need to wait until we have 99 more utxos before we consolidate - then we would have enough utxo.

I can see arguments for both - we should use "size" if we've decided that this is the maximum number of utxos we send period.
where as with threshold, it ensures that we can always find enough utxos for a transaction if we have enough. (though it is unlikely to happen)

what do you think? should we change from threshold to size?

dandanlen: I had this discussion with @msgmaxim and from what I remember, the conclusion was that we should consider the consolidation threshold to be the largest number of utxos we are willing to sign (implying that we would use the threshold here, not the limit). In practice I don't think it makes a big difference, we just need to make a decision and stick to it.

dandanlen:
One last thought here: I think we have an implicit assumption that consolidation_threshold is always greater than the consolidation_limit but we don't enforce this anywhere (we could enforce it in the governance extrinsic).

state-chain/chains/src/btc/utxo_selection.rs
```diff
@@ -107,27 +107,53 @@ pub fn select_utxos_from_pool(
...
+		// Utxos are pre-filtered. Only utxos that belong to current or previous vaults are
+		// remaining.
+		let mut prev_utxos = spendable
+			.extract_if(|utxo: &mut Utxo| utxo.deposit_address.pubkey_x != current_key)
+			.collect::<Vec<_>>();
+
+		let mut remaining = prev_utxos.split_off(sp_std::cmp::min(
+			prev_utxos.len(),
+			consolidation_parameter.consolidation_size as usize,
+		));
```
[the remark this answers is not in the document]
syan095: Yeah this is good. I was overthinking things

syan095:
> One last thought here: I think we have an implicit assumption that consolidation_threshold is always greater than the consolidation_limit but we don't enforce this anywhere (we could enforce it in the governance extrinsic).

I think this is enforced here:
[self.consolidation_size <= self.consolidation_threshold && self.consolidation_size > 1](https://github.com/chainflip-io/chainflip-backend/blob/ea6dbbe1f4094e3c9e8cf3c7dbffff060aa722c2/state-chain/chains/src/btc/utxo_selection.rs#L31)

which is checked when updating the parameter:
https://github.com/chainflip-io/chainflip-backend/blob/ea6dbbe1f4094e3c9e8cf3c7dbffff060aa722c2/state-chain/pallets/cf-environment/src/lib.rs#L342

state-chain/pallets/cf-environment/src/lib.rs
syan095: In this case I should leave the maximum as the threshold.

state-chain/pallets/cf-environment/src/mock.rs
```diff
@@ -175,6 +176,13 @@ impl GetBitcoinFeeInfo for MockBitcoinFeeInfo {
 	}
 }
 
+pub struct MockBitcoinKeyProvider;
+impl KeyProvider<BitcoinCrypto> for MockBitcoinKeyProvider {
+	fn active_epoch_key() -> Option<EpochKey<AggKey>> {
+		CurrentBitcoinKey::get()
+	}
+}
```
dandanlen, line 184: We already have a generic mock for `KeyProvider`, prefer to use that one (cf_traits::mocks::key_provider).

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -663,10 +663,6 @@ pub mod pallet {
 				let (broadcast_id, _) =
 					T::Broadcaster::threshold_sign_and_broadcast(egress_transaction);
 				Self::deposit_event(Event::<T, I>::UtxoConsolidation { broadcast_id });
-				Self::deposit_event(Event::<T, I>::BatchBroadcastRequested {
-					broadcast_id,
-					egress_ids: Default::default(),
-				});
```
dandanlen, line 669: 👌

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -226,3 +153,222 @@ fn update_safe_mode() {
...
+			consolidation_size: 2,
+		});
+
+		BitcoinAvailableUtxos::<Test>::set(vec![
+			utxo(1_000_000, CHANGE_ADDRESS_SALT, Some(epoch_1)),
+			utxo(2_000_000, CHANGE_ADDRESS_SALT, Some(epoch_1)),
+			utxo(3_000_000, CHANGE_ADDRESS_SALT, Some(epoch_1)),
+			utxo(31_000_000, CHANGE_ADDRESS_SALT, Some(epoch_3)),
+			utxo(32_000_000, CHANGE_ADDRESS_SALT, Some(epoch_3)),
+			utxo(33_000_000, CHANGE_ADDRESS_SALT, Some(epoch_3)),
```
dandanlen, line 177: I feel like the tests can be more succinct/to the point. 

For example here, the important details are the epoch numbers, not the amounts or the change addresses. 

We could make this more obvious by not requiring this information to repeat, or by repeating the same amounts (then it's clear that the amounts don't matter).

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -226,3 +153,222 @@ fn update_safe_mode() {
...
+		// Consolidate from storage. Take the first 4 utxos.
+		assert_eq!(
+			Environment::select_and_take_bitcoin_utxos(UtxoSelectionType::SelectForConsolidation),
+			Environment::calculate_utxos_and_change(vec![
+				utxo(11_000_000, CHANGE_ADDRESS_SALT, Some(epoch_1)),
+				utxo(12_000_000, CHANGE_ADDRESS_SALT, Some(epoch_1)),
+				utxo(13_000_000, CHANGE_ADDRESS_SALT, Some(epoch_1)),
+				utxo(21_000_000, CHANGE_ADDRESS_SALT, Some(epoch_2)),
+			])
+		);
```
dandanlen, line 246: Another example: the `4` could be a constant, and then we can just assert that the output contains the expected number of utxos. This test should not be concerned with anything else. 

We can write a separate test for `calculate_utxos_and_change` to make sure it does so correctly.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -568,6 +564,30 @@ impl<T: Config> Pallet<T> {
 				}),
 		}
 	}
+
+	fn calculate_utxos_and_change(spendable_utxos: Vec<Utxo>) -> Option<(Vec<Utxo>, BtcAmount)> {
```
dandanlen, line 568: This method doesn't need to return the spendable_utxos (it's always the same as the input value).

Also the name is a bit misleading - it calculates the change specifically for the consolidation transaction.

We could also just change the `select_utxos_for_consolidation` such this it returns the change amount too.

dandanlen (commented):
Just some minor comments.

state-chain/pallets/cf-environment/src/lib.rs
msgmaxim: Here is how I think about consolidation:
- If we need to make a transfer that will consume all of our UTXOs, we should be able to do it.
- For that, we ensure that the number of UTXO doesn't go above `N` (max we can sign at once), and if it does, we consolidate.
- We don't want to consolidate all at once (leaving not UTXO available for some time), so we consolidate `N' < N`.

In my consolidation PR I chose to call `N` - consolidation threshold and `N'` - consolidation size, and arbitrarily set `N' = N / 2`. We can change these if we want. This does mean that if we require to sign with `N`, we should allow that. If we find that `N` is too big for our engines, we should reduce *it* rather than limiting transactions to some number of UTXOs smaller than `N`.

state-chain/pallets/cf-environment/src/tests.rs
syan095: Simplified the unit tests

state-chain/pallets/cf-environment/src/lib.rs
syan095: Refactored

dandanlen (approved):
Thanks Roy.

PR #4693, opened 2024-03-25
chore: bump subxt to 0.33

kylezs:
Resolves the cargo-audit issue (https://rustsec.org/advisories/RUSTSEC-2022-0093), as this bumps the ed25519-dalek dependency to v2

### DIFF
--- a/engine/src/state_chain_observer/client.rs
+++ b/engine/src/state_chain_observer/client.rs
@@ -59,7 +59,7 @@ const SUBSTRATE_BEHAVIOUR: &str = "Unexpected state chain node behaviour";
 const SYNC_POLL_INTERVAL: Duration = Duration::from_secs(4);
 
 /// Enough time for a state chain transaction to make it into a (unfinalised) block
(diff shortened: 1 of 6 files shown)


engine/src/state_chain_observer/client.rs
```diff
@@ -910,7 +910,7 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 						)?
 						.submit_and_watch()
 						.await?
-						.wait_for_in_block()
```
kylezs, line 913: they removed this in 0.33 as it is unreliable

dandanlen: @AlastairHolmes FYI details are in this PR: https://github.com/paritytech/subxt/pull/1237

Could this be related to the SC Client hash mismatch issues we are seeing?

AlastairHolmes: We should probably increase the timeout its only 30 seconds

AlastairHolmes (approved):
just increase the timeout then we're good I think

PR #4697, opened 2024-03-26
feat: gas swaps use swapping queue

j4m1ef0rd:
# Pull Request

Closes: PRO-1215

## Checklist

## Summary

- Instead of swapping the transaction fee to the gas asset right away, it is now pre-calculated and scheduled via the swapping queue.
	- This means any single ingress/egress of a non-gas asset will cause 2 swaps to be added to the queue.
- Added a `TransactionFeeApi` trait so the fee can be added to the withheld fees once the gas swap is done.
	- New swap type `TransactionFee`.
- Refactored the `convert_asset_to_approximate_output` function to no longer do the swap, it now just calculates the required amount to convert. The ingress-egress-pallet now does the swap (add to the swap queue). 
- Moved the `MockSwapQueueApi` so other pallet tests can use it.
- Added mocks and tests to check the new logic.

## Out of scope

- Because the gas swap is now delayed 2 blocks, the resulting amount may be slightly off from the actual pre-calculated amount.


dandanlen (commented):
Overall shape of this looks good - will take a closer look at tests etc. tomorrow.

state-chain/traits/src/mocks/swap_queue_api.rs
```diff
@@ -0,0 +1,46 @@
...
+
+	fn schedule_swap(
+		from: Asset,
+		to: Asset,
+		amount: AssetAmount,
+		swap_type: SwapType,
+	) -> (u64, Self::BlockNumber) {
+		Self::mutate_value(SWAP_QUEUE, |queue: &mut Option<Vec<MockSwap>>| {
+			match queue {
+				Some(queue) => queue.push(MockSwap { from, to, amount, swap_type }),
```
msgmaxim, line 40: You should be able to use `Option::get_or_insert_default()` and then push the new entry to the result of that.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1391,30 +1395,36 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 				fee_estimate,
 			)
 			.unwrap_or_else(|| {
 				log::warn!(
-					"Unable to convert input to gas for input of {:?} ${:?}. Ignoring transaction fees.",
-					available_amount,
-					asset,
-				);
-				(available_amount, Zero::zero())
+						"Unable to convert input to gas for input of {:?} ${:?}. Ignoring transaction fees.",
```
msgmaxim, line 1411: Looks like unintended change in indentation.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1571,3 +1581,21 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
 		))
 	}
 }
+
+impl<T: Config<I>, I: 'static> TransactionFeeApi<T::TargetChain> for Pallet<T, I> {
+	fn accrue_transaction_fee(
+		_asset: <T::TargetChain as Chain>::ChainAsset,
```
msgmaxim, line 1587: Is asset isn't actually used, can it be removed?

j4m1ef0rd: It is needed because I use the same trait for the handler. See `impl_transaction_fee_api_for_anychain`. Not perfect, but still better than adding another trait I think.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1994,14 +1992,7 @@ impl<T: Config> cf_traits::AssetConverter for Pallet<T> {
...
 		if input_asset == output_asset {
-			if desired_output_amount < available_input_amount {
-				return Some((
-					available_input_amount.saturating_sub(desired_output_amount),
-					desired_output_amount,
-				))
-			} else {
-				return Some((Zero::zero(), available_input_amount))
-			}
+			return Some(available_input_amount.saturating_sub(desired_output_amount))
```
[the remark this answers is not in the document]
j4m1ef0rd: Nice catch. forgot to add the cases where the asset is the same to the unit test, as they were not there before.

state-chain/traits/src/mocks/swap_queue_api.rs
j4m1ef0rd: Its an unstable feature, so just went with `get_or_insert`. Still an improvement.

state-chain/traits/src/lib.rs
```diff
@@ -951,12 +951,14 @@ pub trait AssetConverter {
...
 		input_asset: impl Into<Asset>,
 		available_input_amount: Amount,
 		output_asset: impl Into<Asset>,
 		desired_output_amount: Amount,
-	) -> Option<(Amount, Amount)>;
+	) -> Option<Amount>;
+}
+
+pub trait TransactionFeeApi<C: Chain> {
+	fn accrue_transaction_fee(asset: C::ChainAsset, amount: C::ChainAmount);
```
dandanlen, line 963: Can we adjust the naming of this to be little less ambiguous? Transaction fee overlaps with our own transaction fees and there's a substrate TransactionFeeApi trait too I believe...

IngressEgressFeeApi::accrue_withheld_fee(..) or something?

dandanlen (commented):
Just some nits...

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -849,6 +852,19 @@ pub mod pallet {
 									swap.to
 								);
 							},
+						SwapType::TransactionFee => {
```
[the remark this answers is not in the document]
j4m1ef0rd: Should we put the enum with the extra data like the swap id? it would be repeated information. Or should I make a new enum with the data stripped away?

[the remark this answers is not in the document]
dandanlen: Let's live with the duplication for now - we can address this with PRO-1282.

dandanlen (approved):
LGTM, I just added some minor changes, additional naming updates and a change to one of the mocks.

PR #4707, opened 2024-03-26
fix: continuous adapter

AlastairHolmes:
Basically, its an off by one error.

If the main stream misses the first block in an epoch, the continuous adapter would not fill the missed witness for that first block.

*Note if the engine was restarted it would fill in the gap.*

Looking at the code we were handling the last_processed_index and the start of the epoch in the same way, but the first is an index we don't want to process, but the second is an index we do want to process.

### DIFF
--- a/engine/src/witness/common/chunked_chain_source/chunked_by_vault/continuous.rs
+++ b/engine/src/witness/common/chunked_chain_source/chunked_by_vault/continuous.rs
@@ -77,8 +77,8 @@ where
 					let mut processed_inverse = processed_indices.clone();
 					processed_inverse.invert();
 					processed_inverse.set_range(..epoch.info.1, false);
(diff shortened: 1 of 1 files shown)


msgmaxim:
Looks like something a unit test would have caught?

kylezs (approved):

AlastairHolmes:
We're just waiting to do some extra testing before merging.

kylezs (approved):

AlastairHolmes:
It's just a complex problem whose solution was rushed to meet a release deadline. It's bad code, so I think we'll rewrite it. https://linear.app/chainflip/issue/PRO-1280/re-write-continuous-adapter

kylezs:
Tested manually to expedite release, will create a ticket to unit test.

PR #4710, opened 2024-03-27
chore: remove backcompat asset encoding from rpcs (and make them line up with newer Asset encoding)

AlastairHolmes:
Affects the responses of these rpcs:

lp-api rpcs
- lp_subscribe_order_fills
- lp_order_fills
- lp_get_open_swap_channels
- lp_asset_balances

node rpcs
- cf_pool_price_v2
- cf_subscribe_pool_price_v2
- cf_subscribe_prewitness_swaps
- cf_prewitness_swaps
- cf_supported_assets
- cf_asset_balances

I'll make a doc PR later today.

### DIFF
--- a/state-chain/primitives/src/chains/assets.rs
+++ b/state-chain/primitives/src/chains/assets.rs
@@ -345,6 +345,12 @@ macro_rules! assets {
 						iter.clone().find(|(asset, _t)| *asset == required_asset).ok_or(()).map(|x| x.1)
 					}).ok()
 				}
+
+				pub fn iter(&self) -> impl Iterator<Item = (Asset, &T)> {
+					Asset::all().map(|asset| {
+						(asset, &self[asset])
+					})
+				}
 			}
 
 			impl<T> IndexMut<Asset> for AssetMap<T> {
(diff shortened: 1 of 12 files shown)


state-chain/primitives/src/chains/assets.rs
```diff
@@ -345,6 +345,12 @@ macro_rules! assets {
 						iter.clone().find(|(asset, _t)| *asset == required_asset).ok_or(()).map(|x| x.1)
 					}).ok()
 				}
+
+				pub fn iter(&self) -> impl Iterator<Item = (Asset, &T)> {
+					Asset::all().map(|asset| {
+						(asset, &self[asset])
```
dandanlen, line 351: There's no static guarantee that self[asset] exists (is there?) - it's true as long as the macros are correct, but maybe it's worth including a unit test for each asset map that just does an `iter().collect()` to prevent mistakes being introduced.

dandanlen (approved):

state-chain/primitives/src/chains/assets.rs
AlastairHolmes: it is guaranteed by the function signature the lack of panics in the index impl. But the code could have more unit tests so ill add some more.

dandanlen (approved):

PR #4720, opened 2024-04-03
Update Sdk with Arbitrum support

albert-llimos:
## Checklist

## Summary

Updating SDK with the latest version that has Arbitrum support.

After doing that I found a problem with nonces. This is my hypothesis which turned out to be true:

"The error is nonce too high. I have checcked the nonce counter logic in the bouncer and it looks good, also the prints look like we are not skipping anything. Looks solid.  It even seems to happen when only running all the contract swaps via the SDK in parallel, but I'm pretty sure there is no change on the SDK either on that front.
I believe what is happening is that we have a race condition when running all swaps (normal deposit swaps & contract swaps via SDK). They are all using the same nonces and if some transaction lands in the wrong order the node will reject the transaction, even though it would eventually get mined when the transaction prior to it eventually lands.
My best guess is that we are starting to see this now because the amount of parallel swaps has increased substantially with the addition of USDT and Arbitrum assets."

For contract swaps I have modifed the logic to start each swap from a different account so there can't be issues with nonces. That works.


bouncer/shared/contract_swap.ts
```diff
@@ -87,16 +87,50 @@ export async function performSwapViaContract(
...
+    await send(chainGasAsset(srcChain), wallet.address);
+    await send(sourceAsset, wallet.address);
+
+    if (erc20Assets.includes(sourceAsset)) {
+      // Doing effectively infinite approvals to make sure it doesn't fail.
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      await approveTokenVault(
+        sourceAsset,
+        (
+          BigInt(amountToFineAmount(defaultAssetAmounts(sourceAsset), assetDecimals(sourceAsset))) *
```
kylezs, line 111: we should pull this out into a function that takes a source asset and amount. We use this pattern of `BigInt(amountToFineAmount...` in a number of other places too

bouncer/shared/gaslimit_ccm.ts
```diff
@@ -111,10 +106,10 @@ async function testGasLimitSwap(
 
   let gasSwapScheduledHandle;
 
-  if (CCM_CHAINS_NATIVE_ASSETS[destChain] !== sourceAsset) {
+  if (chainGasAsset(destChain as Chain) !== sourceAsset) {
     gasSwapScheduledHandle = observeSwapScheduled(
       sourceAsset,
-      destChain === 'Ethereum' ? Assets.Eth : ('ArbEth' as Asset),
+      destChain === 'Ethereum' ? Assets.Eth : Assets.ArbEth,
```
kylezs, line 112: just noticed there are lot of other places we still use asset as string, we should move to this Assets type. (ofc, would be another ticket/PR, not directly relevant)

kylezs (approved):

bouncer/shared/gaslimit_ccm.ts
albert-llimos: yes, there is definitely some cleanup to do in multiple places. I'll aim to get it done when I get some time.

PR #4727, opened 2024-04-04
chore: Moved start/stop bidding from Funding to Validator pallet. (#4727)

syan095:
# Pull Request

Closes: PRO-1269

## Checklist

## Summary

Moved the ActiveBidder storage to Validator pallet. Moved corresponding tests, events, error, benchmarking
When trying to redeem while participating an Auction, this will return `Validator::StillBidding` now instead of `Funding::AuctionPhase`.


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -295,6 +295,12 @@ pub mod pallet {
 	pub(super) type MaxAuthoritySetContractionPercentage<T: Config> =
 		StorageValue<_, Percent, ValueQuery>;
 
+	/// Store the list of funded accounts and whether or not they are a active bidder.
+	#[pallet::storage]
+	#[pallet::getter(fn active_bidder)]
+	pub type ActiveBidder<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::AccountId, bool, ValueQuery>;
```
dandanlen, line 302: We can use the Identity hasher here to gain some efficiency.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -367,7 +343,7 @@ pub mod pallet {
 
 			// Not allowed to redeem if we are an active bidder in the auction phase
 			if T::EpochInfo::is_auction_phase() {
-				ensure!(!ActiveBidder::<T>::get(&account_id), Error::<T>::AuctionPhase);
+				T::BidderProvider::ensure_not_bidding(&account_id)?;
```
dandanlen, line 346: Instead of linking the BidderProvider I would suggest creating a `trait RedemptionCheck { fn ensure_can_redeem(&account) -> Result<(), DispatchError>; }`. 

Then we can move this whole check into the validator pallet, and we have more flexibility to add other checks in future. And we can also (I think) delete the BidderProvider.

Should also make it easier to mock/test.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -145,6 +142,9 @@ pub mod pallet {
 		/// This is used to reset the validator's reputation
 		type ReputationResetter: ReputationResetter<ValidatorId = ValidatorIdOf<Self>>;
 
+		/// The Flip token implementation.
+		type Flip: AccountInfo<Self>;
```
dandanlen, line 146: The Flip pallet is already linked via Chainflip::FundingInfo so we don't really need to add it here. 

We could just add the AccountInfo constraint to the Chainflip trait:

```rust
pub trait Chainflip {
	// ...
	type FundingInfo:
		FundingInfo<AccountId = Self::AccountId, Balance = Self::Amount> +
		AccountInfo<Self>;
}

File: state-chain/runtime/src/lib.rs [hunk 4]
```diff
@@ -1118,6 +1120,76 @@ impl frame_support::traits::OnRuntimeUpgrade for FlipToBurnMigration {
 	}
 }
 
+pub struct ActiveBidderMigration;
```
dandanlen, line 1123: Please add this in its own module in `migrations`. The runtime.rs is big enough already, and adding it in a separate module make it easier to resolve conflicts when migrations are added/removed.

dandanlen (commented):
Looking good. I have some further suggestions though :)

state-chain/pallets/cf-validator/src/lib.rs
syan095: Changed this to a BTreeSet.

state-chain/pallets/cf-funding/src/lib.rs
syan095: Yeah the old trait can be removed.

state-chain/pallets/cf-validator/src/lib.rs
syan095: Added this to the mock. Also added AccountInfo to the MockFundInfo

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -343,7 +344,7 @@ pub mod pallet {
 
 			// Not allowed to redeem if we are an active bidder in the auction phase
 			if T::EpochInfo::is_auction_phase() {
-				T::BidderProvider::ensure_not_bidding(&account_id)?;
+				T::RedemptionChecker::ensure_can_redeem(&account_id)?;
```
dandanlen, line 347: EpochInfo::is_auction_phase is defined in the validator pallet so we can move that check inside `ensure_can_redeem` too :)

dandanlen (commented):
Just one more thing... I'll take another look tomorrow.

dandanlen (commented):
Still a couple more things to refactor... 

- There are still some tests in the funding pallet that use Bidders and auction_phase. We don't really need this any more: the funding pallet now doesn't care about auction phases or bidders. We can just use the RedemptionChecker mock. 
- Similarly, we need to make sure that we test the behaviour of the redemption checker implementation in the validator pallet: we need to ensure that the redemption check fails if we are in the auction phase or if the node is bidding. I think is_auction_phase is no longer necessary as part of the EpochInfo trait. It can just be a public function on the Pallet.

dandanlen:
Sorry - still haven't gotten around to re-reviewing this. Will do so tomorrow.

dandanlen (approved):
Looks good, thanks! Will get it merged.

dandanlen:
@GabrielBuragev / @jerryafr does this impact the product in any way? The start-bidding and stop-bidding extrinsics have been moved from the funding pallet to the validator pallet.

PR #4730, opened 2024-04-04
Arbitrum backup rpc

albert-llimos:
# Pull Request

Closes: PRO-1226

## Checklist

## Summary

- Fixed bug in Arbitrum rpc where it wasn't properly parsing the endpoint passed as `opts`.
- Added backup rpc.


charts/chainflip-node/values.yaml
```diff
@@ -42,6 +42,11 @@ engine:
       rpc:
         ws_endpoint: ""
         http_endpoint: ""
+    arb:
+      private_key_file: /etc/chainflip/keys/ethereum_key_file
```
kylezs, line 46: maybe just a comment so it's clear this isn't a mistake but intentional, we want to use the same key here for simplicity

kylezs (approved):
Also you should create a PR on the docs repo to add these new settings (but not merge it yet ofc)

charts/chainflip-node/values.yaml
albert-llimos: true. Done.

PR #4733, opened 2024-04-04
fix: migration for earned fees

dandanlen:
# Pull Request

Closes: PRO-1289

## Checklist

## Summary

Converted the stored AssetMap to a StorageDoubleMap to allow looking up the value directly. This makes storage writes more efficient (and the storage is accessed quite frequently - see the weights files). 

This also makes adding new assets simpler: adding a new asset won't require storage migrations in future since the amount defaults to zero.

I also added utilites to build and AssetMap<T> from a (possibly incomplete) list of (Asset, T) pairs. This allows backward-compatible defaulting of the value in the asset map.


state-chain/pallets/cf-lp/src/migrations/v2.rs
```diff
@@ -0,0 +1,117 @@
...
+	#[derive(Encode, Decode, Default, Clone, PartialEq)]
+	struct BtcMap {
+		btc: AssetAmount,
+	}
+
+	#[derive(Encode, Decode, Default, Clone, PartialEq)]
+	pub struct AssetMap {
+		eth: EthMap,
+		dot: DotMap,
+		btc: BtcMap,
```
kylezs, line 43: the order here in 1.3 is: eth, btc, dot

kylezs (approved):

state-chain/pallets/cf-lp/src/migrations/v2.rs
dandanlen: Hmmm sure? On Berghain it's eth, dot, btc...

kylezs: I guess we're inconsistent elsewhere then...

PR #4737, opened 2024-04-08
refactor/improve-rpc-encode-decode

Janislav:
# Pull Request

Closes: PRO-1157

## Checklist

## Summary

- Primarily removes any usage of NumberOrHex from the return types of the LP-API -> Should be not breaking because in any case LPs have to handle decoding the result as a hex string already.
- Removes also all usage for NumberOrHex and any WrappedType that is using it in a V2 version of the API.
- Provides a backwards compatible API to avoid breaking the LPs

#### Non-Breaking changes

In theory, that should be non-breaking. It still needs some more further checking to ensure that, though. We assume that because NumberOrHex can theoretically decode to a hex string, so LPs also have to handle that case already, which means replacing that with a type that always returns as hex should be none-breaking. Apart from that, during the transition over the next version, we keep the old API as it is and give LPs time to change their implementation. With an upcoming version, we will remove this old API.


msgmaxim:
Not sure if this is important to you PR or not, but just FYI, there is a custom implementation for `BoostPoolDepth` that uses `NumberOrHex` as a way to serialize the amounts as hex: https://github.com/chainflip-io/chainflip-backend/blob/5b074d547d261097468df780a7e998754d966e7f/state-chain/runtime/src/runtime_apis.rs#L76C78-L77C1. Shouldn't really matter since the result will be the same, but for consistency you might want to change that to use `U256` too.

martin-chainflip (approved):

Janislav:
As discussed, putting this here into draft for now @AlastairHolmes.

Janislav:
I think we close this here

PR #4745, opened 2024-04-10
feat: Add prewitness deposit events

j4m1ef0rd:
# Pull Request

Closes: PRO-1295, PRO-1296, PRO-1294, PRO-1293

## Checklist

## Summary

Still not very confident on the details/types in these events, but it's hopefully a good start.

- Added 3 new events
- modified `DepositBoosted` to have the amounts of each pool used.
- updated or added tests to cover the events.

TODO: Test on localnet.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -648,8 +648,9 @@ pub mod pallet {
 			boost_id: BoostId,
 			channel_id: ChannelId,
 			// Ingress fee in the deposit asset. i.e. *NOT* the gas asset, if the deposit asset is
-			// a non-gas asset.
+			// a non-gas asset. The ingress fee is taken *after* the boost fee.
 			ingress_fee: TargetChainAmount<T, I>,
+			// Total fee the user paid for their deposit to be boosted.
 			boost_fee: TargetChainAmount<T, I>,
 			action: DepositAction<T::AccountId>,
 		},
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -58,7 +58,7 @@ pub use weights::WeightInfo;
 
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
 pub enum BoostStatus {
-	Boosted { boost_id: BoostId, pools: Vec<BoostPoolTier> },
+	Boosted { prewitnessed_deposit_id: PrewitnessedDepositId, pools: Vec<BoostPoolTier> },
 	NotBoosted,
 }
 
@@ -91,8 +91,6 @@ pub struct BoostOutput<C: Chain> {
 	used_pools: BTreeMap<BoostPoolTier, C::ChainAmount>,
 	total_fee: C::ChainAmount,
 }
-
-type BoostId = PrewitnessedDepositId;
 
 const SORTED_BOOST_TIERS: [BoostPoolTier; 3] =
 	[BoostPoolTier::FiveBps, BoostPoolTier::TenBps, BoostPoolTier::ThirtyBps];
@@ -275,7 +273,7 @@ pub mod pallet {
 		LiquidityProvision { lp_account: AccountId },
 		CcmTransfer { principal_swap_id: Option<SwapId>, gas_swap_id: Option<SwapId> },
 		NoAction,
-		BoostersCredited { boost_id: BoostId },
+		BoostersCredited { prewitnessed_deposit_id: PrewitnessedDepositId },
 	}
 
 	/// Tracks funds that are owned by the vault and available for egress.
@@ -645,7 +643,7 @@ pub mod pallet {
 			asset: TargetChainAsset<T, I>,
 			amounts: BTreeMap<BoostPoolTier, TargetChainAmount<T, I>>,
 			deposit_details: <T::TargetChain as Chain>::DepositDetails,
-			boost_id: BoostId,
+			prewitnessed_deposit_id: PrewitnessedDepositId,
 			channel_id: ChannelId,
 			// Ingress fee in the deposit asset. i.e. *NOT* the gas asset, if the deposit asset is
 			// a non-gas asset. The ingress fee is taken *after* the boost fee.
@@ -655,19 +653,19 @@ pub mod pallet {
 			action: DepositAction<T::AccountId>,
 		},
 		BoostFundsAdded {
-			account_id: T::AccountId,
+			booster_id: T::AccountId,
 			boost_pool: BoostPoolId<T::TargetChain>,
 			amount: TargetChainAmount<T, I>,
 		},
 		StoppedBoosting {
-			account_id: T::AccountId,
+			booster_id: T::AccountId,
 			boost_pool: BoostPoolId<T::TargetChain>,
 			// When we stop boosting, the amount in the pool that isn't currently pending
 			// finalisation can be returned immediately.
 			unlocked_amount: TargetChainAmount<T, I>,
 			// The ids of the boosts that are pending finalisation, such that the funds can then be
 			// returned to the user's free balance when the finalisation occurs.
-			pending_boosts: BTreeSet<BoostId>,
+			pending_boosts: BTreeSet<PrewitnessedDepositId>,
 		},
 		InsufficientBoostLiquidity {
 			prewitnessed_deposit_id: PrewitnessedDepositId,
@@ -750,11 +748,12 @@ pub mod pallet {
 					let removed_deposits =
 						Self::clear_prewitnessed_deposits(deposit_channel.channel_id);
 
-					if let BoostStatus::Boosted { boost_id, pools } = boost_status {
+					if let BoostStatus::Boosted { prewitnessed_deposit_id, pools } = boost_status {
 						for pool_tier in pools {
 							BoostPools::<T, I>::mutate(deposit_channel.asset, pool_tier, |pool| {
 								if let Some(pool) = pool {
-									let affected_boosters_count = pool.on_lost_deposit(boost_id);
+									let affected_boosters_count =
+										pool.on_lost_deposit(prewitnessed_deposit_id);
 									used_weight.saturating_accrue(T::WeightInfo::on_lost_deposit(
 										affected_boosters_count as u32,
 									));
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1267,20 +1312,22 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			if !boosted_amount.is_zero() {
+				used_pools.insert(boost_tier, boosted_amount);
+			}
+
 			remaining_amount.saturating_reduce(boosted_amount);
 			total_fee_amount.saturating_accrue(fee);
 
 			if remaining_amount == 0u32.into() {
-				return Ok((used_pools, total_fee_amount));
+				return Ok(BoostOutput { contributions: used_pools, total_fee: total_fee_amount });
```
kylezs, line 1326: Would just rename `contributions` here to `used_pools` no need for separate names (and used_pools is a little clearer what it's about exactly imo). Especially since we just name it back to used_pools literally right after.

```suggestion
				return Ok(BoostOutput { used_pools, total_fee: total_fee_amount });
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1351,14 +1400,20 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+							channel_id,
 							deposit_details: deposit_details.clone(),
 							ingress_fee,
 							boost_fee: boost_fee_amount,
 							action,
 						});
 					},
 					Err(err) => {
+						Self::deposit_event(Event::InsufficientBoostLiquidity {
+							boost_id: id,
```
kylezs, line 1413: nit, but this makes it a bit easier to follow
```suggestion
							boost_id: prewitnessed_deposit_id,
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -626,14 +639,31 @@ pub mod pallet {
...
+			amount: TargetChainAmount<T, I>,
+		},
+		AccountStoppedBoosting {
+			account_id: T::AccountId,
+			boost_pool: BoostPoolId<T::TargetChain>,
+			account_available_amount: TargetChainAmount<T, I>,
+			account_in_use_amounts: BTreeMap<BoostId, TargetChainAmount<T, I>>,
+		},
+		InsufficientBoostLiquidity {
+			boost_id: BoostId,
```
kylezs, line 663: Is fair enough to include this boost id, but in the spec we have channel_id here too, so need to add that

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -626,14 +639,31 @@ pub mod pallet {
...
+		BoostFundsAdded {
+			account_id: T::AccountId,
+			boost_pool: BoostPoolId<T::TargetChain>,
+			amount: TargetChainAmount<T, I>,
+		},
+		AccountStoppedBoosting {
+			account_id: T::AccountId,
+			boost_pool: BoostPoolId<T::TargetChain>,
+			account_available_amount: TargetChainAmount<T, I>,
+			account_in_use_amounts: BTreeMap<BoostId, TargetChainAmount<T, I>>,
```
kylezs, line 660: A comment describing why we have two sets of amounts here would be useful for future readers

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -626,14 +639,31 @@ pub mod pallet {
...
 			ingress_fee: TargetChainAmount<T, I>,
 			boost_fee: TargetChainAmount<T, I>,
 			action: DepositAction<T::AccountId>,
 		},
+		BoostFundsAdded {
+			account_id: T::AccountId,
+			boost_pool: BoostPoolId<T::TargetChain>,
+			amount: TargetChainAmount<T, I>,
+		},
+		AccountStoppedBoosting {
```
kylezs, line 656: account is implied
```suggestion
		StoppedBoosting {
```

kylezs: Also the tests don't check for this event

kylezs (commented):
Nice stuff, just a few small mostly cosmetic things

state-chain/pallets/cf-ingress-egress/src/lib.rs
GabrielBuragev: Will there even be a `boost_id` if a boost failed?

GabrielBuragev: Yeah, we will need the `channel_id` here so we can link to the deposit.

kylezs: > Will there even be a `boost_id` if a boost failed?

yes - but your question suggests that maybe this naming is confusing. It's really the `prewitness_deposit_id` so perhaps we should just name it that @j4m1ef0rd

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1019,15 +1055,24 @@ pub mod pallet {
...
+					pool.stop_boosting(booster.clone())
+				})?;
 
 			T::LpBalance::try_credit_account(&booster, asset.into(), unlocked_amount.into())?;
 
+			Self::deposit_event(Event::AccountStoppedBoosting {
+				account_id: booster,
+				boost_pool: BoostPoolId { asset, tier: pool_tier },
+				account_available_amount: unlocked_amount,
+				account_in_use_amounts,
```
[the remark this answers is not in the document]
GabrielBuragev: I agree, maybe we can improve the naming here. 
`unlocked_amount` and `pending_amounts` sounds reasonable.

state-chain/pallets/cf-ingress-egress/src/lib.rs
GabrielBuragev, line 660: Is it possible to have a different key for this map? 
Preferably something that we could use for identifying the channel related to these boosts. `ChannelId` would allow us to do that although it might not be the most ergonomic approach.

Thinking about it a little bit more, we could also just include an array of all boost ids this account contributed to instead of a map with the amounts. We will then need to expose a `boost_id: BoostId` inside `DepositBoosted` & `DepositFinalised.action.BoostersCredited` which will allow us to keep track of these `pending` amounts on our end.
Of course, that is, if the `BoostId` stays the same for a specific deposit throughout its lifecycle (`DepositBoosted -> DepositFinalised.action.BoostersCredited`).

This might be a cleaner approach as we can then also omit the `channel_id` property from this event.

What do you think @kylezs

j4m1ef0rd: >  If so, will this still hold true in the case of a re-org scenario?

Yes. the ID is a counter that will only ever go up that is given when the deposit if prewitnessed.

> This might be a cleaner approach

I agree.

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: Yeah i agree. the prewitness_deposit_id becomes a boost id once it is boosted, so we should draw the line there.

kylezs (approved):

GabrielBuragev (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -626,14 +643,38 @@ pub mod pallet {
 		DepositBoosted {
 			deposit_address: TargetChainAccount<T, I>,
 			asset: TargetChainAsset<T, I>,
-			amount: TargetChainAmount<T, I>,
+			amounts: BTreeMap<BoostPoolTier, TargetChainAmount<T, I>>,
```
dandanlen, line 646: Minor nit: I think we should consider changing this into a `Vec<(BoostPoolTier, TargetChainAmount)>`, it's a bit more json-friendly (although this is SCALE so maybe it doesn't matter). 

But also - doesn't the boost_id already uniquely identify the amounts?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -626,14 +643,38 @@ pub mod pallet {
...
 			deposit_details: <T::TargetChain as Chain>::DepositDetails,
+			boost_id: BoostId,
```
dandanlen, line 648: Also I just realised that boost_id is just the prewitnessed_deposit_id. It's a bit strange that we use these interchangeably despite them having different type aliases... Why do we distinguish these?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -626,14 +643,38 @@ pub mod pallet {
...
 			// Ingress fee in the deposit asset. i.e. *NOT* the gas asset, if the deposit asset is
-			// a non-gas asset.
+			// a non-gas asset. The ingress fee is taken *after* the boost fee.
 			ingress_fee: TargetChainAmount<T, I>,
+			// Total fee the user paid for their deposit to be boosted.
 			boost_fee: TargetChainAmount<T, I>,
 			action: DepositAction<T::AccountId>,
 		},
+		BoostFundsAdded {
+			account_id: T::AccountId,
```
dandanlen, line 658: We sometimes use booster_id and sometimes account_id. Why not use booster_id everywhere we talk about boosting?

dandanlen (commented):
LGTM, Just some nitty comments (not all related to this PR...)

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: `BoostPoolTier` will be unique, that's why i explicitly used`BTreeMap`. I'm pretty sure SCALE will make it output the same as a vec. I prefer to leave it as is. Can change it later if we find its not suitable.

kylezs (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Yeah that's fine. I was just imagining a future where we have more Tiers, or have more complex tiers (for example a tuple of `(Price, MinConfirmations)`). Just something to keep to in mind. WE can change it later if needed.

dandanlen (approved):

PR #4746, opened 2024-04-10
Revert "fix: Wait for ThresholdSignature success before switching to NewKeysActivated (#4534)"

kylezs:
This reverts commit c1271c094f115f3904cddf2a50953a0951322949.

Fixes rotation edge case of
1. Threshold signature request for the Ethereum new key activation after a rotation
2. Threshold signature request for the Eth egress
3. Ethereum signer succeeded for the key rotation and broadcast is sent
4.  Ethereum signer succeeded for the Eth egress but is not sent for broadcast as it's waiting for the key activation (broadcast barrier)
5. Key activation is succesfully witnessed
6. Broadcast requested for the egress was requested now that the key activation has finished. That fails because it was signed with the old key.

### DIFF
--- a/state-chain/cf-integration-tests/src/authorities.rs
+++ b/state-chain/cf-integration-tests/src/authorities.rs
@@ -113,9 +113,6 @@ fn authority_rotates_with_correct_sequence() {
 				RotationPhase::ActivatingKeys(..)
 			));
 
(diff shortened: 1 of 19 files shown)


dandanlen:
Replacing with #4747

PR #4750, opened 2024-04-11
feat: make price impact limit configurable per pool

syan095:
# Pull Request

Closes: PRO-1288

## Checklist

## Summary
MaximumPriceImpact is now configurable per pool instead of a global value, so it can be better customized.


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -971,10 +973,14 @@ pub mod pallet {
 		#[pallet::weight(T::WeightInfo::set_maximum_price_impact())]
 		pub fn set_maximum_price_impact(
 			origin: OriginFor<T>,
+			base_asset: Asset,
+			quote_asset: Asset,
```
dandanlen, line 977: Can we make it `Vec<(Asset, Asset)>` or just `Vec<Asset>` (since we know the quote asset is always `STABLE_ASSET`)? 

I can imagine we might want to update these in batches rather than one at a time.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -971,10 +973,14 @@ pub mod pallet {
...
 			ticks: Option<u32>,
 		) -> DispatchResult {
 			T::EnsureGovernance::ensure_origin(origin)?;
-			MaximumPriceImpact::<T>::set(ticks);
+
+			let asset_pair = AssetPair::try_new::<T>(base_asset, quote_asset)?;
+			MaximumPriceImpact::<T>::mutate(asset_pair, |limit| *limit = ticks);
```
dandanlen, line 983: Let's also add an event `PriceImpactLimitSet { ticks }`.

(Also we can just set the value, we don't need to mutate).

state-chain/pallets/cf-pools/src/lib.rs
syan095: I see. Understood

state-chain/pallets/cf-pools/src/lib.rs
syan095: Since the input ticks is Option<u32>, (support the case where we want to remove the limit), with set(), we have to consider the case where the input is None.
```
match ticks {
	Some(tick) => MaximumPriceImpact::<T>::insert(asset_pair, tick),
	None => MaximumPriceImpact::<T>::remove(asset_pair),
}
```

I just thought that using 1 line of mutate() looks better

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Sorry I didn't explain very well - I should have said `Vec<(Asset, Tick)>` so we can update multiple pools with one governance call.

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: There is also a method `set`: `MaximumPriceImpact::<T>::set(asset_pair, ticks)` which does exactly this :)

dandanlen (approved):

PR #4752, opened 2024-04-11
feat: Authorities should be sorted pseudo-randomly

marcellorigotti:
# Pull Request

Closes: PRO-1201

## Checklist

## Summary

- Changed Storage Items (currentAuthorities, historicalAuthorities) to use a Vec instead of BTreeSet
- Added randomization when creating the new authority set such that authorities are not sorted based on their ID

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -28,7 +28,6 @@ use cf_traits::{
 	MissedAuthorshipSlots, OnAccountFunded, QualifyNode, RedemptionCheck, ReputationResetter,
 	SetSafeMode,
 };
-
 use cf_utilities::Port;
 use frame_support::{
 	pallet_prelude::*,
@@ -39,6 +38,7 @@ use frame_support::{
 	traits::{EstimateNextSessionRotation, OnKilledAccount},
 };
 use frame_system::pallet_prelude::*;
+use nanorand::{Rng, WyRand};
 pub use pallet::*;
 use sp_core::ed25519;
 use sp_std::{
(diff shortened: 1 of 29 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -399,8 +398,6 @@ pub mod pallet {
 					let num_primary_candidates = rotation_state.num_primary_candidates();
 					match T::KeyRotator::status() {
 						AsyncResult::Ready(KeyRotationStatusOuter::KeyHandoverComplete) => {
-							let new_authorities = rotation_state.authority_candidates();
-							HistoricalAuthorities::<T>::insert(rotation_state.new_epoch_index, new_authorities);
```
marcellorigotti, line 403: I removed these 2 lines because I think this is redundant, we add the old set to the HistoricalAuthorities once we complete the rotation inside `initialise_new_epoch` function.

marcellorigotti:
I tried to move the other migration touching the validator pallet and use the VersionedMigration, but I wasn't able to make it works, maybe because it touches 2 different pallet since it is moving the storage item from one to the other?

Still I updated it such that now both migrations runs correctly.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -932,8 +932,17 @@ impl<T: Config> Pallet<T> {
 			old_epoch,
 		);
 
-		let new_authorities = rotation_state.authority_candidates();
+		let mut new_authorities: Vec<<T as Chainflip>::ValidatorId> =
+			rotation_state.authority_candidates().into_iter().collect();
+		let hash = frame_system::Pallet::<T>::block_hash(
+			frame_system::Pallet::<T>::current_block_number(),
+		);
```
dandanlen, line 939: Actually I don't think this will work - the block hash for the current block doesn't exist until *after* it's authored. 

We will have to use the parent hash (`frame_system::Pallet<T>::parent_hash()`).

dandanlen:
LGTM apart from what I mentioned above. I'll sort out the migrations.

dandanlen:
Actually I think there might be another issue... looking.

dandanlen:
Rebased to resolve conflicts.

As discussed - the shuffle needs to be moved in the SessionManager trait impl.

dandanlen (approved):

PR #4753, opened 2024-04-11
fix: set initial arb block

kylezs:
Was failing after upgrade because this wasn't set.

### DIFF
--- a/state-chain/runtime/src/migrations/arbitrum_integration.rs
+++ b/state-chain/runtime/src/migrations/arbitrum_integration.rs
@@ -1,5 +1,5 @@
 use crate::{safe_mode, Runtime};
-use cf_chains::{arb, arb::ArbitrumTrackedData, eth::Address, ChainState};
+use cf_chains::{arb::ArbitrumTrackedData, eth::Address, ChainState};
 use cf_traits::SafeMode;
 use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
 #[cfg(feature = "try-runtime")]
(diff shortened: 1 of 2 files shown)


state-chain/runtime/src/migrations/arbitrum_integration.rs
```diff
@@ -87,36 +91,43 @@ impl OnRuntimeUpgrade for ArbitrumIntegration {
 				address_checker_address,
 				chain_id,
 				usdc_address,
-			): (Address, Address, Address, u64, Address) =
+				start_block_number,
+			): (Address, Address, Address, u64, Address, u64) =
 		match cf_runtime_upgrade_utilities::genesis_hashes::genesis_hash::<Runtime>() {
 			cf_runtime_upgrade_utilities::genesis_hashes::BERGHAIN =>
 			{
+				panic!("Need to set up arbitrum integration for Berghain");
```
kylezs, line 99: these are so we don't forget to set these to the real values when we're ready to set them

state-chain/runtime/src/migrations/arbitrum_integration.rs
```diff
@@ -138,6 +150,12 @@ impl OnRuntimeUpgrade for ArbitrumIntegration {
 		);
 		pallet_cf_environment::ArbitrumChainId::<Runtime>::put(chain_id);
 		pallet_cf_environment::ArbitrumSupportedAssets::<Runtime>::insert(ArbUsdc, usdc_address);
+		pallet_cf_chain_tracking::CurrentChainState::<Runtime, ArbitrumInstance>::put(ChainState {
+			block_height: start_block_number,
+			tracked_data: ArbitrumTrackedData {
+				base_fee: 0,
```
kylezs, line 156: after a single chain tracking operation, this will be filled, and ready for any fee related estimates, won't cause any trouble not setting this "correctly"

albert-llimos (approved):

PR #4758, opened 2024-04-12
fix: clean up broadcast state on success only

dandanlen:
This is a follow-up from the recent issue with the broadcasting on berghain. 

- Added a trait method for re-signing aborted broadcasts
- Use this for CCM re-signing
- Added a governance extrinsic for resigning aborted broadcasts
- more consistent cleanup of broadcast state:
    - When a broadcast *succeeds* do the cleanup, including the callbacks.
    - When it *fails* don't do the cleanup so we can use the state to recover.
    - When re-signing, first do the cleanup of the leftover state.

I want to make some further changes in a future PR:
- i think we can remove the failure callback and move automatic re-signing  a general thing.
- then we can remove CCM resigning from the ingress egress pallet.


dandanlen:
@ramizhasan111 / @syan095 can one of you review this please.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -944,13 +943,33 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
...
 	}
 
-	/// Clean up storage data related to a broadcast ID.
-	fn clean_up_broadcast_storage(broadcast_id: BroadcastId) {
+	fn expire_broadcast(broadcast_id: BroadcastId) {
+		// These would otherwise be cleaned up when the broadacst succeeds or aborts.
+		RequestSuccessCallbacks::<T, I>::remove(broadcast_id);
+		RequestFailureCallbacks::<T, I>::remove(broadcast_id);
 		Self::clean_up_broadcast_storage(broadcast_id);
 	}
```
ramizhasan111, line 974: maybe we should call the ```clean_up_broadcast_storage``` in abort_broadcast instead of calling separately in ```re_sign_broadcast``` and ```expire_broadcast``` which should be equivalent. This would be more readable conceptually since the ```clean_up_broadcast_storage``` clears all the broadcast related storages which become irrelevant when we abort (since we dont broadcast anymoe).

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -944,13 +943,33 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
...
+
+		PendingBroadcasts::<T, I>::try_mutate(|pending| {
+			if pending.contains(&broadcast_id) {
+				Err(Error::<T, I>::BroadcastStillPending)
+			} else {
+				if request_broadcast {
+					pending.insert(broadcast_id);
+				}
+				Ok(())
+			}
```
ramizhasan111, line 964: Having the broadcast in pending_broadcasts or not will not broadcast or not broadcast it. The ```threshold_sign``` function will cause it broadcast anyways at least once.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -811,32 +811,35 @@ pub mod pallet {
...
-							// We are here if the Call needs to be resigned, yet no API call data is
-							// available to use. In this situation, there's nothing else that can be
-							// done.
-							log::error!("Foreign Chain Call message cannot be re-signed: Call data unavailable. Broadcast Id: {:?}", call.broadcast_id);
-						}
+						},
+						Err(e) => {
+							// This can happen if a broadcast is still pending despite
+							// since the previous epoch.
+							// TODO: make sure this can't happen.
```
ramizhasan111, line 831: this case could happen theoretically since the braodcast timeout seems to be 10 mins and for a broadcast to be aborted, it will take more than 150*10 mins which turns out to be 25 hours (a little more than 1 epoch time).

Janislav, line 831: I guess in this case, we have to again re-sign and broadcast, right? Alternative, we could emit an event here and all the data in storage and kick off the process again manually, since this seems to be an edge case.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -633,18 +641,32 @@ pub mod pallet {
 			Self::handle_broadcast_failure(broadcast_id, reporter.into())?;
 			Ok(().into())
 		}
+
+		/// Re-sign and optionally re-send a broadcast request.
+		///
+		/// Requires governance origin.
+		#[pallet::call_index(5)]
+		#[pallet::weight(Weight::zero())]
+		pub fn resign_aborted_broadcast(
```
Janislav, line 650: It feels like there should a particular unit test for this

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -633,18 +641,32 @@ pub mod pallet {
 			Self::handle_broadcast_failure(broadcast_id, reporter.into())?;
 			Ok(().into())
 		}
+
+		/// Re-sign and optionally re-send a broadcast request.
+		///
+		/// Requires governance origin.
+		#[pallet::call_index(5)]
+		#[pallet::weight(Weight::zero())]
```
Janislav, line 649: I think it would be better to benchmark it properly or "over weight" it. Otherwise, this would open a D-Dos attack.

state-chain/traits/src/mocks/broadcaster.rs
```diff
@@ -17,6 +19,49 @@ impl<T> MockPallet for MockBroadcaster<T> {
...
+	fn chain_encoded(&self) -> Vec<u8> {
+		self.encode()
+	}
+
+	fn is_signed(&self) -> bool {
+		self.is_signed
+	}
+
+	fn transaction_out_id(&self) -> <C as cf_chains::ChainCrypto>::TransactionOutId {
+		unimplemented!()
```
Janislav, line 61: Is this not leading to a panic during test execution? 🤔

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: Not quite sure what you mean here.

`threshold_sign` does not update the PendingBroadcasts, so we need to do it here. 

We also only want to allow broadcasts to be re-signed if they are aborted (otherwise we might end up with the same request having multiple simultaneous broadcast requests). 

Does that make sense?

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: If we call it in `abort_broadcast` then we discard the api-call, which we need in order to re-sign.

state-chain/pallets/cf-broadcast/src/lib.rs
ramizhasan111: sorry for the confusing language. what i was saying was not correct anyways so we good here.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Yes I thought so but I wasn't sure if for example the pending ccm would prevent the rotation from completing (due to the broadcast barrier).

dandanlen: Yes - for now we have the option of giving it a bump using governance but of course this isn't ideal. Needs some thought..

dandanlen: I have been thinking we could move this auto-resigning of old txs into the broadcast pallet instead, since the broadcast pallet is more aware of the pending/aborted/succeeded states of the txs.

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: This isn't the only place we do this... see PRO-770.

state-chain/traits/src/mocks/broadcaster.rs
dandanlen: Tests are passing so I guess not 😅  ... will take a look.

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -633,18 +641,32 @@ pub mod pallet {
...
 
 impl<T: Config<I>, I: 'static> Pallet<T, I> {
-	pub fn clean_up_broadcast_storage(broadcast_id: BroadcastId) {
+	pub fn clean_up_broadcast_storage(broadcast_id: BroadcastId) -> Option<ApiCallFor<T, I>> {
 		AwaitingBroadcast::<T, I>::remove(broadcast_id);
 		TransactionMetadata::<T, I>::remove(broadcast_id);
-		RequestSuccessCallbacks::<T, I>::remove(broadcast_id);
-		RequestFailureCallbacks::<T, I>::remove(broadcast_id);
-		if let Some((api_call, _)) = ThresholdSignatureData::<T, I>::take(broadcast_id) {
+		ThresholdSignatureData::<T, I>::take(broadcast_id).map(|(api_call, _)| {
```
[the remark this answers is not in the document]
dandanlen: Well spotted, will remove.

state-chain/pallets/cf-broadcast/src/lib.rs
ramizhasan111: hm.. we could not delete the ```ThresholdSignatureData``` storage in ```clean_up_broadcast_storage``` which contains the apicall. We could then delete it in the expiry and broadcast success. ```ThresholdSignatureData``` is actually just a stored apicall for each broadcast_id. The signature field is not used (see the other comment). It should probably be renamed to something like ```Apicalls```. This way (not deleting the apicall storage and calling cleanup in abort broadcast) feels more intuitive to me but I guess it is a matter of choice so I am ok if you dont want to change it.

state-chain/pallets/cf-broadcast/src/lib.rs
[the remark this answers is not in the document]
ramizhasan111: while we are at it, maybe we should also rename it to something like ```Apicalls``` since now its just storing apicalls indexed by broadcast ids

state-chain/pallets/cf-ingress-egress/src/lib.rs
ramizhasan111: you are right, its true that broadcast barrier will stop this from happening but relying on broadcast barrier is a bit problematic since its chain specific implementation can be different. For example, there can be a chain where there is no broadcast barrier.

dandanlen: I agree we should fix this but will leave it out of scope for this PR.
For now the only chains supporting CCM are EVM and have the same restrictions, so it's ok. I think the correct solution is to trigger re-signing of these calls when we rotate, but this is a larger piece of work... I'll open a Linear ticket.

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: Done.

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: I can see your point, will rename and see if I think it makes sense to restructure further.

dandanlen: I did the renaming.

I also had another look and I don't think one approach is clearly better than the other: when we succeed we want to remove everything, but when we abort we want to keep the api call. So either way there needs to be some flag or special case in the cleanup. Will leave it as is.

state-chain/traits/src/mocks/broadcaster.rs
dandanlen: The broadcast pallet mocks use a different MockApiCall (from cf_chains)... 

This Mock is only required to satisfy the trait requirements of the MockBroadcaster... so in fact this `transaction_out_id` never gets called and that's fine.

dandanlen:
I think this is ready? There are some remaining things (like ccm re-signing) that I think can be addressed separately.

@ramizhasan111 / @Janislav, let me know if there's anything else.

ramizhasan111:
> I think this is ready? There are some remaining things (like ccm re-signing) that I think can be addressed separately.
> 
> @ramizhasan111 / @Janislav, let me know if there's anything else.

looks good to me. Will approve

ramizhasan111 (approved):

PR #4770, opened 2024-04-17
fix: arb vault rotation migration

kylezs:
Fixes:
- setup the arbitrum vault as part of the upgrade
- use correct contract addresses in arbitrum migration
- add arbitrum liquidity and boost liquidity as part of the upgrade

### DIFF
--- a/bouncer/shared/upgrade_network.ts
+++ b/bouncer/shared/upgrade_network.ts
@@ -9,6 +9,7 @@ import { bumpSpecVersionAgainstNetwork } from './utils/spec_version';
 import { compileBinaries } from './utils/compile_binaries';
 import { submitRuntimeUpgradeWithRestrictions } from './submit_runtime_upgrade';
 import { execWithLog } from './utils/exec_with_log';
+import { setupArbVault } from './setup_arb_vault';
 
 async function readPackageTomlVersion(projectRoot: string): Promise<string> {
   const data = await fs.readFile(path.join(projectRoot, '/state-chain/runtime/Cargo.toml'), 'utf8');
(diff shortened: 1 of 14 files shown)


bouncer/shared/upgrade_network.ts
```diff
@@ -68,22 +73,34 @@ async function incompatibleUpgradeNoBuild(
 
   console.log('Engines started');
 
-  await submitRuntimeUpgradeWithRestrictions(runtimePath, undefined, undefined, true);
+  await submitRuntimeUpgradeWithRestrictions(runtimePath, undefined, undefined, false);
```
albert-llimos, line 76: Is this on purpose? I know you told me to change that to run it locally, but is it correct to change it here? Just checking it hasn't accidentally slipped into the PR.

engine/src/witness/arb.rs
```diff
@@ -173,3 +173,97 @@ where
...
+			.await,
+		);
+
+		task_scope(|scope| {
+			async move {
+				let (state_chain_stream, _unfinalised_state_chain_stream, state_chain_client) =
+					state_chain_observer::client::StateChainClient::connect_with_account(
+						scope,
+						"ws://localhost:9944",
+						PathBuf::from("/Users/kylezs/Documents/cf-repos/chainflip-backend/localnet/init/keys/bashful/signing_key_file").as_path(),
```
albert-llimos, line 215: probably a local path that sneaked in there. There is some others.

kylezs: it's an ignored test that will need local paths changed anyway so i left it

albert-llimos (approved):

PR #4773, opened 2024-04-18
fix: use `ubuntu-22.04` for benchmark runner provisioning workflow 🐛

ahasna:


kylezs (approved):
👋 final use of ubuntu 20

PR #4777, opened 2024-04-18
feat: Affiliate Brokers

marcellorigotti:
# Pull Request

Closes: PRO-1273
## Checklist

## Summary

This PR includes the possibility of splitting the broker fee among multiple brokers account, every account specified to receive the fees must be a valid account registered as broker.

This PR is breaking, changing the extrinsic `request_swap_deposit_address` and the event `SwapDepositAddressReady`:
```
SwapDepositAddressReady {
	deposit_address: EncodedAddress,
	source_asset: Asset,
	destination_asset: Asset,
	channel_id: ChannelId,
	broker_commission_rate: BasisPoints,
	channel_metadata: Option<CcmChannelMetadata>,
	source_chain_expiry_block: <AnyChain as Chain>::ChainBlockNumber,
	boost_fee: BasisPoints,
	channel_opening_fee: T::Amount,
	beneficiaries: Vec<Beneficiary<T::AccountId>>,
}
```
where Beneficiary is defined as follow:
```
pub struct Beneficiary<Id> {
	pub account: Id,
	pub bps: BasisPoints,
}
```

Broker API is updated as well to support multiple brokers, it is still backwards compatible, these are two example of the curl commands you can use now to open a swap channel:

- `curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "broker_request_swap_deposit_address", "params": ["ETH", "FLIP","0xabababababababababababababababababababab", 0]}' http://localhost:62378`
Same as the old command, used for a single broker

- `curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "broker_request_swap_deposit_address", "params": ["ETH", "FLIP","0xabababababababababababababababababababab", [{"account":"cFJWxzSP35vz6EnsWfp6tejZfnpeQ1ra3NHXXbrxSEn4GvsJ3","bps":2},{"account":"cFM8kRvLBXagj6ZXvrt7wCM4jGmHvb5842jTtXXg3mRHjrvKy","bps":4}]]}' http://localhost:62378`
New command used for splitting fees among different brokers


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -479,13 +483,42 @@ pub mod pallet {
...
+			broker_commission: BrokerFees<T::AccountId>,
 			channel_metadata: Option<CcmChannelMetadata>,
 			boost_fee: BasisPoints,
 		) -> DispatchResult {
 			ensure!(T::SafeMode::get().deposits_enabled, Error::<T>::DepositsDisabled);
 			let broker = T::AccountRoleRegistry::ensure_broker(origin)?;
-			ensure!(broker_commission_bps <= 1000, Error::<T>::BrokerCommissionBpsTooHigh);
+			let mut broker_fees: Vec<Beneficiary<T::AccountId>> = vec![];
+			let total_bps: BasisPoints;
+			match broker_commission {
```
dandanlen, line 494: This could be more idiomatic. 

Try writing it like this:

```rust
let (beneficiaries, total_bps) = match broker_commission {
	// ...
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -479,13 +483,42 @@ pub mod pallet {
...
+				BrokerFees::Single(broker_bps) => {
+					ensure!(broker_bps <= 1000, Error::<T>::BrokerCommissionBpsTooHigh);
```
dandanlen, line 496: This `ensure` can be applied after we determine the total.

dandanlen (commented):
Nice work. Just some minor suggestions.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1947,7 +1944,7 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
 		source_asset: TargetChainAsset<T, I>,
 		destination_asset: Asset,
 		destination_address: ForeignChainAddress,
-		broker_commission_bps: BasisPoints,
+		broker_commission: Vec<Beneficiary<Self::AccountId>>,
```
acdibble, line 1947: I understand that this is backwards compatible in the broker API, but is it possible to make the extrinsic backwards compatible? There are people building on the protocol that interact directly with these extrinsics.

dandanlen: We were unsure about this and decided to just go with the breaking change. But I guess it can't hurt to avoid this.

We can keep (but deprecate) the old extrinsic and add a new one - this is the simplest way to keep compatibility. @marcellorigotti can you please change this?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -698,6 +675,94 @@ pub mod pallet {
...
+							Error::<T>::BeneficiaryAccountIsNotABroker
+						);
+					}
+					(
+						(*fees_bps)
+							.clone()
+							.into_iter()
+							.filter(|Beneficiary { account: _, bps }| *bps > 0)
+							.collect::<Vec<_>>().try_into().expect("The provided beneficiaries vec cannot have more than MAX_BENEFICIARIES elements"),
+						fees_bps.iter().fold(0, |acc, Beneficiary { account: _, bps }| acc + bps),
```
marcellorigotti, line 725: I am not sure if there is a better way to do this but we are pretty sure this conversion can never fail since the argument passed to the extrinsic is already a BoundedVec.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -698,6 +675,94 @@ pub mod pallet {
...
+							T::AccountRoleRegistry::has_account_role(&account, AccountRole::Broker),
+							Error::<T>::BeneficiaryAccountIsNotABroker
+						);
+					}
+					(
+						(*fees_bps)
+							.clone()
+							.into_iter()
+							.filter(|Beneficiary { account: _, bps }| *bps > 0)
+							.collect::<Vec<_>>().try_into().expect("The provided beneficiaries vec cannot have more than MAX_BENEFICIARIES elements"),
```
acdibble, line 724: it would be more useful to the user (and better for us because we have to spend less time answering this question) to include the actual maximum count in the error message. As it stands, this is just a constant in code, so the user can't look up the correct number either without diving into the backend code afaik.

dandanlen: This isn't the error message - Think of it as more like a code comment about why we never expect to hit this case.
If anyone tries to send too many beneficiaries, it will cause a decoding error and should bubble up to the surface in the broker api (tbc).

marcellorigotti: Yeah can confirm, just tried to submit through the broker api and directly using the new extrinsic and it fails in both cases during decoding for the broker and during verification for the extrinsic.

dandanlen: This isn't an error message that anyone will ever see - Think of it as more like a code comment about why we never expect to hit this case.

If anyone tries to send too many beneficiaries, it will cause a decoding error and should bubble up to the surface in the broker api (i'll check what this actually returns in practice).

dandanlen: And if you try to send an extrinsic with too many values, it will return a decoding error too. If you're using a substrate framework, it should protect against this. 

If you're rolling your own, you should still get a sensible error message about failing to decode the BoundedVec.

acdibble: `data did not match any variant of untagged enum BrokerFees` can be improved in the broker API, I agree that if you are directly submitting the extrinsic yourself, you should be prepared to debug the error messages (even though in practice it means they will ping us)

state-chain/primitives/src/lib.rs
```diff
@@ -322,3 +322,37 @@ fn is_more_recent_semver() {
...
 	ensure_left_is_more_recent(ver(1, 1, 0), ver(1, 0, 2));
 }
+
+pub const MAX_BENEFICIARIES: u32 = 4;
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum BrokerFees<Id> {
+	Single(BasisPoints),
+	Multiple(BoundedVec<Beneficiary<Id>, ConstU32<MAX_BENEFICIARIES>>),
```
acdibble, line 332: Does it make sense to allow for the super(?)broker to also specify a commission rate directly without consuming a beneficiary slot?

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Yeah I also think that the use case of splitting your fees between 5 accounts is remote enough that we can accept the risk that the error message might not be perfect.

BrokerFees is another matter - can definitely be improved.

state-chain/primitives/src/lib.rs
dandanlen: This type exists mainly to allow backwards compatibility in the broker api. It's also more efficient (saves 32 bytes per extrinsic) and therefore cheaper for the user and the network in the base case where you don't want to split fees. This is why we decided to use it in the extrinsic too.

acdibble: I understand, but was wondering if it makes sense to include a separate bps field in the `Multiple` variant that allows the broker opening the channel to specify their own bps without having to consume one of the beneficiary slots

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: See: [cac5f7e](https://github.com/chainflip-io/chainflip-backend/pull/4777/commits/cac5f7e4f3ad02601989e03f40cad9dfec26e175)

marcellorigotti:
Before merging let's wait for @fernandezdavid to confirm what the max number of sub-brokers/beneficiaries should be!

acdibble:
> Before merging let's wait for @fernandezdavid to confirm what the max number of sub-brokers/beneficiaries should be!

We got a response that 3 is great and 5 is "max futureproof mode"

dandanlen:
LGTM, thanks @marcellorigotti !

Would like to also wait for confirmation re. naming before merging this (and depending on the names for eg event processing).

dandanlen:
Ok, I did a final pass, incorporated the name "affiliates" rather than "beneficiaries".

In doing so, I now have the feeling that (particularly in light of the 'affiliate' naming) a slightly clearer public API might be to accept both an explicit `broker_commission` (for yourself) and an optional list of `affiliate_fees` (for others) rather than bundling them into the same `BrokerFees` enum.

I think maybe this is what you were getting at with previous comments @acdibble - do you have an opinion either way? If there are no objections, I feel like the explicit option (separate commission + affiliate fees) is probably better. (sorry for the late u-turn)

acdibble:
> I think maybe this is what you were getting at with previous comments @acdibble - do you have an opinion either way? If there are no objections, I feel like the explicit option (separate commission + affiliate fees) is probably better. (sorry for the late u-turn)

works for me!

dandanlen:
I made some further small (and opinionated) changes:

- removed the deprecation notice and renamed _v2 to _with_affiliates (might as well keep both)
- separate benchmarks for each extrinsic (no affiliates is cheaper)
- remove the Option<..> in the affiliates arg of the extrinsic (empty vec is equivalent) but keep it in the json api (to ensure the apis will accept a missing arg for backwards compat).
- updated the event with a new field `broker_fee` and deprecated `broker_commission`. The idea being that `broker_fee` should refer to the total fee, whereas broker_commission is explicitly reserved for the broker (ie. not including the affiliates), to match the input. `broker_fees` refers to the list of beneficiaries, ie. maybe broker plus affiliates, but could also be empty. `affiliate_fees` means only the affiliates.
- changed the above naming on `ChannelAction`.

Note the above means that for example on the `SwapDepositAddressReady` event, we now have:
broker_commission_rate: same as the old one, ie. the rate the broker gets, *not* the total.
affiliate_fees: the additional fees that will be paid out to affiliates.


Sorry for the back-and-forth. I'm reasonably happy with this now - @acdibble if you're ok with the exposed api / events, please approve and merge.

acdibble:
> Sorry for the back-and-forth. I'm reasonably happy with this now - @acdibble if you're ok with the exposed api / events, please approve and merge.

Looks good. One more thing though after all the refactoring the new extrinsic is the same as the old one except it has one new parameter. Is it possible (and wise) to consolidate them into one extrinsic with the `affiliates` field as an `Option<Affiliates>`?

acdibble (approved):

marcellorigotti:
I unified the two extrinsic into one, I think it makes sense given the directions we took, it feels unnecessary to have an extra extrinsic just for an optional field.
I don't know if there are more "hidden" reason to keep them separate, the changes are there is @dandanlen is happy with it we can merge otherwise feel free to delete the last commit 👍

dandanlen:
The reason for not consolidating was backwards compatibility for anyone using the raw extrinsics rather than the api. 

Backwards compatibility doesn't work the same in SCALE as in JSON - in JSON you can just leave of the extra arg but in SCALE you need an explicit marker byte to know that the value is empty. I would rather we keep the old version for that reason (happy to mark as deprecated, but at the same time it's very little overhead to keep it there).

dandanlen:
> The reason for not consolidating was backwards compatibility for anyone using the raw extrinsics rather than the api.
> 
> Backwards compatibility doesn't work the same in SCALE as in JSON - in JSON you can just leave of the extra arg but in SCALE you need an explicit marker byte to know that the value is empty. I would rather we keep the old version for that reason (happy to mark as deprecated, but at the same time it's very little overhead to keep it there).

I made this change (reverted the commit that consolidated the extrinsics) and merged with latest main - once CI passes this is good to go.

dandanlen (approved):

dandanlen:
I'll fix the bouncer issue.

PR #4779, opened 2024-04-19
feat: added safe mode for ingress egress pallet

j4m1ef0rd:
# Pull Request

Closes: PRO-1300

## Checklist

## Summary

- Added 3 safe mode enable flags for the boost features (add, stop, boost). These are implemented for each chain.
- Changed the eth and btc mocks to work with the chain instance's as it was now needed.
- Added unit tests for the 3 flags.
- Added the new safe mode settings to the 1.4 arbitrum migration and set them to CODE_RED by default, so for testnet/mainnet we can turn the boost features on when we are ready.
	- For localnets, the boost features will be defaultly enabled at genesis.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/mock_eth.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/mock_eth.rs
@@ -1,5 +1,5 @@
 pub use crate::{self as pallet_cf_ingress_egress};
-use crate::{DepositBalances, DepositWitness};
+use crate::{DepositBalances, DepositWitness, Instance1, PalletSafeMode};
 
 use cf_chains::eth::EthereumTrackedData;
 pub use cf_chains::{
@@ -14,7 +14,7 @@ pub use cf_primitives::{
 };
 use cf_test_utilities::{impl_test_helpers, TestExternalities};
 use cf_traits::{
-	impl_mock_callback, impl_mock_chainflip,
+	impl_mock_callback, impl_mock_chainflip, impl_mock_runtime_safe_mode,
 	mocks::{
 		address_converter::MockAddressConverter,
 		api_call::{MockEthereumApiCall, MockEvmEnvironment},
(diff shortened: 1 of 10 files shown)


kylezs (approved):
Just made one small change, to rename from swap to deposit, which is more accurate

state-chain/pallets/cf-ingress-egress/src/mock_eth.rs
```diff
@@ -109,15 +109,17 @@ impl NetworkEnvironmentProvider for MockNetworkEnvironmentProvider {
 	}
 }
 
-impl crate::Config for Test {
+impl_mock_runtime_safe_mode! { ingress_egress_ethereum: PalletSafeMode<Instance1> }
```
dandanlen, line 112: Doing this would avoid having to add the `Instance1` in all the tests.

```suggestion
impl_mock_runtime_safe_mode! { ingress_egress_ethereum: PalletSafeMode<()> }
```

dandanlen (approved):

state-chain/pallets/cf-ingress-egress/src/mock_eth.rs
dandanlen: (The actual value of `Instance` only matters if there are multiple instances in the runtime).

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
```diff
@@ -787,3 +802,152 @@ fn test_add_boost_funds() {
 		}));
 	});
 }
+
+#[track_caller]
+fn turn_safe_mode_on() {
+	<MockRuntimeSafeMode as SetSafeMode<MockRuntimeSafeMode>>::set_code_red();
```
msgmaxim, line 808: Would be good to also check the state before it is updated (in the test below too). Otherwise we might get a false positive if the asserted value is the default one.

msgmaxim:
Didn't we want to have separate switches for each of the capabilities (e.g. see https://linear.app/chainflip/issue/PRO-1300/boost-safe-mode#comment-40578bf2)? Right now it looks like it is only possible to either disable or enable everything at once.

j4m1ef0rd:
> Right now it looks like it is only possible to either disable or enable everything at once

I'm not exactly sure, but I think you can set the safe mode using the defined struct that has the 3 enable flags in it, sending the struct via an extrinsic,. Allowing for granular control. But for the tests we just use the SafeMode impl that sets them all on or off. @kylezs or @syan095 might be able to confirm this?

msgmaxim (approved):

PR #4780, opened 2024-04-19
feat: cf_boost_pool_details rpc

msgmaxim:
# Pull Request

Closes: PRO-1324

## Checklist

## Summary

Added `cf_boost_pool_details` rpc call which takes asset as the sole parameter and returns details (amounts, pendings boosts and pending withdrawals) of all existing pools for that asset. I wasn't sure if we wanted pool tier as a parameter as well (see my question in PRO-1324), but the approach I took here made the most sense to me (changing this wouldn't be difficult, so happy to do so if necessary).

I considered simply making the fields of `BoostPool` public and read them directly when constructing `BoostPoolDetails` in the rpc handler, but in the end I thought it was important to keep `ScaledAmount` an implementation detail of BoostPool, so I added getters which convert all usages of `ScaledAmount` to ChainAsset first (which is consitent with the rest of BoostPool's interface.)

Usage:
```
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "cf_boost_pool_details", "params": ["ETH", "10"]}' http://localhost:9944
```

See sample response here: https://github.com/chainflip-io/chainflip-backend/blob/716a4c9e0bcfb85ea3f3bfac6d8304a87457eed7/state-chain/custom-rpc/src/snapshots/custom_rpc__test__boost_details_serialization.snap


kylezs (approved):
Looks good to me, would await @GabrielBuragev 's approval before merging

j4m1ef0rd (approved):

GabrielBuragev:
Are the parameters (asset, tier) optional? 
I think it would be nice to be able to call this endpoint without providing any parameters and it would return the details for all the pools.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -66,6 +69,14 @@ pub struct BoostPoolDepth {
 	pub available_amount: AssetAmount,
 }
 
+#[derive(Encode, Decode, Eq, PartialEq, TypeInfo, Serialize, Deserialize)]
+pub struct BoostPoolDetails {
+	pub available_amount: AssetAmount,
+	pub amounts: BTreeMap<AccountId32, AssetAmount>,
```
GabrielBuragev, line 75: I think we can simply return one property `available_amounts: BTreeMap<AccountId32, AssetAmount>` instead of 2 separate.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -66,6 +69,14 @@ pub struct BoostPoolDepth {
...
+	pub pending_boosts: BTreeMap<PrewitnessedDepositId, BTreeMap<AccountId32, AssetAmount>>,
```
GabrielBuragev, line 76: `pending_boosts` might be a bit confusing here. Maybe `deposits_pending_finalization` or `pending_amounts` ?

GabrielBuragev: Also, can we adjust it so it follows the specs defined?
```
deposits_pending_finalization: [{
			channel_id,
			source_chain,
			asset,
			booster_shares:[{
				account_id,
				amount
			}]	
		}] // We can derive the total in_use amount from this
```

GabrielBuragev:
> "result": {
    "5": {
      "available_amount": 1000,
      "amounts": {
        "cFPdef3hF5zEwbWUG6ZaCJ3X7mTvEeAog7HxZ8QyFcCgDVGDM": 1000
      },
      "pending_boosts": {},
      "pending_withdrawals": {}
    },
    "10": {
      "available_amount": 0,
      "amounts": {},
      "pending_boosts": {
        "1": {
          "cFHsUq1uK5opJudRDczhdPVj6LGoVTqYsfj71tbHfKsTAzkJJ": 500
        }
      },
      "pending_withdrawals": {
        "cFHsUq1uK5opJudRDczhdPVj6LGoVTqYsfj71tbHfKsTAzkJJ": [
          1
        ]
      }
    },
    "30": {
      "available_amount": 0,
      "amounts": {},
      "pending_boosts": {},
      "pending_withdrawals": {}
    },
  },
    
Can we return an array instead of an object with `feeTier` keys? The defined schema spec should give you an idea of the preferred format for the response.

state-chain/runtime/src/runtime_apis.rs
msgmaxim: Not sure how I missed this part of the spec on Notion (I find Notion updates hard to track sometimes...). I thought I'd just start by doing the easiest thing, which is exposing all information (which is more or less how it is described in the linear issue), and make any adjustments if requested.

Now that I see the spec, I have questions:
- Why is it necessary to include `source_chain`/`asset`? This RPC for a specific asset already, so it is already known by the caller of this RPC.
- Do we actually want `channel_id` rather than the deposit id (as I'm providing currently)? Channels can have more than one deposit, so I think the deposit id might be necessary to uniquely identify the deposit, and it is more the question whether we want to *additionally* provide channel_id.

GabrielBuragev: > Why is it necessary to include source_chain/asset? This RPC for a specific asset already, so it is already known by the caller of this RPC.

Initially, i thought this RPC would have the ability to return the details for all the existing boost pools and the asset/tier params would be optional. The preferred way would be to allow the following:
- No params provided - return the details for all boost pools
- `asset` param provided - return the details for all boost pools for the provided asset
- `asset` & `tier` provided - return the details only for the specific asset/tier pool 

In the first case we will need the asset/source_chain included in the specific object. So lets include it and enable these 3 use cases above.

> Do we actually want channel_id rather than the deposit id (as I'm providing currently)? Channels can have more than one deposit, so I think the deposit id might be necessary to uniquely identify the deposit, and it is more the question whether we want to additionally provide channel_id.

Thats a good point. Lets include the `prewitnessed_deposit_id` here. But lets also keep the other fields for convenience please (`channel_id`, `source_chain`, `asset`). We can do a clean up in the end based on what ended up being used or not. I really can not recall the exact reason why we included `channel_id` here and not `prewitnessed_deposit_id`.

GabrielBuragev: > Does this mean that we are only interested in the amounts (the sum of them?) rather than which deposits we are waiting for? Seems like both are important and if you have the deposit id (as I provide currently), you'd be able to derive the amount, but not the other way around.

Its a good point, we might further improve this rpc by including the deposit ids besides the other fields too (`account_id`, `amount`).
I would rather we expose as much details as possible on this RPC. Exposing only the ids means that we have to look elsewhere to get the details related to that deposit.

GabrielBuragev:
> Can I ask how this will be used?

It will be used differently in different places. 
- On the LP app, we will probably use it on the client side (web) to show the stats for the specific LP
- In the processor/indexer we will have to use this RPC (specifically the `deposits_pending_finalization`) to associate the shares per account for each boost (this was decided as an alternative to emitting these amounts in an event which would be a lot of data)
- In our cache, we will probably use this to keep in sync with the current state chain state so we can use it in the block explorer and other apps (displaying non-historical data, only present)

Lets ignore the filtering by `account_id` for now and include it at a later point if really needed. Its not a hard requirement.

msgmaxim:
> Lets ignore the filtering by account_id for now and include it at a later point if really needed. Its not a hard requirement.

OK. Though I still want to understand the motivation behind this.

> On the LP app, we will probably use it on the client side (web) to show the stats for the specific LP

Makes sense if it will be used by clients' browsers. Though my understanding was that node's RPC server was to be used by the app's backend, and the users' browsers aren't making direct rpc requrests to a node. If the calls go through the backend, then you'd already have all the data for each LP on the backend, and you could filter internally. Is there a reason for clients to call node RPCs directly (other than this being potentially more work for the backend to expose this)?

> In the processor/indexer we will have to use this RPC (specifically the deposits_pending_finalization) to associate the shares per account for each boost


> In our cache, we will probably use this to keep in sync

It seems like for these two the filtering by LP isn't needed? I was specifically asking for when we'd need to filter by account_id (though I appreciate your comprehensive response).

state-chain/runtime/src/runtime_apis.rs
msgmaxim: >  Exposing only the ids means that we have to look elsewhere to get the details related to that deposit.

But the amounts are already exposed this very response under `pending_boosts`. I don't think we should return the amounts twice (similar to how you suggested that total_available_amount wasn't necessary).

msgmaxim: > But lets also keep the other fields for convenience please (channel_id, source_chain, asset). We can do a clean up in the end based on what ended up being used or not. I really can not recall the exact reason why we included channel_id here and not prewitnessed_deposit_id.

There is no problem adding `source_chain` and `asset`, but the reason I asked about `channel_id` specifically is that we don't currently store that information in the boost pool. Let's think whether it is actually required (or might be required in the future), and if so, think of the best way to provide it.

msgmaxim:
Updates:

- Included insta tests (I will now close https://github.com/chainflip-io/chainflip-backend/pull/4792). You can now see the [snap file](https://github.com/chainflip-io/chainflip-backend/blob/716a4c9e0bcfb85ea3f3bfac6d8304a87457eed7/state-chain/custom-rpc/src/snapshots/custom_rpc__test__boost_details_serialization.snap) to check how the resonse is formatted.
- Asset is now an optional parameter; if not specificed, pools for all assets will be returned. For this I have added chain/asset into every boost details entry.
- All maps are now lists with the "key" as a extra field (to match the schema in notion)
- All amounts are now encoded as hex (updated the boost_depth rpc too)
- It seemed strange to see chain/asset nested inside another "asset" field, so I instead flattened it for both boost rpcs (see the snap file); happy to revert if the nesting was somehow desirable.

@GabrielBuragev @kylezs

state-chain/runtime/src/runtime_apis.rs
GabrielBuragev: > > Exposing only the ids means that we have to look elsewhere to get the details related to that deposit.
> 
> But the amounts are already exposed this very response under `pending_boosts`. I don't think we should return the amounts twice in the same response (similar to how you suggested that total_available_amount wasn't necessary).

You are right. After seeing the snap i have a better understanding of your comments and the current schema looks good!

GabrielBuragev:
> Makes sense if it will be used by clients' browsers. 

To give you a bit more context on how the LP app was structured. We ended up doing some of the RPC calls directly on the client side (LP app) back in the days so we could ship it out faster and in time (without having to setup a whole backend for it). Having that in mind, i thought it might be useful to have that filtering baked in so we could utilize it there without any extra effort. But when i think about it deeper now, we anyways want to start moving most of the stuff on the LP app to its own backend, so this might be a good time to start that. We will still be able to utilize the cache and filter by account ids there. 

We can move forward without that.

GabrielBuragev (approved):

PR #4781, opened 2024-04-19
Arbitrum gas limit multiplier

albert-llimos:
## Checklist

## Summary

Gas required for transactions in Arbitrum changes over time depending on several factors, mainly Ethereum calldata price. It's not that Arbitrum gas price for a transaction changes over time, which it also does, but the amount of gas needed per transaction changes over time.

Therefore, I am implementing a `gas_limit_multiplier` that is used to track the gas limit variability. For now this is only needed for ingress/egress fee calculations.

For context, Arbitrum have NodeInterface call that returns parameters related to fees. That includes, among others, a gas estimation that I'm using to calculate the multiplier, and the current l2 base fee. Since we are anyway not doing anything fancy with `fee_history` we can instead replace it for the value returned by the NodeInterface. I have checked that it behaves correctly.


dandanlen (commented):
LGTM - will wait for @kylezs and/or @AlastairHolmes  to give the all-clear on the witnessing side.

engine/src/evm/rpc/node_interface.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::str::FromStr;
+
+use anyhow::{Ok, Result};
+
+use super::{EvmRpcClient, EvmRpcSigningClient};
+
+abigen!(NodeInterface, "$CF_ARB_CONTRACT_ABI_ROOT/INodeInterface.json");
+
+// This is a kind of precompile on Arbitrum (although not deployed on chain)
+const NODE_INTERFACE_ADDRESS: &str = "0x00000000000000000000000000000000000000C8";
```
kylezs, line 12: What do you mean? Is there some kind of reference we can provide a link to here, or somewhere we can check this?

albert-llimos: https://docs.arbitrum.io/build-decentralized-apps/nodeinterface/reference
It's an implementation at a node level as if it were a contract deployed at that address but it's not really deployed. I'll improve that comment with the link.

kylezs (approved):

PR #4784, opened 2024-04-19
feat: LP Asset rebalancing (PRO-1259)

Janislav:
# Pull Request

Closes: PRO-1259

## Checklist

## Summary

This PR adds an extrinsic that allows LPs to move assets between accounts without the need to withdrawal first. The existing extrinsic and LP-API endpoints doesn't change, which should make it none-breaking for the LPs. The changes in the LP-pallet could affect the engine, though, which make it breaking. I think we have to check this.

#### Non-Breaking changes


Janislav:
For now, there is only a unit-test. If we are happy with the PR, I would also add a bouncer test.

dandanlen (commented):
Looking good - will take a closer look on Monday. Only comment so far is that I think we should come up with a better name than `move_asset`. Will think about it....

dandanlen:
I took another look, changed the naming to 'transfer' instead of 'move'.

As discussed, I think a bouncer test would be useful for regression prevention.

dandanlen:
Apart from that this is ready to go. Nice work.

api/lib/src/lp.rs
```diff
@@ -256,6 +256,29 @@ pub trait LpApi: SignedExtrinsicApi + Sized + Send + Sync + 'static {
...
+		let (tx_hash, ..) = self
+			.submit_signed_extrinsic(RuntimeCall::from(pallet_cf_lp::Call::transfer_asset {
+				amount,
+				asset,
+				destination,
+			}))
+			.await
+			.until_in_block()
+			.await
+			.context("Registration for Liquidity Refund Address failed.")?;
```
dandanlen, line 277: Just noticed this:

```suggestion
			.context("Unable to transfer asset.")?;
```

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -347,6 +349,10 @@ impl<T: Config> Pallet<T> {
 		if amount > 0 {
 			match destination {
 				AccountOrAddress::Internal(destination_account) => {
+					ensure!(
+						account_id != destination_account,
+						Error::<T>::CannotTransferToOriginAccount
```
dandanlen, line 354: Good idea 👍

Janislav:
Docs are updated here: https://github.com/chainflip-io/chainflip-docs/pull/251

Janislav:
I think this should be ready now. I added some more things:

- Ensure that destination_account != source_account
- Bouncer test that is testing a balance transfer between LP accounts
- Updated the docs
- Changed amount parameter to hex string only

bouncer/shared/lp_api_test.ts
```diff
@@ -167,11 +175,17 @@ async function testTransferAsset() {
...
 
-  assert(newBalance > oldBalance, `Failed to observe balance increase after transfer`);
+  assert(
+    newBalanceDestination > oldBalanceDestination,
+    `Failed to observe balance increase after transfer for destination account!`,
+  );
+  assert(
+    newBalancesSource < oldBalanceSource,
+    `Failed to observe balance decrease after transfer for source account!`,
+  );
```
dandanlen, line 188: Should the balances be equal before and after?

Ie. can we improve this check even further and test for `oldBalanceSource + oldBalanceDestination === newBalancesSource + newBalanceDestination` ?

Janislav:
I guess this is ready.

dandanlen (approved):

PR #4785, opened 2024-04-21
fix: dot metadata update

kylezs:
Fixes a bug introduced here: https://github.com/chainflip-io/chainflip-backend/pull/4388/files#diff-1015fc028d6f372919e563c645da89d8d3dd0fce1ec9ef4ef0400a5774b18541


j4m1ef0rd (approved):

engine/src/dot/http_rpc.rs
```diff
@@ -235,6 +248,8 @@ mod tests {
 				.await;
 
 		for block_hash in block_hash_of_runtime_updates {
+			println!("TRYING BLOCK: {:?}", block_hash);
```
dandanlen, line 251: Was this left here intentionally?

dandanlen (approved):

PR #4792, opened 2024-04-23
test: snapshot testing for new rpcs

msgmaxim:
# Pull Request

Closes: PRO-1336

## Checklist

## Summary

Added snapshot tests for the boost RPCs.

I also used `assert_json_snapshot`, which seems to work better than `assert_snapshot` with converting to JSON manually used for other RPCs (it also pretty-prints, which means on conflict you will see its precise location because serialization is no longer a single line).

Note that this is on top of the other open PR (if both are approved at the same time, it might be easier to merge this one first).

### DIFF
--- a/state-chain/custom-rpc/src/snapshots/custom_rpc__test__boost_depth_serialization.snap
+++ b/state-chain/custom-rpc/src/snapshots/custom_rpc__test__boost_depth_serialization.snap
@@ -1,0 +1,22 @@
+---
+source: state-chain/custom-rpc/src/lib.rs
+expression: val
+---
+[
+  {
+    "asset": {
+      "chain": "Ethereum",
+      "asset": "FLIP"
+    },
+    "tier": 10,
+    "available_amount": 1000000000000000000000000000000000000
+  },
+  {
+    "asset": {
+      "chain": "Ethereum",
+      "asset": "FLIP"
+    },
+    "tier": 30,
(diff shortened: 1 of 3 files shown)


state-chain/custom-rpc/src/snapshots/custom_rpc__test__boost_depth_serialization.snap
```diff
@@ -0,0 +1,22 @@
...
+expression: val
+---
+[
+  {
+    "asset": {
+      "chain": "Ethereum",
+      "asset": "FLIP"
+    },
+    "tier": 10,
+    "available_amount": 1000000000000000000000000000000000000
```
dandanlen, line 12: Should we not serialize amounts using hex? ( @GabrielBuragev ?)

msgmaxim: Yeah, was thinking about that too as I was writing this test. Don't think JS will be able to read this number accurately. I thought if the PRs are merged quickly, I'd make the change in a separate PR (since the PR for the other RPC method is already merged anyway), but since there are some other comments to address anyway, I will update it to use the hex amounts too.

msgmaxim:
Closing this as now I've added these tests directly in https://github.com/chainflip-io/chainflip-backend/pull/4780

PR #4800, opened 2024-04-25
Test/bouncer broadcast barrier test

albert-llimos:
Add test for https://github.com/chainflip-io/chainflip-backend/pull/4794 .

Even though it's time sensitive, it run well (hit the bug) both locally and on CI consistently.

It was failing when adding the test and before having those fixes:
https://github.com/chainflip-io/chainflip-backend/actions/runs/8831672390/job/24247514001

Passing when merging the fixes from that PR:
https://github.com/chainflip-io/chainflip-backend/actions/runs/8831971085/job/24248558879

After testing it I put it in the nightly run since it's quite slow.


ramizhasan111:
LGTM. would let someone with more bouncer knowledge to have a look.

PR #4814, opened 2024-04-30
chore: clean up shared libs

kylezs:
# Pull Request

## Checklist

## Summary

- Removes the old shared libs
- change the `LC_ID_DYLIB` to not contain the system path of the build system (which was responsible for seeing my local system paths in people's build errors)


kylezs:
Will do in another PR

PR #4816, opened 2024-05-01
feat: Governance extrinsic to create boost pools

j4m1ef0rd:
# Pull Request

Closes: PRO-1351

## Checklist

## Summary

- Change the BoostPoolTier enum into a `u16`.
	- Boost events now all use a `u16` for the pool bps instead of a `FiveBps` string.
- New extrinsic that takes a list of boot pools and creates them all. The list can contain any assets of the chain at any bps (except 0).
	- Emmits a boost pool created event for each.
- Removed the creation of the boost pools from the pallet migration.
- New boost pool setup command for the bouncer. This is called in the "setup_for_test.sh" script. I creates the 5,10 and 30 bps boost pools for all assets on all chains.
- Added a test and benchmark for the new extrinsic.

## Note
Now that we can control the release of boost features on networks by simply not creating the pools, we no longer need to set the ingress egress pallet safe mode to red (see `ArbitrumIntegration`). Since @msgmaxim is making changes to the ingress egress safe mode in his PR (to also make it into 1.4), I will leave out the change here for now unless his PR will not make it in to 1.4.


state-chain/pallets/cf-ingress-egress/src/migrations/add_boost_pools.rs
```diff
@@ -1,153 +0,0 @@
...
-		old::DepositChannelDetails<T, I>,
-		OptionQuery,
-	>;
-}
-
-pub struct Migration<T: Config<I>, I: 'static>(PhantomData<(T, I)>);
-
-impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
-	fn on_runtime_upgrade() -> Weight {
-		// Add (default) boost status to existing deposit channels:
```
msgmaxim, line 39: We still want this, right? We no longer need to create pools, but we still need to add new fields to all existing `DepositChannelDetails` records.

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
```diff
@@ -942,13 +936,70 @@ fn stop_boosting_is_disabled_by_safe_mode() {
...
+			}),
+			RuntimeEvent::IngressEgress(Event::BoostPoolCreated {
+				boost_pool: BoostPoolId { asset: eth::Asset::Eth, tier: TIER_10_BPS },
+			}),
+			RuntimeEvent::IngressEgress(Event::BoostPoolCreated {
+				boost_pool: BoostPoolId { asset: eth::Asset::Flip, tier: TIER_5_BPS },
+			})
+		);
+
+		// Should not be able to create the same pool again
```
msgmaxim, line 987: Let's also check that we didn't accidentally removed the existing pool (e.g. that we didn't set the pool to None in `try_mutate_exists`).

state-chain/pallets/cf-ingress-egress/src/benchmarking.rs
```diff
@@ -462,6 +478,23 @@ mod benchmarks {
 		assert_eq!(PrewitnessedDeposits::<T, I>::iter().count(), 0);
 	}
 
+	#[benchmark]
+	fn create_boost_pools(n: Linear<1, 255>) {
```
msgmaxim, line 482: If we know that the relationship is trivially linear, i.e. `total_weight = weight * count`, we can just benchmark creation of 1 pool and multiply by `process_deposits.len()` (i.e. how it is done in `process_deposit`). By using `Linear` as paramter here, you are asking substrate to run this a large number of times (which is a bit of a waste) to approximate the linear relationship that we already know.

bouncer/commands/setup_boost_pools.ts
```diff
@@ -0,0 +1,79 @@
...
+        );
+        const observeBoostPoolAlreadyExists = observeEvent(`governance:FailedExecution`, chainflip);
+
+        boostPoolEvents.push(
+          Promise.race([observeBoostPoolCreated, observeBoostPoolAlreadyExists]),
+        );
+      }
+    }
+
+    const ingressEgress = await chainIngressEgress(chain);
```
msgmaxim, line 57: How about renaming like this? (mostly about having "pallet" somewhere in the name)
```suggestion
    const pallet = await ingressEgressPalletForChain(chain);
```

bouncer/commands/setup_boost_pools.ts
```diff
@@ -0,0 +1,79 @@
...
+    await submitGovernanceExtrinsic(ingressEgress.createBoostPools(newPools));
```
msgmaxim, line 58: Does this need to be awaited? I think you can just remove the `await` (so they can be executed in parallel), and then rely on `await Promise.all(boostPoolEvents);` to await their completion.

bouncer/setup_for_test.sh
```diff
@@ -1,6 +1,7 @@
 set -e
 ./commands/setup_vaults.ts
 ./commands/setup_swaps.ts
+./commands/setup_boost_pools.ts
```
kylezs, line 4: this is going to increase setup time further - i don't think there's a reason for this, we should run this in parallel with the swaps setup, using Promise.all()

kylezs (commented):
Looks good, but yeah we definitely don't want to increase bouncer setup time if we can avoid it

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1335,8 +1346,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		let mut used_pools = BTreeMap::new();
 
-		for boost_tier in SORTED_BOOST_TIERS {
-			if boost_tier as u16 > max_boost_fee_bps {
+		let sorted_boost_tiers = BoostPools::<T, I>::iter_prefix(asset)
+			.map(|(tier, _)| tier)
+			.collect::<BTreeSet<_>>();
```
kylezs, line 1351: Adding a debug_assert! here that they are sorted in ascending order would be helpful

msgmaxim (approved):

PR #4818, opened 2024-05-01
feat: store boost fees explicitly + rpc

msgmaxim:
# Pull Request

Closes: PRO-1356

## Checklist

## Summary

I went ahead and implemented this based on the discussion on discord without waiting for PRO-1355 considering we want to merge this for 1.4. Hopefully this is close to what we want.

- Extended boost pool storage to additionally store fees earned by every booster for each pending boost.
- Added `cf_boost_pool_pending_fees` rpc to expose those fees (internally, we still use the same runtime api call, cf_boost_pool_details, which now also returns the fee).
- The rpc works similarly to `cf_boost_pool_details`, in that it takes an optional asset, and returns data for all fee tiers for the asset; when the asset is not specified, data for all assets is returned.
- Added insta test for the new RPC's response.

Example:
```
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "cf_boost_pool_pending_fees", "params": []}' http://localhost:9944
```
(see the insta file for how the response is formatted)

I also realised (from the discord discussion) that we'd need to query historical data by specifying a block hash. I added an optional block hash parameter to the new RPC for this. Note however, that the existing boost RPC don't let you specify the hash, since we never talked about this. Let me know if we want this parameter in the remaining boost RPCs too @GabrielBuragev. (Update: already did this.)

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -1513,7 +1513,7 @@ where
 	}
 
 	fn cf_boost_pool_details(&self, asset: Option<Asset>) -> RpcResult<BoostPoolDetailsResponse> {
-		let get_boost_details_for_asset = |asset| {
+		execute_for_all_or_one_asset(asset, |asset| {
 			self.client
 				.runtime_api()
 				.cf_boost_pool_details(self.client.info().best_hash, asset)
@@ -1524,16 +1524,7 @@ where
 						.collect()
 				})
 				.map_err(to_rpc_error)
-		};
-
-		if let Some(asset) = asset {
-			get_boost_details_for_asset(asset)
-		} else {
-			let results_for_each_asset: RpcResult<Vec<_>> =
-				Asset::all().map(get_boost_details_for_asset).collect();
-
-			results_for_each_asset.map(|inner| inner.into_iter().flatten().collect())
-		}
+		})
 	}
 
 	fn cf_boost_pool_pending_fees(
@@ -1541,7 +1532,7 @@ where
 		asset: Option<Asset>,
 		at: Option<state_chain_runtime::Hash>,
 	) -> RpcResult<BoostPoolFeesResponse> {
-		let get_boost_fees_for_asset = |asset| {
+		execute_for_all_or_one_asset(asset, |asset| {
 			self.client
 				.runtime_api()
 				.cf_boost_pool_details(self.unwrap_or_best(at), asset)
@@ -1552,16 +1543,7 @@ where
 						.collect()
 				})
 				.map_err(to_rpc_error)
-		};
-
-		if let Some(asset) = asset {
-			get_boost_fees_for_asset(asset)
-		} else {
-			let results_for_each_asset: RpcResult<Vec<_>> =
-				Asset::all().map(get_boost_fees_for_asset).collect();
-
-			results_for_each_asset.map(|inner| inner.into_iter().flatten().collect())
-		}
+		})
 	}
 }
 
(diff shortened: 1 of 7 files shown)


state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
```diff
@@ -273,7 +292,7 @@ where
 			amount.saturating_accrue(excess_contributed);
 
 			if let Some(amount) = boosters_to_receive.get_mut(lucky_id) {
-				amount.saturating_accrue(remaining_to_receive)
+				amount.total.saturating_accrue(remaining_to_receive)
```
j4m1ef0rd, line 295: Will this eventually create a discrepancy between the total amount and the fees collected? maybe we should just add the `remaining_to_receive` to the fee here as well?

j4m1ef0rd (approved):

state-chain/custom-rpc/src/lib.rs
```diff
@@ -357,7 +359,13 @@ mod boost_pool_details {
 					.into_iter()
 					.map(|(deposit_id, owed_amounts)| PendingBoost {
 						deposit_id,
-						owed_amounts: map_to_vec(owed_amounts),
+						owed_amounts: owed_amounts
```
kylezs, line 362: can inline the other map_to_vec usage now

state-chain/custom-rpc/src/lib.rs
```diff
@@ -371,10 +379,48 @@ mod boost_pool_details {
...
+	#[derive(Serialize, Deserialize)]
+	pub struct BoostPoolFeesRpc {
+		fee_tier: u16,
+		#[serde(flatten)]
+		asset: Asset,
+		pending_fees: Vec<PendingFees>,
+	}
+
+	impl BoostPoolFeesRpc {
+		pub fn new(asset: Asset, fee_tier: u16, details: BoostPoolDetails) -> Self {
```
kylezs, line 398: Jamie has a alias in his PR for this BoostPoolTier which is u16 @j4m1ef0rd .  Whichever PR gets in 2nd should update this.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1484,6 +1537,34 @@ where
 			results_for_each_asset.map(|inner| inner.into_iter().flatten().collect())
 		}
 	}
+
+	fn cf_boost_pool_pending_fees(
+		&self,
+		asset: Option<Asset>,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<BoostPoolFeesResponse> {
+		let get_boost_fees_for_asset = |asset| {
```
kylezs, line 1546: we use an identical pattern for the boost pool details, we could factor this out

state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
kylezs: Yeah I think it makes sense to add to the fee here too. Any extra ought to be considered as part of their earned fees.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -371,10 +379,48 @@ mod boost_pool_details {
...
+				pending_fees: details
+					.pending_boosts
+					.into_iter()
+					.map(|(deposit_id, owed_amounts)| PendingFees {
+						deposit_id,
+						fees: owed_amounts
+							.into_iter()
+							.map(|(account_id, amount)| AccountAndAmount {
+								account_id,
+								amount: NumberOrHex::from(amount.fee),
```
kylezs, line 411: In the LP API we're moving towards using U256 instead of NumberOrHex - we probably want to be consistent here too. I think there was another place in another boost PR we used NumberOrHex too, so would check all of them.

state-chain/custom-rpc/src/lib.rs
msgmaxim: FWIW I did consider inlining it, but decided to keep because `map_to_vec` has a very limited scope and it seemed to have improved readability by de-cluttering the caller IMO. It is subjective, but I will inline so we can make progress.

state-chain/custom-rpc/src/lib.rs
msgmaxim: Are you talking about the mechanism that executes a lambda for all assets if asset is `None`? I factored out `execute_for_all_or_one_asset` now.

If you are talking about the fact that both of them call `cf_boost_pool_details`, then I don't think it is as important. Note that the lambda passed to `map` is different, and whatever code is repeated isn't worth factoring out further IMO.

msgmaxim:
Addressed all comments now @kylezs. I also added an optional block hash parameter to other boost RPCs (to match the one added here).

state-chain/custom-rpc/src/lib.rs
kylezs: Yeah the `execute_for_all_or_one_asset` 👍

kylezs (approved):

kylezs:
I had Andrew look at the response shape yesterday and he said all looked good, so merging this. Nice one 👍

PR #4819, opened 2024-05-02
Feat: per chain safe mode to restrict deposits

msgmaxim:
# Pull Request

Closes: PRO-1218

## Checklist

## Summary

Note that this is per chain rather than per asset (but per chain is easier to implement and after chatting with @kylezs it seems that per chain is what we actually want here).

- Added `deposits_enabled` to ingress-egress pallet's safe mode, which can be toggled to disable/enable openning of deposit channels for the target chain.
- Added a unit test to show that disabling `deposits_enabled` results in Errors on trying to open a channel.
- Anabled CODE_GREEN on ingress-egress pallets since we no longer want to use safe mode to prevent premature interactions with boost pools (see https://github.com/chainflip-io/chainflip-backend/pull/4816), and CODE_RED would now disable all deposits, which we don't want.


state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
```diff
@@ -795,31 +795,20 @@ fn test_add_boost_funds() {
...
-	);
-	<MockRuntimeSafeMode as SetSafeMode<MockRuntimeSafeMode>>::set_code_green();
-	assert!(
-		<MockRuntimeSafeMode as sp_core::Get<PalletSafeMode<()>>>::get() ==
-			PalletSafeMode::CODE_GREEN
-	);
+	let boost_mode = if enable { PalletSafeMode::CODE_GREEN } else { PalletSafeMode::CODE_RED };
+
+	let new_mode =
+		PalletSafeMode { deposits_enabled: get_safe_mode().deposits_enabled, ..boost_mode };
```
j4m1ef0rd, line 806: This is ok, but i think it would be better to just set the individual bool's in the test, like what you did in the `safe_mode_prevents_deposit`. That way we can make sure we are using the correct flag as well. To keep it clean we can still use the old `turn_safe_mode_off` function.

j4m1ef0rd (approved):

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
msgmaxim: I think right now it is not clear which approach is better. When we have more bools in `PalletSafeMode`, we will have a better idea, and maybe we could revisit then.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -718,6 +721,8 @@ pub mod pallet {
 		AddBoostFundsDisabled,
 		/// Retrieving boost funds disabled due to safe mode.
 		StopBoostingDisabled,
+		/// Disabled due to safe mode for the chain
+		DepositsDisabledForChain,
```
kylezs, line 725: DepositChannelCreationDisabled - technically deposits are still possible via already open channels or via contract initiated swaps

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1794,3 +1794,34 @@ fn ingress_fee_is_withheld_or_scheduled_for_swap() {
 
 	test_ingress_or_egress_fee_is_withheld_or_scheduled_for_swap(ingress_function)
 }
+
+#[test]
+fn safe_mode_prevents_deposit() {
```
kylezs, line 1799:
```suggestion
fn safe_mode_prevents_deposit_channel_creation() {
```

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
```diff
@@ -836,7 +825,7 @@ fn boosting_deposits_is_disabled_by_safe_mode() {
 			BoostPoolTier::FiveBps
 		));
 
-		turn_safe_mode_on();
+		toggle_boosting_with_safe_mode(false);
```
kylezs, line 828: nit, toggle suggests if it's false, then make it true and vice vesrsa, rather than providing a value to set it
```suggestion
		boosting_with_safe_mode(false);
```

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1794,3 +1794,34 @@ fn ingress_fee_is_withheld_or_scheduled_for_swap() {
...
+	new_test_ext().execute_with(|| {
+		assert_ok!(IngressEgress::open_channel(
+			&ALICE,
+			eth::Asset::Eth,
+			ChannelAction::LiquidityProvision { lp_account: 0 },
```
kylezs, line 1804: Would be another PR - but perhaps it makes sense to block deposit channels for swaps, but not LPs e.g. in a liquidity shortage or something we might want to do that? Something we can think about

kylezs (approved):

dandanlen (changes requested):
This new safe mode now shadows the `deposits_enabled` safe mode in the swapping pallet. 

I think we should remove the one in the swapping pallet.

msgmaxim:
> This new safe mode now shadows the `deposits_enabled` safe mode in the swapping pallet.
> 
> I think we should remove the one in the swapping pallet.

Removed the deposits_enabled field from the swapping pallet now (and the corresponding Error). Also updated the migration code to account for the deleted field. @dandanlen

dandanlen (approved):

PR #4822, opened 2024-05-03
Realistic BTC egress fees

martin-chainflip:
# Pull Request

Closes: PRO-1360

## Checklist

## Summary

Egress fees only included the base transaction and a single output, but a typical egress also needs to spend a vault utxo and send some change back to the vault. This PR includes these missing aspects into the egress fee

### DIFF
--- a/state-chain/chains/src/btc.rs
+++ b/state-chain/chains/src/btc.rs
@@ -119,7 +119,10 @@ impl Default for BitcoinTrackedData {
 	}
 }
 
+// A Bitcoin transaction consists of some base data, a list of input UTXOs and a list of output UTXOs
 impl FeeEstimationApi<Bitcoin> for BitcoinTrackedData {
+	// When a user deposits some BTC, they create an input UTXO that we need to spend as part of a transaction
+	// some time in the future, so we charge them the costs of spending such an input UTXO
 	fn estimate_ingress_fee(
 		&self,
 		_asset: <Bitcoin as Chain>::ChainAsset,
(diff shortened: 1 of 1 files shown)


kylezs (approved):

state-chain/chains/src/btc.rs
```diff
@@ -119,20 +119,30 @@ impl Default for BitcoinTrackedData {
...
 	fn estimate_ingress_fee(
 		&self,
 		_asset: <Bitcoin as Chain>::ChainAsset,
 	) -> <Bitcoin as Chain>::ChainAmount {
 		self.btc_fee_info.fee_per_input_utxo()
 	}
 
+	// When a user wants to receive some BTC, we need to create a transaction which typically spends
+	// a UTXO from the vault and creates two output UTXOs: One going to the user and one sending the
+	// remaining BTC back into the vault.
```
dandanlen, line 137: Does similar reasoning not apply to the ingress fee? Ie. the ingress needs to be wrapped in a tx, therefore the ingress fee could also justifiably include the base tx fee and the output to the vault.

I feel like we should at least add a note about whether imposing these costs on the egress fee is a conscious decision or not, and if so, why (for example, are we consciously encouraging btc inputs? can/do we intend to improve this further in future? ...).

dandanlen (approved):
LGTM, with a single comment.

state-chain/chains/src/btc.rs
martin-chainflip: Deposited BTC will remain in the deposit address until they get included into a transaction that would happen 
anyways, so we only charge the cost of that. That *does* make deposits much cheaper than egress, but this is because of how BTC transactions work, not so much a conscious choice from us. We could try to share the costs more equally between deposits and egress, but that would mean making assumptions about the ratio of deposits vs egresses on BTC.

PR #4823, opened 2024-05-03
chore: better dylib path finding

kylezs:
I want to test the apt and docker packaging is correct on a linux machine before merging this.

# Pull Request

To test it works:
1. `unset DYLD_LIBRARY_PATH`
2. `cargo clean`
3. `cargo cf-test`

Tests should pass.

@martin-chainflip has tested this and it works.

I've tested the linux and docker builds on a linux machine to ensure the builds and path finding works correctly.

## Checklist

## Summary

- Use `@rpath` instead of `DYLD_LIBRARY_PATH` to find the dylibs for mac - so `DYLD_LIBRARY_PATH` doesn't need to be set.
- For linux and docker we supply the correct paths `/usr/lib/chainflip-engine` or `/usr/local/lib` respectively. 
- Used install_name_tool to set the path for the dylib
- $ORIGIN doesn't work the way `@executable_path` does for mac, so leaving that to another PR, as this is already an improvement
- Add some extra log and debug lines

### DIFF
--- a/engine-runner-bin/Cargo.toml
+++ b/engine-runner-bin/Cargo.toml
@@ -25,6 +25,8 @@ assets = [
     # manually.
     [
         "target/release/libchainflip_engine_v1_4_1.so",
+        # This is the path where the engine dylib is searched for on linux.
+        # As set in the build.rs file.
         "usr/lib/chainflip-engine/libchainflip_engine_v1_4_1.so",
         "755",
     ],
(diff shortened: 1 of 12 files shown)


engine-runner-bin/Cargo.toml
```diff
@@ -25,13 +25,17 @@ assets = [
     # manually.
     [
         "target/release/libchainflip_engine_v1_4_1.so",
+        # This is the path where the engine dylib is searched for on linux.
+        # As set in the build.rs file.
         "usr/lib/chainflip-engine/libchainflip_engine_v1_4_1.so",
```
martin-chainflip, line 30: is it correct that this is a relative path and doesn't start with a slash?

kylezs: this is just how the cargo-deb library does it, agree it's a little strange

martin-chainflip (approved):

PR #4833, opened 2024-05-06
refactor: remove storage for prewitnessed deposits

msgmaxim:
# Pull Request

Closes: PRO-1291

## Checklist

## Summary

Since we removed active boosting, the only purpose of storing PrewitnessedDeposits was to store the boosted amount, so that when we see a finalised deposit we can correctly respond depending on the amount. However, we already have BoostStatus recorded on the channel, and having boost info split this way resulted in unnecessary complexity, making some dubious states possible (when they aren't consistent).

I moved the boosted amount inside BoostStatus, which means I could remove `PrewitnessedDeposits` altogether, as well as some related code (e.g. cleanup). This also led to a considerable simplification of `process_single_deposit`.


state-chain/pallets/cf-ingress-egress/src/migrations/add_boost_pools.rs
```diff
@@ -58,6 +79,13 @@ impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
 		let number_of_channels_in_lookup =
 			old::DepositChannelLookup::<T, I>::iter_keys().count() as u32;
 
+		// Ensure that test is set up correctly:
+		#[cfg(test)]
```
j4m1ef0rd, line 83: Strange to do this conditionally in the pre and post upgrade functions. Why not just do both checks in the test itself?

j4m1ef0rd (approved):

state-chain/pallets/cf-ingress-egress/src/migrations/add_boost_pools.rs
msgmaxim: updated

state-chain/pallets/cf-ingress-egress/src/migrations/add_boost_pools.rs
```diff
@@ -50,6 +69,8 @@ impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
 			},
 		);
 
+		let _ = old::PrewitnessedDeposits::<T, I>::clear(u32::MAX, None);
```
kylezs, line 72: this will have to be in another migration, since this migration will be applied in 1.4 - unless we cherry pick into 1.4. But think I'd rather not, and we try to focus on stabilising the release now with only essential commits. So yeah, will need a separate migration, and pallet version bump etc.

msgmaxim: Done. I also bumped the spec version as discussed.

kylezs (approved):

PR #4834, opened 2024-05-06
feat/pro-1066/fee-calc-wrong-results

Janislav:
# Pull Request

Closes: PRO-1066

## Checklist

## Summary

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -19,7 +19,6 @@ mod boost_pool;
 
 use boost_pool::BoostPool;
 pub use boost_pool::OwedAmount;
-use cf_primitives::chains::assets::get_simulation_amount;
 
 use frame_support::{pallet_prelude::OptionQuery, transactional};
 
@@ -2005,5 +2004,16 @@ impl<T: Config<I>, I: 'static> IngressEgressFeeApi<T::TargetChain> for Pallet<T,
 				tracker.register_transfer(fee);
 			});
 		}
+	}
+}
+
+/// Returns an amount for an none-gas asset we use for transaction fee estimation.
+pub fn get_simulation_amount(asset: Asset) -> Option<u128> {
+	match asset {
+		Asset::Flip => Some(5_000_000_000_000_000_000),
+		Asset::Usdc => Some(10_000_000),
+		Asset::Usdt => Some(10_000_000),
+		Asset::ArbUsdc => Some(10_000_000),
+		_ => None,
 	}
 }
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -19,6 +19,7 @@ mod boost_pool;
 
 use boost_pool::BoostPool;
 pub use boost_pool::OwedAmount;
+use cf_primitives::chains::assets::get_simulation_amount;
 
 use frame_support::{pallet_prelude::OptionQuery, transactional};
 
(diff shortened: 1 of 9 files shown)


Janislav:
There might be more testing needed. Current test suite is passing.

state-chain/primitives/src/chains/assets.rs
```diff
@@ -531,6 +533,17 @@ macro_rules! assets {
 	}
 }
 
+/// Returns an amount for an none-gas asset we use for transaction fee estimation.
+pub fn get_simulation_amount(asset: Asset) -> Option<u128> {
```
dandanlen, line 537: I don't think this needs to be in the primitives crate. It's not a primitive and it's not used anywhere except the ingress/egress pallet. (in fact not sure it even needs to be a function).

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1858,26 +1859,37 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 		} else {
+			// In case of an none-gas asset, we need to use the simulation amount. If the simulation
+			// amount is not available, we use the available amount.
+			let simulation_amount = if let Some(simulation_amount) =
+				get_simulation_amount(asset.into())
+			{
+				simulation_amount
+			} else {
+				log::warn!("No simulation amount available for {asset:?}. Using available amount.");
+				available_amount.into()
```
dandanlen, line 1870: We could avoid this branch and potential error case if we update the asset macro, adding a `NonNativeAsset` type for example (not sure how simple this would be). 

I think it's worth trying. Leaving it as it is, it's easy to forget to add a variant to `get_simulation_amount` if we add a new asset. And then we need a runtime upgrade to fix it. 

I would suggest we either add the new type as suggested above (-> compile time guarantees) OR if that proves impractical, add the simulation amounts as storage items with a governance update extrinsics (-> flexibility).

dandanlen:
I'm going to make some changes - I noticed that there's an edge case when the swap amount is less than the simulation amount (it would be possible to swap more in fees than is available from the swap). 

Am going to try a couple of alternatives, will push some changes later this morning.

Janislav:
Yeah, good point. Maybe have to just do `min(simulation_amount, actual_amount)` in this case?

dandanlen:
> Yeah, good point. Maybe we have to just do `min(simulation_amount, actual_amount)` in this case?

I moved the check into the estimation logic - makes more sense IMO, and allows it to be tested.

Can you please review my changes?

Janislav (commented):
Looks good from my point of view. Unfortunately, I can not approve because it's my PR, I think :D.

dandanlen (approved):

dandanlen:
@AlastairHolmes since this touches the assets/pools pallet, would be good if you can take a quick look too please.

AlastairHolmes (commented):
This function kind of doesn't make any sense (Not because of this change, it previously didn't either). But with this change it makes less sense because now it is specific to fees (And the naming and description don't suggest that).

Instead of just changing the name and description I think we should put this change outside of the function, instead of inside (i.e. just pass the estimation_amount in as the `available_input_amount`). And importantly also move the checks that the returned amount is less than the available_input_amount (`.min(vailable_input_amount.into())` and `if available_input_amount.is_zero()`) outside the function. And change the parameter name to `input_amount`. Because then it is just an estimated_asset_conversion, and doesn't need to "know" anything about fees etc.

Should be a small change.

AlastairHolmes:
I see @dandanlen's previous comment, either go the full way, and update the function name, description, and parameters (to reflect that the function is specifically intend to "calculate_gas_fee") or put the checks outside (Probably the better way) or move them out.

AlastairHolmes:
Sorry I realize I wrote that pretty badly. Basically, I think the intent of the function isn't very clear, and so I think just a renaming, and updating the description will do the trick.

Also we could remove `desired_output_amount` and `output_asset` by adding a generic to the function for `Chain`, and then move the `fee_estimation_cap` into `calculate_asset_conversion`.

I think that is best solution.

AlastairHolmes (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1796,21 +1796,21 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			IngressOrEgress::Egress => T::ChainTracking::estimate_egress_fee(asset),
 		};
 
-		let transaction_fee = if asset == <T::TargetChain as Chain>::GAS_ASSET {
+		let fees_withheld = if asset == <T::TargetChain as Chain>::GAS_ASSET {
 			// No need to schedule a swap for gas, it's already in the gas asset.
 			Self::accrue_withheld_fee(asset, sp_std::cmp::min(fee_estimate, available_amount));
 			fee_estimate
```
AlastairHolmes, line 1802: you need to do min here as well, right? @dan

dandanlen: I added the min in the transaction fee in the `else` branch. I figured even if we swallow the deposit, we still want to convert the input to our gas asset (because we might need to pay to fetch it).

dandanlen: (hope that makes sense...)

PR #4835, opened 2024-05-06
Update Arbitrum migrations and chainspec

albert-llimos:
# Pull Request

Closes: PRO-1224

## Checklist

## Summary

Update chainspecs and migrations with the correct addresses.

The `eth_address_checker_address` and `arb_vault_address` have the same address, that is not a mistake. It's just that we use the same deployer wallet.

### DIFF
--- a/state-chain/runtime/src/migrations/arbitrum_integration.rs
+++ b/state-chain/runtime/src/migrations/arbitrum_integration.rs
@@ -102,12 +102,12 @@ impl OnRuntimeUpgrade for ArbitrumIntegration {
 				cf_runtime_upgrade_utilities::genesis_hashes::BERGHAIN => {
 					log::warn!("Need to set up arbitrum integration for Berghain");
 					(
-						[0u8; 20].into(),
-						[0u8; 20].into(),
-						[0u8; 20].into(),
+						hex_literal::hex!("BFe612c77C2807Ac5a6A41F84436287578000275").into(),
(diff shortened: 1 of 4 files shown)


state-chain/runtime/src/migrations/arbitrum_integration.rs
```diff
@@ -102,38 +102,38 @@ impl OnRuntimeUpgrade for ArbitrumIntegration {
...
-						[0u8; 20].into(),
-						[0u8; 20].into(),
+						hex_literal::hex!("BFe612c77C2807Ac5a6A41F84436287578000275").into(),
+						hex_literal::hex!("79001a5e762f3bEFC8e5871b42F6734e00498920").into(),
+						hex_literal::hex!("c1B12993f760B654897F0257573202fba13D5481").into(),
 						arb::CHAIN_ID_MAINNET,
-						[0u8; 20].into(),
-						0,
+						hex_literal::hex!("af88d065e77c8cC2239327C5EDb3A432268e5831").into(),
+						208460974,
```
dandanlen, line 110: Do we need to consider changing this to a more recent block before we push the upgrade, to prevent engines from scanning weeks' worth of block? Or will the witnessing only kick off once the new key (with the new epoch start) is generated? ( @kylezs ? )

kylezs: Only if we have a vault will this be used - when we execute `witness_initialize_arbitrum_vault` we would provide a more recent one, as that then sets it.

PR #4842, opened 2024-05-07
chore: rename boost methods

msgmaxim:
# Pull Request

Closes: PRO-1302

## Checklist

## Summary

Wouldn't normally open a PR this small, but wanted to close PRO-1302 (not sure why I didn't bundle this in some other PR).

Although the names are still quite generic, at least they no longer resemble "hooks", and something more specific could be misleading (as they do more than one thing).

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
@@ -306,7 +306,7 @@ where
 		Ok(())
 	}
 
-	pub(crate) fn on_finalised_deposit(
+	pub(crate) fn process_deposit_as_finalised(
 		&mut self,
 		prewitnessed_deposit_id: PrewitnessedDepositId,
 	) -> Vec<(AccountId, C::ChainAmount)> {
@@ -339,7 +339,10 @@ where
 	}
 
 	// Returns the number of boosters affected
-	pub fn on_lost_deposit(&mut self, prewitnessed_deposit_id: PrewitnessedDepositId) -> usize {
+	pub fn process_deposit_as_lost(
+		&mut self,
+		prewitnessed_deposit_id: PrewitnessedDepositId,
+	) -> usize {
 		let Some(booster_contributions) = self.pending_boosts.remove(&prewitnessed_deposit_id)
 		else {
 			log_or_panic!(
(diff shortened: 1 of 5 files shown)


kylezs:
> Wouldn't normally open a PR this small

Please keep opening PRs this small! Smaller is better (ofc, within reason)

msgmaxim:
> > Wouldn't normally open a PR this small
> 
> Please keep opening PRs this small! Smaller is better (ofc, within reason)... and it makes more sense to do larger PRs when you're not in the same timezone too ofc

Next time I might open a "fixed typo in a comment" PR ;)

state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
```diff
@@ -339,7 +339,10 @@ where
 	}
 
 	// Returns the number of boosters affected
-	pub fn on_lost_deposit(&mut self, prewitnessed_deposit_id: PrewitnessedDepositId) -> usize {
+	pub fn process_deposit_as_lost(
```
kylezs, line 342: why not something like `socialise_lost_deposit` ? this name isn't much better at describing what it actually does imo. Same for the finalised

msgmaxim: > this name isn't much better at describing what it actually does

I'm not sure that this should be the goal necessarily. From my perspective, we have a boost pool, and we can interact with it in 3 ways: (a) add/withdraw funds, (b) use funds for boosting, and (c) notifying the pool of the boost's outcome. The fact that the loses are socialised can be seen as an implementation detail. While you may argue that being more specific doesn't hurt, I would say "socialise lost deposit" may suggest that maybe there is another method that also needs to be called when a deposit is lost (there isn't, which a more generic name communicates better). When you see "handle_deposit_as_lost", you immediately know that everything to do with a lost deposit (at least from the pool's perspective) is done in this one method.

Happy to discuss more in person if you want.

kylezs: > you immediately know that everything to do with a lost deposit

I'm not sure I thought this tbh. I think it's in general more useful for it be clear when reading the calling code what is being done than saying under what conditions it's being called, which we already know since we're in the calling code. So if we wanted to go this way, then the rename doesn't achieve much imo, doesn't provide more information than `on_lost_deposit` naming for example.

Will wait for @dandanlen to take a look too

dandanlen: I've discussed this with Maxim previously and I don't have a very strong opinion either way here, I can see both sides of the argument. It's true the renaming doesn't seem to achieve much, at least it makes it clearer to the reader that it's not a callback/hook.

kylezs (approved):

PR #4849, opened 2024-05-08
feat: Solana chain integration

syan095:
Implements the transaction builder logic and chain APIs for Solana:
- Rotations
- Fetches
- Batch Fetches
- Transfers
- Ccms - the cf_parameter is now passed into the API call so it can be decoded by Solana CCMs (needed for CcmExtraAccounts)
- Usdc support 

Closes PRO-1328
Closes PRO-1361


kylezs:
Moving to draft since it's do not merge, just to be safe :)

state-chain/chains/src/sol/instruction_builder.rs
dandanlen, line 1: As a general comment on this - I feel like the builder pattern is not really the right choice for this. 

For example, the api allows this: 

```rust
InstructionBuilder::default().fetch_from(..).rotate_agg_key(..).finalize();
```

When in fact the above is (i guess?) invalid. We don't really need this flexibility. We only ever need one step.

Another thing is that I don't think this code should be concerned with the Environment trait. That trait is an implementation detail of how we get data from the chain. However at this level, we want to implement a 'pure' instruction builder that is independent of this - note how we have to implement Environment in order to test it. I think at this level we would be better off passing the arguments explicitly. 

I think a better api would be to define an enum that contains, for each instruction type, exactly the arguments that are required to build the tx:

```rust
enum SolanaTransaction {
	Transfer {
		to: SolanaAccount,
		agg_key: SoalanAggKey,
	},
	Fetch {
		from: Vec<DepositChannel<Solana>>,
		vault_program_account: SolanaAddress,
	},
	// ...
}
```

And use it in a simple function like this:

```rust
fn build_transaction(
	agg_key: SolananAggKey,
	nonce_account: SolananNonce,
	compute_unit_price: SolanaAmount,
	tx: SolanaTransaction::Transfer { .. }
) -> Result<Vec<SolanaInstruction>, TransactionBuilderError>;
```

Not even sure the error type will be needed since most (all?) of the possible errors were due to environment lookup issues. 

Then it's the calling code's responsibility to pull all the right values out of the environment / storage / whatever.

state-chain/chains/src/sol.rs
```diff
@@ -210,3 +196,11 @@ impl address::ToHumanreadableAddress for SolAddress {
 		self.to_string()
 	}
 }
+
+#[derive(Clone, Encode, Decode, TypeInfo, Eq, PartialEq, RuntimeDebug)]
+pub struct SolanaDepositChannelState {
+	pub seed: Vec<u8>,
```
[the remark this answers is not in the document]
dandanlen: (looks like it's just the channel_id - in that case can we not just use u64?).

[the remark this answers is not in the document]
albert-llimos: the seeds in Solana are of variable length up to 32 bytes. We could use some fixed length if we wanted as long as we make sure we encode it correctly in the api call.

state-chain/chains/src/sol/api.rs
```diff
@@ -74,44 +198,44 @@ impl<Env: 'static> ConsolidateCall<Solana> for SolanaApi<Env> {
...
-		fetch_params: vec::Vec<FetchAssetParams<Solana>>,
-		transfer_params: vec::Vec<TransferAssetParams<Solana>>,
+		_fetch_params: vec::Vec<FetchAssetParams<Solana>>,
+		_transfer_params: vec::Vec<TransferAssetParams<Solana>>,
 	) -> Result<Self, AllBatchError> {
-		Ok(Self::AllBatch { fetch_params, transfer_params })
+		// Create a BatchFetch for all deposit_channels
+		// Figure out how to derive deposit_channels from ChannelId (or pass DepositChannel in)
+		// Create a Transfer for each Transfer
+		todo!("PRO-1348 This should be implemented after allowing Multiple transactions to be returned by this trait.")
```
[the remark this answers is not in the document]
syan095: I'll change the the todo() to Dispatch errors instead.
Though to note that currently, we have todo!() littered all over the place for Solana (but I think they are unreachable, as a Solana API call needs to be crafted first)

state-chain/chains/src/sol.rs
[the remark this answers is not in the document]
syan095: I realized that since the seed is just channel ID, we can only use the "bump" u8 as the ChannelState. We can get the seed from the channel ID when we need to.

state-chain/chains/src/sol/instruction_builder.rs
syan095: Refactored. Environments are queried in api.tx, and instruction builder itself is now infallable.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -0,0 +1,313 @@
...
+	/// Create an instruction set to `transfer` from our Vault account to a target account.
+	pub fn transfer(
+		to: TransferAssetParams<Solana>,
+		agg_key: SolAddress,
+		nonce_account: SolAddress,
+		compute_price: SolAmount,
+	) -> Vec<SolInstruction> {
+		let instructions = match to.asset {
+			SolAsset::Sol =>
+				vec![SystemProgramInstruction::transfer(&agg_key.into(), &to.to.into(), to.amount)],
```
dandanlen, line 95: Is the aggkey account not touched by the transfer call? ( @albert-llimos ?)

I would have expected a `SolAccountMeta::new(agg_key.into(), true),` here, a bit like in the other calls.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -0,0 +1,313 @@
...
+				],
+			),
+			set_upgrade_authority(
+				upgrade_manager_program_data_account.into(),
+				&agg_key.into(),
+				Some(&new_agg_key.into()),
+			),
+		];
+		instructions.extend(all_nonce_accounts.into_iter().map(|nonce_account| {
+			SystemProgramInstruction::nonce_authorize(
```
dandanlen, line 129: We call nonce_authorize on all nonce accounts here, and then in `finalize`, the `nonce_authorize` instruction is added again. Is this correct? (I guess 'all_nonce_accounts' means 'all except the nonce that is used to sign'?)

dandanlen (commented):
I had another look through this - just a couple more comments. I'm happy for these to be addressed in this branch directly, but after that let's please keep this branch free of new commits, and only add to it via PRs.

state-chain/chains/src/sol/instruction_builder.rs
albert-llimos: If I recall properly the SystemProgramInstruction::transfer adds that.

state-chain/chains/src/sol/instruction_builder.rs
albert-llimos: Finalize calls `advance_nonce_account`, not `nonce_authorize`.

state-chain/chains/src/sol/instruction_builder.rs
dandanlen: Ok yeah it's a system call, not a contract call.

state-chain/chains/src/sol/sol_tx_core.rs
```diff
@@ -756,28 +801,94 @@ impl FromStr for Hash {
...
+	DepositChannel::<Solana> {
+		channel_id,
+		address: pda,
+		asset: crate::assets::sol::Asset::Sol,
+		state: bump,
+	}
+}
+
+/// Values used for testing purposes
+pub mod sol_test_values {
```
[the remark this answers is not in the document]
syan095: This is done in the Address derivation PR, where I explicitly made the distinction clearer between test values (only used in tests, can change) and consts (provided by Solana chain, will not change)

dandanlen:
LGTM - shall we merge this into the upstream Solana PR? Any preference @albert-llimos @ramizhasan111 @syan095

ramizhasan111:
> LGTM - shall we merge this into the upstream Solana PR? Any preference @albert-llimos @ramizhasan111 @syan095

Maybe its better if we merge the upstream PR first into main, and then merge this to main and so on. Would help especially because this PR uses some of the files from another solana PR that was merged to main and so once the upstream is merged to main, we can resolve the conflicts this has with main and view the net changes more cleanly before merging this to main.

state-chain/chains/src/sol/api.rs
```diff
@@ -1,70 +1,233 @@
...
+{
+	fn compute_price() -> Result<SolAmount, SolanaTransactionBuildingError> {
+		<Self as ChainEnvironment<ComputePrice, SolAmount>>::lookup(ComputePrice)
+			.ok_or(SolanaTransactionBuildingError::CannotLookupComputePrice)
+	}
+
+	fn durable_nonce() -> Result<SolHash, SolanaTransactionBuildingError> {
+		<Self as ChainEnvironment<DurableNonce, SolHash>>::lookup(DurableNonce)
+			.ok_or(SolanaTransactionBuildingError::CannotLookupDurableNonce)
+	}
```
ramizhasan111, line 44: it looks like we are not using this?

albert-llimos:
I'd also merge them one by one to make it cleaner, as Ramiz said.

state-chain/chains/src/sol/api.rs
```diff
@@ -1,70 +1,233 @@
...
+	pub fn batch_fetch(
+		deposit_channels: Vec<DepositChannel<Solana>>,
+	) -> Result<Self, SolanaTransactionBuildingError> {
+		// Lookup the current Aggkey
+		let agg_key = Environment::lookup_account(SolanaEnvAccountLookupKey::AggKey)?;
+		let vault_program_data_account =
+			Environment::lookup_account(SolanaEnvAccountLookupKey::VaultProgramDataAccount)?;
+		let system_program_id = SolAddress::from_str(SYSTEM_PROGRAM_ID)
+			.expect("Preset System Program ID account must be valid.");
+		let next_nonce =
```
ramizhasan111, line 127: same here

state-chain/chains/src/sol/api.rs
```diff
@@ -1,70 +1,233 @@
...
+			Environment::lookup_account(SolanaEnvAccountLookupKey::AvailableNonceAccount)?;
+		let compute_price = Environment::compute_price()?;
+
+		// Build the instruction_set
+		let instruction_set =
+			SolanaInstructionBuilder::transfer(transfer_param, agg_key, next_nonce, compute_price);
+
+		let transaction =
+			SolTransaction::new_unsigned(SolMessage::new(&instruction_set, Some(&agg_key.into())));
```
ramizhasan111, line 164: we need to add the durable nonce here while creating the ```SolMessage``` using the function ```new_with_blockhash```. Moreover, the nonce_account and the hash inside it should be coupled in the type and for each nonce account, we should also store its nonce and then use it here.

state-chain/chains/src/sol/api.rs
```diff
@@ -1,70 +1,233 @@
...
-		gas_budget: <Solana as Chain>::ChainAmount,
-		message: vec::Vec<u8>,
-	},
-	ApiCall {
-		threshold_signature: <SolanaCrypto as ChainCrypto>::ThresholdSignature,
-	},
+	) -> Result<Self, SolanaTransactionBuildingError> {
+		// Lookup the current Aggkey
+		let agg_key = Environment::lookup_account(SolanaEnvAccountLookupKey::AggKey)?;
+		let next_nonce =
```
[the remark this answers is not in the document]
syan095: Refactored the trait and the way the TX is built

state-chain/chains/src/sol/api.rs
syan095: Updated this. Also made the SolMessage::new() test only. Other functions in the Message struct had its visibility updated as well. This way test-only function won't accidentally make its way into production/non-test.

dandanlen:
Can we please avoid simply committing to this branch? If we are treating this as the 'base branch' for the Solana implementation, let's please stick to PR reviews for adding further updates? We have no idea if the latest commit conflicts with downstream changes in the other PRs. 

Otherwise I would suggest we progressively merge the dependent branches into this. I really think this would make it easier to keep track of all the changes.

syan095:
> The changes pushed by Roy were still part of the PR review, as we found out that some things were not fully correct during the review: [#4849 (comment)](https://github.com/chainflip-io/chainflip-backend/pull/4849#discussion_r1611666414) [#4849 (comment)](https://github.com/chainflip-io/chainflip-backend/pull/4849#discussion_r1611645449) But we should not push more things on this PR unless it's to address review comments.

Yeah I though the base branch would be the best place to fix this as it applies to all transactions we build. 
I've propagated this change to other changes downstream and resolved all conflicts.

dandanlen:
Yes I was referring to the final extra commit added yesterday - there are a lot of other PRs depending on this, and it's already been reviewed, so we should avoid direct updates.

albert-llimos:
Some notes passed from the PRs just merged:

1. The AssetWithDerivedAddress and therefore decomposed_fetch_params was added as a way pass in the ATA account if and only if the fetched asset is Usdc. This way the ATA derivation must be done in the API layer (as to doing it in the Instruction builder) and ensures the Instruction Builder is infallible.

2. The complicated logic in SolanaApi::transfer() is to ensure that we only fetch the extra environment variables (which will include storage lookups) when we have at least 1 Token transfer, and that we only do 1 set of lookup for any number of token transfers. This is optimized so that we don't do a bunch of storage lookups for each transfer

Comments that needs to be addressed:

- Currently when building the instructions it's just using the USDC token mint pubkey. For example, in the `transfer` api build it assumes that if the asset is not SOL then it's USDC. That should be generalizable if in the future we add another token. Imo if that is a significant refactor I'd just leave it like it is right now and we do that work later, as it's low priority and doesn't add any value until we don't have everything else working.

- This is not part of this PR itself but we will build the Solana environment lookup so we can get several of the values in one single lookup instead of getting them individually.

- Add documentation where needed for hardcoded values.

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
```diff
@@ -0,0 +1,134 @@
...
+use sol_prim::AccountBump;
+
+use crate::sol::{consts::*, AddressDerivationError, DerivedAddressBuilder, SolAddress};
+use core::str::FromStr;
+
+/// Derive address for a given channel ID
+pub fn derive_deposit_address(
+	channel_id: ChannelId,
+	vault_program: SolAddress,
+) -> Result<(SolAddress, AccountBump), AddressDerivationError> {
```
albert-llimos, line 14: Passing a comment from Dan:
"We could create a DerivedAddress struct instead of using a raw tuple."

syan095: THis is now addressed.

albert-llimos:
For clarity, this PR (API Calls) is considered completed and reviewed. This PR is now used as the base PR for Solana where we'll keep merging all the rest of Solana PRs until Solana is in a functional state and we have branched for 1.5 release.

ramizhasan111:
Now that the build issue is solved and the env PR merged, I guess its mostly feature complete? Can we get a list of remaining tasks, if any? @albert-llimos

albert-llimos:
> Now that the build issue is solved and the env PR merged, I guess its mostly feature complete? Can we get a list of remaining tasks, if any? @albert-llimos

Yes but there is still a bunch of small tickets. The remaining tasks are in linear.

ramizhasan111:
just pushed a fix of silly mistake I made in the limit fetches/transfers PR.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -36,7 +36,7 @@ pub mod weights;
 pub use weights::WeightInfo;
 pub mod migrations;
 
-pub const PALLET_VERSION: StorageVersion = StorageVersion::new(11);
+pub const PALLET_VERSION: StorageVersion = StorageVersion::new(12);
```
kylezs, line 39: where is the migration that moves this to v12? The placeholder migration is only at 11.

ramizhasan111: ok, so the storage version should be at 11. I traced the change and I remember that this was bumped in https://github.com/chainflip-io/chainflip-backend/pull/5023 which was incorrect since the storage version didnt need to be bumped.

kylezs: perfect thanks for checking and resolving 👌

albert-llimos (approved):

PR #4763, opened 2024-04-15
feat: instantiate solana

ramizhasan111:
# Pull Request

Closes: PRO-1329

## Checklist

Please conduct a thorough self-review before opening the PR.

- [x] I am confident that the code works.
- [x] I have updated documentation where appropriate.

## Summary

This PR includes the following tasks:

- define sol types and impl them
- instantiate pallets for sol using correct types
- add sol storage items in env pallet
- migrations
- address derivation for sol
- impl Chain and ChainCrypto traits for sol
- other sol related required chores

#### Non-Breaking changes

If this PR includes non-breaking changes, select the `non-breaking` label. On merge, CI will automatically cherry-pick the commit to a PR against the release branch.

PR #4851, opened 2024-05-08
feat: Solana address derivation with ATA logic

albert-llimos:
# Pull Request

Closes: PRO-1265

## Checklist

## Summary
Formally integrated Address derivation module into the Sol::sol_tx_core

Provide derivation for:
* derived account
* derived Associated Token Account (ATA)
* derived Deposit Channel, depending on the Asset type

Address derivation is used when crafting TXs such as Fetch and Transfer, as well as when creating deposit channels. 
ATA derivation is used with Token (which will be supported later in a separate PR).

Pre-emptively added partial token support for address derivation
Pre-emptively added BatchAll impl for Solana API

Clarified the structural distinction between Test and Production code.
Test values and code are contained with Test module with #[cfg(test)] to ensure a clear distinction between these and production code.


albert-llimos:
We won't merge this PR until the two PRs underneath are merged (#4763 & #4817)

state-chain/chains/src/sol/api.rs
```diff
@@ -222,14 +237,32 @@ impl<Env: 'static> ExecutexSwapAndCall<Solana> for SolanaApi<Env> {
...
 	) -> Result<Self, AllBatchError> {
-		// Create a BatchFetch for all deposit_channels
-		// Figure out how to derive deposit_channels from ChannelId (or pass DepositChannel in)
-		// Create a Transfer for each Transfer
+		let vault_program = Env::lookup_account(SolanaEnvAccountLookupKey::VaultProgram)
+			.map_err(|e| AllBatchError::DispatchError(e.into()))?;
+		let deposit_channels = fetch_params
+			.into_iter()
+			.map(|fetch_param| {
+				address_derivation::derive_deposit_channel::<Env>(
```
dandanlen, line 250: We shouldn't have to re-derive the deposit channel here. 

I think maybe we need to change the definition of the `DepositFetchId` for Solana. 

At the moment it's just `ChannelId` - we can make it `(ChannelId, Bump)`, and then we have all the information we need to build the fetch tx.

syan095: We still need the derived account ID for fetching as well. What we need is (AccountID, Bump), I feel like it is easier to derive it from a channelID than storing the accountID + bump and passing them in

dandanlen: I would prefer if we follow existing patterns: we already have derived all the information and it's already stored in the ingress-egress pallet where we trigger the fetch. For all other chains we define the DepositFetchId such that it contains all the information required for a fetch, and we extract this from the DepositChannel (hence this constraint on the DepositFetchId: 
```
		+ for<'a> From<&'a DepositChannel<Self>>;
```

I think the correct approach here is to use the DepositFetchId for this, ie. create a SolanaDepositFetchId with the address, channel_id and bump.

Then the FetchAssetParams will contain the required information and we don't need this extra conversion into a DepositChannel.

dandanlen: Alternatively we could refactor fetching such that it always passes a list of deposit channels, and remove the whole layer of abstraction around FetchId etc. I think this might make things clearer and easier to implement. But this would be a much bigger refactor. 

Either way, we should use an approach that is consistent across chains.

syan095: Done. We only pass in FetchParam, instead of re-generating the deposit_channel

syan095:
Just realized when I started Token support that for token, we will need both the derived deposit address as well as ATA derived from the said deposit address (therefore it will not be a single address, but 2) so the return value from the address derivation will need to be changed further.

Further work will be done in the Token Support PR as we need the new asset.

albert-llimos:
Fwiw, depending on the witnessing solution it might be that we need to derive an extra address for every deposit channel. That is regardless of being a native asset or an spl-token.

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
```diff
@@ -0,0 +1,149 @@
...
+	let seed = channel_id.to_le_bytes();
+	derive_address(seed, vault_program)
+		// .and_then(|(address, bump)| match asset {
+		// 	SolAsset::Sol => Ok((address, bump)),
+		// 	SolAsset::Usd => {
+		// 	    let mint_pubkey =
+		// 	Env::lookup_account(crate::sol::api::SolanaEnvAccountLookupKey::TokenMintPubkey)?;
+		// 	    derive_associated_token_account(address, mint_pubkey)
+		// 	},
+		// })
```
dandanlen, line 27: Can this be removed?

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
```diff
@@ -0,0 +1,149 @@
...
+	DerivedAddressBuilder::from_address(vault_program)?.chain_seed(seed)?.finish()
+}
+
+#[allow(dead_code)]
+/// Derive a Associated Token Account (ATA) of a main target account.
+fn derive_associated_token_account(
+	target: SolAddress,
+	mint_pubkey: SolAddress,
+) -> Result<(SolAddress, AccountBump), AddressDerivationError> {
+	let associated_token_program_id = SolAddress::from_str(ASSOCIATED_TOKEN_PROGRAM_ID)
```
dandanlen, line 50: Not related specifically to this PR, but we have a lot of constants that are stored as a string and we convert them at runtime - there is probably a way to do the conversion at compile time...

dandanlen: Update: it looks like bs58 allows this - we just need to bump the version from 0.4 to 0.5.1. (we can consider this for later). This would have the advantage of compile-time checking all the base58 strings to ensure they are valid base58 of the correct length.

I just tested this locally:

```rust
const fn const_pubkey(s: &'static str) -> Pubkey {
	Pubkey(bs58::decode(s.as_bytes()).into_array_const_unwrap())
}

// This is fine
const VALID_PUBKEY: Pubkey = const_pubkey("274BzCz5RPHJZsxdcSGySahz4qAWqwSDcmz1YEKkGaZC");

// This causes a compile-time error
const INVALID_PUBKEY: Pubkey = const_pubkey("--invalid chars--");
```

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
syan095: This is partial work for token support that has been implemented properly in the Token support PR.
This part has been cleaned up

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
dandanlen: I added an issue to Linear.

PR #4853, opened 2024-05-09
Boost bouncer commands and test

j4m1ef0rd:
# Pull Request

Closes: PRO-1189, PRO-1364

## Checklist

## Summary

- Added bouncer commands to add boost funds and stop boosting.
- Added optional boosting for swaps in bouncer (eg, via the new_swap command)
- Added a boost test that does a boosted swap and checks the fees collected.
- Refactored the creation of the boost pools so the new test could create one.
- Cleaned up a few other things in the bouncer.


bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+keyring.setSS58Format(2112);
+
+async function lpSubmitExtrinsic(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  extrinsic: any,
+  keyringPair: KeyringPair,
+  errorOnFail: boolean = true,
+) {
+  let extrinsicResult;
+  await lpMutex.runExclusive(async () => {
```
msgmaxim, line 35: It seems strange to me that we hardcode `lpMutex` here, while a different account can passed in as a parameter. We should either hardcode both the acount and the mutex, or provide a way to set a custom mutex too.

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+  boostTier: number,
+  amount: number,
+  lpUri = '//LP_BOOST',
+): Promise<Event> {
+  await using chainflip = await getChainflipApi();
+  const lp = keyring.createFromUri(lpUri);
+
+  assert(boostTier > 0, 'Boost fee must be greater than 0');
+
+  await provideLiquidity(asset, amount, true, lpUri);
```
msgmaxim, line 90: I know this functions already existed, but this just adds funds to the LP account, right? I think we use "provide liquidity" to mean creating a range/limit order, which isn't done here. Maybe "deposit_liquidity" (or just "add_lp_funds") would be a better name?

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+  return observeBoostFundsAdded;
+}
+
+/// Adds boost funds to the boost pool and does a swap with boosting enabled, then stops boosting and checks the fees collected are correct.
+async function testBoostingForAsset(asset: Asset, boostFee: number, lpUri: string, amount: number) {
+  await using chainflip = await getChainflipApi();
+  console.log(`Testing boosting for ${asset} at ${boostFee}bps`);
+
+  // Start with a clean slate by stopping boosting before the test
+  const preTestStopBoostingEvent = await stopBoosting(asset, boostFee, lpUri, false);
```
msgmaxim, line 121: Seems weird to "stop boosting" just to get the event to check there aren't any pending boosts. Maybe better to use the new boost rpcs for this? Or just inspect storage?

j4m1ef0rd: It also cleans up at the end of the test so we don't have and boosting funds left in this tier for other tests that might happen after this.

bouncer/shared/boost.ts
j4m1ef0rd: That's why i called the function `lpSubmitExtrinsic`. Because it uses the LP mutex. Seemed ok to use the same mutex for different LP accounts. It's not perfect. If we end up using it somewhere else, I would move it to utils and make the mutex a param. Unless you think its worth doing that now?

bouncer/shared/boost.ts
j4m1ef0rd: I didn't know that was an implied meaning. can someone else confirm? @kylezs .
Also might leave it to another PR, this one is big enough.

kylezs: I think it's ok as is, we have terms like "setRangeOrder" "setLimitOrder" for the other things. Maybe `depositLiquidity` is a little more explicit though. Think it can be left for another PR though, will leave that decision to you @j4m1ef0rd

bouncer/commands/stop_boosting.ts
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env -S pnpm tsx
+// INSTRUCTIONS
+//
+// This command takes 3 arguments:
+// 1 - Asset
+// 2 - Tier
+// 3 (optional) - Account URI (Default: "//LP_1")
```
kylezs, line 7: Maybe we should default to "//LP_2" since that's what we use for boost in general? or `LP_BOOST` ?

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+export async function stopBoosting(
+  asset: Asset,
+  boostTier: number,
+  lpUri = '//LP_BOOST',
+  errorOnFail: boolean = true,
+): Promise<Event | undefined> {
+  await using chainflip = await getChainflipApi();
+  const lp = keyring.createFromUri(lpUri);
+
+  assert(boostTier > 0, 'Boost fee must be greater than 0');
```
kylezs, line 51:
```suggestion
  assert(boostTier > 0, 'Boost tier must be greater than 0');
```

bouncer/shared/boost.ts
kylezs, line 121: We can be a bit clearer here. It makes sense for someone to potentially want to run this test with `boostFee: 22`, however, this test will fail immediately since stopBoosting will try to stopBoosting for tier 22 which doesn't exist. We should be able to provide any boostFee to the test (since we can provide any fee to the channels).

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+  // Start with a clean slate by stopping boosting before the test
+  const preTestStopBoostingEvent = await stopBoosting(asset, boostFee, lpUri, false);
+  assert.strictEqual(
+    preTestStopBoostingEvent?.data?.pendingBoosts?.length ?? 0,
+    0,
+    'Stopped boosting but, the test cannot start with pending boosts.',
+  );
+
+  // Add boost funds
+  console.log(`Providing liquidity for boosting`);
```
kylezs, line 129: I think @msgmaxim is proven right here 😉 
```suggestion
  console.log(`Adding {amount} {asset} to boost tier {tier}`);
```

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+  await addBoostFunds(asset, boostFee, amount, lpUri);
```
kylezs, line 130: same issue exists here as above - if a non tier fee is provided this will fail

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  // TODO: Check that we are the only booster in the pool
```
kylezs, line 132: you were going to do this here or?

bouncer/shared/boost.ts
kylezs: Also this means that we'll only be testing the boost through one pool and not through several pools? Several pools gets us more coverage with the same test duration. So, would be ideal to have one that tests that

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  // Do a swap
+  const swapAsset = asset === Assets.Usdc ? Assets.Flip : Assets.Usdc;
+  const destAddress = await newAddress(swapAsset, 'LP_1');
```
kylezs, line 136: Not sure we should use 'LP_1' here, quite likely to conflict in other tests

bouncer/shared/boost.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  // Check that the swap was boosted
+  const depositEvent = await Promise.race([observeSwapBoosted, observeDepositFinalised]);
+  if (depositEvent.name.method === 'DepositFinalised') {
+    throw new Error('Deposit was finalised without seeing the DepositBoosted event');
+  } else if (depositEvent.name.method !== 'DepositBoosted') {
+    throw new Error(`Unexpected event ${depositEvent.name.method}`);
+  }
+
+  const depositFinalisedEvent = await observeDepositFinalised;
```
kylezs, line 171: I'm not sure how js handles this - have you tried to make it deliberately fail in this way? mainly thinking about the fact we have to copy the promise the here that was used above

bouncer/shared/broker_fee_collection.ts
```diff
@@ -204,12 +198,8 @@ async function testBrokerFees(asset: Asset, seed?: string): Promise<void> {
   // Check that the balance after withdrawal is correct after deducting withdrawal fee
   const balanceAfterWithdrawal = await getBalance(asset, withdrawalAddress);
   console.log(`${asset} Balance after withdrawal:`, balanceAfterWithdrawal);
-  const balanceAfterWithdrawalBigInt = BigInt(
-    amountToFineAmount(balanceAfterWithdrawal, assetDecimals(asset)),
-  );
-  const balanceBeforeWithdrawalBigInt = BigInt(
-    amountToFineAmount(balanceBeforeWithdrawal, assetDecimals(asset)),
-  );
+  const balanceAfterWithdrawalBigInt = amountToFineAmountBigInt(balanceAfterWithdrawal, asset);
```
kylezs, line 201: 👍

bouncer/shared/setup_boost_pools.ts
```diff
@@ -1,19 +1,78 @@
...
 
-type BoostPoolId = {
+export type BoostPoolId = {
   asset: Asset;
   tier: number;
 };
 
 // These are the tiers of boost pools that will be created for each asset
 const boostPoolTiers = [5, 10, 30];
 const chains = ['Ethereum', 'Polkadot', 'Bitcoin', 'Arbitrum'] as const;
```
kylezs, line 19: We have a `chainConstants` in the chainflip-cli - perhaps we can use that? Seems hard to believe this is the first time we've required such a constant

bouncer/shared/setup_boost_pools.ts
```diff
@@ -1,19 +1,78 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const observeBoostPoolEvents: Promise<any>[] = [];
+
+  for (const pool of newPools) {
+    if (chainFromAsset(pool.asset) !== chain) {
+      throw new Error(`All assets must be from the same chain`);
+    }
+
+    if (pool.tier <= 0) {
+      throw new Error(`Invalid tier value: ${pool.tier}`);
```
kylezs, line 38: No reason not to be explicit
```suggestion
      throw new Error(`Tier value: ${pool.tier} cannot be less than 0`);
```

bouncer/shared/setup_boost_pools.ts
```diff
@@ -1,19 +1,78 @@
...
+
+    const observeBoostPoolCreated = observeEvent(
+      `${chain.toLowerCase()}IngressEgress:BoostPoolCreated`,
+      {
+        test: (event) =>
+          event.data.boostPool.asset === pool.asset &&
+          event.data.boostPool.tier === pool.tier.toString(),
+      },
+    );
+    const observeBoostPoolAlreadyExists = observeEvent(`governance:FailedExecution`);
```
kylezs, line 49: I mean, you're observing governanceFailedExecution

bouncer/shared/setup_boost_pools.ts
```diff
@@ -1,19 +1,78 @@
...
+  }
+
+  submitGovernanceExtrinsic((api) =>
+    api.tx[ingressEgressPalletForChain(chain)].createBoostPools(newPools),
+  );
+
+  const boostPoolEvents = await Promise.all(observeBoostPoolEvents);
+  for (const event of boostPoolEvents) {
+    if (event.name.method !== 'BoostPoolCreated') {
+      // TODO: decode error here
```
kylezs, line 63: We have a `handleSubstrateError` function

bouncer/shared/boost.ts
j4m1ef0rd: Yes I tested it and it works as intended.

bouncer/shared/setup_boost_pools.ts
j4m1ef0rd: I was unable to use that function because it only works for dispatch errors, not error codes in the event data like this. So i create a new function that can decode it instead.

bouncer/shared/boost.ts
j4m1ef0rd: Now that we are creating default boost pools and filling them with funds, running the test on any tier lager than 5 will break the test because it will use that tier instead. So I thought it would be better to just hard code the test to use tier 4.

bouncer/shared/boost.ts
j4m1ef0rd: These details are already logged inside of the addBoostFunds function. The log was there because the `provideLiquidity` function does not log a simple message like this.

bouncer/shared/boost.ts
j4m1ef0rd: > Several pools gets us more coverage 

Unit tests cover this, so i think its not worth the added complexity here to test it. Maybe in the future.

j4m1ef0rd: > if a non tier fee is provided this will fail

This is intended. I wanted to keep the scope of this utility function limited.

bouncer/shared/boost.ts
j4m1ef0rd: The main point of the function is just to make it easier to get the extrinsic result. Because of the mutex, I am unable to get the result directly and must store it in a variable first. This makes it messy.

bouncer/shared/boost.ts
j4m1ef0rd: Happy to rename it in a different PR.

bouncer/shared/boost.ts
kylezs: It's more about the fact you used the term "provide liquidity" rather than "add boost funds"

bouncer/shared/boost.ts
j4m1ef0rd: I like to make the tests resistant to failing halfway and getting stuck in a state that needs to be cleaned up manually before the test can be re-run. So i automate the cleanup at the start of the test, instead of just an assert. This also helps a lot when first building the test. The test will only work sequentially for each asset because of the nature of the pools, so i think it's fine.

bouncer/shared/boost.ts
kylezs: https://linear.app/chainflip/issue/PRO-1382/rename-provideliquidity-in-bouncer-to-depositliquidity

bouncer/shared/boost.ts
msgmaxim: I feel like this was perhaps more useful when you were deveoping this, and not so much anymore (particularly since this doesn't guarantee full reset). I don't really mind keeping it though.

bouncer/shared/boost.ts
msgmaxim: I don't really mind the way it is written now. It is a little strange, but it does work (I was initially concerned about using a wrong mutex, but you are right that different accounts can share the same mutex).

I do like Kyle's suggestion re creating a class (which I imagine would take a mutex and a keyringPair in its constructor). Sounds like this might then be reused in other places too. Unless we are in the rush to merge this, it might be worth doing it here?

bouncer/shared/boost.ts
```diff
@@ -81,13 +82,19 @@ export async function addBoostFunds(
...
   await using chainflip = await getChainflipApi();
   const lp = keyring.createFromUri(lpUri);
+  console.log(`address: ${lp.address}`);
 
-  assert(boostTier > 0, 'Boost fee must be greater than 0');
+  assert(boostTier > 0, 'Boost tier must be greater than 0');
 
-  await provideLiquidity(asset, amount, true, lpUri);
+  if (depositLiquidity) {
+    // Provide a little more than the amount because some will be lost due to ingress fee
```
msgmaxim, line 96: I think this could be a little surprising to the caller of `addBoostFunds`. Can we not change the caller to instead provide a bit more than necessary?

Also, do we really need this function to fund the account internally? Shouldn't be too difficult to call `provideLiquidity` first if necessary. Then `depositLiquidity` parameter won't be necessary.

j4m1ef0rd: > do we really need this function to fund the account internally?

I was trying to make things simple for the tests. But yeah, turned out messy like this. I'll try removing it.

>Can we not change the caller to instead provide a bit more than necessary

I Wanted the added boost funds to still be exact. Thought this was the best way to do it.

kylezs: Relying on the boost funds being exact is probably going to end up in a flaky test, bouncer should be considered to test the overall end-to-end with the exact values of things tested in unit tests in the individual components

bouncer/shared/utils.ts
```diff
@@ -951,6 +951,26 @@ export async function submitChainflipExtrinsic(
...
+
+  private mutex: Mutex;
+
+  constructor(keyringPair: KeyringPair, mutex: Mutex = new Mutex()) {
+    this.keyringPair = keyringPair;
+    this.mutex = mutex;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async Submit(extrinsic: any, errorOnFail: boolean = true) {
```
msgmaxim, line 965: Why is this capitalised? I don't think we use this style for any methods (both in js and rust).

msgmaxim (approved):

bouncer/shared/utils.ts
j4m1ef0rd: woops. strange that prettier or something did complain about this.

kylezs (approved):

PR #4870, opened 2024-05-17
Implemented CCM transfer API call for Solana

syan095:
# Pull Request

Closes: PRO-1358

## Checklist

Implemented CCM transfer API call for Solana
the cf_parameter is now passed into the API call so it can be decoded
by the Soalana CCM

Added encoding test for CcmExtraAccounts - required to be included in the
ccm call.


state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -139,6 +138,55 @@ impl SolanaInstructionBuilder {
 
 		Self::finalize(instructions, nonce_account.into(), agg_key.into(), compute_price)
 	}
+
+	pub fn ccm_transfer(
+		transfer_param: TransferAssetParams<Solana>,
+		source_chain: cf_primitives::ForeignChain,
+		source_address: Option<ForeignChainAddress>,
+		message: Vec<u8>,
+		extra_accounts: SolCcmExtraAccounts,
```
albert-llimos, line 147: as before, I'd name this `ccm_accounts` instead.

state-chain/chains/src/sol/sol_tx_core.rs
```diff
@@ -772,6 +772,61 @@ impl From<Hash> for SolHash {
...
+
+#[derive(Encode, Decode, TypeInfo, Serialize, Deserialize, Debug, Clone)]
+pub struct CcmExtraAccounts {
+	pub cf_receiver: CcmExtraAddress,
+	pub remaining_accounts: Vec<CcmExtraAddress>,
+}
+
+impl CcmExtraAccounts {
+	pub fn remaining_account_metas(self) -> Vec<AccountMeta> {
+		self.remaining_accounts.into_iter().map(|acc| acc.into()).collect::<Vec<_>>()
```
albert-llimos, line 798: AccountMeta has an extra parameter, `isSigner` that is not passed as part of the ccm accounts. Will this default that value to false?

state-chain/chains/src/lib.rs
```diff
@@ -496,6 +496,7 @@ pub trait ExecutexSwapAndCall<C: Chain>: ApiCall<C::ChainCrypto> {
 		source_address: Option<ForeignChainAddress>,
 		gas_budget: C::ChainAmount,
 		message: Vec<u8>,
+		cf_parameter: Vec<u8>,
```
albert-llimos, line 499: Rename to `cf_parameters` to match the State Chain code. Same for other instances of this.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -139,6 +138,55 @@ impl SolanaInstructionBuilder {
 
 		Self::finalize(instructions, nonce_account.into(), agg_key.into(), compute_price)
 	}
+
+	pub fn ccm_transfer(
```
albert-llimos, line 142: Rename this to `ccm_transfer_native`

state-chain/chains/src/sol/sol_tx_core.rs
```diff
@@ -1227,49 +1318,62 @@ mod tests {
...
-					AccountMeta::new_readonly(agg_key_pubkey, true),
-					AccountMeta::new(to_pubkey, false),
-					AccountMeta::new_readonly(cf_receiver, false),
-					AccountMeta::new_readonly(Pubkey::from_str(SYSTEM_PROGRAM_ID).unwrap(), false),
-					AccountMeta::new_readonly(
-						Pubkey::from_str(SYS_VAR_INSTRUCTIONS).unwrap(),
-						false,
-					),
-				],
+					vec![
```
albert-llimos, line 1345: This looks incorrect to me but I could be wrong.
This should be a vector of AccountMetas where the extra accounts are appended at the end:
vec![a,b...,remainingAccount0, remainingAccount1...]
However, what I'm reading here is this:
vec![vec![a,b...],vec![remainingAccount0, remainingAccount1...]]

state-chain/chains/src/sol/sol_tx_core.rs
syan095: Yes. in the conversion trait impl:
```impl From<CcmAddress> for AccountMeta {```

the is_signer is hard-coded to be false

File: state-chain/chains/src/sol/instruction_builder.rs (hunk 4 above)
Comment (syan095) [reply state-chain/chains/src/sol/instruction_builder.rs:142]: This will support other assets within itself - so we can just call this function for both native and token asset.

File: state-chain/chains/src/sol/sol_tx_core.rs (hunk 5 above)
Comment (syan095) [reply state-chain/chains/src/sol/sol_tx_core.rs:1345]: the flatten call below flattens the 2 vec into a single one. But I understands that this is not obvious. For the tests in the Core, we should hardcode the value/structure to ensure the instruction builder builds the correct tx hash. 
I'll change it to a more direct value.

File: state-chain/chains/src/sol/instruction_builder.rs (hunk 4 above)
Comment (albert-llimos) [reply state-chain/chains/src/sol/instruction_builder.rs:142]: right, makes sense :+1:

File: state-chain/chains/src/sol/sol_tx_core.rs (hunk 5 above)
Comment (albert-llimos) [reply state-chain/chains/src/sol/sol_tx_core.rs:1345]: if the flatten does exactly that then it's fine as is.

File: state-chain/chains/src/btc/api.rs [hunk 6]
```diff
@@ -133,6 +133,7 @@ impl<E: ReplayProtectionProvider<Bitcoin>> ExecutexSwapAndCall<Bitcoin> for Bitc
 		_source_address: Option<ForeignChainAddress>,
 		_gas_budget: <Bitcoin as Chain>::ChainAmount,
 		_message: Vec<u8>,
+		_cf_parameter: Vec<u8>,
```
albert-llimos, line 136: There is still some sneaky `cf_parameter` missing an `s` at the end.

albert-llimos:
Overall this looks good to me. So, as usual, time to leave it for @dandanlen for further review.
For context, in Solana for CCM we will require the user to pass an array of addresses that we need to include in the egress transaction. The first one will be the address to receive the CCM call. This is because in Solana the address to recieve the call is different than the address to receive the asset.
We have used the `cf_parameters` for that purpose.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -139,6 +138,54 @@ impl SolanaInstructionBuilder {
...
+				&transfer_param.to.into(),
+				transfer_param.amount,
+			),
+			ProgramInstruction::get_instruction(
+				&VaultProgram::ExecuteCcmNativeCall {
+					source_chain: source_chain as u32,
+					source_address: codec::Encode::encode(&source_address),
+					message,
+					amount: transfer_param.amount,
+				},
```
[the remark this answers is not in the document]
albert-llimos: that's right.

state-chain/chains/src/sol/sol_tx_core.rs
```diff
@@ -773,6 +773,61 @@ impl From<Hash> for SolHash {
...
+			false => AccountMeta::new_readonly(from.pubkey, false),
+		}
+	}
+}
+
+#[derive(Encode, Decode, TypeInfo, Serialize, Deserialize, Debug, Clone)]
+pub struct CcmAccounts {
+	pub cf_receiver: CcmAddress,
+	pub remaining_accounts: Vec<CcmAddress>,
+}
```
ramizhasan111, line 795: I think we should not have the ```cf_receiver``` as part of the CcmAccounts type. The ```cf_receiver``` here is the same as the ```transfer_param.to```that is also passed in when creating the apicall so there is duplication. Moreover, we dont seem to use the ```cf_receiver``` when we construct the call. We use the ```transfer_param.to```. Also, its conceptuallz easier to follow if this type only contains the other accounts and so it can be renamed to the ```CcmTouchedAccounts``` or something.

albert-llimos: That's not right. The `cf_receiver` is not and should not be the same as the `transfer_params.to` because programs should not hold funds in Solana. It's therefore two different addresses.
And the `cf_receiver` is indeed used when crafting the `VaultProgram::ExecuteCcmNativeCall` instruction.
The naming `remaining_accounts` comes from the Anchor programs. Using a matching name was easy to understand (for me) but looking only at the State Chain it might not be the best naming. `CcmAdditionalAccounts` or something like that might be nicer.

PR #4887, opened 2024-05-23
feat: swap input excludes network fee

msgmaxim:
# Pull Request

Closes: PRO-1387

## Checklist

## Summary

This fixes inconsistent behavior where network fee shown in the `SwapExecuted` event was (depending on the swap direction/assets) deducted from either intermediate or output amount, but never from the input amount (in case the swap is from USDC). Now when the fee is deducted from some amount, it is always reflected in the appropriate field of `SwapExecuted`.

All existing tests used a 0 network fee. I added a new test with a non-zero network fee to test this updated behaviour.

FIY @acdibble


dandanlen (commented):
Sorry, took me a while to get round to this. 

I feel like the way to fix this is to fix the behaviour of the `Swap` methods (ie. stable_amount should not be set in `new()`) rather than compensating for a bad implementation by adjusting the amount when we emit it in the event. I've pushed a commit that does something like this - let me know what you think (and feel free to revert/edit).

Btw. We could also name the struct members a bit more accurately and consistently, like `amount` -> `input_amount`, `from` -> `input_asset` etc. This would make things easier to read and reason about. 
(Also it looks like fee_taken is unused? Maybe we can remove it.)

Ultimately we should change this so that we distinguish between:
- first leg input/output asset/amount
- network fee
- second leg input/output asset/amount

But that can probably wait for later.

Also @acdibble do you mind if we remove the deprecated event members?

msgmaxim:
@dandanlen I don't think your commit worked as intended, so I reverted it. It broke not just the new unit test that I added, but also some of the existing ones 😅. 

I removed `fee_taken` and renamed `amount -> input_amount` as it is the most confusing one. I chose not to rename the rest as I'd rather keep this small and merge it quickly and leave more contentious changes for another PR (maybe the one where we do the other changes that you suggested).

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1827,6 +1818,49 @@ fn max_swap_amount_can_be_removed() {
...
+				TO_ASSET,
+				AMOUNT,
+				destination_address.clone(),
+				bounded_vec![],
+				0,
+			);
+		})
+		.then_process_blocks_until(|_| System::block_number() == 3)
+		.then_execute_with(|_| {
+			let expected_swap_input = AMOUNT - NETWORK_FEE * AMOUNT;
```
dandanlen, line 1846: I think this might be where I got confused.

In a two-legged swap, the swap input (of the whole swap) is pre-network-fee. 
But in a one-legged swap we take the network fee and then emit the swap input... 

IMO this is another symptom of conflating 'user swaps' and 'amm swaps'/'swap legs'.

I think we should structure it more like this going forward: 

swap_input
first_leg: Optional [
	swap_leg_input
		-> LP Fee
	swap_leg_output
]
	-> network_fee
second_leg: Optional [
	swap_leg_input
		-> LP Fee
	swap_leg_output
]
swap_output

dandanlen (approved):
I think my commit did work as intended, but then intention was wrong: I was confused again between the notion of Swaps vs Swap Legs...

PR #4888, opened 2024-05-23
feat: NetworkFeeTaken includes swap_id

msgmaxim:
# Pull Request

Closes: PRO-1383

## Checklist

## Summary

As @dandanlen suggested, I moved NetworkFeeTaken into the swapping pallet so it can be emitted from there with the `swap_id`. This means we can still call `take_network_fee` where `swap_id` doesn't exist such as when simulating swaps. I also updated `take_network_fee` to return the fee in addition to the remaining amount, so the caller can use it in the event. Normally I would be concerned about making it the caller's responsibility to emit the event, but in this case the event is only emitted in a single place (and it will likely remain this way), so this should be fine.

Note that this changes event indexes in the pools pallet, so depending on how you process them this might be a breaking change @acdibble.


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -826,7 +830,16 @@ pub mod pallet {
 					"All swaps should have Stable amount set here"
 				);
 				let stable_amount = swap.stable_amount.get_or_insert_with(Default::default);
-				*stable_amount = T::SwappingApi::take_network_fee(*stable_amount).remaining_amount;
+
+				let NetworkFeeTaken { remaining_amount, network_fee } =
+					T::SwappingApi::take_network_fee(*stable_amount);
```
dandanlen, line 835: I was actually thinking we could move the whole implementation into the swapping pallet, including the storage and swapping logic. Keeping it in the pools pallet seems like it adds an unnecessary layer of indirection.

dandanlen: But this is a good start 😅

msgmaxim: Created https://linear.app/chainflip/issue/PRO-1404/move-collectednetworkfee-and-related-logic-into-the-swapping-pallet (which you can feel free to edit). Let me know if this PR can be merged.

dandanlen (approved):

acdibble:
> Note that this changes event indexes in the pools pallet, so depending on how you process them this might be a breaking change @acdibble. Do we need a placeholder event to keep the indexes unchanged?

nope, the order in the pallet doesn't affect us

PR #4889, opened 2024-05-23
feat: scale to json decoder

dandanlen:
# Pull Request

## Checklist

## Summary

This is the first step in PRO-1246 

I added a crate `scale-json`, which provides a utility for decoding scale-encoded bytes to json. It uses the scale-decoder crate, which allows writing custom decoders for scale-encoded values.

In the utilities crate, there is a new `dynamic-events` module that uses the scale-json crate to provide an `EventDecoder` specifically for decoding runtime events. This takes metadata and raw scale-encoded bytes and decodes them to a serde_json::Value. 

I had to make some decisions wrt. the decoding format, and in particular the decoder uses some heuristics to detect byte sequences so it can decode them as hex strings instead of sequences of numbers (which would be the default for json). 

There's a binary included in the PR that uses the EventDecoder. You can `cargo build` then run it at `./target/debug/scale-json-event-logger` to find out how it works.


utilities/scale-json/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+				// Heuristic to determine if the array is a byte array:
+				// - Array must be non-empty
+				// - AND Each value decodes to an unsigned json Number (ie. not an enum)
+				// - AND each value was encoded as a single byte
+				let (is_bytes, all_values) = values.try_fold(
+					(values.remaining() > 0, Vec::with_capacity(values.remaining())),
+					|(all_bytes, mut values), res| {
+						res.and_then(|item| {
+							let decoded_item = item.decode_with_visitor(self)?;
+							let is_byte = item.bytes().len() == 1 && decoded_item.is_u64();
```
kylezs, line 234: not sure I follow what this is about. We check that the item has 1 byte, but it's decoded into 8 bytes? Why is a single byte value not decoded into a u8 here? i.e. why are we not doing `&& decoded_item.is_u8()`

utilities/scale-json/src/test.rs
```diff
@@ -0,0 +1,173 @@
...
+impl Chain for Bitcoin {
+	type Asset = btc::Asset;
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Encode, Decode, TypeInfo)]
+struct InstanceLike<T: Chain> {
+	inner: T::Asset,
+}
+
+fn make_type_resolver<T: TypeInfo + 'static>() -> (u32, PortableRegistry) {
```
kylezs, line 83: just checking my understanding, when using this in a real implementation, we would just have one registry, and all the types would be registered to it? because currently we create a registry for every type separately, one for each test (which is fine for convenience).

utilities/scale-json/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+		value.0
+	}
+}
+impl AsRef<serde_json::Value> for ScaleDecodedToJson {
+	fn as_ref(&self) -> &serde_json::Value {
+		&self.0
+	}
+}
+
+// Implementing this is what gives us `DecodeAsType` for `ScaleDecodedToJson`.
```
kylezs, line 81: And i'm assuming this is why we also need the wrapper type, because we otherwise can't implement a foreign trait on a foreign type?

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
+use core::ops::Deref;
```
kylezs, line 1: A brief file heading, describing what this is about would be helpful.

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+		self.0
+			.pointer("/phase/ApplyExtrinsic")
+			.and_then(|index| index.as_u64())
+			.filter(|index| *index == extrinsic_index)
+			.map(|_| self.event())
+	}
+
+	/// Returns the inner event and it's extrinsic index if the event was triggered during the
+	/// ApplyExtrinsic phase.
+	pub fn indexed_extrinsic_event(&self) -> Option<(u64, DynamicEvent)> {
```
kylezs, line 71: nit: type alias, we possibly already have one for this iirc

```suggestion
	pub fn indexed_extrinsic_event(&self) -> Option<(ExtrinsicIndex, DynamicEvent)> {
```

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	/// Returns the inner event if it matches the given extrinsic index.
+	pub fn extrinsic_event(&self, extrinsic_index: u64) -> Option<DynamicEvent> {
+		self.0
+			.pointer("/phase/ApplyExtrinsic")
+			.and_then(|index| index.as_u64())
+			.filter(|index| *index == extrinsic_index)
+			.map(|_| self.event())
+	}
```
kylezs, line 67: This is so if we have an iterator of these DynamicEventRecords we can filter search them with something like this?

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+										)
+									})
+									.collect::<ErrorInfo>(),
+								_ => panic!("Inner error type is not an Enum"),
+							}
+						}),
+					)
+				})
+				.collect::<NamedVariantLookup<_>>(),
+			_ => panic!("Outer error type is not an Enum"),
```
kylezs, line 180: nit: maybe slightly more explicit / context specific, like we expect it be an enum because we expect it to specifically be the highest level RuntimeError enum
```suggestion
			_ => panic!("Outer error type is not an Enum"),
```

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+	events_type_id: u32,
+	types: PortableRegistry,
+	errors: NamedVariantLookup<ErrorInfo>,
+}
+
+impl EventDecoder {
+	/// Create a new event decoder.
+	///
+	/// The arguments to this constructor can be extracted from substrate metadata.
+	pub fn new(mut types: PortableRegistry, errors_type_id: u32) -> Option<Self> {
```
kylezs, line 148: What's the reasoning for requiring this separate type id for the errors? Assuming this is some substrate specific way of doing things? why is that?

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+				.variants
+				.into_iter()
+				.map(|pallet_error_type| {
+					(
+						pallet_error_type.index,
+						(pallet_error_type.name, {
+							let type_id = pallet_error_type
+								.fields
+								.first()
+								.expect("error variant has exactly one field")
```
kylezs, line 160: nit: technically this is asserting that we have *at least* one field. if we wanted to asset exactly one field, we could do a `.next()` and then check is_empty() or something.

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+			.find(|t| t.ty.path.segments.join("::") == SYSTEM_EVENTS_TYPE_NAME)
+			.map(|t| t.id)
+			.expect("System events type must exist in the metadata");
+
+		let events_type_id = *types
+			.retain(|id| events_type_id == id)
+			.get(&events_type_id)
+			.expect("outer event enum must exist in the metadata");
+
+		debug_assert_eq!(
```
kylezs, line 195: any reason not to make this an `assert_eq` instead of debug assert? if this isn't the case it's likely something is up

utilities/scale-json/src/test.rs
```diff
@@ -0,0 +1,173 @@
...
+			num_i128: -5,
+		}),
+		array: RuntimeEvent::Other(OtherEvent::Array(vec![1, 2, 3])),
+		array_empty: RuntimeEvent::Other(OtherEvent::Array(vec![])),
+		eth_instance: RuntimeEvent::Other(OtherEvent::Ethereum(InstanceLike { inner: eth::Asset::Usdc })),
+		btc_instance: RuntimeEvent::Other(OtherEvent::Bitcoin(InstanceLike { inner: btc::Asset::Btc })),
+		byte_array: RuntimeEvent::Other(OtherEvent::BytesArray([0xcf; 32])),
+		byte_vec: RuntimeEvent::Other(OtherEvent::ByteVec(vec![0xcf; 42])),
+		byte_vec_empty: RuntimeEvent::Other(OtherEvent::ByteVec(vec![])),
+	}
```
kylezs, line 151: maybe a test with some nested structs so we ensure the composite recursion works correctly

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+		let decoded = (0..event_count)
+			.map(|_| {
+				ScaleDecodedToJson::decode_as_type(
+					events_data_cursor,
+					self.events_type_id,
+					&self.types,
+				)
+				.and_then(|decoded| {
+					let mut json: JsonValue = decoded.into();
+					if let Some(dispatch_error) =
```
kylezs, line 235: little comment that we're converting the error to a more useful/readable format here

utilities/src/with_std/dynamic_events.rs
kylezs, line 148: we only ever return Some() here, can remove the Option

utilities/src/with_std/dynamic_events.rs
```diff
@@ -0,0 +1,419 @@
...
+			DynamicEventRecord::new(json)
+				.expect("Valid json event record")
+				.event()
+				.extrinsic_outcome()
+				.expect("outcome should not be None")
+		};
+
+		assert!(outcome(success).is_ok());
+		assert!(matches!(outcome(failed_1).unwrap_err(), DispatchError::DispatchError(_)));
+		assert!(matches!(outcome(failed_2).unwrap_err(), DispatchError::KnownModuleError { .. }));
```
kylezs, line 417: think we also check the module is correct?

utilities/scale-json-event-logger/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+				hex::decode(s.trim_start_matches("0x"))?
+					.try_into()
+					.map_err(|_| anyhow::anyhow!("Invalid option or hash: {s}."))?,
+			))),
+		}
+	}
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
kylezs: If so, maybe worthwhile adding it to CI to run against bouncer or something? Then we can see if any errors are hit etc? though if you're close with the actual usage of it, then probably not worth it

utilities/scale-json/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+		),+
+		$(,)?
+	) => {
+		$(
+			fn $name<'scale, 'resolver>(
+				self,
+				value: $t,
+				_type_id: scale_decode::visitor::TypeIdFor<Self>,
+			) -> Result<Self::Value<'scale, 'resolver>, Self::Error> {
+				Ok(serde_json::Value::String(trimmed_hex(value$( .$convert() )?)))
```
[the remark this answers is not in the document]
dandanlen: Yes. For example the number 1 should be encoded as 0x1 not 0x01.

utilities/scale-json/src/lib.rs
dandanlen: `is_u8` doesn't exist unfortunately 😓 .
`decoded_item` here is a `serde_json::Value`. So this is checking if the original scale value was encoded in a single byte *and* that it was decoded to a json number. (actually I need to double-check the single-byte assumption, ironically enough I don't think the assumption holds for compact encoding...).

Think of `is_u64` as meaning `is_unsigned_integer`.

utilities/scale-json/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+		self,
+		value: &mut scale_decode::visitor::types::Composite<'scale, 'resolver, Self::TypeResolver>,
+		_type_id: scale_decode::visitor::TypeIdFor<Self>,
+	) -> Result<Self::Value<'scale, 'resolver>, Self::Error> {
+		if value.has_unnamed_fields() {
+			if value.remaining() == 1 {
+				// Simple wrapper struct. No need for array.
+				let field = value.next().unwrap()?;
+				assert!(field.name().is_none(), "Checked for unnamed fields.");
+				Ok(field.decode_with_visitor(RawJsonVisitor::<R>::default())?)
```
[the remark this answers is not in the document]
dandanlen: Composite could be a struct or it could be contents of an enum variant (see below, this is called from visit_variant).

A composite with a single unnamed field would be, for example: 

```rust
struct This(..)
```

or:

```rust
enum X {
	A(This),
}
```

In terms of how this is encoded in Scale, an enum is equivalent to struct (just encode the fields). It's the metadata that gives context about how to *decode* it and that is what determines if it calls visit_composite (which has a list of fields that *might* be named) vs visit_tuple (just list of fields with no names) vs visit_array or visit_sequence (lists of unnamed fields that all have the same type). Array and sequence are actually treated the same in scale_decode.

utilities/scale-json/src/test.rs
dandanlen: Yes normally we would get the registry from the metadata and we only need to refresh it on a runtime upgrade. This will become clearer with the upcoming PR #4905

utilities/scale-json-event-logger/src/main.rs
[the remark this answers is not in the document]
dandanlen: For now I just used it for sanity checking the decoder: running it for a long time to ensure it doesn't panic for example. 

It's a nice tool to have in and of itself (You can easily print / filter all events for a given block hash for example, and it's trivially extendable to eg. scan a range of blocks to look for an event). 

I'm not sure adding it to CI would be useful though.

utilities/scale-json-event-logger/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+) -> anyhow::Result<()> {
+	let events_data = client
+		.storage()
+		.at(hash)
+		.fetch_raw(EVENTS_STORAGE_KEY)
+		.await?
+		.expect("No events in block");
+
+	let event_decoder = EventDecoder::new(
+		client.metadata().types().clone(),
```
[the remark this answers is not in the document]
dandanlen: True - this doesn't refresh the metadata, I didn't really write this with much other than debugging in mind. The next PR does this for the rpc call (see #4905 ) and it should be simple to port that solution across to here if we want.

utilities/src/with_std/dynamic_events.rs
dandanlen: Yes - this will be useful in the sc client, you can filter_map the list of event records to extract all the events for an extrinsic.

utilities/src/with_std/dynamic_events.rs
dandanlen: The errors type id is available from the metadata directly. The metadata contains the type ids for `RuntimeError`, `RuntimeEvent`, `RuntimeCall` but sadly not the `EventRecord<RuntimeEvent>`, which is why the latter needs to be searched for.

utilities/src/with_std/dynamic_events.rs
dandanlen: The fact it has exactly one field is what guarantees that first() is not None. ('at least one field' is implicit in the expect, 'exactly one field' is what explains why we assume there to be at least one... does that make sense?)

Also I'm not sure why i used first instead of next(). Maybe fields is not an iterator...

utilities/src/with_std/dynamic_events.rs
dandanlen: I updated the diff in your comment with an alernative error message - is that better?

utilities/src/with_std/dynamic_events.rs
dandanlen: I guess it just seemed gratuitous given that we expect this to always hold - it was more of a sanity check for development. I don't have a strong feeling either way though... could even just remove it tbh.

utilities/src/with_std/dynamic_events.rs
dandanlen: The idea of these types is just to augment the inner json with some event-specific methods, ie. you can use them to wrap some json that you 'know' contains eg. an EventRecord, and use the methods to do useful things, rather than having to manipulate raw json.

utilities/src/with_std/dynamic_events.rs
dandanlen: This isn't testing any decoding, it's really just passing through the json strings verbatim, so there's not much value in testing this... but I suppose sure, why not.

I agree there should be a test for the error decoding though. 

FWIW the error conversion code was mostly lifted from the scc, I can pull across the tests from there (and maybe add a few extra, we'll see).

utilities/src/with_std/dynamic_events.rs
kylezs: works for me

utilities/src/with_std/dynamic_events.rs
kylezs: > The fact it has exactly one field is what guarantees that first() is not None

I'm not sure what you mean. If there are two fields, then first() would be some, but there would be two fields and not exactly one?

utilities/scale-json-event-logger/src/main.rs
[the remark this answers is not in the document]
kylezs: Yeah, maybe just a comment in case someone uses it as reference code

utilities/src/with_std/dynamic_events.rs
dandanlen: Maybe clearer: this code doesn't assert that we have exactly one field. It asserts that we have at last one field, and it does so confidently because we assume that there is exactly one field.

AlastairHolmes: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one

utilities/scale-json/src/lib.rs
dandanlen: Update - pretty sure the compact encoding thing is fine. Arrays/Vecs of bytes are not encoded using compact encoding, which is exactly what we expect. I imagine it's possible for someone to explicitly encode an array of compact-encoded bytes, and in that case, they would be decoded to a json array, not a hex string. I think this is fine: Compact encoding is only useful for numbers wider than a u8.

kylezs (approved):

utilities/scale-json/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+		if value.fields().remaining() == 0 {
+			Ok(serde_json::Value::String(value.name().to_string()))
+		} else {
+			Ok(serde_json::Value::Object(
+				std::iter::once(value)
+					.map(|value| {
+						Ok::<_, Self::Error>((
+							value.name().to_string(),
+							RawJsonVisitor::<R>::default()
+								.visit_composite(value.fields(), R::TypeId::default())?,
```
AlastairHolmes, line 384: What is the type id for in scale-decode?

AlastairHolmes (approved):

AlastairHolmes:
I did notice this is forcing us to down grade our rustls version, and us multiple versions of scale in the lock file. I couldn't quickly see where this was coming from in the dependencies, but maybe worth correcting if possible.

utilities/scale-json/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+		visit_i256, &'scale [u8; 32],
+	}
+
+	decode_to_array! {
+		visit_tuple, Tuple,
+	}
+
+	decode_to_array_or_hex! {
+		visit_sequence, Sequence,
+		visit_array, Array,
```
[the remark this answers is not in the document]
dandanlen: It's same as the difference between a vec and an array. In the former case, the length is encoded as part of the payload. In the latter case the length is known statically (ie. encoded in the metadata). 

The underlying implementation of Array and Sequence are almost identical, in fact Sequence is defined like this:

```rust
pub struct Sequence<'scale, 'resolver, R: TypeResolver> {
    bytes: &'scale [u8],
    // Mostly we just delegate to our Array logic for working with sequences.
    // The only thing we need to do otherwise is decode the compact encoded
    // length from the beginning and keep track of the bytes including that.
    values: Array<'scale, 'resolver, R>,
}
```

utilities/scale-json/src/lib.rs
dandanlen: You can use it in the Visitor implementation's TypeResolver implementation to resolve additional information about a type. (See: https://docs.rs/scale-type-resolver/0.1.1/scale_type_resolver/trait.TypeResolver.html)

For example, the `PortableRegistry` type [implements this](https://docs.rs/scale-type-resolver/0.2.0/src/scale_type_resolver/portable_registry.rs.html#59), where type_id is just an integer and maps to metadata about the structure of the type. This metadata is used to delegate to the appropriate methods like visit_composite etc.

At *this* level of the decoding 'stack', we don't have access to the resolver so we can't resolve any information. The assumption is that we've reached the 'lowest resolution' and have to decide based on the available structural/primitive information.

Tbh. I'm not really sure why the type-id is passed without any means of resolving it... it does seem to defeat the purpose.

dandanlen:
> I did notice this is forcing us to down grade our rustls version, and to use multiple versions of the scale deps in the lock file. I couldn't quickly see where this was coming from in the dependencies, but maybe worth correcting if possible.

I think this was related to the old event decoder definition that was still in the runtime/custom rpcs crates. I removed them and bumped the version no rustls.

Will rebase against main too.

utilities/scale-json-event-logger/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+	client: &subxt::OnlineClient<SubstrateConfig>,
+	hash: H256,
+	filter: &PathFilter,
+) -> anyhow::Result<()> {
+	let events_data = client
+		.storage()
+		.at(hash)
+		.fetch_raw(EVENTS_STORAGE_KEY)
+		.await?
+		.expect("No events in block");
```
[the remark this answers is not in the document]
dandanlen: I admit I didn't think about this too hard - I think it's just to express that we always expect some return value since events are stored a Vec<_> (ie. it's not an Option<Vec<_>> so even if there are no values, you always). I guess the justification could be clearer.

PR #4890, opened 2024-05-23
feat: Solana USDC token support 

syan095:
# Pull Request

Closes: PRO-1359

## Checklist

## Summary
Added SolUsdc token to the list of Assets under the Solana chain
Added Token support for the Transfer and BatchFetch transaction for the Instruction builder
Added token support for the Solana API


syan095:
Notes: 
1. The `AssetWithDerivedAddress` and therefore `decomposed_fetch_params` was added as a way pass in the ATA account if and only if the fetched asset is Usdc. This way the ATA derivation must be done in the API layer and ensures the Instruction Builder is infallible. This design might be improved.

2. The complicated logic in SolanaApi::transfer() is to ensure that we only fetch the extra environment variables (which will include storage lookups) when we have at least 1 Token transfer, and that we only do 1 set of lookup for any number of token transfers. This is optimized so that we don't do a bunch of storage lookups for each transfer

albert-llimos:
Logically this looks good to me.

state-chain/chains/src/sol/api.rs
```diff
@@ -148,13 +158,28 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
+			.map(|param| AssetWithDerivedAddress::decompose_fetch_params(param, token_mint_pubkey))
+			.collect::<Result<Vec<_>, _>>()?;
 
 		// Build the instruction_set
 		let instruction_set = SolanaInstructionBuilder::fetch_from(
-			fetch_params,
+			decomposed_fetch_params,
+			token_mint_pubkey,
+			token_vault_ata,
+			token_program_id,
```
ramizhasan111, line 182: I guess ```token_mint_pubkey``` and ```token_program_id``` (or atleast one) are Usdc related addresses and not general token realted addresses @albert-llimos? because if so, we should rename it to indicate its Usdc specific.

state-chain/chains/src/sol/api.rs
```diff
@@ -176,32 +201,94 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
+							nonce_account,
+							compute_price,
+						))
+					})
+					.collect::<Result<Vec<_>, _>>()?
+			},
+		]
+		.into_iter()
+		.flatten()
+		.map(|instruction_set| Self {
```
ramizhasan111, line 282: This assumes that any asset in the FetchParams that is not Sol, is Usdc. I know we might not add any future tokens other than Usdc but would be nice for it to be extendible to other assets. 

In general, the way code is written assumes (also see fetch_from() function ) that when we are talking about token, we mean Usdc so it seems it wont be extendible to others potential tokens we might wanna support in the future

state-chain/chains/src/sol/api.rs
albert-llimos: `token_program_id` is the native Solana token program
`token_mint_pubkey` would be the `usdc_mint_pubkey`. I wouldn't rename anything to `usdc` because we should have the token code generic in case we decide to support new tokens in the future.

state-chain/chains/src/sol/api.rs
albert-llimos: I agree with that. It should be generalized so we can support any token by just passing a different `mint_pub_key` for the token.

state-chain/chains/src/sol/api.rs
ramizhasan111: The thing is that this is assumed to be usdc specific address in the code since when we call the ```fetch_from```, this is used to create a Usdc fetch tx so either we should make the code generic such that we read from environment in a generic way or we should rename it for now since the code treates it as it is the usdc address. This ties in to my other comment regarding us having generic code that is extendible to other sol assets. 

Also, in the CCM for tokens PR we also have a ```ccm_transfer_usdc_token``` that uses this key and there even the function name is usdc specific

albert-llimos: yes you're right. I'd prefer to have it generic, as everything expect the `usdc_mint_pubkey` would be exactly the same code for another token.

state-chain/chains/src/sol/api.rs
```diff
@@ -148,13 +158,28 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
+			.expect("Preset Solana System Program ID account must be valid.");
 		let (nonce_account, durable_nonce) = Environment::nonce_account()?;
 		let compute_price = Environment::compute_price()?;
+		let token_mint_pubkey =
+			Environment::lookup_account(SolanaEnvAccountLookupKey::TokenMintPubkey)?;
+		let token_program_id = SolAddress::from_str(TOKEN_PROGRAM_ID)
+			.expect("Preset Solana Token program ID must be valid.");
+		let token_vault_ata = Environment::lookup_account(
+			SolanaEnvAccountLookupKey::TokenVaultAssociatedTokenAccount,
+		)?;
```
dandanlen, line 170: I think Roy mentioned this in another PR: it seems like it would make sense for the Solana environment lookup to just expose a single `struct SolanaAccounts { token_program, vault_account, etc etc }` rather than looking up each account individually. 

Doing it individually is (a) hard to read and (b) probably inefficient since we incur the cost of ~1 read per account instead of a single read for all accounts.

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
```diff
@@ -1,49 +1,23 @@
...
 
-/// Derive address/deposit channels for a channel ID
-pub fn derive_deposit_channel(
+/// Derive address for a given channel ID
+pub fn derive_deposit_address(
 	channel_id: ChannelId,
-	_asset: SolAsset,
 	vault_program: SolAddress,
-) -> Result<DepositChannel<Solana>, AddressDerivationError> {
+) -> Result<(SolAddress, AccountBump), AddressDerivationError> {
```
dandanlen, line 14: We could create a `DerivedAddress` struct instead of using a raw tuple.

state-chain/chains/src/sol/api.rs
albert-llimos: yes definitely. When we build the environment we can do that.

state-chain/chains/src/sol/sol_tx_core/address_derivation.rs
```diff
@@ -121,29 +103,32 @@ mod tests {
 	}
 
 	#[test]
-	fn can_derive_deposit_channel_for_sol() {
+	fn can_derive_deposit_address_native() {
 		let vault_program = SolAddress::from_str(sol_test_values::VAULT_PROGRAM).unwrap();
+		assert_eq!(
+			derive_deposit_address(0u64, vault_program).unwrap(),
+			(SolAddress::from_str("JDtAzKWKzQJCiHCfK4PU7qYuE4wChxuqfDqQhRbv6kwX").unwrap(), 254u8),
```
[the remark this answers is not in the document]
albert-llimos: these ones are just deriving the accounts using the Solana SDK or anything that derives accounts according to the Solana specs.
I think these one to check address derivation is fine as is. There are others in the code that could use more documentation indeed.

PR #4895, opened 2024-05-23
feat: declarative Solana program macro

dandanlen:
# Pull Request

I created this because I was finding it hard to find my way around the Solana program definitions. 

I think this is an improvement because it keeps all the definitions in one place and ensures the calls are constructed correctly (ie. we don't rely on the caller passing the correct vec of account metas). 

I realise after writing it that there is already a json abi - we could parse that instead, it would have the same effect, but would require a bit more work. 

Creating as a draft for now as I finished it in a bit of a rush and want to make sure the test pass before opening for review.


albert-llimos:
I think the idea of this is good. The question, as you wrote, is whether we can use the json abi for the same purpose. I think I'm happy with this approach as it is very readable and should be easy enough to match against the abi.

dandanlen:
Yeah - I'll compare the current implementation with the json abi before removing draft status.

albert-llimos:
> This broke some unit tests that compare the generated tx to some expected output

I'll need to check this thoroughly (very manual testing) so we don't break anything. Please don't merge this on the other PR.

state-chain/chains/src/sol/sol_tx_core.rs
```diff
@@ -1090,30 +1066,18 @@ mod tests {
...
-					AccountMeta::new_readonly(agg_key_pubkey, true),
-					AccountMeta::new_readonly(deposit_channel, false),
-					AccountMeta::new(deposit_channel_ata, false),
-					AccountMeta::new(
-						Pubkey::from_str(TOKEN_VAULT_ASSOCIATED_TOKEN_ACCOUNT).unwrap(),
-						false,
-					),
-					AccountMeta::new_readonly(Pubkey::from_str(MINT_PUB_KEY).unwrap(), false),
-					AccountMeta::new_readonly(Pubkey::from_str(TOKEN_PROGRAM_ID).unwrap(), false),
-					AccountMeta::new_readonly(Pubkey::from_str(SYSTEM_PROGRAM_ID).unwrap(), false),
```
dandanlen, line 1115: FYI I think the difference might come from here. The old code had the system_program_id, the idl definitions don't contain this. Everything else is the same.

albert-llimos: I think this is related to a gotcha that Solana has, where the System program is required for some calls but it's not required anywhere, so you need to know that it's required and pass it, otherwise it will fail. The new Anchor IDL probably messed that one up.
In any case, fetch tokens is exactly the one that needs several changes so I'll have to review it. Thanks for the info.

albert-llimos:
These are the changes I have pushed:
- Updated to the latest programs (0.5.0) where the Upgrade Manager is removed. I have therefore removed the related logic.
- Update function calls and test them locally in a live network. Found a couple of bugs but now they're all working.
- Update localnet tag, image and bouncer initialization to match that version of the programs. I want to make sure they're always aligned. 
- Added the address derivation (with tests) for the fetch_account address calculation (`derive_fetch_account`). 
- Use that fetch account derivation in `decompose_fetch_params` function, which is where it seemed to make more sense to have it, as the derivation depends also on the asset.
- "Hardcoded" the solana native programs (system_program, token_program...) inside the instruction builder so we don't have to pass them around.

Things that are missing to have this done: (@dandanlen )
- Add support for the ccm_remaining_accounts so they can be passed as extra accounts to the ccm calls. Dan said he already had some ideas for doing that.
- Some new functions of the programs have no arguments and that seems to be an issue for the macro. We need to update it so it works with no arguments. I've left a "TODO" in `program_instructions.rs`  and the commented out interfaces already there.

I think those are the only things missing for this to be ready. Regarding the new function calls for the 0.5.0 program, these are "governance" calls and I wouldn't bother implementing them for now, especially not in this PR. So I'd just instantiate them in the `program_instructions.rs` as mentioned above so the macro is complete and that's all.

dandanlen:
I've added the missing features mentioned above. Tests are now failing because the ccm transaction outputs don't match - I guess this is to be expected?

albert-llimos:
> I've added the missing features mentioned above. Tests are now failing because the ccm transaction outputs don't match - I guess this is to be expected?

Yes. I have now updated those, added some missing ccm accounts in the tests and fixed the merge conflicts. This looks complete now from my point of view.

dandanlen:
One final improvement I would like to make is that if the contract addresses are deterministic  (which seems to be the case(?): they are included in the idl file), we can hard-code them via the macro definition and avoid a storage lookup and extra parameter when building the call.

albert-llimos:
> One final improvement I would like to make is that if the contract addresses are deterministic (which seems to be the case(?): they are included in the idl file), we can hard-code them via the macro definition and avoid a storage lookup and extra parameter when building the call. WDYT?

I think I'd rather avoid that. For the contract addresses to be deterministic it requires generating a new keypair upon deployment. If we do that now we'd need to keep the keypair safely stored until the deployment of all contracts.
Maybe it's an optimization worth doing before deployment but doing it now sounds like an unnecessary headache to have right now.

albert-llimos (approved):

PR #4897, opened 2024-05-23
feat: Usdc token support for Solana CCM transaction

syan095:
# Pull Request

Closes: PRO-1390

## Checklist

## Summary

Added Usdc Token support for CCM messages in the Solana instruction builder and Solana API


syan095:
Something to note here: 
Because of the current ExecutexSwapAndCall trait, we only craft 1 ccm transaction at a time.
This means we will perform a set of Environment lookups for each CCM tx we craft. 

One way to optimize, if we so choose is to create ccms in batch. But this may not be something that we care about - because ccm are by nature more expensive, so we may want to leave it this way.

PR #4901, opened 2024-05-24
refactor: bouncer cleanup

j4m1ef0rd:
# Pull Request

Closes: PRO-1382, PRO-785

## Checklist

## Summary

- Renamed `provideLiquidity` -> `depositLiquidity`
- Changed the new swap command to log the deposit address.
- Fixed the use of the chainflip api in the lp api test.
	- Made the observe swap function error if the api becomes disconnected.
- Added a new function `executeWithTimeout` to use instead of `runWithTimeout` that logs the execution time and handles the process exit.
	- Refactored most commands and tests to use the new timeout function.
	- Changed some timeouts that were too large compared to the actual execution time.

### DIFF
--- a/bouncer/commands/get_usdc_balance.ts
+++ b/bouncer/commands/get_usdc_balance.ts
@@ -15,4 +15,5 @@ async function getUsdcBalanceCommand(ethereumAddress: string) {
   console.log(await getErc20Balance('Ethereum', ethereumAddress, contractAddress));
 }
 
-await executeWithTimeout(getUsdcBalanceCommand(process.argv[2] ?? '0'), 5);
+const ethereumAddress = process.argv[2] ?? '0';
+await executeWithTimeout(getUsdcBalanceCommand(ethereumAddress), 5);
--- a/bouncer/commands/get_usdc_balance.ts
+++ b/bouncer/commands/get_usdc_balance.ts
@@ -7,19 +7,12 @@
 // For example: ./commands/get_usdc_balance.ts 0xcf1dc766fc2c62bef0b67a8de666c8e67acf35f6
 // might print: 100.2
 
-import { runWithTimeout, getContractAddress } from '../shared/utils';
+import { executeWithTimeout, getContractAddress } from '../shared/utils';
 import { getErc20Balance } from '../shared/get_erc20_balance';
 
 async function getUsdcBalanceCommand(ethereumAddress: string) {
   const contractAddress = getContractAddress('Ethereum', 'Usdc');
-  const balance = await getErc20Balance('Ethereum', ethereumAddress, contractAddress);
-  console.log(balance);
-  process.exit(0);
+  console.log(await getErc20Balance('Ethereum', ethereumAddress, contractAddress));
 }
 
-const ethereumAddress = process.argv[2] ?? '0';
-
-runWithTimeout(getUsdcBalanceCommand(ethereumAddress), 5000).catch((error) => {
-  console.error(error);
-  process.exit(-1);
-});
+await executeWithTimeout(getUsdcBalanceCommand(process.argv[2] ?? '0'), 5);
--- a/bouncer/commands/upgrade_network.ts
+++ b/bouncer/commands/upgrade_network.ts
@@ -119,4 +119,4 @@ async function main(): Promise<void> {
 }
 
 // Quite a long timeout, as the sequence of try-runtime runs takes some time.
-await executeWithTimeout(main(), 30 * 60 * 1000);
+await executeWithTimeout(main(), 30 * 60);
--- a/bouncer/commands/upgrade_network.ts
+++ b/bouncer/commands/upgrade_network.ts
@@ -29,7 +29,7 @@ import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
 
 import { upgradeNetworkGit, upgradeNetworkPrebuilt } from '../shared/upgrade_network';
-import { runWithTimeout } from '../shared/utils';
+import { executeWithTimeout } from '../shared/utils';
 
 async function main(): Promise<void> {
   await yargs(hideBin(process.argv))
@@ -116,14 +116,7 @@ async function main(): Promise<void> {
     )
     .demandCommand(1)
     .help().argv;
-
-  process.exit(0);
 }
 
 // Quite a long timeout, as the sequence of try-runtime runs takes some time.
-runWithTimeout(main(), 30 * 60 * 1000).catch((error) => {
-  console.error('upgrade_network exiting due to error: ', error);
-  if (process.exitCode === 0) {
-    process.exitCode = -1;
-  }
-});
+await executeWithTimeout(main(), 30 * 60 * 1000);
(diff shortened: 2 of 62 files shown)


bouncer/shared/utils.ts
```diff
@@ -344,12 +344,16 @@ export async function observeEvent(
   eventQuery?: EventQuery,
   stopObserveEvent?: () => boolean,
   finalized = false,
+  debug = false,
```
msgmaxim, line 367: I know that you added something similar in a few other places already (e.g. `doPerformSwap`), but I wonder if instead of passing parameters for this we should use and enviornment variable instead (similar to RUST_LOG). Then maybe (in a separate PR) we could write a wrapper around console.log to check that env variable. Wouldn't be too difficult to have different levels as well. Alternatively, we could look into using a third party library for this. The way we do it right now is awkward and inconsistent.

bouncer/shared/lp_api_test.ts
```diff
@@ -70,6 +70,8 @@ async function provideLiquidityAndTestAssetBalances() {
 }
 
 async function testRegisterLiquidityRefundAddress() {
+  await using chainflip = await getChainflipApi();
```
msgmaxim, line 72: Why is this necessary when we have `chainflip` in the global scope?

bouncer/commands/get_usdc_balance.ts
```diff
@@ -7,19 +7,12 @@
...
+  console.log(await getErc20Balance('Ethereum', ethereumAddress, contractAddress));
 }
 
-const ethereumAddress = process.argv[2] ?? '0';
-
-runWithTimeout(getUsdcBalanceCommand(ethereumAddress), 5000).catch((error) => {
-  console.error(error);
-  process.exit(-1);
-});
+await executeWithTimeout(getUsdcBalanceCommand(process.argv[2] ?? '0'), 5);
```
msgmaxim, line 18: IMO the code was more clear with `const ethereumAddress = process.argv[2] ?? '0';` uninlined (otherwise it took me a while to understand what we are reading from command line arguments). Same in other places.

bouncer/shared/utils.ts
kylezs: We could also leverage a lightweight js logging library, which likely does this plus some more potentially useful features

bouncer/shared/lp_api_test.ts
kylezs: When using `await using` I think best not to put in the global scope, as it works like `Drop` in rust. Becomes much harder to reason about when it's in the global scope. So if there is a global usage, we should change that. 

Note that this function is doing something like an Arc.clone() and isn't re-initialising the whole client.

bouncer/commands/get_usdc_balance.ts
martin-chainflip: agreed

bouncer/shared/utils.ts
msgmaxim: That's what I meant by
> Alternatively, we could look into using a third party library for this.

😉

bouncer/shared/lp_api_test.ts
msgmaxim: Ah, I just realised that the global one was deleted. I was mostly saying that it didn't make sense to keep both. Must have looked at my local branch.

bouncer/shared/utils.ts
kylezs: oh lol, missed that completely

kylezs: Created a ticket: https://linear.app/chainflip/issue/PRO-1415/use-a-js-logging-library-in-bouncer

bouncer/commands/get_usdc_balance.ts
msgmaxim: I was suggesting that we uninline all other instances where process.argv was inlined by this PR, not just this particular line @kylezs 🙂

kylezs:
Told @j4m1ef0rd I'd take this over while he's away, but for now have deliberately left it to reduce bouncer changes that aren't directly addressing the flaky issues we've been seeing.

bouncer/shared/utils.ts
j4m1ef0rd: Looks like the function is getting removed in #4958 anyway.

kylezs (approved):

bouncer/shared/utils.ts
kylezs: I'll deal with it in that PR

kylezs:
It looks like this PR introduces some new flakiness, I've only seen the connection issues on this PR. Should investigate this before merging.

j4m1ef0rd:
Seems like the problem is with the connection to the BTC node. The error is happening when the swapping.ts test first tries to connect to the BTC node, no connection?. The false timeout message lead me astray for a while.
What could be causing this? @ahasna
I don't see how anything i did in this PR would affect that.

bouncer/commands/upgrade_network.ts
```diff
@@ -116,14 +116,7 @@ async function main(): Promise<void> {
...
 }
 
 // Quite a long timeout, as the sequence of try-runtime runs takes some time.
-runWithTimeout(main(), 30 * 60 * 1000).catch((error) => {
-  console.error('upgrade_network exiting due to error: ', error);
-  if (process.exitCode === 0) {
-    process.exitCode = -1;
-  }
-});
+await executeWithTimeout(main(), 30 * 60 * 1000);
```
kylezs, line 122: Isn't this off by a factor of 1000 now?

kylezs:
> Seems like the problem is with the connection to the BTC node.

I would first try rebasing not just to resolve conflicts but there have been a couple of changes to localnets recently

bouncer/full_bouncer.sh
```diff
@@ -9,6 +9,7 @@ echo "Running nightly tests 🧪"
 ./tests/rotates_through_btc_swap.ts
 ./tests/btc_utxo_consolidation.ts
 ./tests/rotation_barrier.ts
+./tests/multiple_members_governance.ts
```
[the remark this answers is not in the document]
j4m1ef0rd: I thought it might of been interfering with the boost test, but now i think it was fine. But it still makes sense to me, A test that changes the environment like that should not run at the same time as other tests, so i moved it to full bouncer instead.

bouncer/tests/all_concurrent_tests.ts
```diff
@@ -65,7 +63,6 @@ runWithTimeout(runAllConcurrentTests(), 2000000)
     process.exit(0);
   })
   .catch((error) => {
-    console.error!('All concurrent tests timed out. Exiting.');
```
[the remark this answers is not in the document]
j4m1ef0rd: This error message was logging whenever an error happened, not just a timeout, so it was very misleading. We already print a timeout log message and print the error here otherwise, so this error message was redundant.

bouncer/shared/utils.ts
```diff
@@ -216,6 +216,26 @@ export const runWithTimeout = async <T>(promise: Promise<T>, millis: number): Pr
...
+export async function executeWithTimeout<T>(promise: Promise<T>, seconds: number): Promise<void> {
+  const start = Date.now();
+  await runWithTimeout(promise, seconds * 1000).catch((error) => {
+    console.error(error);
+    process.exit(-1);
+  });
+  const executionTime = (Date.now() - start) / 1000;
+
+  if (executionTime > seconds * 0.9) {
+    console.log(
```
kylezs, line 229: console.warn ?

kylezs (approved):

PR #4912, opened 2024-05-28
fix: more robust governance bouncer test

kylezs:
# Pull Request

## Checklist

## Summary

Possible the GovernanceExecuted is actually picking up the executed of another test, and therefore maybe running the check too early. This just protects against that.

### DIFF
--- a/bouncer/shared/utils.ts
+++ b/bouncer/shared/utils.ts
@@ -429,15 +429,15 @@ export async function observeSwapEvents(
             if ('DepositChannel' in expectedEvent.data.origin) {
               if (
                 Number(expectedEvent.data.origin.DepositChannel.channelId) === channelId &&
-                sourceAsset === (expectedEvent.data.sourceAsset as Asset) &&
-                destAsset === (expectedEvent.data.destinationAsset as Asset) &&
-                swapType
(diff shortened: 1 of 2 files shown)


bouncer/shared/utils.ts
```diff
@@ -977,3 +977,30 @@ export function calculateFeeWithBps(fineAmount: bigint, bps: number): bigint {
...
+          clearInterval(intervalId);
+          reject(new Error('Maximum attempts reached'));
+        }
+      } catch (error) {
+        clearInterval(intervalId);
+        reject(error);
+      }
+    }, pollTime);
+  });
+}
```
acdibble, line 1006: wouldn't bother w/ set interval here honestly:
```suggestion
export async function tryUntilSuccess(
  closure: () => Promise<boolean>,
  pollTime: number,
  maxAttempts: number,
) {
  for (let attempts = 0; attempts < maxAttempts; attempts++) {
    await sleep(pollTime);
    if (await closure()) return;
  }
  
  throw new Error('Maximum attempts reached');
}
```

kylezs: haha yeah, just added that and pushed without seeing this, definitely over complicated (and it didn't work)

albert-llimos (approved):

PR #4796, opened 2024-04-25
chore: Update Solana Infra

albert-llimos:
## Checklist

Please conduct a thorough self-review before opening the PR.

- [x] I am confident that the code works.
- [x] I have updated documentation where appropriate.

## Summary

- Update Solana contracts to the latest with it's corresponding initialization change.
- Update Solana infra with the 1.18 client.
- A bunch of CI updates to to the new Solana client having a symlink that causes issues in Linux.

PR #4916, opened 2024-05-29
fix: add missing debug.log on recreate in localnet/manage.sh

msgmaxim:
# Pull Request

## Checklist

## Summary

I noticed that "recreate" option stopped working on localnets, and realised that the fix I've merged in the past for this (https://github.com/chainflip-io/chainflip-backend/pull/4487) got reverted in https://github.com/chainflip-io/chainflip-backend/pull/4796. I'm assuming this wasn't intentional, but let me know if it was @albert-llimos.

### DIFF
--- a/localnet/manage.sh
+++ b/localnet/manage.sh
@@ -136,6 +136,9 @@ build-localnet() {
     fi
   done
 
+  mkdir -p /tmp/chainflip/
+  touch /tmp/chainflip/debug.log
+
   echo "🪢 Pulling Docker Images"
   docker compose -f localnet/docker-compose.yml -p "chainflip-localnet" pull --quiet >>$DEBUG_OUTPUT_DESTINATION 2>&1
   echo "🔮 Initializing Network"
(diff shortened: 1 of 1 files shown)


localnet/manage.sh
```diff
@@ -136,6 +136,9 @@ build-localnet() {
     fi
   done
 
+  mkdir -p /tmp/chainflip/
+  touch /tmp/chainflip/debug.log
```
kylezs, line 140: We should use the variable we have `$DEBUG_OUTPUT_DESTINATION`

msgmaxim: updated

albert-llimos:
Pretty sure it was @ahasna as part of the debugging. Probably not on purpose but he maybe can confirm.

localnet/manage.sh
```diff
@@ -137,7 +137,7 @@ build-localnet() {
   done
 
   mkdir -p /tmp/chainflip/
```
kylezs, line 139: we should also use the variable we have for this too - if we use the variable below, hardcoding this doesn't make sense

kylezs: CHAINFLIP_BASE_PATH="/tmp/chainflip"

msgmaxim: updated

kylezs (approved):

PR #4919, opened 2024-05-29
feat: add monitoring RPCs

marcellorigotti:
# Pull Request

Closes: PRO-1274

## Checklist

## Summary

Added a new list of RPCs which will be used for monitoring purposes such that if something is going to change in the state-chain it is not breaking the compatibility with our monitoring system

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -29,7 +29,7 @@ use pallet_cf_swapping::SwapLegInfo;
 use sc_client_api::{BlockchainEvents, HeaderBackend};
 use serde::{Deserialize, Serialize};
 use sp_api::ApiError;
-use sp_core::{bounded_vec::BoundedVec, ConstU32, U256};
+use sp_core::U256;
 use sp_runtime::{
 	traits::{Block as BlockT, Header as HeaderT, UniqueSaturatedInto},
 	Permill,
@@ -37,12 +37,14 @@ use sp_runtime::{
 use state_chain_runtime::{
 	chainflip::{BlockUpdate, Offence},
 	constants::common::TX_FEE_MULTIPLIER,
+	monitoring_apis::{
+		AuthoritiesInfo, BtcUtxos, EpochState, ExternalChainsBlockHeight, FeeImbalance, FlipSupply,
+		LastRuntimeUpgradeInfo, MonitoringData, OpenDepositChannels, PendingBroadcasts,
+		PendingTssCeremonies, RedemptionsInfo,
+	},
 	runtime_apis::{
-		AuthoritiesInfo, BoostPoolDepth, BoostPoolDetails, BrokerInfo, BtcUtxos, CustomRuntimeApi,
-		DispatchErrorWithMessage, EpochState, EventFilter, ExternalChainsBlockHeight,
-		FailingWitnessValidators, FeeImbalance, FlipSupply, LastRuntimeUpgradeInfo,
-		LiquidityProviderInfo, MonitoringData, MonitoringRuntimeApi, OpenDepositChannels,
-		PendingBroadcasts, PendingTssCeremonies, RedemptionsInfo, ValidatorInfo,
+		BoostPoolDepth, BoostPoolDetails, BrokerInfo, CustomRuntimeApi, DispatchErrorWithMessage,
+		EventFilter, FailingWitnessValidators, LiquidityProviderInfo, ValidatorInfo,
 	},
 	NetworkFee,
 };
@@ -51,7 +53,7 @@ use std::{
 	marker::PhantomData,
 	sync::Arc,
 };
-
+pub mod monitoring;
 mod type_decoder;
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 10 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -84,7 +85,18 @@ pub enum RotationPhase<T: Config> {
 	NewKeysActivated(RuntimeRotationState<T>),
 	SessionRotating(Vec<ValidatorIdOf<T>>, <T as Chainflip>::Amount),
 }
-
+impl<T: pallet::Config> Display for RotationPhase<T> {
```
dandanlen, line 88: Better to make this std-gated and use core::fmt instead of importing via scale.

```suggestion
#[cfg(feature = std)]
impl<T: pallet::Config> core::fmt::Display for RotationPhase<T> {
	fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
	//...
	}
}
```

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1744,6 +1851,212 @@ where
...
+			.cf_authorities(self.unwrap_or_best(at))
+			.map_err(to_rpc_error)
+	}
+	fn cf_external_chains_block_height(
+		&self,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<ExternalChainsBlockHeight> {
+		self.client
+			.runtime_api()
+			.cf_external_chains_block_height(self.unwrap_or_best(at))
```
dandanlen, line 1936: Most of these methods are just passed through to the runtime api.

We could write a simple macro that reduces the repetition here. 

Not sure about the exact syntax but something like this:

```rust
macro_rules! pass_through {
	$( $name:ident -> $result_type:ty ),+
} {
	$(
		fn $name(&self, at: Option<state_chain_runtime::Hash>) -> RpcResult<$result_type> {
			self.client
				.runtime_api()
				.$name(self.unwrap_or_best(at))
				.map_err(to_rpc_error)
		}
	)+
}
```

Would let us write this for all of the simple cases: 

```
pass_through! {
	cf_authorities -> AuthoritiesInfo,
	cf_external_chains_block_height -> ExternalChainsBlockHeight,
	// ...
}
```

dandanlen (commented):
Looking good, just some refactoring suggestions.

I would also move the monitoring-specific code into dedicated modules/files just because the main custom-rpc/lib file is getting a bit unwieldly. Same for the decl_runtime and ideally the impl_runtime (if possible - not sure the macros support it). It would be nice to start keeping the individual modules separate instead of lumping everything in the same file. 

I want to take a closer look at the individual metrics, will do so next week.

dandanlen (approved):

state-chain/runtime/src/lib.rs
```diff
@@ -1870,6 +1876,166 @@ impl_runtime_apis! {
...
+			}
+		}
+		fn cf_pending_tss_ceremonies() -> PendingTssCeremonies {
+			PendingTssCeremonies {
+				evm: pallet_cf_threshold_signature::PendingCeremonies::<Runtime, EvmInstance>::iter().collect::<Vec<_>>().len() as u32,
+				bitcoin: pallet_cf_threshold_signature::PendingCeremonies::<Runtime, BitcoinInstance>::iter().collect::<Vec<_>>().len() as u32,
+				polkadot: pallet_cf_threshold_signature::PendingCeremonies::<Runtime, PolkadotInstance>::iter().collect::<Vec<_>>().len() as u32,
+			}
+		}
+		fn cf_pending_swaps() -> u32 {
```
dandanlen, line 1970: nit: could be cf_pending_swap_count

state-chain/runtime/src/lib.rs
```diff
@@ -1870,6 +1876,166 @@ impl_runtime_apis! {
...
+			let swaps: Vec<_> = pallet_cf_swapping::SwapQueue::<Runtime>::iter().collect();
+			swaps.iter().fold(0u32, |acc, elem| acc + elem.1.len() as u32)
+		}
+		fn cf_open_deposit_channels() -> OpenDepositChannels {
```
dandanlen, line 1974: Same - could be cf_open_deposit_channels_count

state-chain/runtime/src/lib.rs
```diff
@@ -1870,6 +1876,166 @@ impl_runtime_apis! {
...
+				arbitrum: arb_channels,
+			}
+		}
+		fn cf_fee_imbalance() -> FeeImbalance {
+			let eth = pallet_cf_ingress_egress::WithheldTransactionFees::<Runtime, EthereumInstance>::iter().fold(0, |acc, elem| acc+ elem.1) -
+				pallet_cf_broadcast::TransactionFeeDeficit::<Runtime, EthereumInstance>::iter().fold(0, |acc, elem| acc + elem.1);
+			let dot = pallet_cf_ingress_egress::WithheldTransactionFees::<Runtime, PolkadotInstance>::iter().fold(0, |acc, elem| acc+ elem.1) -
+				pallet_cf_broadcast::TransactionFeeDeficit::<Runtime, PolkadotInstance>::iter().fold(0, |acc, elem| acc + elem.1);
+			let btc = pallet_cf_ingress_egress::WithheldTransactionFees::<Runtime, BitcoinInstance>::iter().fold(0, |acc, elem| acc+ elem.1) -
+				pallet_cf_broadcast::TransactionFeeDeficit::<Runtime, BitcoinInstance>::iter().fold(0, |acc, elem| acc + elem.1);
```
dandanlen, line 1993: Does this make sense for bitcoin?

marcellorigotti: Oh yeah forgot for btc doesn't work, removed!

PR #4925, opened 2024-05-31
chore: log chainflip header in runner rather than in dylib

kylezs:
# Pull Request

## Checklist

## Summary

This would log inside the dylib, which is strange when an upgrade occurs and it's handed over to the new dylib, as this would print again. This PR just prints it on binary startup, rather than dylib startup.

### DIFF
--- a/engine-runner-bin/src/main.rs
+++ b/engine-runner-bin/src/main.rs
@@ -31,6 +31,17 @@ mod new {
 // 4. If this new version completes, then we're done. The engine should be upgraded before this is
 //    the case.
 fn main() -> anyhow::Result<()> {
+	println!(
+		"
+		 ██████╗██╗  ██╗ █████╗ ██╗███╗   ██╗███████╗██╗     ██╗██████╗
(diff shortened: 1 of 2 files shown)


AlastairHolmes:
But I'm not sure this is a problem. Also this removes the log from the node.

kylezs:
>  Also this removes the log from the node.

`print_start_and_end` is only called in the engine.

Yeah, wouldn't say it's a problem. But it makes more sense to me to have it log when the binary is started and not when the inner dylibs start (which is more an implementation detail).

AlastairHolmes:
> >  Also this removes the log from the node.
> 
> `print_start_and_end` is only called in the engine though?
> 
> Yeah, wouldn't say it's a problem. But it makes more sense to me to have it log when the binary is started and not when the inner dylibs start (which is more an implementation detail).

Guess someone removed it from the node, not sure why though Ill have a look

martin-chainflip (approved):

AlastairHolmes:
I guess I just disagree with this change.

kylezs:
not a strong opinion tbh - will close to clear up some github space.

PR #4928, opened 2024-05-31
fix/refresh-replay-protection

Janislav:
# Pull Request

Closes: PRO-1396

## Checklist

## Summary

#### Non-Breaking changes


state-chain/chains/src/dot.rs
```diff
@@ -1119,6 +1123,16 @@ mod test_polkadot_extrinsics {
...
+
+		let new_replay_protection = PolkadotReplayProtection {
+			nonce: 13,
+			signer: account_id_1,
+			genesis_hash: Default::default(),
+		};
+
+		extrinsic_builder.refresh_replay_protection(new_replay_protection.clone());
+
+		assert_eq!(new_replay_protection, extrinsic_builder.replay_protection);
```
dandanlen, line 1135: Careful - this test has the #[ignore] annotation so is never actually run.

I think it was used just to output some test payloads (that's why there are all the print statements).

Better to create a dedicated test for refreshing the replay protection. 

Also we should test that the signature is removed:

```suggestion
		assert_eq!(new_replay_protection, extrinsic_builder.replay_protection);
		assert!(!extrinsic_builder.is_signed());
```

state-chain/chains/src/dot.rs
```diff
@@ -484,6 +484,10 @@ impl PolkadotExtrinsicBuilder {
 	pub fn is_signed(&self) -> bool {
 		self.signature.is_some()
 	}
+
+	pub fn refresh_replay_protection(&mut self, replay_protection: PolkadotReplayProtection) {
+		self.replay_protection = replay_protection;
```
dandanlen, line 489: We should also remove the old signature if there is one.

state-chain/chains/src/evm/api.rs
```diff
@@ -168,6 +172,42 @@ impl<C: EvmCall> EvmTransactionBuilder<C> {
...
+	pub fn transaction_out_id(&self) -> <EvmCrypto as ChainCrypto>::TransactionOutId {
+		let sig_data = self.sig_data.expect("Unsigned transaction_out_id is invalid.");
+		SchnorrVerificationComponents {
+			s: sig_data.sig.into(),
+			k_times_g_address: sig_data.k_times_g_address.into(),
+		}
+	}
+
+	pub fn refresh_replay_protection(&mut self, replay_protection: EvmReplayProtection) {
+		self.replay_protection = replay_protection;
```
dandanlen, line 209: Same here, we need to remove the signature.

state-chain/chains/src/evm/benchmarking.rs
```diff
@@ -1,19 +1,66 @@
...
+	fn key_manager_address() -> Address {
+		unimplemented!()
+	}
+
+	fn chain_id() -> api::EvmChainId {
+		unimplemented!()
+	}
+
+	fn next_nonce() -> u64 {
+		unimplemented!()
```
dandanlen, line 60: Will this not cause the benchmarks to panic? 

Should we not return `BenchmarkValue::benchmark_value()` for each of these instead of `unimplemented!()`?

state-chain/chains/src/mocks.rs
```diff
@@ -377,6 +382,12 @@ impl<C, Call> MockTransactionBuilder<C, Call> {
 	pub fn set_requires_refresh() {
 		REQUIRES_REFRESH.with(|is_valid| *is_valid.borrow_mut() = true)
 	}
+	pub fn set_refreshed_replay_protection() {
+		REFRESHED_REPLAY_PROTECTION.with(|is_valid| *is_valid.borrow_mut() = false)
+	}
+	pub fn get_set_refreshed_replay_protection_state() -> bool {
```
dandanlen, line 388:
```suggestion
	pub fn get_refreshed_replay_protection_state() -> bool {
```

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -703,22 +704,27 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 	}
 
 	/// Signs a API call, use `Call::on_signature_ready` as the callback, and returns the signature
 	/// request ID.
 	fn threshold_sign(
-		api_call: <T as Config<I>>::ApiCall,
+		mut api_call: <T as Config<I>>::ApiCall,
 		broadcast_id: BroadcastId,
 		should_broadcast: bool,
+		refresh_replay_protection: bool,
```
dandanlen, line 716: This value is always false except when used from the governance call.

I would suggest we simple move the conditional into the body the extrinsic instead, then we don't need to thread the boolean value through all the other functions where it isn't used.

dandanlen (commented):
Mostly LGTM, just one issue (need to reset the signature) and some suggestions.

state-chain/chains/src/evm/benchmarking.rs
Janislav: No - I tried it. But yeah, I agree, returning a proper value makes sense.

dandanlen (approved):
LGTM thanks. I added one extra commit that removes some expects - these were there already, but now that we have an additional code path for signing/sending transactions, I feel a little less confident about the assertions. I also remove some unused code.

dandanlen:
I'll avoid merging this until the underlying PR is reviewed.

PR #4946, opened 2024-06-07
WIP: feat: return deposit tx_hashes where possible

kylezs:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

Where possible return the deposit tx hash to make it easier for frontends / integrations to show the end-to-end flow.


niklasnatter (approved):

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -556,7 +556,7 @@ mod tests {
 						deposit_address: polkadot_account_id,
 						amount: 100u128,
 						asset: cf_chains::assets::dot::Asset::Dot,
-						deposit_details: (),
+						deposit_details: 1,
```
niklasnatter, line 559: i think there is no (external chain) block number on the `DepositFinalised`/`DepositBoosted`, so it would be nice to have that in the deposit detail (or on the event)

state-chain/chains/src/evm.rs
```diff
@@ -21,6 +21,14 @@ use serde::{Deserialize, Serialize};
 use sp_core::ConstBool;
 use sp_std::{convert::TryFrom, str, vec};
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, Default)]
+pub struct DepositDetails {
+	// In the case of EVM Native Deposits (ETH or arbETH), because we need to detect ingresses by
+	// checking balances rather than using events, there can be more than one hash associated with
+	// a deposit amount.
+	pub tx_hashes: Option<Vec<H256>>,
```
niklasnatter, line 29: would it makes sense to add the (external) block number here for the case where we dont have a tx hash?

state-chain/chains/src/dot.rs
```diff
@@ -331,7 +331,7 @@ impl Chain for Polkadot {
 	type ChainAsset = assets::dot::Asset;
 	type DepositFetchId = PolkadotChannelId;
 	type DepositChannelState = PolkadotChannelState;
-	type DepositDetails = ();
+	type DepositDetails = PolkadotExtrinsicIndex;
```
niklasnatter, line 334: i think there is no (external chain) block number on the `DepositFinalised`/`DepositBoosted`, so it would be nice to have that in the deposit detail (or on the event)

state-chain/chains/src/evm.rs
kylezs: We already have that as part of the extrinsic

state-chain/chains/src/dot.rs
kylezs: yeah it's on the extrinsic but not the event... probably does make sense to have it on the event

niklasnatter: yes it would be nice to have it on the event (at least for polkadot, maybe in general)

niklasnatter (approved):

dandanlen (approved):

PR #4952, opened 2024-06-11
chore (bouncer): add no-floating-promises lint

kylezs:
# Pull Request

## Checklist

## Summary

After adding the lint, fixed all the places we have them. Please review the places I've said to explicitly ignore.

### DIFF
--- a/bouncer/shared/utils.ts
+++ b/bouncer/shared/utils.ts
@@ -276,8 +276,9 @@ function getCachedSubstrateApi(defaultEndpoint: string) {
                 setTimeout(() => {
                   if (connections === 0) {
                     api = undefined;
-                    /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
-                    Reflect.get(target, 'disconnect', receiver).call(target);
+                    Reflect.get(target, 'disconnect', receiver)
+                      .call(target)
+                      .catch(() => null);
                   }
                 }, 5_000).unref();
               }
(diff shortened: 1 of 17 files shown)


bouncer/shared/utils.ts
```diff
@@ -276,6 +276,7 @@ function getCachedSubstrateApi(defaultEndpoint: string) {
                 setTimeout(() => {
                   if (connections === 0) {
                     api = undefined;
+                    /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
                     Reflect.get(target, 'disconnect', receiver).call(target);
```
kylezs, line 280: @acdibble what's the idea behind this?

bouncer/shared/utils/substrate.ts
```diff
@@ -149,6 +149,7 @@ export function observeEvent<T = any>(
   const it = subscribeHeads({ chain, finalized });
 
   controller?.signal.addEventListener('abort', () => {
+    /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
     it.return();
```
kylezs, line 153: @acdibble same here, what's the idea behind ths?

bouncer/shared/utils.ts
acdibble, line 280: I'd suggest adding a dummy catch, because this is in a `setTimeout` there's no way to handle it throwing. The idea here is that between tests we want to wait 5 seconds to actually disconnect to prevent all the subscribers from unsubscribing and then immediately opening a new subscription
```suggestion
                    Reflect.get(target, 'disconnect', receiver).call(target).catch(() => null);
```

bouncer/shared/utils/substrate.ts
acdibble: This should be fine, this function should never throw afaik.

acdibble (approved):

PR #4960, opened 2024-06-13
chore: don't panic in scheduled swaps

j4m1ef0rd:
# Pull Request

Closes: PRO-1410

## Checklist

## Summary

Changed the unwrap to an `unwrap_or_default` if the `get_scheduled_swap_legs` returns error. This is fine for now, but when the more complex FoK logic is in, it will not mirror the same behaviour as the actual `process_swaps_for_block` function and may miss swaps.

### DIFF
--- a/state-chain/runtime/src/lib.rs
+++ b/state-chain/runtime/src/lib.rs
@@ -1745,7 +1745,8 @@ impl_runtime_apis! {
 				let execute_at = core::cmp::max(block, current_block.saturating_add(1));
 
 				let swaps: Vec<_> = swaps_for_block.iter().filter(|swap| swap.from == base_asset || swap.to == base_asset).cloned().collect();
-				Swapping::get_scheduled_swap_legs(swaps, base_asset).unwrap().into_iter().map(move |swap| (swap, execute_at))
+				// TODO: Use the FoK logic when this fails instead of unwrap_or_default.
+				Swapping::get_scheduled_swap_legs(swaps, base_asset).unwrap_or_default().into_iter().map(move |swap| (swap, execute_at))
 			}).collect()
 		}
 
--- a/state-chain/runtime/src/lib.rs
+++ b/state-chain/runtime/src/lib.rs
@@ -44,7 +44,7 @@ use cf_chains::{
 	Arbitrum, Bitcoin, CcmChannelMetadata, DefaultRetryPolicy, ForeignChain, Polkadot, Solana,
 	TransactionBuilder,
 };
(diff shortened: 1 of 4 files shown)


state-chain/runtime/src/lib.rs
```diff
@@ -1745,7 +1745,8 @@ impl_runtime_apis! {
 				let execute_at = core::cmp::max(block, current_block.saturating_add(1));
 
 				let swaps: Vec<_> = swaps_for_block.iter().filter(|swap| swap.from == base_asset || swap.to == base_asset).cloned().collect();
-				Swapping::get_scheduled_swap_legs(swaps, base_asset).unwrap().into_iter().map(move |swap| (swap, execute_at))
+				// TODO: Use the FoK logic when this fails instead of unwrap_or_default.
```
kylezs, line 1748: can we create a ticket for this and put it in the FoK project

msgmaxim: Yeah, we should return the swaps for the first leg.

dandanlen: PRO-1437

kylezs: > [PRO-1437](https://linear.app/chainflip/issue/PRO-1437)

and in the meantime we will return the first leg, or nothing? Either way, I think that ticket should be done soon anyway

dandanlen: Not sure - yes first leg is better than nothing. But it would be good to communicate somehow that the second legs are missing.

j4m1ef0rd: I did think of that but it seemed quite complicated. @msgmaxim do you think it's doable?

msgmaxim: I like the suggestion in [PRO-1437](https://linear.app/chainflip/issue/PRO-1437). Let's use price estimates for calculating the second leg. (I'm not yet sure whether we shoud alwasy do it this way, or only fall back when the simulation fails; both are estimates anyway, so I'm leaning towards the simpler approach).

I don't mind merging this (after we fix fix not communicating the first leg). However, we might save a bit of time if we implement [PRO-1437](https://linear.app/chainflip/issue/PRO-1437) directly.

dandanlen: I agree the simpler approach is better. LPs are unlikely to blindly use our output values directly anyway: they will (or at least, they *should*) decide on their own price and apply that.

j4m1ef0rd:
rebased onto the FoK PR. Will create a new PR.

PR #4961, opened 2024-06-13
chore: handle linux dylib copying 🖨️

ahasna:


localnet/manage.sh
```diff
@@ -95,6 +97,14 @@ build-localnet() {
   mkdir -p $CHAINFLIP_BASE_PATH
   touch $DEBUG_OUTPUT_DESTINATION
 
+  if [ "$OS_TYPE" == "Linux" ]; then
```
kylezs, line 100: I would leave a comment saying that these `.so` files are built on ubuntu 22 - so if they don't work, it's clear why

kylezs (approved):

PR #4962, opened 2024-06-13
feature/PRO-1378/validator-gas-refund

Janislav:
# Pull Request

Closes: PRO-1378, PRO-1494, PRO-1496

## Checklist

## Summary

Adds a new pallet that is responsible for organizing and coordinating refunds. Also takes care of the migrations from the storage items in other pallets where we bookkeep the fees until now.

#### Non-Breaking changes


state-chain/traits/src/lib.rs
```diff
@@ -952,3 +954,16 @@ pub trait AssetConverter {
 pub trait IngressEgressFeeApi<C: Chain> {
 	fn accrue_withheld_fee(asset: C::ChainAsset, amount: C::ChainAmount);
 }
+
+pub trait Refunding<C: Chain> {
+	fn record_gas_fees(account_id: C::ChainAccount, asset: C::ChainAsset, amount: C::ChainAmount);
+	fn with_held_transaction_fees(asset: C::ChainAsset, amount: C::ChainAmount);
```
dandanlen, line 960: It might be simpler to define the trait in terms of `AssetAmount` and `cf_primitives::Asset` instead of making it chain-generic. Then we can just implement it on the pallet, and there's no need to define a macro for all the trait defs. We can just use `into()` where we call the methods.

state-chain/pallets/cf-refunding/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		let chains = WithheldTransactionFees::<T>::iter_keys().collect::<Vec<_>>();
+
+		for chain in chains {
+			let mut available_funds = WithheldTransactionFees::<T>::get(chain);
+			// Integrity check before we start refunding
+			if available_funds < RecordedFees::<T>::get(chain).unwrap().values().sum() {
+				log::warn!(
+                    "Integrity check for refunding failed for asset {:?}. Refunding will be skipped.", chain);
```
dandanlen, line 120: I don't think we should skip in this case: it's possible that we (currently) don't enough funds to fully refund what we owe, but we still want to process refunds if we can.

dandanlen (commented):
I know this is still in draft, but still, some early-bird comments.

Janislav:
Will extend testing a bit tomorrow, but in general I think It's ready to review.

dandanlen (commented):
Still a fair few issues - I've been changing most of this as I was reviewing. Will push changes. I've also created some Linear sub-issues inside PRO-1378 to track some additional things.

I would like us to use this pallet for tracking/withdrawing/swapping asset balances more generally (for example, broker/lp balances), so will also rename it accordingly.

One thing we should address before pushing this to mainnet is that we should consider adding a limit to the number of refunds that are processed. There are ~250 validators in the refund list currently. I checked with @albert-llimos and that's quite a big batch. IMO we should (a) add a limit to the number of batches and (b) add a minimum refund amount that is some multiple of the egress fee. I added Linear issues for both of these.

state-chain/cf-integration-tests/src/new_epoch.rs
```diff
@@ -124,12 +134,26 @@ fn epoch_rotates() {
...
+				);
+				Refunding::withhold_transaction_fee(ForeignChain::Ethereum, 100);
+				Refunding::record_gas_fee(
+					cf_chains::ForeignChainAddress::Dot(PolkadotAccountId::from_aliased(
+						DOT_AGG_KEY,
+					)),
+					ForeignChain::Polkadot,
+					10,
+				);
+				Refunding::withhold_transaction_fee(ForeignChain::Polkadot, 10);
```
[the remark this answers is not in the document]
dandanlen: Update: there's already a Linear issue for bouncer tests.

state-chain/pallets/cf-asset-balances/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+		assert!(recorded_fees_eth.is_empty());
+		assert!(recorded_fees_arb.is_empty());
+
+		assert_eq!(WithheldAssets::<Test>::get(ForeignChain::Ethereum.gas_asset()), 0);
+		assert_eq!(WithheldAssets::<Test>::get(ForeignChain::Arbitrum.gas_asset()), 0);
+	});
+}
+
+#[test]
+fn skip_refunding_if_safe_mode_is_disabled() {
```
kylezs, line 69:
```suggestion
fn skip_refunding_if_safe_mode_is_enabled() {
```

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+			Self::deposit_event(Event::VaultDeficitDetected {
+				chain,
+				amount_owed: *amount_owed,
+				available: *available,
+			});
+		}
+		let amount_reconciled = match chain {
+			ForeignChain::Ethereum | ForeignChain::Arbitrum => match owner {
+				ExternalOwner::Account(address) =>
+					if *amount_owed > *available {
```
kylezs, line 162: we've already checked this above

state-chain/runtime/src/lib.rs
```diff
@@ -1140,14 +1161,14 @@ type PalletMigrations = (
 	pallet_cf_broadcast::migrations::PalletMigration<Runtime, PolkadotInstance>,
 	pallet_cf_broadcast::migrations::PalletMigration<Runtime, BitcoinInstance>,
 	pallet_cf_broadcast::migrations::PalletMigration<Runtime, ArbitrumInstance>,
-	// pallet_cf_broadcast::migrations::PalletMigration<Runtime, SolanaInstance>,
+	pallet_cf_broadcast::migrations::PalletMigration<Runtime, SolanaInstance>,
```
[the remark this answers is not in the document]
ramizhasan111: yeah, i think we should uncomment all solana migrations.
I guess the two uncommented were for the pallets whose new migrations were written in this PR but yeah we should uncomment all.

state-chain/pallets/cf-asset-balances/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::{mock::*, ExternalOwner, Liabilities, Pallet, WithheldAssets};
+
+fn payed_gas(chain: ForeignChain, amount: AssetAmount, account: ForeignChainAddress) {
+	Pallet::<Test>::record_liability(account, chain.gas_asset(), amount);
+	Pallet::<Test>::withhold_assets(chain.gas_asset(), amount);
+}
+
+fn assert_egress(
+	number_of_egresses: usize,
+	maybe_additional_conditions: Option<fn(egresses: Vec<MockEgressParameter<AnyChain>>)>,
```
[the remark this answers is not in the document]
Janislav: For now. Doesn't hurt to have it more configurable.

state-chain/pallets/cf-asset-balances/src/lib.rs
Janislav: And we need to somehow know it here again. The only solution would be is remembering the result of the condition and then using this instead, but I don't really see big benefit.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+				.map_err(Into::into)?;
+		Self::deposit_event(Event::RefundScheduled {
+			egress_id: egress_details.egress_id,
+			destination: address,
+			amount,
+		});
+		Ok(())
+	}
+
+	/// Reconciles the amount owed with the amount available for distribution.
```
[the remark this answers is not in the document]
Janislav: It's not really fitting for `trigger_reconciliation`. I think we just need to add one for `trigger_reconciliation` would not remove this one. Doesn't hurt to have some explaining comments if they are already there.

state-chain/pallets/cf-asset-balances/src/tests.rs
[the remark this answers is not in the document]
kylezs: I'd argue it does hurt, it's harder to understand what's going on. Wouldn't block the PR if you really want to leave it, but it is more complex than necessary.

state-chain/pallets/cf-asset-balances/src/lib.rs
kylezs: Why do we need to know it here again? If this condition were true, we'd have already thrown an event above?

Janislav: Because we want to fire this event for any chain under this condition. For the same condition we decide if we want to send funds but only for specific chain (ETH, ARB). We need to know this because of two different reasons at two different areas of the code.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		for chain in ForeignChain::iter() {
+			let owed_assets = Liabilities::<T>::take(chain.gas_asset());
+			WithheldAssets::<T>::mutate(chain.gas_asset(), |total_withheld| {
+				let mut owed_assets = owed_assets.into_iter().collect::<Vec<_>>();
+				owed_assets.sort_by_key(|(_, amount)| core::cmp::Reverse(*amount));
+
+				for (destination, amount) in &mut owed_assets {
+					debug_assert!(*amount > 0);
```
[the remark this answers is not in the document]
Janislav: Dan added this, but I think the idea this that there can not be a key/validator for which we recorded that it has paid 0 fees. That should be logical, not possible. If it is, it's a bug. So I guess a debug_assert makes sense here.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -201,7 +203,7 @@ impl<T: Config> Pallet<T> {
 					},
 				other => {
 					log_or_panic!(
-						"{:?} Liabilities are not supported for chain {:?}.",
+						"{:?} Liabilities are only supported for ExternalOwner::Account types, not {:?} for EVM chains.",
 						other,
 						chain
```
kylezs, line 208: These variables don't match the message now

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+impl<T: Config> AssetWithholding for Pallet<T> {
+	fn withhold_assets(asset: Asset, amount: AssetAmount) {
+		WithheldAssets::<T>::mutate(asset, |fees| {
+			fees.saturating_accrue(amount);
+		});
+	}
+
+	#[cfg(feature = "try-runtime")]
```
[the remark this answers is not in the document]
Janislav: I think because we need to access the storage in a migration of another pallet for the verification (also the reason for the try-runtime feature). I think that's the easiest we to do it also in terms of removing the references to the other pallet after the migration is done.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+mod benchmarking;
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
+
+pub const PALLET_VERSION: StorageVersion = StorageVersion::new(0);
+
+// TODO: Figure out adequate values for these.
+pub const MAX_ETH_REFUND_PER_EPOCH: usize = 25;
```
Janislav, line 29: @kylezs and @albert-llimos happy to get some opinion on what would be a reasonable value for this.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+pub const MAX_ARB_REFUND_PER_EPOCH: usize = 10;
```
Janislav, line 30: same here

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+pub const REFUND_FEE_MULTIPLE: AssetAmount = 100;
```
Janislav, line 31: and here

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+						chain.gas_asset(),
+						owed_assets.into_iter().collect::<BTreeMap<_, _>>(),
+					);
+				}
+			});
+		}
+	}
+
+	pub fn vault_imbalance(asset: Asset) -> VaultImbalance<AssetAmount> {
+		let owed = Liabilities::<T>::get(asset).values().sum::<u128>();
```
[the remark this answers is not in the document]
Janislav: I think it's the same reason again... We just need this for the migration and then want to remove it after the migration is done. Most likely we will end up with an inline solution again then, because we argue that it's only used a single time inside the pallet itself and there is no point in having it implemented in the trait or even in an own method.

state-chain/pallets/cf-asset-balances/src/lib.rs
kylezs: Why do we need a max? If we have the ability to refund them and we've kept the funds aside, why not do it?

state-chain/pallets/cf-asset-balances/src/lib.rs
kylezs: Should be `MAX_ARB_ETH_REFUND_PER_EPOCH` (if we have it at all - see above question)

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+			Ok(ScheduledEgressDetails { egress_id, .. }) => {
+				Self::deposit_event(Event::RefundScheduled {
+					egress_id,
+					destination: address,
+					amount,
+				});
+				Ok(())
+			},
+			Err(err) => {
+				Self::deposit_event(Event::RefundSkipped { reason: err });
```
kylezs, line 157: I don't think this is true right? If we've scheduled it, it's already been scheduled. This error bubbles up to just a break out of a loop. We're not in a transactional block of code here.

state-chain/pallets/cf-asset-balances/src/lib.rs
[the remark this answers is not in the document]
kylezs: Ok, we should create a ticket to not forget this then, can label it with "1.6" so we address it before 1.6 is released.

state-chain/pallets/cf-asset-balances/src/lib.rs
kylezs: I think 100 is reasonable

state-chain/pallets/cf-asset-balances/src/lib.rs
Janislav: Primarily because we assume that for the first refund, the batch could become quite massive ~250. I think we fear a bit the unknowns of this.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+					match Self::reconcile(chain, destination, amount, total_withheld) {
+						Ok(_) =>
+							if *total_withheld == 0 {
+								break;
+							},
+						Err(_) => break,
+					}
+				}
+
+				owed_assets.retain(|(_, amount)| *amount > 0);
```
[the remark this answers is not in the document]
Janislav: That has been added by Dan, but I think the idea is to clean up the map entries after we refunded. I don't think we want to take over the old empty entries over to the next refund round and then clean it up right before we start the new refund. Apart from that, I don't think it would make a difference in terms of removing the debug_assert. With the current implementation it should also actually never happen, and if it does, we know there is a logical problem.

state-chain/pallets/cf-asset-balances/src/lib.rs
Janislav: Ahhh good catch. You mean, we have to do this inside a `with_storage_layer` or `with_transaction` to make this work.

state-chain/pallets/cf-asset-balances/src/lib.rs
kylezs: Ooh, it's not the MAX_ETH_REFUND, but it's `MAX_REFUNDED_VALIDATORS_ETH_PER_EPOCH` or something like that. 

Even still, if we're only refunding 25, then it'll take over 2 weeks to refund everyone, seems too long. Keen to hear @albert-llimos 's opinion here on the batch size. Ideally we can just submit everything in one batch but maybe there is a practical limitation there.

albert-llimos: Dan looped me into the conversation here: https://github.com/chainflip-io/chainflip-backend/pull/4962#pullrequestreview-2160793703.
The batch size for ~250 transfers might be problematic for gas limits. According to some tests we shouldn't be close to the gas limit but it's still "risky" to do that massive tx.
It looked like the best solution was to either limit it (like done here) or to use the new logic added on Solana to limit the amount of transfers that can be put together in a single `AllBatch`. My preference was to cherry pick that logic for Solana (small) into here and set it to a reasonable amount of transfers so it would automatically be split up. Especially since we want to include a limit for Ethereum anyway. That's what I last told to Dan  but I'm not sure if there was any other new information after that.
That would still be me preferred solution.

albert-llimos: FWIW, we should be able to easily get away with at least 50 or even 100 refunds in one go, so if we want to just increase that hardcoded number instead of adding new logic for simplicity, that's also fine. Then the refund will take a shorter time.

kylezs: Ok makes sense. In that case we definitely want to split it up. Makes sense. 
Not sure about splitting it per epoch though, only 25 per epoch would take forever. We could just construct all the batches up front and schedule them 10 SC blocks apart or something, rather than wait potentially over 2 weeks before the next validators can get their refund. 

Where can I see the Solana AllBatch limit code? If we can code share that would be nice.

kylezs: Ok that looks fairly involved. Given the size of this PR already, I'd rather make the batch 50 and create an issue to address it in a better way.

Janislav: I think the design would work quite well and easy for the future use-case where we don't collect fees other more than half a year. I would not make the implementation unnecessary complicated for a single edge case. I think an epoch is 3 days, at the moment, right? If you say 100 should be doable, and we go with let's say 75 per epoch (to be safe) all current validators should be refunded in about 10 days, right? Would sound acceptable to me.

kylezs: https://linear.app/chainflip/issue/PRO-1513/better-refund-batching

Janislav: If I think more about it, I think it would make sense to split it up by blocks over the entire epoch. I don't think we should hit 75-100 on a single epoch right now, but it's actually not so unrealistic (should check the numbers) especially when we expect volume to increase in the future.

kylezs: Yeah there are plenty of ways we could do it, but I think we should think about it separately. Doing 50 or so per epoch for now is good enough, and we can continue thinking about it after this is in.

state-chain/pallets/cf-asset-balances/src/lib.rs
Janislav: Yes, we need one. @albert-llimos tries to figure out a number and the result will most likely surprise you ^^

albert-llimos: If we go for 50 as above for Ethereum let's do the same for Arbitrum. The way that Arbitrum's gas consumption works is very weird and I'd rather be on the same side. I know maybe you were expecting we would 10x that number in Arbitrum, sorry to dissapoint.

state-chain/pallets/cf-asset-balances/src/lib.rs
Janislav: I implemented it now, the same way we do it when we burn flip.

state-chain/pallets/cf-asset-balances/src/lib.rs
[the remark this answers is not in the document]
Janislav: https://linear.app/chainflip/issue/PRO-1515/remove-migration-feature-gated-functions-after-15

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -136,15 +137,19 @@ impl<T: Config> Pallet<T> {
...
-		match T::EgressHandler::schedule_egress(chain.gas_asset(), amount, address.clone(), None)
-			.map_err(Into::into)
-			.and_then(|result @ ScheduledEgressDetails { egress_amount, fee_withheld, .. }| {
-				if egress_amount < REFUND_FEE_MULTIPLE * fee_withheld {
-					Err(Error::<T>::RefundAmountTooLow.into())
-				} else {
-					Ok(result)
-				}
-			}) {
+		match with_storage_layer(|| {
```
kylezs, line 139: we can add a unit test for this case too

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -26,9 +26,8 @@ mod tests;
 
 pub const PALLET_VERSION: StorageVersion = StorageVersion::new(0);
 
-// TODO: Figure out adequate values for these.
-pub const MAX_ETH_REFUND_PER_EPOCH: usize = 50;
-pub const MAX_ARB_REFUND_PER_EPOCH: usize = 10;
+pub const MAX_REFUNDED_VALIDATORS_ETH_PER_EPOCH: usize = 50;
+pub const MAX_REFUNDED_VALIDATORS_ARB_PER_EPOCH: usize = 50;
```
kylezs, line 30: ARB_ETH

state-chain/pallets/cf-asset-balances/src/lib.rs
kylezs: to be clear, I mean the case that if we can't process, that it reverts, there's no egress scheduled, and we still have the amounts etc. in storage.

Janislav: That would be nice, but I think it doesn't make a lot of sense. The only storage that is manipulated here is in the ingress/egress pallet, but in the case of a unit-test it's just the mock that is not using real storage, so I don't think there is a lot to test here. There is already a unit test called `do_not_refund_if_amount_is_too_low` where we check the event and the storage inside the pallet, but that has actually nothing to do with the `with_storage_layer` functionality. I added a check that verifies that the egress mock remains 0.

kylezs: Having a deeper look it is testing the case, but accidentally - and is not testing another case - the case where the schedule_egress itself fails - so we should add one for that case. 

Just to be clear, there are two cases for failure here:
1. The egress amount is less than or equal to the fee (fails inside schedule egress) - though because the mock isn't perfect we would instead just do a `MockEgressHandler::return_failure(true)` to emulate this.
2. The egress *could* be scheduled but we deem it not worthwhile since it's not 100x the amount - this is the case currently hit accidentally. It's accidentally hit because the test sets the fee to double the refund fee - which would normally fail inside the schedule_egress in the usual case.

kylezs: I see it's tested in another place `keep_fees_in_storage_if_egress_fails` so we should be good here actually 👍

kylezs (approved):

kylezs:
Nice one 🚀

PR #4964, opened 2024-06-14
fix: old dylib should be pulled from the previous release by tag

kylezs:
WIP: Also need to handle PRs to release branches.

# Pull Request

Closes: PRO-1443

## Checklist

## Summary

This fixes an issue where, if we had a release branch that had been tagged, causing release-berghain to run, then the next time we released off that branch, the "latest" release would actually be the previous patch release, which is not what we want.

Instead, if we're on release/1.4 we want to get the latest 1.3.x tag and use that.

Will need to cherry-pick this across to release too.

### DIFF
--- a/.github/workflows/_25_package.yml
+++ b/.github/workflows/_25_package.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   FORCE_COLOR: 1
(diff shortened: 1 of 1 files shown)


kylezs:
#4966

PR #4967, opened 2024-06-14
feat: RPC safe mode statuses

marcellorigotti:
# Pull Request

Closes: PRO-1332

## Checklist

## Summary

Added an RPC to expose the current safe mode status of out pallets

#### Non-Breaking changes


marcellorigotti:
Question for @dandanlen, is there a particular reason we weren't using serde (Serialize, Deserialize) on the pallets structs, I added it to avoid duplicating the PalletSafeMode struct for every pallet (Such that PalletSafeMode is what is returned by the RPC).

dandanlen:
> Question for @dandanlen, is there a particular reason we weren't using serde (Serialize, Deserialize) on the pall�ets structs, I added it to avoid duplicating the PalletSafeMode struct for every pallet (Such that PalletSafeMode is what is returned by the RPC).

I believe serde previously did not support `no-std`. Also I guess we just didn't need it...

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -166,6 +172,32 @@ pub enum EventFilter {
 	SystemOnly,
 }
 
+#[derive(Serialize, Deserialize, Encode, Decode, Eq, PartialEq, TypeInfo, Debug)]
+pub struct SafeModeStatuses {
+	pub swapping: pallet_cf_swapping::PalletSafeMode,
+	pub emissions: pallet_cf_emissions::PalletSafeMode,
```
dandanlen, line 178: This struct looks equivalent to safe_mode::RuntimeSafeMode. 

If so, better to re-use that definition rather than create another one. 

(we probably just have to add Serialize/Deserialize in the impl_safe_mode macro definition).

marcellorigotti:
> Sorry - a bit slow to re-review this. LGTM now, just one small improvement.
> 
> I'm not sure though if there's that much value in adding this rpc - it's effectively just returning the RuntimeSafeMode storage item... and it's doing so in such a way that it can break if there's a change to the datatype. It might be better to just read the storage item directly and decode it using the metadata (which the polkaJS client does anyway).

Yeah I agree on that, it's the same as reading the storage item, I think the idea was to have an RPC to avoid breaking compatibility without knowing. 
I'll ask the product team why they needed this 👍

dandanlen:
I don't have any great objections to adding this as long as we understand the limitations.

marcellorigotti:
Ok the idea is to expose this to external parties, and to avoid to use polkadot libraries to do so.
I'll implement the last suggested changes!

dandanlen (approved):

PR #4973, opened 2024-06-17
feat: generic sol token apicalls

ramizhasan111:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

#### Non-Breaking changes


state-chain/chains/src/sol/api.rs
```diff
@@ -60,35 +71,56 @@ pub trait SolanaEnvironment:
...
-					SolanaTransactionBuildingError::CannotLookupTokenVaultAssociatedTokenAccount,
-				SolanaEnvAccountLookupKey::TokenVaultPdaAccount =>
-					SolanaTransactionBuildingError::CannotLookupTokenVaultPdaAccount,
 			},
 		)
 	}
 
+	fn lookup_account_retain(
+		key: SolanaEnvAccountLookupKey,
+		sol_environment: &mut HashMap<SolanaEnvAccountLookupKey, SolAddress>,
```
dandanlen, line 80: Maybe this hasn't been linked with the runtime yet, but i'm pretty sure HashMaps are not supported. Prefer to use BTreeMap.

dandanlen: Also, not quite sure what the purpose of this is - I guess some sort of cache? If so then I would suggest making it an implementation detail rather than baking it into the trait definition.

But in fact I would suggest just removing this - I think substrate already caches storage reads. And there is a separate issue to change the environment lookups. If necessary we can tackle any optimisations then.

state-chain/chains/src/sol/api.rs
```diff
@@ -60,35 +71,56 @@ pub trait SolanaEnvironment:
...
 		<Self as ChainEnvironment<AllNonceAccounts, Vec<(SolAddress, SolHash)>>>::lookup(
 			AllNonceAccounts,
 		)
 		.map(|nonces| nonces.into_iter().map(|(addr, _hash)| addr).collect::<Vec<_>>())
 		.ok_or(SolanaTransactionBuildingError::NoNonceAccountsSet)
 	}
+
+	fn get_token_environment(
+		token_environments: &mut HashMap<SolAsset, TokenEnvironment>,
+		asset: SolAsset,
```
dandanlen, line 101: Same here, I would suggest either abstracting away the cache implementation behind the call to `lookup`, or just removing the cache. Seems like premature opotimisation.

state-chain/chains/src/sol/api.rs
```diff
@@ -201,94 +228,80 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
-				)?;
-				let system_program_id = SolAddress::from_str(SYSTEM_PROGRAM_ID)
-					.expect("Preset Solana System Program ID account must be valid.");
-
-				// Build the Transfer instruction set for Token transfers.
-				token
-					.into_iter()
-					.map(|transfer_param| {
+					),
+					SolAsset::SolUsdc => {
```
dandanlen, line 251: Should this not match on any asset? 

Ie:

```suggestion
					token_asset => {
						// ..
						let sol_token_environment = Environment::get_token_environment(
							&mut token_environments,
							token_asset,
						)?;
```

Otherwise we still only support USDC 😅

state-chain/chains/src/sol/api.rs
```diff
@@ -201,94 +228,80 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
-			call_type: SolanaTransactionType::Transfer,
-			transaction: SolTransaction::new_unsigned(SolMessage::new_with_blockhash(
-				&instruction_set,
-				Some(&agg_key.into()),
-				&durable_nonce.into(),
-			)),
-			_phantom: Default::default(),
-		})
-		.collect::<Vec<_>>())
+							SOL_USDC_DECIMAL,
```
dandanlen, line 289: And should this be stored in the token environment lookup too, instead of hard-coding it for USDC?

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -24,26 +26,35 @@ use crate::{
...
 		fetch_params: FetchAssetParams<Solana>,
-		token_mint_pubkey: SolAddress,
+		token_environments: &mut HashMap<SolAsset, TokenEnvironment>,
 	) -> Result<(SolanaDepositFetchId, AssetWithDerivedAddress), SolanaTransactionBuildingError> {
 		match fetch_params.asset {
 			SolAsset::Sol => Ok((fetch_params.deposit_fetch_id, AssetWithDerivedAddress::Sol)),
-			SolAsset::SolUsdc =>
+			SolAsset::SolUsdc => {
+				let sol_usdc_environment =
+					Environment::get_token_environment(token_environments, SolAsset::SolUsdc)?;
```
dandanlen, line 50:
```suggestion
			token_asset => {
				let sol_token_environment =
					Environment::get_token_environment(token_environments, token_asset)?;
```


no?

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -121,9 +130,28 @@ impl SolanaInstructionBuilder {
 						SolAccountMeta::new_readonly(agg_key.into(), true),
 						SolAccountMeta::new_readonly(fetch_id.address.into(), false),
 						SolAccountMeta::new(ata.address.into(), false),
-						SolAccountMeta::new(token_vault_ata.into(), false),
-						SolAccountMeta::new_readonly(token_mint_pubkey.into(), false),
-						SolAccountMeta::new_readonly(token_program_id.into(), false),
+						SolAccountMeta::new(
+							token_environments
+								.get(&SolAsset::SolUsdc)
```
dandanlen, line 135: This is also still usdc-specific. Shouldn't we update `AssetWithDerivedAddress` to have a token variant? `AssetWithDerivedAddress::Token(token_ata)` or something?

dandanlen (commented):
This feels incomplete - the code is still quite USDC-specific in parts. Is this intentional?

state-chain/chains/src/sol/api.rs
ramizhasan111: I wasn't sure that substrate caches storage reads, I thought it doesnt. If we are sure that is the case then yeah it is not needed. In case it doesnt, then I think it makes sense to do this since 10 fetches would mean 10 storage reads for example.

state-chain/chains/src/sol/api.rs
ramizhasan111: I dont think it makes sense to store it in the storage since this is a constant.

state-chain/chains/src/sol/instruction_builder.rs
ramizhasan111: The idea was to have a variant for each token in ```AssetWithDerivedAddress``` and so each token would have a match arm here. But I can see that all tokens would have the same code so we could just have one variant. Will update

state-chain/chains/src/sol/instruction_builder.rs
ramizhasan111: will update

state-chain/chains/src/sol/api.rs
ramizhasan111: same here, the idea was to have a match arm for each of the future assets we could have. But I can see it would be basically the same code so we can reuse. will change

state-chain/chains/src/sol/api.rs
dandanlen: If we read the whole environment in one go (as a single struct), instead of storing all the addresses separately, then we don't need caching. 

We can just get the environment and then we have all the addresses we need already available.

dandanlen:
Is there a Linear issue for this btw?

state-chain/chains/src/sol/api.rs
ramizhasan111: It could be different for different tokens so we would need to have a different value for different tokens. 

See the latest commit i pushed where i made the token arms generic, the SOL_USDC_DECIMAL is now retrieved via a function that can return other constants depending on the token.

ramizhasan111:
> Is there a Linear issue for this btw?

i dont think so since this was treated as part of addressing comments on the underlaying solana apicalls PR.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -27,15 +26,15 @@ use crate::{
...
+	api::{get_token_decimals, SolanaEnvironment, TokenEnvironment},
 	consts::TOKEN_PROGRAM_ID,
 };
 use sp_std::str::FromStr;
 
 /// Internal enum type that contains SolAsset with derived ATA
-pub enum AssetWithDerivedAddress {
-	Sol,
-	Usdc(DerivedAta),
+pub struct AssetWithDerivedAddress {
```
dandanlen, line 35: Doc comment is stale.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -44,16 +43,24 @@ impl AssetWithDerivedAddress {
...
-			SolAsset::Sol => Ok((fetch_params.deposit_fetch_id, AssetWithDerivedAddress::Sol)),
-			SolAsset::SolUsdc => {
-				let sol_usdc_environment =
-					Environment::get_token_environment(token_environments, SolAsset::SolUsdc)?;
+			SolAsset::Sol => Ok((
+				fetch_params.deposit_fetch_id,
+				AssetWithDerivedAddress { asset: SolAsset::Sol, ata: None },
+			)),
+			token_asset => {
+				let sol_environment =
```
dandanlen, line 51: nit: `token_environment`

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -27,15 +26,15 @@ use crate::{
...
 };
 use sp_std::str::FromStr;
 
 /// Internal enum type that contains SolAsset with derived ATA
-pub enum AssetWithDerivedAddress {
-	Sol,
-	Usdc(DerivedAta),
+pub struct AssetWithDerivedAddress {
+	pub asset: SolAsset,
+	pub ata: Option<DerivedAta>,
```
dandanlen, line 37: I feel like the previous enum might have been a better choice than a struct. 

The struct doesn't prevent someone from building this for example:

```rust
AssetWithDerivedAddress {
	asset: Sol,
	ata: Some(ata)
}
```

which would be invalid.

dandanlen: In fact I followed this code through to where it's used an I think `decompose_fetch_params` and `AssetWithDerivedAddress` are not necessary. 

We could just call `derive_associated_token_account` in `fetch_from` and then remove all this code.

dandanlen:
Btw not sure if you saw this comment re. HashMaps in the runtime:

https://github.com/chainflip-io/chainflip-backend/pull/4973#discussion_r1647095207

state-chain/chains/src/sol/api.rs
syan095: Yeah I agree, I also think the caching design might be a bit over-complicated here. Since we only do 1 egress per block, by getting all the environment in the AllBatch layer, we are already doing only 1 storage read per block. 
By hiding the caching logic in the environment implementation, it is not obvious that caching is taking place, and it looks like we are doing repeated reads per environment variable per Transfer, which is not the case. I tend to think that it would be better to make it immediately clear that we are retrieving the environments _once_ per AllBatch (hence my original design of getting all the variables, then loop over each Transfer).

ramizhasan111: Regarding using HashMaps: I am not sure if Hashmaps are supported by runtime or not, but happy to make it BTreeMap anyways.

Regarding reading the whole env at once: Ok, in the interest of simplicity i will see how to lump together the different environment elements into 1, remove caching and minimize reads.

state-chain/chains/src/sol/instruction_builder.rs
ramizhasan111: I changed it to make the ```ata``` available in the generic token_asset arm.

I also thought that ```decompose_fetch_params``` is not needed but didnt want to change it as part of this PR but I guess I'll just do it. makes it much simpler

state-chain/chains/src/sol/api.rs
dandanlen: > Regarding using HashMaps: I am not sure if Hashmaps are supported by runtime or not, but happy to make it BTreeMap anyways.

HashMap is imported from std (`use std::collections::HashMap;`), which is not available in the runtime. 

I don't think we need to lump *everything* into one. I can also see why caching *might* make sense here once we support multiple assets. But either way, this is not the way to implement the cache: it should be invisible to the trait methods. For example, it should be possible to implement the environment *without* a cache (for example in the mocks).

dandanlen:
FYI. I merged the changes from the #4895 .

The tests don't compile because of the HashMap thing.

state-chain/chains/src/sol/api.rs
ramizhasan111: > HashMap is imported from std (`use std::collections::HashMap;`), which is not available in the runtime.

Oh yes, didn't see that. Thats what happens when you use vs code's auto-import :)

> I don't think we need to lump _everything_ into one. I can also see why caching _might_ make sense here once we support multiple assets. But either way, this is not the way to implement the cache: it should be invisible to the trait methods. For example, it should be possible to implement the environment _without_ a cache (for example in the mocks).

I don't think I agree with this. Its a matter of perspective on what the SolanaEnvironment trait is. I don't see why a general Environment trait cant have two different functions, one a normal lookup function and the another that lookup while caching in a provided argument.  Sometimes, you might wanna cache, sometimes you might not want to cache. We could have another trait specifically for caching that makes it explicit but that would be just splitting the trait. The caching cant be done inside the ```lookup``` functions since we need to pass the caching variable from the apicall construction inwards into the lookup which we cant do with the current ```ChainEnvironment``` trait.

Moreover, the confusion for me comes from the fact that i think the SolanaEnvironment trait is unnecessary since 1) its a single use trait and 2) it implements all the functions inside by default. We could have all the trait functions as normal functions without and outside the trait. This is what I did at first when I implemented the caching but then I realized that the function is similar to the other functions inside SolanaEnvironment trait and to keep things similar I put it inside.

albert-llimos:
@ramizhasan111  Unfortunately there is a downside in the storage optimizations I came up with that affect this PR.
See: https://github.com/chainflip-io/chainflip-backend/pull/5009
That impacts this PR in a sense that the TokenEnvironment now only needs to have the `token_mint_pubkey`, there's no need to have the `token_vault_ata` nor the `token_vault_pda_account` as they'll be derived on the fly. Sorry for that.

albert-llimos:
> @ramizhasan111 Unfortunately there is a consequence in the storage optimizations I came up with that affect this PR. See: #5009 That impacts this PR in a sense that the TokenEnvironment now only needs to have the `token_mint_pubkey`, there's no need to have the `token_vault_ata` nor the `token_vault_pda_account` as they'll be derived on the fly.

@ramizhasan111 As explained in this comment https://github.com/chainflip-io/chainflip-backend/pull/5009#issuecomment-2194953233 , I think you can disregard my previous comment. Let's implement this with the variables in storage and we'll see later if the optimization actually makes sense.

state-chain/chains/src/sol/api.rs
```diff
@@ -21,20 +22,36 @@ use crate::{
 
 use cf_primitives::{EgressId, ForeignChain};
 
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo)]
+pub struct TokenEnvironment {
+	pub token_mint_pubkey: SolAddress,
+	pub token_vault_ata: SolAddress,
+	pub token_vault_pda_account: SolAddress,
```
albert-llimos, line 29: The `token_vault_pda_account` is the same across tokens, only the `token_vault_pda_account` account changes on a per token basis. Not sure if you did that so we get all values in one lookup but I'd take it away as it's confusing.

dandanlen: FYI. You wrote `token_vault_pda_account` twice. 

Maybe it would help if we have some doc comments explaining what each of these addresses are.

albert-llimos: ups, just fixed the typo, thanks.
It's documented in the Solana architecture doc, both explained and in the diagram.
https://www.notion.so/chainflip/Solana-Architecture-955481ad1afa468b8bb1a0af0897d18e?pvs=4#2626b37dac5e4cbb8312015e1759bc97

ramizhasan111: will update

syan095:
This has been merged into #5016 the Environment PR. This PR can now be closed.

PR #4975, opened 2024-06-18
feat: Solana Allbatch Support

syan095:
# Pull Request

Closes: PRO-1348

## Checklist

## Summary

Allbatch now returns a Vec<ApiCall> instead of a single one.
Added support for Solana AllBatch call, which returns 1 tx for all fetches, and 1 tx for each Transfer.
No test was added, but a new linear issue was created to capture the work needed to test this PRO-1449

NOTE: The event `cf_ingress_egress::BatchBroadcastRequested`'s broadcast ID field is now a Vec<BroadcastId> instead of a single value: `broadcast_ids: Vec<BroadcastId>,`. This will be a breaking change for Engine or other Binary that monitors this event.


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -579,7 +579,7 @@ pub mod pallet {
 			asset: TargetChainAsset<T, I>,
 		},
 		BatchBroadcastRequested {
-			broadcast_id: BroadcastId,
+			broadcast_ids: Vec<BroadcastId>,
```
dandanlen, line 582: This will need a bit more thought, I don't think it's as simple as it seems. 

The event is used to correlate each individual broadcast with a swap egress. A many-to-many mapping doesn't really work for this. So this event needs to remain unchanged. 

We might have to pass in the egress id for each transfer with the transfer params, and the `new_unsigned` would return a `Vec<(Vec<EgressId>, Transaction)>`...

syan095: Good point, I also wondered about this.

syan095: Updated the AllBatch signature accordingly. 
The broadcast event now only include EgressIds that a associated with the broadcast

dandanlen (commented):
LGTM - I'll take another look after the merge conflicts are resolved.

albert-llimos:
We'll take care of the conflicts when merging #4973 but for now we just keep merging PR's as they are ready avoid getting conflicts in multiple open PRs.

PR #4978, opened 2024-06-18
feat: try to submit previous btc broadcasts before our own

kylezs:
# Pull Request

Closes: PRO-1256

## Checklist

## Summary

See linked issue.

### DIFF
--- a/engine/src/state_chain_observer/sc_observer.rs
+++ b/engine/src/state_chain_observer/sc_observer.rs
@@ -24,7 +24,7 @@ use std::{
 	},
 	time::Duration,
 };
-use tracing::{debug, error, info, info_span, Instrument};
+use tracing::{debug, error, info, info_span, warn, Instrument};
 
 use crate::{
 	btc::retry_rpc::BtcRetryRpcApi,
@@ -433,7 +433,7 @@ where
                                                                     Err(error) => {
                                                                         // Just log any errors, don't throw, as this one is not technically our responsibility to broadcast.
                                                                         // and we want to get to attempting broadcasting our one.
-                                                                        error!("Error on Bitcoin PendingBroadcast {awaiting_broadcast_id:?}: {error:?}");
+                                                                        warn!("Error on Bitcoin PendingBroadcast {awaiting_broadcast_id:?}: {error:?}");
                                                                     }
                                                                 }
                                                             }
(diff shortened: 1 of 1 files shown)


AlastairHolmes (approved):

engine/src/state_chain_observer/sc_observer.rs
```diff
@@ -417,6 +418,25 @@ where
                                             let btc_rpc = btc_rpc.clone();
                                             let state_chain_client = state_chain_client.clone();
                                             scope.spawn(async move {
+
+                                                // We check for AwaitingBroadcasts for Bitcoin specifically because if the previous broadcast was not broadcast,
+                                                // it can cause ours to fail, as we could be using a change UTXO that's only created in the previous broadcast.
+                                                for (awaiting_broadcast_id, awaiting_broadcast_data) in state_chain_client
+                                                    .storage_map::<pallet_cf_broadcast::AwaitingBroadcast<Runtime, BitcoinInstance>, Vec<_>>(current_block.hash)
+                                                    .await?
+                                                    .into_iter()
```
dandanlen, line 427: I think AwaitingBroadcast might include aborted broadcasts. 

Better to iterate over PendingBroadcasts and then fetch the information from AwaitingBroadcasts by broadcast+id.

engine/src/state_chain_observer/sc_observer.rs
```diff
@@ -417,6 +416,29 @@ where
...
+                                                            .storage_map_entry::<pallet_cf_broadcast::AwaitingBroadcast<Runtime, BitcoinInstance>>(current_block.hash, &awaiting_broadcast_id)
+                                                            .await? {
+                                                                match btc_rpc.send_raw_transaction(awaiting_broadcast_data.transaction_payload.encoded_transaction).await {
+                                                                    Ok(tx_hash) => info!("Bitcoin PendingBroadcast {awaiting_broadcast_id:?} success: tx_hash: {tx_hash:#x}"),
+                                                                    Err(error) => {
+                                                                        // Just log any errors, don't throw, as this one is not technically our responsibility to broadcast.
+                                                                        // and we want to get to attempting broadcasting our one.
+                                                                        error!("Error on Bitcoin PendingBroadcast {awaiting_broadcast_id:?}: {error:?}");
```
dandanlen, line 436: would a `warn!` be more appropriate here? The code didn't error.

(I see this is the same in the 'normal' code path below)

dandanlen (approved):

engine/src/state_chain_observer/sc_observer.rs
kylezs: Could be a warn I guess - we don't really expect this to happen though. in the case we are submitting broadcasts, if they have already been submitted, they should return as a success - so the RPC would have errored in some way. But you're right it's not critical to the code path, so warn might be a little better. I'm indifferent, so will make the change

PR #4984, opened 2024-06-19
feat: smart retrier endpoint selection

kylezs:
# Pull Request

Closes: PRO-1409

## Checklist

## Summary

The idea here is that we want to keep trying the secondary endpoint until a specified time (2 minutes) has elapsed. We then try the primary endpoint again to check it's online. If not, we try the backup again. Note that this is global to all requests for the chain - previously every request would start by trying the primary, but not it will try the currently preferred endpoint, and occassionally re-attempt the primary, as we want to bias towards the primary.

### DIFF
--- a/engine/src/retrier.rs
+++ b/engine/src/retrier.rs
@@ -310,17 +310,15 @@ impl<Client: Send + Sync + Clone + 'static> ClientSelector<Client> {
 	}
 
 	pub fn request_failed(&mut self, failed_client: PrimaryOrSecondary) {
-		// If we don't have a second endpoint, then we can only try the primary.
-		if self.secondary_signal.is_none() {
-			return;
+		// If we have a second endpoint, then we should switch to the other one.
+		if self.secondary_signal.is_some() {
+			self.prefer = if failed_client == PrimaryOrSecondary::Primary {
+				self.last_failed_primary = Some(tokio::time::Instant::now());
+				PrimaryOrSecondary::Secondary
+			} else {
+				PrimaryOrSecondary::Primary
+			};
 		}
-
-		self.prefer = if failed_client == PrimaryOrSecondary::Primary {
-			self.last_failed_primary = Some(tokio::time::Instant::now());
-			PrimaryOrSecondary::Secondary
-		} else {
-			PrimaryOrSecondary::Primary
-		};
 	}
 }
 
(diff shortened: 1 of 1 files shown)


engine/src/retrier.rs
```diff
@@ -297,6 +308,20 @@ impl<Client: Send + Sync + Clone + 'static> ClientSelector<Client> {
 		.await
 		.0
 	}
+
+	pub fn request_failed(&mut self, failed_client: PrimaryOrSecondary) {
+		// If we don't have a second endpoint, then we can only try the primary.
+		if self.secondary_signal.is_none() {
```
AlastairHolmes, line 314: Seems reasonable to invert the condition and move the code into the if?

engine/src/retrier.rs
```diff
@@ -297,6 +308,20 @@ impl<Client: Send + Sync + Clone + 'static> ClientSelector<Client> {
...
+			return;
+		}
+
+		self.prefer = if failed_client == PrimaryOrSecondary::Primary {
+			self.last_tried_primary = Some(tokio::time::Instant::now());
```
AlastairHolmes, line 319: tried is wrong, the time is when the primary last "failed".

engine/src/retrier.rs
```diff
@@ -269,22 +270,32 @@ impl<Client: Send + Sync + Clone + 'static> ClientSelector<Client> {
...
+	pub async fn select_client(&mut self) -> (Client, PrimaryOrSecondary) {
 		futures::future::select_all(
 			utilities::conditional::conditional(
 				&self.secondary_signal,
 				|secondary_signal| {
 					// If we have two clients, then we should bias the requested one, but if it's
 					// not ready, request from the other one.
-					match primary_or_secondary {
-						PrimaryOrSecondary::Secondary => [secondary_signal, &self.primary_signal],
+					match self.prefer {
```
AlastairHolmes, line 289: You don't truely need this `prefer` member, you could only have the option and set it to None, once the duration has past. If you did this you could us a [Timer](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) instead of a Instant.

AlastairHolmes (approved):

engine/src/retrier.rs
AlastairHolmes: I see this would slightly change the behaviour actually so probably not worth it, but it does seem odd to have both members.

kylezs: Agreed, that it seems odd, thought about it for a bit. But settled for this as the intention is at least clear.

engine/src/retrier.rs
kylezs: guess it's last "tried" assuming we're currently using the backup - but yeah failed is more explicit, will change

AlastairHolmes (approved):

PR #4987, opened 2024-06-20
feat: scheduled swap sub price estimation fallback

j4m1ef0rd:
# Pull Request

Closes: PRO-1410, PRO-1437

## Checklist

## Summary

- Changed the `get_scheduled_swap_legs` function to not return a result. In case of swap simulation failure, it will use the given pool_price (of the input asset) to estimate the swap amount (in Usdc) for the second part of the leg.
	- The first leg will now always be returned, even if the swap simulation fails. 
	- The only reason the second leg would not be returned, is if there is no pool_price (no liquidity in the pool at all)
- Added a unit test to cover this new fallback code.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -239,6 +239,7 @@ pub mod pallet {
 		SwapDepositHandler,
 	};
 	use frame_system::WeightInfo as SystemWeightInfo;
+	use sp_runtime::SaturatedConversion;
 
 	use super::*;
 
@@ -951,7 +952,7 @@ pub mod pallet {
 									cf_amm::common::Amount::from(swap.input_amount()),
 									sqrt_price_to_price(pool_sell_price?),
 								)
-								.as_u128(),
+								.saturated_into(),
 							)
 						})?;
 
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -917,6 +917,7 @@ pub mod pallet {
 		) -> Vec<SwapLegInfo> {
 			let mut swaps: Vec<_> = swaps.into_iter().map(SwapState::new).collect();
 
+			// Can ignore the result here because we use pool price fallback below
 			let _res = Self::swap_into_stable_taking_network_fee(&mut swaps);
 
 			swaps
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -942,32 +942,28 @@ pub mod pallet {
 							(None, None)
 						};
 
-						if swap.stable_amount.is_some() || pool_sell_price.is_some() {
+						let amount = swap.stable_amount.or_else(|| {
 							// If the swap into stable asset failed, fallback to estimating the
 							// amount via pool price.
-							let amount = swap.stable_amount.unwrap_or_else(|| {
+							Some(
 								output_amount_ceil(
 									cf_amm::common::Amount::from(swap.input_amount()),
-									sqrt_price_to_price(
-										pool_sell_price.expect("Safe because of check above"),
-									),
+									sqrt_price_to_price(pool_sell_price?),
 								)
-								.as_u128()
-							});
+								.as_u128(),
+							)
+						})?;
 
-							Some(SwapLegInfo {
-								swap_id: swap.swap_id(),
-								base_asset,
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -941,23 +942,37 @@ pub mod pallet {
...
-							source_amount,
-						})
+						if swap.stable_amount.is_some() || pool_sell_price.is_some() {
+							// If the swap into stable asset failed, fallback to estimating the
+							// amount via pool price.
+							let amount = swap.stable_amount.unwrap_or_else(|| {
+								output_amount_ceil(
+									cf_amm::common::Amount::from(swap.input_amount()),
+									sqrt_price_to_price(
+										pool_sell_price.expect("Safe because of check above"),
```
dandanlen, line 952: Not a big deal but we can probably remove a level of nesting and the `expect` by using `?` here.

```rust
let amount = swap.stable_amount.or_else(|| {
	Some(
		//...
			pool_sell_price?
		//
	)
})?;

**Review by dandanlen — COMMENTED** [review]
Looks fine to me.

File: state-chain/pallets/cf-swapping/src/lib.rs (hunk 1 above)
Comment (j4m1ef0rd) [reply state-chain/pallets/cf-swapping/src/lib.rs:952]: I tried 3 versions of this logic trying to make it better, and never thought of that. 😅

File: state-chain/pallets/cf-swapping/src/lib.rs [hunk 2]
```diff
@@ -912,13 +913,13 @@ pub mod pallet {
...
 			swaps: Vec<Swap>,
 			base_asset: Asset,
-		) -> Result<Vec<SwapLegInfo>, ()> {
+			pool_sell_price: Option<SqrtPriceQ64F96>,
+		) -> Vec<SwapLegInfo> {
 			let mut swaps: Vec<_> = swaps.into_iter().map(SwapState::new).collect();
 
-			Self::swap_into_stable_taking_network_fee(&mut swaps)
-				.map_err(|_| log::error!("Failed to simulate swaps"))?;
+			let _res = Self::swap_into_stable_taking_network_fee(&mut swaps);
```
msgmaxim, line 920: I know have a comment further below, but might be worth leaving a comment here as well to explain why we can ignore the result.

msgmaxim (approved):

state-chain/runtime/src/lib.rs
```diff
@@ -1735,7 +1735,8 @@ impl_runtime_apis! {
 			all_prewitnessed_swaps
 		}
 
-		fn cf_scheduled_swaps(base_asset: Asset, _quote_asset: Asset) -> Vec<(SwapLegInfo, BlockNumber)> {
+		fn cf_scheduled_swaps(base_asset: Asset, quote_asset: Asset) -> Vec<(SwapLegInfo, BlockNumber)> {
+			assert_eq!(quote_asset, STABLE_ASSET, "Only USDC is supported as quote asset");
```
kylezs, line 1739: Why is this assert necessary? It's not necessarily always going to be the case that we'll only allow USDC as the quote asset.

Also not sure about panicking here anyway, or below on something that's totally possible since it's user input. Anyone could provide the wrong asset. I think we should throw an error here instead, that way the user can actually get the error in the response too, instead of "server error" or something

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -941,23 +943,33 @@ pub mod pallet {
...
 
+						let amount = swap.stable_amount.or_else(|| {
+							// If the swap into stable asset failed, fallback to estimating the
+							// amount via pool price.
+							Some(
+								output_amount_ceil(
+									cf_amm::common::Amount::from(swap.input_amount()),
+									sqrt_price_to_price(pool_sell_price?),
+								)
+								.as_u128(),
```
kylezs, line 954: this can panic - `.unwrap_or(u128::MAX)`

kylezs (changes requested):

state-chain/runtime/src/lib.rs
j4m1ef0rd: > Why is this assert necessary?

Because the `get_scheduled_swap_legs` function (and entire swapping pallet) assumes the stable asset is usdc. I wanted to sanity check it here.

> totally possible since it's user input

The parent functions do check that the pool exists and return an error. Not exactly confirming that quote asset is usdc, but I am unsure how to get `cf_scheduled_swaps` to return an error.

kylezs (approved):

PR #4991, opened 2024-06-20
chore: improve docker compatibility and dylib path resolution 🚀

ahasna:
Inspired by: https://github.com/chainflip-io/chainflip-backend/pull/4989

### DIFF
--- a/localnet/manage.sh
+++ b/localnet/manage.sh
@@ -116,15 +116,6 @@ build-localnet() {
   mkdir -p $CHAINFLIP_BASE_PATH
   touch $DEBUG_OUTPUT_DESTINATION
 
-  if [ "$OS_TYPE" == "Linux" ]; then
-    echo "🕵🏻‍♂️  Detected OS: $OS_TYPE. configuring LD_LIBRARY_PATH"
-    echo "ℹ️ Note: The .so files in 'old-engine-dylib' are built for Ubuntu 22.04."
-    cp ./old-engine-dylib/libchainflip_engine_v*.so $BINARY_ROOT_PATH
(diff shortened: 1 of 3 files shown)


localnet/manage.sh
```diff
@@ -98,24 +117,24 @@ build-localnet() {
   touch $DEBUG_OUTPUT_DESTINATION
 
   if [ "$OS_TYPE" == "Linux" ]; then
-    echo "🕵🏻‍♂️  Detected OS: $OS_TYPE. Copying .so files..."
+    echo "🕵🏻‍♂️  Detected OS: $OS_TYPE. configuring LD_LIBRARY_PATH"
     echo "ℹ️ Note: The .so files in 'old-engine-dylib' are built for Ubuntu 22.04."
-    sudo cp $BINARY_ROOT_PATH/libchainflip_engine_v*.so /usr/lib/
-    sudo cp ./old-engine-dylib/libchainflip_engine_v*.so /usr/lib/
+    cp ./old-engine-dylib/libchainflip_engine_v*.so $BINARY_ROOT_PATH
+    export LD_LIBRARY_PATH=$PWD/$BINARY_ROOT_PATH
```
kylezs, line 123: Why this change? Dylibs are normally stored under /usr/lib 

Also, even with the change to where the files are moved, I don't think the LD_LIBRARY_PATH is necessary after the PR I just merged earlier: #4983 - would make sure you're rebased on main and check that.

kylezs: want to confirm the exact reason it doesn't work on fedora, I have a feeling it was something else that's off but happy to be proven wrong

kylezs (approved):

PR #4994, opened 2024-06-21
fix: eagerly prune ExtraCallData

dandanlen:
# Pull Request

Closes: PRO-1391

## Checklist

## Summary

The ExtraCallData storage currently takes up ~300Mb of storage on mainnet. 

This PR makes two changes:

1. remove the ExtraCallData when we recombine it to construct the dispatch calldata.
2. prevent it from being inserted if we have already processed it (ie. if the call has already passed the voting threshold).

This is a non-breaking runtime change.

### DIFF
--- a/state-chain/pallets/cf-witnesser/src/lib.rs
+++ b/state-chain/pallets/cf-witnesser/src/lib.rs
@@ -446,7 +446,13 @@ pub mod pallet {
 					*vote = true;
 
 					if let Some(extra_data) = extra_data {
-						ExtraCallData::<T>::append(epoch_index, call_hash, extra_data);
+						// Extra data is no longer needed if we have reached the threshold since it
+						// has already been consumed.
+						if vote_count <=
+							success_threshold_from_share_count(num_authorities) as usize
+						{
+							ExtraCallData::<T>::append(epoch_index, call_hash, extra_data);
+						}
 					}
 
 					Ok(vote_count)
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-witnesser/src/lib.rs
```diff
@@ -461,7 +467,7 @@ pub mod pallet {
 				(last_expired_epoch..=current_epoch)
 					.all(|epoch| CallHashExecuted::<T>::get(epoch, call_hash).is_none())
 			{
-				if let Some(mut extra_data) = ExtraCallData::<T>::get(epoch_index, call_hash) {
+				if let Some(mut extra_data) = ExtraCallData::<T>::take(epoch_index, call_hash) {
```
kylezs, line 470: We might as well delete the clean up code now then? And just write a migration to clean up anything that's currently there?

state-chain/pallets/cf-witnesser/src/tests.rs
```diff
@@ -676,3 +680,59 @@ fn can_punish_failed_witnesser_in_previous_epochs() {
...
+		assert_eq!(pallet_dummy::Something::<Test>::get(), Some(2u32));
+
+		// Extra call data is empty.
+		assert!(ExtraCallData::<Test>::get(
+			current_epoch,
+			CallHash(frame_support::Hashable::blake2_256(&*call))
+		)
+		.is_none(),);
+
+		// Can vote again, but not call data is inserted and result is unaffected.
```
kylezs, line 725:
```suggestion
		// Can vote again, but call data is not inserted and result is unaffected.
```

state-chain/pallets/cf-witnesser/src/lib.rs
dandanlen: It will clean itself up over the course of a couple of epochs, and finding the values that can be deleted now will require iterating over all the votes, checking the vote count etc. I'd rather save the time it would take develop and test this.

kylezs: ok, let's create an issue at least so we don't forget about

kylezs (approved):

state-chain/pallets/cf-witnesser/src/lib.rs
dandanlen: What I mean is that it's ok to forget about it - the values get pruned after 2 epochs anyway.

AlastairHolmes (approved):

PR #4996, opened 2024-06-21
feat: Solana Durable Nonce Accounts logic

ramizhasan111:
# Pull Request

Closes: PRO-1209

## Checklist

## Summary

#### Non-Breaking changes


state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -393,6 +407,39 @@ pub mod pallet {
...
+
+			if let Some(_nonce) = SolanaUnAvailableNonceAccounts::<T>::take(nonce_account) {
+				SolanaAvailableNonceAccounts::<T>::append((nonce_account, durable_nonce));
+				Self::deposit_event(Event::<T>::DurableNonceSetForAccount {
+					nonce_account,
+					durable_nonce,
+				});
+				Ok(().into())
+			} else {
+				Err(Error::<T>::NonceAccountNotBeingUsedOrDoesntExist.into())
```
dandanlen, line 440: I'm not sure we need this error case. 

We witness the nonce accounts, so if the engines witness a new nonce account (imagine we add a new one via governance - not sure if this is possible/likely?) then we have to 'believe' them, right?

albert-llimos: If we were to add a nonce via governance it's unclear to me if we'd do it that way or we'd add another governance function that adds a new nonce, not just have the engine suddenly witness a new account that the SC is not aware of. But we haven't thought about that yet, so it might make sense to not throw an error as there might be a use for it.

albert-llimos:
Other than addressing Dan's comments above, LGTM.

state-chain/pallets/cf-environment/src/lib.rs
ramizhasan111: We have a predifined list of nonce accounts and so so if we witness a nonce for a nonce account that does is not registered with the state chain, we should ignore it. If we add a new nonce account, it should be via governance. We should have the state chain control what the nonce account list is and the cfes should witness those. We shouldnt have the cfes deciding that we now need a new nonce account and so lets start witnessing that. We should beleive the cfes for what the nonce is, not what nonce accounts to witness.

dandanlen: Ok. Which account has the authority to add new durable nonce accounts? The aggkey I guess?

dandanlen: Ok then I agree it makes sense to restrict this on the SC 👌

albert-llimos:
@ramizhasan111 do you plan to add the initialization of nonce accounts and values as part of this PR or will you do it all together when implementing https://linear.app/chainflip/issue/PRO-1362/implement-solana-environment?

ramizhasan111:
> @ramizhasan111 do you plan to add the initialization of nonce accounts and values (chainspec and migration) as part of this PR or will you do it all together when implementing [linear.app/chainflip/issue/PRO-1362/implement-solana-environment](https://linear.app/chainflip/issue/PRO-1362/implement-solana-environment)?

I think its better to wait for the Environment PR to go in (https://github.com/chainflip-io/chainflip-backend/pull/4861), which already has the migrations and stuff. The migration for this can then be easily added.

albert-llimos:
> > @ramizhasan111 do you plan to add the initialization of nonce accounts and values (chainspec and migration) as part of this PR or will you do it all together when implementing [linear.app/chainflip/issue/PRO-1362/implement-solana-environment](https://linear.app/chainflip/issue/PRO-1362/implement-solana-environment)?
> 
> I think its better to wait for the Environment PR to go in (#4861), which already has the migrations and stuff. The migration for this can then be easily added.

I'm not sure @syan095 will use that PR, but even then the PR doesn't have much regarding the nonces. All the nonces will come from this PR so I think for completeness' sake I'd just add the chainspec and migration logic for the nonces here.

ramizhasan111:
> I'm not sure @syan095 will use that PR, but even then, the PR doesn't have much regarding the nonces. All the nonces will come from this PR so I think for completeness' sake I'd just add the chainspec and migration logic for the nonces here.

Yeah I planned to add it as part of this PR. just wanted to wait for that one to go in, rebase this one onto that and add the nonce related entries in migration and chain spec. But in any case, I will do it as part of this PR.

albert-llimos:
> Yeah I planned to add it as part of this PR. just wanted to wait for that one to go in, rebase this one onto that and add the nonce related entries in migration and chain spec. But in any case, I will do it as part of this PR.

Actually, after some light bulb moment, I've just figured out an optimization so we don't need to add anything in the environment in the other PR. Or in other words, that the other PR is effectively canceled. Therefore, the current environment is everything we'll have, so you can already go ahead and do the migration and chainspec changes.

state-chain/chains/src/sol/api.rs
```diff
@@ -25,25 +25,28 @@ use cf_primitives::ForeignChain;
...
 {
 	fn compute_price() -> Result<SolAmount, SolanaTransactionBuildingError> {
 		<Self as ChainEnvironment<ComputePrice, SolAmount>>::lookup(ComputePrice)
 			.ok_or(SolanaTransactionBuildingError::CannotLookupComputePrice)
 	}
 
-	fn nonce_account() -> Result<(SolAddress, SolHash), SolanaTransactionBuildingError> {
-		<Self as ChainEnvironment<NonceAccount, (SolAddress, SolHash)>>::lookup(NonceAccount)
+	fn nonce_account() -> Result<DurableNonceAndAccount, SolanaTransactionBuildingError> {
+		<Self as ChainEnvironment<DurableNonce, DurableNonceAndAccount>>::lookup(DurableNonce)
```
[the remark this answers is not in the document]
ramizhasan111: will do this as part of https://github.com/chainflip-io/chainflip-backend/pull/4973 since it wasnt introduced in this PR.

state-chain/node/src/chain_spec.rs
```diff
@@ -97,6 +97,8 @@ pub struct StateChainEnvironment {
 	sol_vault_address: SolAddress,
 	sol_genesis_hash: Option<SolHash>,
 	sol_usdc_address: SolAddress,
+	sol_durable_nonces_and_accounts: [DurableNonceAndAccount; 7], /* we inject 7 nonce accounts
```
dandanlen, line 100: Do we always need 7 accounts? Event on a localnet?

state-chain/node/src/chain_spec/berghain.rs
```diff
@@ -52,6 +53,37 @@ pub const ENV: StateChainEnvironment = StateChainEnvironment {
 	sol_vault_address: SolAddress([0; 32]), // TODO: fill in the valid Solana address,
 	sol_genesis_hash: Some(SolHash(bs58_array("5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d"))),
 	sol_usdc_address: SolAddress(bs58_array("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v")),
+	// todo: put correct values for nonces and accounts
```
dandanlen, line 56: Better to make this a todo!() 

That way if we ever do create a new mainnet chainspec (unlikely), we *can't* forget to set these values.

state-chain/node/src/chain_spec/perseverance.rs
```diff
@@ -52,6 +53,37 @@ pub const ENV: StateChainEnvironment = StateChainEnvironment {
 	sol_vault_address: SolAddress([0; 32]), // TODO: fill in the valid Solana address,
 	sol_genesis_hash: Some(SolHash(bs58_array("EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG"))),
 	sol_usdc_address: SolAddress(bs58_array("4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU")),
+	// todo: put correct values for nonces and accounts
+	sol_durable_nonces_and_accounts: [
```
dandanlen, line 57: Same here.

state-chain/node/src/chain_spec/sisyphos.rs
```diff
@@ -52,6 +54,37 @@ pub const ENV: StateChainEnvironment = StateChainEnvironment {
 	sol_vault_address: SolAddress([0; 32]), // TODO: fill in the valid Solana address,
 	sol_genesis_hash: Some(SolHash(bs58_array("EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG"))),
 	sol_usdc_address: SolAddress(bs58_array("4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU")),
+	// todo: put correct values for nonces and accounts
+	sol_durable_nonces_and_accounts: [
```
dandanlen, line 58: Same

state-chain/node/src/chain_spec/testnet.rs
```diff
@@ -48,6 +49,37 @@ pub const ENV: StateChainEnvironment = StateChainEnvironment {
 	sol_vault_address: SolAddress([0; 32]), // TODO: fill in the valid Solana address,
 	sol_genesis_hash: None,
 	sol_usdc_address: SolAddress(bs58_array("24PNhTaNtomHhoy3fTRaMhAFCRj4uHqhZEEoWrKDbR5p")),
+	// todo: put correct values for nonces and accounts
+	sol_durable_nonces_and_accounts: [
```
dandanlen, line 53: Same - better for this to fail before we launch the network.

state-chain/runtime/src/migrations/solana_integration.rs
```diff
@@ -86,68 +86,107 @@ impl OnRuntimeUpgrade for SolanaIntegration {
...
-					log::warn!("Need to set up Solana integration for Berghain");
-					(
+		let (vault_address, genesis_hash, usdc_address, durable_nonces_and_accounts): (
+			SolAddress,
+			Option<SolHash>,
+			SolAddress,
+			Vec<DurableNonceAndAccount>,
+		) = match cf_runtime_upgrade_utilities::genesis_hashes::genesis_hash::<Runtime>() {
+			cf_runtime_upgrade_utilities::genesis_hashes::BERGHAIN => {
+				log::warn!("Need to set up Solana integration for Berghain");
```
dandanlen, line 97: log::error! , we don't want to miss this.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -688,7 +688,13 @@ impl<T: Config> Pallet<T> {
 	}
 
 	pub fn get_sol_nonce_and_account() -> Option<DurableNonceAndAccount> {
-		SolanaAvailableNonceAccounts::<T>::mutate(|nonce_and_accounts| nonce_and_accounts.pop())
+		let nonce_and_account = SolanaAvailableNonceAccounts::<T>::mutate(|nonce_and_accounts| {
+			nonce_and_accounts.pop()
+		});
+		nonce_and_account.map(|(account, nonce)| {
+			SolanaUnAvailableNonceAccounts::<T>::insert(account, nonce);
```
dandanlen, line 695: Do we need to store the nonce value here? It looks like it's never used - we could just store a map of `<Account, ()>`?

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -352,3 +356,114 @@ fn test_consolidation_change_amount() {
...
+		let (account2, nonce2) = Environment::get_sol_nonce_and_account().unwrap();
+		assert_eq!((account2, nonce2), (SolAddress([4; 32]), SolHash([40; 32])));
+		assert_eq!(SolanaUnAvailableNonceAccounts::<Test>::get(account2).unwrap(), nonce2);
+		assert_eq!(
+			SolanaUnAvailableNonceAccounts::<Test>::iter_keys().collect::<Vec<_>>().len(),
+			2
+		);
+
+		// put back the first nonce
+		Environment::update_sol_nonce(RuntimeOrigin::root(), account1, nonce1).unwrap();
```
dandanlen, line 390: It would be better to add a new *different* nonce than the initial one instead of adding the same on back again.

state-chain/node/src/chain_spec.rs
ramizhasan111: Yes, this is the fixed number of nonce accounts we want to use. We'll use the same number of accounts on all networks to make testing similar to real world scenario.

state-chain/runtime/src/migrations/solana_integration.rs
ramizhasan111: i dont understand, this is not an error. its just an info log. if anything, we should make it log::info!

state-chain/node/src/chain_spec/berghain.rs
ramizhasan111: we can't have a todo!() here since this is a const and so it is evaluated at compile time and so it errors during compilation

state-chain/pallets/cf-environment/src/lib.rs
ramizhasan111: The reason we need to store the used nonce here is because of how current nonce witnessing works. the cfe checks if the witnessed nonce is the same as the current in-use nonce.

state-chain/runtime/src/migrations/solana_integration.rs
dandanlen: We have missed this kind of thing in the past. It's easy to skim over info and warnings. An error shows up clearly. It doesn't prevent the migration checks from passing.

dandanlen (approved):

PR #5001, opened 2024-06-25
test: engine-runner command line args passing filters correctly

kylezs:
# Pull Request

Closes: PRO-1455

## Checklist

## Summary

A bug slipped through into main, fixed here: https://github.com/chainflip-io/chainflip-backend/pull/4982 - which is hard to pick up without a test like this one.

### DIFF
--- a/engine-runner-bin/src/main.rs
+++ b/engine-runner-bin/src/main.rs
@@ -164,7 +164,7 @@ mod tests {
 		// There should be no error reading settings error - it will likely error due to no
 		// connection, but settings should be ok since all args are provided in command line args.
 		assert_ne!(
-			old::cfe_entrypoint(
+			new::cfe_entrypoint(
 				CStrArray::from_rust_strings(&rust_args).unwrap(),
 				engine_upgrade_utils::NO_START_FROM,
 			)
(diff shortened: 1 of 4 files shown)


engine-runner-bin/src/main.rs
```diff
@@ -77,3 +80,100 @@ fn main() -> anyhow::Result<()> {
...
+
+	// This tests:
+	// 1. We can run by providing all necessary settings via command line.
+	// 2. That args are filtered out correctly before passing them into the old version.
+	// 3. That the old version can run with the filtered args. implying that the settings in the new
+	//    version are backwards compatible.
+	#[test]
+	fn incompatible_args_should_be_filtered_out() {
+		// create temporary file
+		let tempdir = tempfile::TempDir::new().unwrap();
```
dandanlen, line 99: Does `tempfile` automatically clean up the temporary file that is created? Maybe we should delete it?

dandanlen (approved):

engine-runner-bin/src/main.rs
kylezs: Should be cleaned up on Drop: https://docs.rs/tempfile/latest/tempfile/#resource-leaking

AlastairHolmes (approved):

PR #5002, opened 2024-06-26
chore: build solana consts in compile time

syan095:
# Pull Request

Closes: PRO-1402

## Checklist

## Summary
Solana const addresses are now built during compile time, instead of at runtime. This removes the wraps and expects


state-chain/chains/Cargo.toml
```diff
@@ -50,7 +50,7 @@ generic-array = { version = "0.14.7", default-features = false, features = [
   "serde",
   "more_lengths",
 ] }
-bs58 = { version = "0.4.0", default-features = false, features = ["alloc"] }
+bs58 = { version = "^0.5", default-features = false, features = ["alloc"] }
```
albert-llimos, line 53: I'd advocate for using exact versions. For another PR in the engine we will use `0.5.1` so I'd use that one

albert-llimos:
One thing worth mentioning is that we'll have to do more conversions like this on top of the CFE PR, as there is several instances like that.
https://github.com/chainflip-io/chainflip-backend/pull/4933

state-chain/chains/Cargo.toml
dandanlen: Also FYI the "^" is unnecessary. `"^0.x.y"` is the same as `"0.x.y"`

dandanlen (approved):

dandanlen:
> One thing worth mentioning is that we'll have to do more conversions like this on top of the CFE PR, as there is several instances like that. #4933

Maybe we rebase this on top of the cfe PR and then make the changes there?

albert-llimos:
> > One thing worth mentioning is that we'll have to do more conversions like this on top of the CFE PR, as there is several instances like that. #4933
> 
> Maybe we rebase this on top of the cfe PR and then make the changes there?

Yes I think that'd be great, since the CFE PR shouldnt have many changes.

albert-llimos:
The CFE PR has now been merge into the main branch. Therefore there are some merge conflicts to fix.
@syan095 One thing worth noting is that in the CFE PR the constants were moved from the State Chain constants file into the sol-prim (foreign-chain) constants since they are also being used by the engine. You'll need to fix that that when fixing the merge conflicts.
The good thing though is that now you can turn into constants all the SolAddresses intruduced by the CFE PR in this PR itself so we don't break it into two parts.

state-chain/chains/Cargo.toml
syan095: Updated to 0.5.1

PR #5004, opened 2024-06-26
chore: remove deprecated fields

kylezs:
# Pull Request

Closes: PRO-1431

## Checklist

## Summary

Removes deprecated fields. Confirmed with @niklasnatter they use the new fields now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes included in each Chainflip release will be documented in thi
 ## [1.5.0]
 
 - Remove some deprecated fields from Swapping events
-  - `egress_amount` in `SwapExecuted` event - use `swap_amount` instead
+  - `egress_amount` in `SwapExecuted` event - use `swap_output` instead
   - `deposit_amount` in `SwapExecuted` event - use `swap_input` instead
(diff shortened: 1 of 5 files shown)


dandanlen:
It's worth thinking about how to communicate this externally (since there are some integrators that depend on the raw events) - for one, we should include this in the changelog, with a list of the fields that have been removed.

dandanlen (approved):

msgmaxim:
All of these have been removed/reworked in this upcoming PR: https://github.com/chainflip-io/chainflip-backend/pull/5025. I don't know if it makes sense to have a separate PR for this (e.g. it is not like this needs to be urgently done before the larger PR is ready).

> It's worth thinking about how to communicate this externally

Btw, this applies to all other changes we have make to the swapping events (see https://www.notion.so/chainflip/Updated-Swapping-Events-9c1ac17013c84b50b2037c4ca7c5268e). I'm not leaving any "deprectated" fields in my PR, since the changes are so fundamental, I'm not making any attempts to keep the events backwards compatible.

dandanlen:
Ok I agree - no need to make this change separately if it's going to be superseded soon anyway.

PR #5009, opened 2024-06-27
Feat: Solana environment optimization

albert-llimos:
## Checklist

## Summary

Derive `vault_data_account`, `token_vault_pda` and `token_vault_ata` on the fly when building transactions so we don't have to store them in the environment.
They shouldn't be needed for anything else than building calls, not even anywhere in the CFE, so it seems totally worth it.

### DIFF
--- a/state-chain/chains/src/sol/api.rs
+++ b/state-chain/chains/src/sol/api.rs
@@ -55,14 +55,8 @@ pub trait SolanaEnvironment:
 					SolanaTransactionBuildingError::CannotLookupAggKey,
 				SolanaEnvAccountLookupKey::VaultProgram =>
 					SolanaTransactionBuildingError::CannotLookupVaultProgram,
-				SolanaEnvAccountLookupKey::VaultProgramDataAccount =>
-					SolanaTransactionBuildingError::CannotLookupVaultProgramDataAccount,
 				SolanaEnvAccountLookupKey::TokenMintPubkey =>
 					SolanaTransactionBuildingError::CannotLookupTokenMintPubkey,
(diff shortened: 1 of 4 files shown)


ramizhasan111:
not sure about the optimisation. The account derivation is a const derivation and so will be exactly same for every tx construction so it might still be better to store rather than derive? Not a strong opinion though.

albert-llimos:
> not sure about the optimisation. The account derivation is a const derivation and so will be exactly same for every tx construction so it might still be better to store rather than derive? Not a strong opinion though.

I don't have a very strong opinion either but we discussed something similar around token accounts for deposit channels and everyone (Dan, Alastair, Roy) seemed very eager to store the minimum data on-chain. I think given that we only build transactions once in a while (only in fetches, transfers and so on), it seems better to derive it on the fly rather than store it. 
Also, in this case it has the nice property of directly supporting any token, given that the `token_vault_ata` will be derived from whichever `token_mint_pubkey` is passed.

albert-llimos:
> not sure about the optimisation. The account derivation is a const derivation and so will be exactly same for every tx construction so it might still be better to store rather than derive? Not a strong opinion though.

Actually upon further thinking it might still make sense to not jump the gun yet. My proposal therefore is to just disregard this for now, implement the nonces and the environment pallet with the variables in storage. We will be packing the lookups according to how the'll be read so we can minimize the number of reads. When we have that together we can assess better if this PR will introduce a significant improvement (mainly reducing the number of lookups, besides reducing the variables in storage).

albert-llimos:
Closing this as we are not aiming to implement it.

PR #5011, opened 2024-06-27
fix: record_fees

AlastairHolmes:
record_fees function was broken recently, as mutate gives a mut reference and doesn't use the closures return value to set the storage item.


dandanlen (approved):
Thanks 🙈

PR #5014, opened 2024-07-01
refactor: move collected network fee to swapping pallet

j4m1ef0rd:
# Pull Request

Closes: PRO-1404

## Checklist

## Summary

- Moved the `FlipBuyInterval` and `CollectedNetworkFee` storage items from the pools pallet to the swapping pallet.
	- Moved related functions and genesis config item.
	- Moved related tests and benchmarks.
	- Added migration to runtime.
- Changed the pools tests to no longer know anything about a network fee, this changed some of the values from swaps a little (no more network fee).
- Refactored the tests that got moved to the swapping pallet. Might be hard to see the changes because of the move, I can try and split the move into a separate commit, but it didn't seem worth the time unless requested.
- Added the `test_fee_estimation_basis` test, not sure why it was just stilling as a function description.


state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2498,3 +2497,124 @@ fn swap_output_amounts_correctly_account_for_fees() {
...
+fn test_buy_back_flip() {
+	new_test_ext().execute_with(|| {
+		const INTERVAL: BlockNumberFor<Test> = 5;
+		const SWAP_AMOUNT: AssetAmount = 1000;
+		const NETWORK_FEE: Permill = Permill::from_percent(2);
+
+		NetworkFee::set(NETWORK_FEE);
+
+		// The default buy interval is zero, and this means we don't buy back.
+		assert_eq!(FlipBuyInterval::<Test>::get(), 0);
```
msgmaxim, line 2511: What is the purpose of this line? If you want to test the behaviour when interval = 0, you need to  ensure the relavant code gets called (calling on_initialize etc). Otherwise you might as well remove this.

state-chain/runtime/src/lib.rs
```diff
@@ -1165,6 +1165,96 @@ type PalletMigrations = (
 type MigrationsForV1_5 =
 	(migrations::housekeeping::Migration, migrations::reap_old_accounts::Migration);
 
+pub struct NetworkFeesMigration;
```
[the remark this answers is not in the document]
j4m1ef0rd: hmm, strange, it was there, I tested it, must of lost the change. Nice catch.

state-chain/pallets/cf-swapping/src/tests.rs
j4m1ef0rd: I think the purpose was to just check that it is set to 0 by default at genesis. Not very important, but worth leaving in i think.

msgmaxim: I think it is worth testing the behaviour when set to 0 in that case. It is important that we have logic for handling this separately, i.e. we don't divide by 0, for example.

dandanlen (approved):
LGTM. 

Just a minor thing, no need to change this now, but in future please prefer to put the runtime-level migrations in their own modules - makes the diffing/merging much easier when the time comes to remove the migrations after they've passed.

msgmaxim (approved):

j4m1ef0rd:
@dandanlen the upgrade test is still failing bouncer. I thought this was fixed. This branch is updated with main.

dandanlen:
> @dandanlen the upgrade test is still failing bouncer. I thought this was fixed. This branch is updated with main.

Yeah I thought it was fixed too... will take a look.

dandanlen:
It must be something that's in main - this PR is failing the upgrade check too, and it doesn't change any code: 
https://github.com/chainflip-io/chainflip-backend/pull/5033

PR #5016, opened 2024-07-01
feat: solana API environment

syan095:
# Pull Request

Closes: PRO-1362
Closes: PRO-1472

## Checklist

## Summary
Created a struct that contains Solana addresses to be stored in the Environment pallet.
Initial values are added to the storage via Migration and Genesi config
Hooked up the storage with SolEnvironment
Solana API and Instruction build can now pull these addresses in a single storage read to build API calls


albert-llimos:
I've taken a quick look (will review further tomorrow) but the direction of the changes seem good.
It's just a bit weird because this PR pulls the code from the generic USDC PR but then the changes don't take it into account, as all the new code has the hardcoded USDC values.
@syan095  Was your idea that we merge this first into the base PR and then we make more changes to the Generic Token PR? It might be an option to merge this to the Generic Token branch instead, this way we can merge that one as is, and we "just" need to add the corresponding generalized code to this PR adequately. 
Also tagging @ramizhasan111 as he has worked on the other PR.

albert-llimos:
>  Seems like you guys discussed implementing this instead of the other generic USDC PR for simplicity reasons 👍
Assuming the above, this LGTM. As usual, we'll wait for Dan/Ramiz to check it out.

PR #5017, opened 2024-07-01
feat: fix for polkadot CheckMetadataHash update

dandanlen:
# Pull Request

## Checklist

## Summary

Fixes the encoding of polkadot extrinsics by adding the new item to SignedExtra. The unit test can be used to print a payload and this payload can be verified in polkaJS to decode correctly. 

(`0xdd02840022fd62bede1c2c45822a5007893b90c1d6c3407d9eb275aeb4890541cd893f6e01c215c3a4cae054916d032cbf162efe92ca5fbe9953b686a5e1fe734db7d3f31503598c2fbbded5940b2cb40a7ae459d5cd769bb9460771e5aec5ae53367b268e004400001d00006d07de61ff7c24ff7e3688df5e0e20e619454b8c82975dbd2afda98efa281f1301001a04041a010e000504006d07de61ff7c24ff7e3688df5e0e20e619454b8c82975dbd2afda98efa281f1300`)

I also added a new governance extrinsic to resign and optionally re-broadcast transactions (taken from #4758). However, on second thought, it might be better to trigger the re-broadcast from the the runtime upgrade itself, since adding a new extrinsic is a breaking change. 

I'll remove this and add the required code to re-sign the transactions during the runtime upgrade. 

I also bumped all the version to 1.4.5 and added a Changelog entry.


state-chain/chains/src/dot.rs
```diff
@@ -982,6 +987,8 @@ pub struct PolkadotReplayProtection {
 	pub nonce: PolkadotIndex,
 }
 
+pub type PolkadotCheckMetadataHash = Option<()>;
```
AlastairHolmes, line 990: I didn't suggest not using this, as the `Mode` enum they use is like [this](https://docs.rs/frame-metadata-hash-extension/0.3.0/src/frame_metadata_hash_extension/lib.rs.html#53):

```rust
/// The mode of [`CheckMetadataHash`].
#[derive(Decode, Encode, PartialEq, Debug, TypeInfo, Clone, Copy, Eq)]
enum Mode {
	Disabled,
	Enabled,
}
```

And is on reason to believe the Option encoding would be the same as that.

AlastairHolmes (approved):
Approved, but I think a review of our handling of this should be done, as I highly doubt Polkadot will make improvements to their release processes. We could, for example nightly spin up our own testnets using the latest Polkadot runtime release tag/branch, and run our test suite.

state-chain/chains/src/dot.rs
dandanlen: My understanding is that the Mode is what is used in the runtime (ie. in Polkadot) to tell the runtime whether or not it supports this. 

The associated `AdditionalSigned` is what is actually signed over, which is an `Option<[u8; 32>`. See [here](https://github.com/paritytech/polkadot-sdk/blob/8dbe4ee80734bba6644c7e5f879a363ce7c0a19f/substrate/frame/metadata-hash-extension/src/lib.rs). 

What's a bit crazy is that even if the runtime *doesn't* support it, you need to add `None` to the signed payload...

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -686,7 +686,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 	/// Signs a API call, use `Call::on_signature_ready` as the callback, and returns the signature
 	/// request ID.
-	fn threshold_sign(
+	// NOTE: Function made Temporarily public - do not merge this to main.
```
AlastairHolmes, line 689: ?

state-chain/chains/src/dot.rs
dandanlen: As discussed - you were right. I've updated it.

AlastairHolmes (approved):

martin-chainflip (approved):

PR #5018, opened 2024-07-02
chore: add compute limits

albert-llimos:
# Pull Request

Closes: PRO-1357

## Checklist

### DIFF
--- a/state-chain/chains/src/sol.rs
+++ b/state-chain/chains/src/sol.rs
@@ -96,15 +96,16 @@ pub const LAMPORTS_PER_SIGNATURE: SolAmount = 5000u64;
 
 // This is to be used both for ingress/egress estimation and for setting the compute units
 // limit when crafting transactions by the State Chain.
-mod compute_units_costs {
-	use super::SolAmount;
-	pub const BASE_COMPUTE_UNITS_PER_TX: SolAmount = 450u64;
-	pub const COMPUTE_UNITS_PER_FETCH_NATIVE: SolAmount = 7_500u64;
-	pub const COMPUTE_UNITS_PER_TRANSFER_NATIVE: SolAmount = 300u64;
-	#[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


state-chain/chains/src/sol.rs
```diff
@@ -107,16 +107,16 @@ impl ChainCrypto for SolanaCrypto {
...
-	pub const COMPUTE_UNITS_PER_TRANSFER_NATIVE: SolAmount = 300u64;
-	#[allow(dead_code)]
-	pub const COMPUTE_UNITS_PER_FETCH_TOKEN: SolAmount = 31_000u64;
-	#[allow(dead_code)]
-	pub const COMPUTE_UNITS_PER_TRANSFER_TOKEN: SolAmount = 41_200u64;
+// limit when crafting transactions by the State Chain. Using a 50% buffer to ensure we'll
+// have enough compute units to cover the actual cost.
+pub mod compute_units_costs {
+	use super::SolComputeLimit;
+	pub const BASE_COMPUTE_UNITS_PER_TX: SolComputeLimit = 300u32 * 3 / 2;
```
dandanlen, line 114: Nit: we could define a function `const fn compute_limit_with_buffer(u32) -> u32` just for consistency and so that it's clear where the 3/2 comes from.

dandanlen (approved):

state-chain/chains/src/sol.rs
albert-llimos: I thought about that, it'd look cleaner, but then it will be computed at runtime every time, right?

albert-llimos:
Fyi, I won't be merging this until the bouncer PR (#5029 ) is merged to ensure that this compute limits work as intended.

state-chain/chains/src/sol.rs
dandanlen: Functions defined with `const fn` run at compile time :) (they can only access other const fns though).

PR #5019, opened 2024-07-03
feat: Solana transaction builder implementation

syan095:
# Pull Request

Closes: PRO-1475

## Checklist

## Summary

Implements the Solana's transaction builder - which is used to actually broadcast the Solana Call

### DIFF
--- a/state-chain/runtime/src/chainflip.rs
+++ b/state-chain/runtime/src/chainflip.rs
@@ -315,21 +315,29 @@ impl TransactionBuilder<Bitcoin, BitcoinApi<BtcEnvironment>> for BtcTransactionB
 pub struct SolanaTransactionBuilder;
 impl TransactionBuilder<Solana, SolanaApi<SolEnvironment>> for SolanaTransactionBuilder {
 	fn build_transaction(
-		_signed_call: &SolanaApi<SolEnvironment>,
+		signed_call: &SolanaApi<SolEnvironment>,
 	) -> <Solana as Chain>::Transaction {
-		todo!()
+		signed_call.transaction.clone()
 	}
+
 	fn refresh_unsigned_data(_tx: &mut <Solana as Chain>::Transaction) {
-		todo!()
+		// Transaction fee on SOlana is a fixed 5k lamports per signature. Since we always have
+		// a single signature, therefore the transaction fee does not change.
 	}
+
 	fn calculate_gas_limit(_call: &SolanaApi<SolEnvironment>) -> Option<U256> {
-		todo!()
+		// Solana sets computation limits via the `compute_limit` attribute, instead of
+		// via transaction fee. Therefore this mechanism is not used for Solana.
+		None
 	}
+
 	fn requires_signature_refresh(
 		_call: &SolanaApi<SolEnvironment>,
 		_payload: &<<Solana as Chain>::ChainCrypto as ChainCrypto>::Payload,
 	) -> bool {
-		todo!()
+		// We use Durable Nonce mechanism to avoid the 150 blocks expiry period.
+		// This means the transaction won't expire and therefore no need to be resigned.
+		false
 	}
 }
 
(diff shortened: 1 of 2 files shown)


state-chain/runtime/src/chainflip.rs
```diff
@@ -315,21 +315,29 @@ impl TransactionBuilder<Bitcoin, BitcoinApi<BtcEnvironment>> for BtcTransactionB
...
 	fn refresh_unsigned_data(_tx: &mut <Solana as Chain>::Transaction) {
-		todo!()
+		// Transaction fee on SOlana is a fixed 5k lamports per signature. Since we always have
+		// a single signature, therefore the transaction fee does not change.
 	}
+
 	fn calculate_gas_limit(_call: &SolanaApi<SolEnvironment>) -> Option<U256> {
-		todo!()
+		// Solana sets computation limits via the `compute_limit` attribute, instead of
+		// via transaction fee. Therefore this mechanism is not used for Solana.
```
albert-llimos, line 330: This is true for regular swaps but not for CCM Swaps. For the later, it will be similar to Ethereum/Arbitrum where we should calculate the gas_limit from the output of the gas swap and set that as the gas limit (`compute_limit`).
However, that also requires changes in the Transaction Builder. I'd first leave this in a state that it works for regular swaps and I'll open a PR where we tackle the gas limit for CCM swaps.

state-chain/runtime/src/chainflip.rs
```diff
@@ -315,21 +315,29 @@ impl TransactionBuilder<Bitcoin, BitcoinApi<BtcEnvironment>> for BtcTransactionB
...
+		signed_call: &SolanaApi<SolEnvironment>,
 	) -> <Solana as Chain>::Transaction {
-		todo!()
+		signed_call.transaction.clone()
 	}
+
 	fn refresh_unsigned_data(_tx: &mut <Solana as Chain>::Transaction) {
-		todo!()
+		// Transaction fee on SOlana is a fixed 5k lamports per signature. Since we always have
+		// a single signature, therefore the transaction fee does not change.
```
albert-llimos, line 325: I'm not sure about how this works. From the name it sounds like it's all the data before signing, but there is this comment in the code: `Note that calldata cannot be updated, or it would invalidate the signature.`
For Solana, if this is before any signature is done, then we should refresh the priority fee (as you say the base fee doesn't need any update).
However, if the comment is correct and this is after we have a signature, then we shouldn't change the priority fee, as that would require a new signature.
Do you know which one is it?

state-chain/runtime/src/chainflip.rs
albert-llimos: Reason being that the priority is getting regular swaps tested and we can implement CCM-specific logic separately.

syan095: Ok sounds good!

state-chain/runtime/src/chainflip.rs
syan095: This is currently only used by EVM chains where gas fee and priority fee is updated.

To answer your question - once the transaction content is changed (fees updated), the next function: `requires_signature_refresh` will return true, and we will then re-sign the call.

This refresh fn is called just before when a call is broadcasted(in the `fn start_broadcast_attempt()` - so it ensures the call always have the correct gas fee setting. If the gas fee changed, the call is re-signed will be tried again.

state-chain/runtime/src/chainflip.rs
albert-llimos: Added it here: https://linear.app/chainflip/issue/PRO-1479/solana-ccm-gas-limit-calculation
Please add some reference to that linear ticket in the code.

state-chain/runtime/src/chainflip.rs
syan095: So the question here would be - how fast do fees change in Solana and how accurate do we have to be? Basically we are taking the lesser evil here:
1. risk of inaccurate gas setting causing the transaction to fail 
2. cost of resigning a call

albert-llimos: Gotcha, thanks for the explanation.
Then we shouldn't use this for Solana but not for the reason explained in the comment. We can't use it because refreshing the priority fee would require resigning, which we don't want to do.

albert-llimos (approved):

PR #5021, opened 2024-07-03
Feat/nonce witnessing cfe

ramizhasan111:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

#### Non-Breaking changes

### DIFF
--- a/engine/src/witness/sol/nonce_witnessing.rs
+++ b/engine/src/witness/sol/nonce_witnessing.rs
@@ -1,4 +1,9 @@
-use crate::witness::common::{RuntimeCallHasChain, RuntimeHasChain};
+use crate::{
+	state_chain_observer::client::{
+		chain_api::ChainApi, extrinsic_api::signed::SignedExtrinsicApi, storage_api::StorageApi,
+	},
+	witness::common::{RuntimeCallHasChain, RuntimeHasChain},
+};
 use cf_chains::{
 	sol::{SolAddress, SolHash},
 	Chain,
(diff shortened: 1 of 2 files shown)


engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -43,41 +48,48 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
-			let nonce_accounts = nonce_accounts.clone();
+			let state_chain_client = state_chain_client.clone();
 			async move {
-				let nonce_addresses: Vec<SolAddress> = nonce_accounts
-					.clone()
-					.into_iter()
-					.map(|(nonce_address, _)| nonce_address)
-					.collect();
+				let nonce_accounts: HashMap<SolAddress, SolHash> = state_chain_client
+					.storage_map::<pallet_cf_environment::SolanaUnAvailableNonceAccounts<
```
albert-llimos, line 60: Can we think about any scenario in which also checking the AvailableNonceAccounts could avoid an edge case? Of course that shouldn't happen in normal circumstances, ideally never, but just throwing it out there in case anyone can think about one.

albert-llimos:
LGTM. One of the things I realize with witnessing the nonces with this approach (periodically) is that we have to periodically pull the Durable Nonce values from storage, which is not ideal. If we were witnessing only after a broadcast is witnessed we could get away with just reading the NonceAccounts one time and not reading storage again.
However, we went for this approach due to:
- Getting durable nonces from Solana ensures we will always be in sync regardless of broadcasts
- The logic for to figure out which nonce was updated might be a bit combersome, which is why we went for this approach.

I'd definitely go for this approach now but it's worth thinking if we should do that as an optimization. I'll think more about it.

albert-llimos (approved):

PR #5024, opened 2024-07-05
Feat: Make FoK constants configurable

j4m1ef0rd:
# Pull Request

Closes: PRO-1436

## Checklist

## Summary

Changed the FoK constants `swap retry delay` and `max swap retry duration` into storage items.
- an entry in the `PalletConfigUpdate` to set the value.
- an event that is emitted when the updated.
- a default value for the storage item and genesis config.
- Added to the `PalletConfigUpdate` test.
- Removed the update flip buy interval extrinsic
- Changed the buy flip interval set event to match the naming of the other events.


j4m1ef0rd:
Do they need a migration to set the values on a live chain? Or do they somehow use the value i set in the new genesis config? @dandanlen

msgmaxim:
> Do they need a migration to set the values on a live chain? Or do they somehow use the value i set in the new genesis config? @dandanlen

I would expect that you need to set them during migration. You can also test this by trying runtime upgrade against the current release.

dandanlen:
Yes - either set the value during a migration or set a default value if there's an obvious and sensible default. (You can set defaults in the storage definition)

dandanlen:
Just to follow up on this - you can declare storage defaults like this:

```rust
// Declares a storage value of type u32. If not set, will return 5 as the default.
pub type X<T: Config> = StorageValue<_, u32, ValueQuery, Const<5>>;

// Instead of Const<_> the default value can be implemented with a Get impl:
struct MyDefault;
impl Get<u32> for MyDefault {
	fn get() -> u32 {
		42
	}
}
pub type Y<T: Config> = StorageValue<_, u32, ValueQuery, MyDefault>;
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1144,6 +1155,25 @@ pub mod pallet {
 			})?;
 			Ok(())
 		}
+
+		/// Updates the max swap retry duration .
+		///
+		/// ## Events
+		///
+		/// - [MaxSwapRetryDurationBlocks](Event::MaxSwapRetryDurationBlocks)
```
kylezs, line 1163:
```suggestion
		/// - [UpdatedMaxSwapRetryDuration](Event::UpdatedMaxSwapRetryDuration)
```

state-chain/pallets/cf-swapping/src/tests/fill_or_kill.rs
```diff
@@ -61,7 +61,7 @@ fn both_fok_and_regular_swaps_succeed_first_try() {
 fn price_limit_is_respected_in_fok_swap() {
 	const SWAPS_ADDED_BLOCK: u64 = 1;
 	const SWAPS_SCHEDULED_FOR_BLOCK: u64 = 3;
-	const SWAP_RETRIED_AT_BLOCK: u64 = SWAPS_SCHEDULED_FOR_BLOCK + SWAP_RETRY_DELAY_BLOCKS as u64;
+	const SWAP_RETRIED_AT_BLOCK: u64 = SWAPS_SCHEDULED_FOR_BLOCK + DEFAULT_SWAP_RETRY_DELAY_BLOCKS;
```
kylezs, line 64: why are we using the default for these tests instead of the retry delay that's in storage?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -613,7 +636,8 @@ pub mod pallet {
 							}
 						};
 
-						let retry_block = current_block + SWAP_RETRY_DELAY_BLOCKS.into();
+						let retry_block =
+							current_block + max(SwapRetryDelay::<T>::get(), 1u32.into());
```
kylezs, line 640: we could dedup this into a variable

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1000,6 +1024,31 @@ pub mod pallet {
...
+		///
+		/// - [UpdatedSwapRetryDelay](Event::UpdatedSwapRetryDelay)
+		///
+		/// ## Errors
+		///
+		/// - [BadOrigin](frame_system::BadOrigin)
+		/// - [ZeroSwapRetryDelayNotAllowed](pallet_cf_pools::Error::ZeroSwapRetryDelayNotAllowed)
+		#[pallet::call_index(13)]
+		#[pallet::weight(T::WeightInfo::update_buy_interval())]
+		pub fn update_swap_retry_delay(
```
kylezs, line 1040: I don't think we need a new function here, we have an `update_pallet_config` function already - this seems like a good candidate to add there - and ofc add some test cases for it

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1144,6 +1155,25 @@ pub mod pallet {
...
+		#[pallet::call_index(10)]
+		#[pallet::weight(T::WeightInfo::update_max_swap_retry_duration())]
+		pub fn update_swap_retry_delay(
```
kylezs, line 1166: same here, we can just add this to the update_pallet_config function that exists already

state-chain/pallets/cf-swapping/src/tests/fill_or_kill.rs
j4m1ef0rd: I did try that, but to get the value from storage it needs to be done within a `execute_with` context, but if we do that we cant use the value as an argument in the `then_execute_at_block`. So it becomes very difficult and messy. Best option was to just use a default constant, since the constant sets the mock genesis, it should not get out of sync unless the test directly changes the storage.

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: Great idea, did know it was a thing. I'll move the flip buy interval in there as well while im at it.

kylezs: Yeah sounds good

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2000,7 +2020,7 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
 	> {
 		if let Some(refund_params) = &refund_params {
 			ensure!(
-				refund_params.retry_duration <= MAX_SWAP_RETRY_DURATION_BLOCKS,
+				refund_params.retry_duration <= MaxSwapRetryDurationBlocks::<T, I>::get(),
```
kylezs, line 2023: Can be done in a separate issue - but since we can have retry_duration set to 0, I think this is a good case to test on its own. That if it's 0, we still attempt to swap and then refund if it fails.

kylezs (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: PRO-1511

PR #5025, opened 2024-07-05
Swapping Pallet Redesign

msgmaxim:
# Pull Request

Closes: PRO-1431, PRO-1282

## Checklist

## Summary

- In general, swapping pallet events have changed according to https://www.notion.so/chainflip/Updated-Swapping-Events-9c1ac17013c84b50b2037c4ca7c5268e
- Introduced the concept of `SwapRequest` for swaps that have one or more parts. Removed storage accociated with the need to handle multiple swaps for CCM (e.g. `CcmOutputs`, `PendingCcms`), since all requets are stored in the unified `SwapRequests` storage.
- There is now a unified interface for initiating swaps, which is done through `init_swap_request`.
- CCM swaps are now done in up to two phases: (1) principal swap is scheduled/executed, and (2) if principal swap is successful the gas swap is scheduled/executed. This is a prerequisite for FoK CCM, and DCA will likely follow a similar approach.
- Swaps now have `swap_request_id` in addition to `swap_id` to link them to the original request. Swap request ids start from the value `SwapIdCouter + CcmIdCounter` to ensure we leave room for the product to reassign existing swaps_ids and ccm_ids to matching `swap_request_id`s.
- `MaximumSwapAmountSet` now applies to the deposit amount rather than the amounts used in individual “subswaps” (which has been the case for ccm gas and ccm principle swaps, and would also be the case for individual subswaps in DCA). The reason is (1) it makes sense, and (2) I needed to move this check from `schedule_swap` either way, and putting it in `init_swap_request` made the most sense.
- Removed `NetworkFeeTaken`. Instead, network fee is reported in `SwapExecuted` in the `network_fee` field.
- After discussion with @dandanlen, we agreed that migrating any pending swaps would be a lot of work (and error prone), and we should instead make sure that we don't have any swaps during migration. I have extended `UpgradeCondition` to include this extra condition.


state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -241,17 +238,9 @@ fn test_buy_back_flip() {
...
-				.expect("Should have scheduled a swap usdc -> flip"),
-			&MockSwap {
-				from: STABLE_ASSET,
-				to: FLIP,
-				amount: EXPECTED_COLLECTED_FEES,
-				swap_type: SwapType::NetworkFee,
-			}
-		);
+
+		// TODO: check that a swap is scheduled once we network fee stuff
```
niklasnatter, line 242:
```suggestion
		// TODO: check that a swap is scheduled once the network fee stuff
```

state-chain/pallets/cf-swapping/src/migrations/swapping_redesign.rs
```diff
@@ -0,0 +1,96 @@
...
+	fn on_runtime_upgrade() -> Weight {
+		// Note: we don't migrate items from SwapQueue because we will
+		// ensure that it is empty during the upgrade.
+
+		log::info!("Swapping redesign migration!");
+
+		let latest_used_ccm_id = old::CcmIdCounter::<T>::get();
+		let latest_used_swap_id = SwapIdCounter::<T>::get();
+
+		SwapRequestIdCounter::<T>::put(latest_used_ccm_id + latest_used_swap_id);
```
niklasnatter, line 50: i think from the product side it would be okay to only use `latest_used_swap_id` here, but passing the sum of them should be fine too (maybe im missing something though @acdibble)

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -366,28 +410,26 @@ pub mod pallet {
 		},
 		/// A swap has been executed.
 		SwapExecuted {
+			swap_request_id: SwapRequestId,
```
niklasnatter, line 413: if i understand things right, we would not be able to determine if this is the `gas` part of a `ccm` swap atm, right?

niklasnatter: i guess we could read the `gasBudget` from the `SwapRequested` and try to match that

acdibble: i assume the swap `swap_id` would correlate to the gas or principal swap id

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -366,28 +410,26 @@ pub mod pallet {
...
-			swap_id: SwapId,
+			swap_request_id: SwapId,
 			egress_id: EgressId,
 			asset: Asset,
 			amount: AssetAmount,
 			fee: AssetAmount,
 		},
 		RefundEgressScheduled {
-			swap_id: SwapId,
+			swap_request_id: SwapId,
```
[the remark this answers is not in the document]
acdibble: or does the type also need to change?

state-chain/pallets/cf-swapping/src/lib.rs
niklasnatter: but i think we dont know the `swap_id` of the gas/principle swap (bc there is no event that announces them)?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1269,87 +1359,108 @@ pub mod pallet {
...
-			to: Asset,
-			amount: AssetAmount,
-			destination_address: ForeignChainAddress,
-			broker_commission: Beneficiaries<Self::AccountId>,
+		fn init_swap_request(
+			input_asset: Asset,
+			input_amount: AssetAmount,
+			output_asset: Asset,
+			request_type: SwapRequestType,
+			broker_fees: Beneficiaries<Self::AccountId>,
```
dandanlen, line 1410: Broker fees only apply to swaps from channels, we know that they will never apply to other swap types - why not embed them in the SwapRequestType?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1269,87 +1359,108 @@ pub mod pallet {
...
 			refund_params: Option<ChannelRefundParameters>,
-			channel_id: ChannelId,
-		) -> SwapId {
-			// Permill maxes out at 100% so this is safe.
-			let fee: u128 = Permill::from_parts(
-				broker_commission.iter().fold(0, |acc, entry| acc + entry.bps) as u32 *
-					BASIS_POINTS_PER_MILLION,
-			) * amount;
+			origin: SwapOrigin,
```
dandanlen, line 1412: Same with origin I guess, it only applies to user swaps, and its only use seems to be to add to events. Could just be embedded in the SwapType.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1269,87 +1359,108 @@ pub mod pallet {
...
+			});
 
-			let encoded_destination_address =
-				T::AddressConverter::to_encoded_address(destination_address.clone());
-			let swap_origin = SwapOrigin::DepositChannel {
-				deposit_address: T::AddressConverter::to_encoded_address(deposit_address),
-				channel_id,
-				deposit_block_height,
+			// Do not limit the maximum swap amount for network fee swaps.
+			let net_amount = if matches!(request_type, SwapRequestType::NetworkFee) {
```
dandanlen, line 1443: Not just network swaps? What about gas fee swaps for example?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1363,222 +1474,200 @@ pub mod pallet {
...
+						},
+					);
 
-			let ccm_id = CcmIdCounter::<T>::mutate(|id| {
-				id.saturating_accrue(1);
-				*id
-			});
+					RequestTypeEncoded::NetworkFee
+				},
+				SwapRequestType::IngressEgressFee => {
```
dandanlen, line 1507: It feels like there's a lot of repetition in these branches. Could we not just define schedule_swap as `schedule_swap(request_id, &swap_request)` and then we can avoid having this huge match statement?.

Ideally though, schedule_swap should not have to care about the originating swap request. We should probably add a `fees: Vec<Fee>` member that is stored with every request. Then that can be propagated when we schedule the swap, and the request *type* doesn't matter.

This will make it easier to take the broker fees in USDC, as is the plan. (we just add a `Fee::BrokerFee {id,rate}`).

dandanlen (commented):
I think this is good, although I think we can try to reduce the number of almost-but-not-quite-the-same data types. 

FWIW I also think using min_output works for individual sub-swaps but we might have to store the price limit with the request anyway. Consider the case of a streamed swap: at the level of an individual swap 'chunk', knowing the min_output of the whole swap is not very useful. In order to calculate the min_output of each chunk, we probably need the price limit of the overall request.

Related to this - refunding failed FOK orders is still handled inside the execute_batch method. I'm not sure this will be compatible with DCA (at the very least, this 'inner' part of the swapping pipeline will need access to the 'outer' request context; it might also need to finalise the DCA swap in some way so that the partially executed amount can be egressed). I feel like this isn't fully covered here (maybe it's planned for a future PR).

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: There is `SwapScheduled { swap_type, .. }` where swap_type is: 

```rust
pub enum SwapType {
	Swap,
	CcmPrincipal,
	CcmGas,
	NetworkFee,
	IngressEgressFee,
}
```

niklasnatter: you are right - missed that! can resolve this then

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -146,27 +150,47 @@ pub struct SwapLegInfo {
...
+		let destination_chain: ForeignChain = destination_asset.into();
+		if !destination_chain.ccm_support() {
+			return Err(CcmFailReason::UnsupportedForTargetChain)
+		} else if deposit_amount < gas_budget {
+			return Err(CcmFailReason::InsufficientDepositAmount)
+		}
+
+		// if the gas asset is different.
+		let output_gas_asset = ForeignChain::from(destination_asset).gas_asset();
+		let other_gas_asset = if source_asset == output_gas_asset || gas_budget.is_zero() {
```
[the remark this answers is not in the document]
msgmaxim: "Other" in the name is from the original code. My understanding is that it was to communicate that we only provide value in this Option when the asset is *other* than the input asset. Note that None for output_gas_asset could suggest that there is no output asset for some reason, which would be confusing/misleading.

state-chain/pallets/cf-swapping/src/migrations/swapping_redesign.rs
msgmaxim: I thought in our discussion we decided to reserve ids in the range `(latest_used_swap_id, latest_used_swap_id + latest_used_ccm_id]` for existing CCM swaps (since requets ids will be replacing ccm_ids)?

state-chain/pallets/cf-pools/src/tests.rs
msgmaxim: Jamie's network fee PR has been moved, so this is no longer relevant :)

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: The intention was to use `SwapRequestId` (i.e. we will only egress/refund once per request, rather than egressing for every failed chunk).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -146,27 +150,47 @@ pub struct SwapLegInfo {
 impl Swap {
 	fn new(
 		swap_id: SwapId,
+		swap_request_id: SwapId,
```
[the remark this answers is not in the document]
msgmaxim: No strong opinion, but I don't think it is necessary in this case. Both `SwapRequestId` and `SwapId` are just u64, so the same type (and aren't expected to change). Mixing them up creates some confusion and in general not ideal, but it doesn't break anything.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -146,27 +150,47 @@ pub struct SwapLegInfo {
...
+		let principal_swap_amount = deposit_amount.saturating_sub(gas_budget);
+
+		let destination_chain: ForeignChain = destination_asset.into();
+		if !destination_chain.ccm_support() {
+			return Err(CcmFailReason::UnsupportedForTargetChain)
+		} else if deposit_amount < gas_budget {
+			return Err(CcmFailReason::InsufficientDepositAmount)
+		}
+
+		// if the gas asset is different.
```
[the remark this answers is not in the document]
msgmaxim: The comment is from the original code, but I updated it now to be more useful.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -182,37 +206,58 @@ impl From<DispatchError> for BatchExecutionError {
...
+		other_gas_asset: Option<Asset>,
+	},
+	GasSwapScheduled {
+		gas_swap_id: SwapId,
+		principal_output_amount: AssetAmount,
+	},
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+enum SwapRequestState {
```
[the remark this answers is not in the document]
msgmaxim: > the latter might be little less verbose but has fewer static guarantees.

Actually depending on how you define "verbose" the latter might end up being even more verbose due to less type safety, e.g. when you need to get output address for a "regular" swap and have to assert/error our when it is None, even though in theory this shouldn't happen.

> Either we embed all the data in the enum variants (ie. put SwapRefundParameters in the SwapRequestState) 

I think I prefer this option. I can just put refund parameters next to output_address in Ccm And Regular variants.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -182,37 +206,58 @@ impl From<DispatchError> for BatchExecutionError {
...
+	NetworkFee,
+	IngressEgressFee,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+
+struct SwapRequest {
+	id: SwapRequestId,
+	input_asset: Asset,
+	output_asset: Asset,
```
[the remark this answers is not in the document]
msgmaxim: My approach has been to store the data that may be needed at some point later in the request's lifecycle. `input_amount` is never used again after the initial processing. Do you want to add it just for completeness?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -331,6 +368,19 @@ pub mod pallet {
...
 	pub enum Event<T: Config> {
+		/// New swap has been requested
+		SwapRequested {
+			swap_request_id: SwapRequestId,
+			input_asset: Asset,
+			input_amount: AssetAmount, // includes broker fee
+			output_asset: Asset,
+			broker_fee: AssetAmount,
+			origin: SwapOrigin,
+			request_type: RequestTypeEncoded,
```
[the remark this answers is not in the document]
msgmaxim: I'm not sure about this. `SwapRequest` also contains state that changes throughout the request's lifecycle, and I the event should only expose the parameters rather than its dynamic state.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: We can also have swaps initiated from the contract, where broker fees don't apply. Otherwise would be a clear improvement, but I'm not sure now.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: How do you mean exactly? Whether the deposit comes from the vault contract or a deposit channel seems orthogonal to whether it is "regular"/ccm.

> and its only use seems to be to add to events.

To me that's the more reason to keep them separate, since it won't be needed after the initial event is emitted.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Added exception for `IngressEgressFee` too.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: > Could we not just define schedule_swap as schedule_swap(request_id, &swap_request)

not sure how to reconcile this with 

` schedule_swap should not have to care about the originating swap request`

maybe we could discuss this in a synchronous conversation when you are back.

> We should probably add a fees: Vec<Fee> member that is stored with every request.

Yeah, that sounds like a good idea, I will give it a go.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -182,37 +206,58 @@ impl From<DispatchError> for BatchExecutionError {
...
-
 pub struct CcmSwapAmounts {
 	pub principal_swap_amount: AssetAmount,
 	pub gas_budget: AssetAmount,
 	// if the gas asset is different to the input asset, it will require a swap
 	pub other_gas_asset: Option<Asset>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub enum RequestTypeEncoded {
```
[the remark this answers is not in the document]
kylezs: This seems to still be the case in the code @msgmaxim - is there a reason you didn't decide to do this?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -182,37 +206,58 @@ impl From<DispatchError> for BatchExecutionError {
...
+	NetworkFee,
+	IngressEgressFee,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+enum CcmSwapState {
+	PrincipalSwapScheduled {
+		principal_swap_id: SwapId,
+		gas_budget: AssetAmount,
+		other_gas_asset: Option<Asset>,
```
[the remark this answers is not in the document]
kylezs: same here, this is unaddressed

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1156,131 +1194,202 @@ pub mod pallet {
...
+		fn process_ccm_outcome(
+			swap_request_id: SwapRequestId,
+			output_asset: Asset,
+			output_address: ForeignChainAddress,
+			principal_amount: AssetAmount,
+			gas_amount: AssetAmount,
+			ccm_deposit_metadata: CcmDepositMetadata,
+		) {
+			Self::deposit_event(Event::<T>::SwapRequestCompleted { swap_request_id });
+			// Schedule the given ccm to be egressed and deposit an event.
```
kylezs, line 1206: nit: would remove comment, is pretty clear

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1156,131 +1194,202 @@ pub mod pallet {
...
-										swap_id: swap.swap_id(),
-										asset: swap.output_asset(),
-										amount: swap_output,
-										reason: err.into(),
-									});
-								},
+				let swap_request_id = swap.swap.swap_request_id;
+
+				let Some(mut request) = SwapRequests::<T>::take(swap_request_id) else {
+					log_or_panic!("swap request not found");
```
kylezs, line 1262:
```suggestion
					log_or_panic!("swap request {swap_request_id} not found");
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1156,131 +1194,202 @@ pub mod pallet {
...
+					continue;
+				};
+
+				let Some(output_amount) = swap.final_output else {
+					log_or_panic!("Swap is not completed yet!");
```
kylezs, line 1267:
```suggestion
					log_or_panic!("Swap {} is not completed yet!", swap.swap_id());
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1156,131 +1194,202 @@ pub mod pallet {
...
+										request.input_asset,
+										*other_gas_asset,
+										*gas_budget,
+										output_amount,
+									);
+
+									// We are not done with the request, so put it back
+									SwapRequests::<T>::insert(swap_request_id, request);
+								} else {
+									Self::process_ccm_outcome(
```
kylezs, line 1304: should we Self::deposit_event(Event::<T>::SwapRequestCompleted { swap_request_id }); here too?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1532,172 +1673,199 @@ pub mod pallet {
...
 
-			Self::deposit_event(Event::<T>::CcmDepositReceived {
-				ccm_id,
-				principal_swap_id,
-				gas_swap_id,
-				deposit_amount,
-				destination_address: encoded_destination_address,
-				deposit_metadata: deposit_metadata.clone(),
-			});
+						let swap_request = SwapRequest {
```
kylezs, line 1826: nit: inline for consistency

kylezs (commented):
Haven't finished yet, basically finished reviewing the core code, but haven't looked at tests or migrations, will do that tomorrow

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -194,37 +235,50 @@ impl From<DispatchError> for BatchExecutionError {
...
 pub struct CcmSwapAmounts {
 	pub principal_swap_amount: AssetAmount,
 	pub gas_budget: AssetAmount,
 	// if the gas asset is different to the input asset, it will require a swap
 	pub other_gas_asset: Option<Asset>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+enum CcmSwapState {
+	PrincipalSwapScheduled {
```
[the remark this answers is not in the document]
msgmaxim: I'm not sure I agree. When writing a state machine there will always be some place in code where externally the state transition happened and while we are updating the state variable, it is "out of date" for a few lines of code, and it is OK. If you think of `process_swap_outcomes` as the function that handles state transition (which it is), then the state we read is "previous" state, and what we write is the "new" state, and then it makes a lot of sense.

On the other hand, renaming to `PrincipalSwap` etc makes the code less clear IMO, as then it is less obvious that this is a step in the process, and starts to look more like some info about the principal swap (raising further questions like why things like `gas_budget` are stored on the principle swap).

I used `Regular` because there is only one state, so I don't think of it as a state machiine really, but more like the swaps's type, and in that sense it is more similar to (and consistent with) `NetworkFee` and `IngressEgressFee`.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1156,131 +1194,202 @@ pub mod pallet {
...
-							} else {
-								log_or_panic!(
-									"NetworkFee burning should not be in asset: {:?}",
-									swap.output_asset()
+							CcmSwapState::GasSwapScheduled {
+								gas_swap_id,
+								principal_output_amount,
+							} => {
+								debug_assert!(swap.swap_id() == *gas_swap_id);
+
```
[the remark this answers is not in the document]
msgmaxim: It is done inside `process_ccm_outcome` to avoid duplication.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: As, above this is done in `process_ccm_outcome`.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -467,34 +508,30 @@ pub mod pallet {
...
-			swap_id: SwapId,
-			fee_amount: AssetAmount,
-		},
 		BuyIntervalSet {
 			buy_interval: BlockNumberFor<T>,
 		},
 		SwapRetryDelaySet {
 			swap_retry_delay: BlockNumberFor<T>,
 		},
+		// TODO: add SwapFailed?
```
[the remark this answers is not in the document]
msgmaxim: I left it as a todo since we haven't quite agreed on how this should work. For example, do we emit it for every swap in a batch every time the batch fails even it will be retried (seems a bit excessive), or do we only emit when we "give up" on the swap and refund it (which means it will never be emitted for non-FoK swaps)?

IMO it is OK to add this in a future PR (while we get some clarity on it) since this event it "new", i.e. it doesn't replace any existing one that the product is expecting.

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
kylezs: sure not a strong opinion

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
kylezs: Ok, is there a ticket to follow up on this? Is this for the product specifically or?

state-chain/pallets/cf-swapping/src/migrations/swapping_redesign.rs
```diff
@@ -0,0 +1,96 @@
...
+	pub type PendingCcms<T: Config> = StorageMap<Pallet<T>, Twox64Concat, u64, CcmSwap>;
+
+	#[frame_support::storage_alias]
+	pub type CcmOutputs<T: Config> = StorageMap<Pallet<T>, Twox64Concat, u64, CcmSwapOutput>;
+}
+
+impl<T: Config> OnRuntimeUpgrade for Migration<T> {
+	fn on_runtime_upgrade() -> Weight {
+		// Note: we don't migrate items from SwapQueue because we will
+		// ensure that it is empty during the upgrade.
```
kylezs, line 43: This is pretty critical - will make note a note.

state-chain/pallets/cf-swapping/src/migrations/separate_swap_state.rs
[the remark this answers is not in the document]
kylezs: done here: https://github.com/chainflip-io/chainflip-backend/pull/5090

state-chain/runtime/src/migrations/move_network_fees.rs
```diff
@@ -34,7 +34,7 @@ impl frame_support::traits::OnRuntimeUpgrade for NetworkFeesMigration {
 			>(b"CollectedNetworkFee");
 
 			// Bump the version of both pallets
-			StorageVersion::new(5).put::<pallet_cf_swapping::Pallet<Runtime>>();
+			StorageVersion::new(6).put::<pallet_cf_swapping::Pallet<Runtime>>();
```
kylezs, line 37: as above, this is unnecessary - will be fixed once rebased onto main after https://github.com/chainflip-io/chainflip-backend/pull/5090

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1748,61 +1908,19 @@ pub mod pallet {
...
-	}
-}
-
 impl<T: Config> cf_traits::FlipBurnInfo for Pallet<T> {
 	fn take_flip_to_burn() -> AssetAmount {
 		FlipToBurn::<T>::take()
 	}
 }
 
+pub struct NoPendingSwaps<T: Config>(PhantomData<T>);
```
kylezs, line 1917: If we introduce DCA, and a swap is spread over 12 hours or more, into its subswaps which are in the queue, then we can't submit the upgrade until it's done.  We'll likely want to migrate the swaps instead... but if we include this condition, even if we do choose to migrate them, we won't be able to push the upgrade to do so... and if we still allow contract swaps to come through someone could halt our ability to upgrade entirely... which would be near irrecoverable, since we would need to do an upgrade to reverse it.

Note also that this condition will *not* be applied to this current runtime upgrade, since we submit the upgrade using the extrinsic of the *old*/ current runtime.

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: Started a thread on discord.

bouncer/shared/utils.ts
```diff
@@ -294,84 +294,107 @@ export async function observeSwapEvents(
...
+          continue;
+        }
+
         const expectedEvent = {
           data: event.toHuman().data,
         };
 
+        const data = expectedEvent.data;
+
         switch (expectedMethod) {
```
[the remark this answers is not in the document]
msgmaxim: Didn't write this code, but I'm guessing it is because the name is the event is exposed as "method" through substrate's API. I will change to `expectedEvent` to be consistent with the other vars.

bouncer/shared/utils.ts
```diff
@@ -294,84 +294,107 @@ export async function observeSwapEvents(
...
+              egressId = data.egressId as EgressId;
               console.log(`${tag} swap egress scheduled with id: (${egressId[0]}, ${egressId[1]})`);
             }
             break;
           case batchBroadcastRequested:
-            expectedEvent.data.egressIds.forEach((eventEgressId: EgressId) => {
+            for (const eventEgressId of data.egressIds) {
               if (egressId[0] === eventEgressId[0] && egressId[1] === eventEgressId[1]) {
-                broadcastId = [egressId[0], Number(expectedEvent.data.broadcastId)] as BroadcastId;
+                broadcastId = [egressId[0], Number(data.broadcastId)] as BroadcastId;
```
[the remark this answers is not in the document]
msgmaxim: Renamed to `BroadcastChainAndId`.

bouncer/shared/utils.ts
```diff
@@ -294,84 +294,107 @@ export async function observeSwapEvents(
...
+              Number(channel.channelId) === channelId &&
+              Object.values(channel.depositAddress)[0] === depositAddress &&
+              sourceAsset === (data.inputAsset as Asset) &&
+              destAsset === (data.outputAsset as Asset)
+            ) {
+              swapRequestId = data.swapRequestId;
+              expectedMethod = swapScheduledEvent;
+
+              // This ensures that we loop through all events again to find SwapScheduled:
+              keepSearching = true;
```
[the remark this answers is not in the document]
msgmaxim: What do you mean by "recursion" exactly in this case? What would the signature of the function look exactly? One of my goals for this test was to fix it with minimal changes. IMO the concept of repeating the search in case we found something is not too difficult to understand.

Also the reason I don't want to spend time improving this code is that `observeSwapEvents` (along with `performAndTrackSwap` and others) only exist to serve one test and should probably be removed in favour of more general approach for tracking events.

state-chain/cf-integration-tests/src/swapping.rs
```diff
@@ -500,68 +572,14 @@ fn can_process_ccm_via_direct_deposit() {
...
-
-		assert_events_match!(
-			Runtime,
-			RuntimeEvent::EthereumIngressEgress(
-				pallet_cf_ingress_egress::Event::CcmBroadcastRequested {
-					egress_id: actual_egress_id,
-					..
-				},
-			) if actual_egress_id == egress_id => ()
+		// It is sufficient to check that swap is "requested", the rest is
```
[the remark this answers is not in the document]
msgmaxim: Sure, open to doing this in a future PR.

That said, I think it is not unreasonable to have one comprehensive 'happy path" test and letting other tests focus on how their diverge from that core scenario. If we require that all tests assert everything that can be asserted, they will be harder to write/maintain (also reason about) for what I would consider a limited benefit.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1156,131 +1180,202 @@ pub mod pallet {
...
+				// for swaps from STABLE_ASSET:
+				let swap_input = if swap.input_asset() == STABLE_ASSET {
+					swap.stable_amount.unwrap_or_else(|| {
+						log_or_panic!("stable amount must be set for swaps from STABLE_ASSET");
 						swap.input_amount()
-					};
+					})
+				} else {
+					swap.input_amount()
+				};
```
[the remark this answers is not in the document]
msgmaxim: OK, the reason I didn't inline it at first is to have a place for the comment, but looks like it is still works OK inlined.

bouncer/shared/utils.ts
```diff
@@ -391,28 +414,60 @@ export enum SwapType {
...
+
+        return channelMatches && sourceAssetMatches && destAssetMatches && requestTypeMatches;
       }
       // Otherwise it was a swap scheduled by interacting with the Eth smart contract
       return false;
     },
   }).event;
 }
 
+export async function observeSwapScheduled(swapRequestId?: number, swapType?: SwapType) {
```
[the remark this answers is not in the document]
msgmaxim: deleted

kylezs (approved):

PR #5027, opened 2024-07-05
Fix/limit sol batches

ramizhasan111:
# Pull Request

Closes: PRO-1401

## Checklist

## Summary

This introduces a general functionality of limiting the number of fetches and transfers while creating the allbatch call. It also implements the limits for the case of solana. For solana, we want the limit on transfers to be 1 less than the available nonces. The remaining nonce is reserved for fetch tx. We put a limit on fetches so that there is only one fetch tx.

#### Non-Breaking changes

### DIFF
--- a/foreign-chains/solana/sol-prim/src/consts.rs
+++ b/foreign-chains/solana/sol-prim/src/consts.rs
@@ -38,3 +38,7 @@ pub const MAX_TRANSACTION_LENGTH: usize = 1_232;
 pub const NONCE_ACCOUNT_LENGTH: u64 = 80;
 
 pub const SOL_USDC_DECIMAL: u8 = 6u8;
+
+// Due to transaction size limit in Solana, we have a limit on number of fetches in a solana hetch
+// tx
+pub const MAX_SOL_FETCHES_PER_TX: usize = 10;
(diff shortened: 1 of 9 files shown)


albert-llimos (approved):
The thing that I was trying to explain on Friday was that a thing this implementation is not capable of handling is a maximum amount of actions. For example, let's say we have a chain that batches fetches and transfers in AllBatch and that it has a transaction limit so we can't just batch an arbitrary amount of them. With this code we can only limit the specific amount of batches and fetches, while we'd want to limit the total amount of actions. This would be the case, for example, for Solana if we could actually batch fetches and transfers.
Having said that, so far we don't have a chain that has that limitation so I'm happy to merge this.

ramizhasan111:
> The thing that I was trying to explain on Friday was that a thing this implementation is not capable of handling is a maximum amount of actions. For example, let's say we have a chain that batches fetches and transfers in AllBatch and that it has a transaction limit so we can't just batch an arbitrary amount of them. With this code we can only limit the specific amount of batches and fetches, while we'd want to limit the total amount of actions. This would be the case, for example, for Solana if we could actually batch fetches and transfers.
> Having said that, so far we don't have a chain that has that limitation so I'm happy to merge this.

I think that could be done in addition to this by adding a chain-specific limit to the total number of actions (if needed in a future chain).

PR #5028, opened 2024-07-05
feat(custom-rpc): add boost balances to account info

acdibble:
# Pull Request

## Checklist

## Summary

It's currently quite cumbersome for an LP to find their individual boost balances across the pools. This adds that information directly to the account info RPC for easier and direct consumption.

I just named things so they have a name but they can be changed to whatever. The important thing is to make it easier for LPs to interact with the boost feature.

Tried to test it in a localnet but it wouldn't get past the healthcheck phase :(


state-chain/custom-rpc/src/lib.rs
```diff
@@ -207,7 +220,8 @@ pub enum RpcAccountInfo {
 		balances: any::AssetMap<NumberOrHex>,
 		refund_addresses: HashMap<ForeignChain, Option<ForeignChainAddressHumanreadable>>,
 		flip_balance: NumberOrHex,
-		earned_fees: any::AssetMap<AssetAmount>,
```
acdibble, line 210: this is a breaking change but we don't use `u128`s with the custom rpcs

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1943,6 +1966,20 @@ mod test {
...
+							fee_tier: 5,
+							total_balance: 100_000_000,
+							available_balance: 50_000_000,
+							in_use_balance: 50_000_000,
+						}],
+					},
+					dot: dot::AssetMap { dot: vec![] },
+					arb: arb::AssetMap { eth: vec![], usdc: vec![] },
+					sol: sol::AssetMap { sol: vec![] },
+				},
```
dandanlen, line 1982:
```suggestion
				boost_balances: any::AssetMap {
					btc: btc::AssetMap {
						btc: vec![LiquidityProviderBoostPoolInfo {
							fee_tier: 5,
							total_balance: 100_000_000,
							available_balance: 50_000_000,
							in_use_balance: 50_000_000,
						}],
					},
					..Default::default(),
				},
```

state-chain/runtime/src/lib.rs
```diff
@@ -1600,6 +1600,39 @@ impl_runtime_apis! {
...
+					(
+						asset,
+						pool_details.into_iter().filter_map(|(fee_tier, details)| {
+							let available_balance = details.available_amounts.into_iter().find_map(|(id, amount)| {
+								if account_id == id {
+									Some(amount)
+								} else {
+									None
+								}
+							})?;
```
dandanlen, line 1615: I think you might want to use `unwrap_or_default` here. Otherwise if the whole balance is tied up in pending boosts, BoostPoolInfo won't show up.

state-chain/runtime/src/lib.rs
```diff
@@ -1600,6 +1600,39 @@ impl_runtime_apis! {
 					(asset, pallet_cf_lp::FreeBalances::<Runtime>::get(&account_id, asset).unwrap_or(0))
 				).collect(),
 				earned_fees: AssetMap::from_iter(pallet_cf_lp::HistoricalEarnedFees::<Runtime>::iter_prefix(&account_id)),
+				boost_balances: AssetMap::from_iter(Asset::all().map(|asset| {
```
dandanlen, line 1603: This can be condensed a little, AssetMap has a constructor for this: 

```suggestion
				boost_balances: AssetMap::from_fn(|asset| {
```

state-chain/runtime/src/lib.rs
acdibble: I checked that case, it returns an available amount of `0`

dandanlen:
@acdibble fyi it looks like there are some conflicts and some clippy issues - LMK if you want me to take a look.

state-chain/runtime/src/lib.rs
```diff
@@ -1635,6 +1635,43 @@ impl_runtime_apis! {
...
+						if total_balance == 0 {
+							return None
+						}
+
+						Some(LiquidityProviderBoostPoolInfo {
+							fee_tier,
+							total_balance,
+							available_balance,
+							in_use_balance: owed_amount,
+							is_boosting: details.pending_withdrawals.keys().any(|id| *id == account_id),
```
kylezs, line 1671:
```suggestion
							is_withdrawing: details.pending_withdrawals.keys().any(|id| *id == account_id),
```

PR #5029, opened 2024-07-08
Feat/solana bouncer testing

albert-llimos:
# Pull Request

Closes: PRO-1264

## Checklist

## Summary

- Add support for bs58 encoded addreses when requesting a deposit channel
- Update SDK to 1.4.0-solana.5
- Update bouncer accordingly
- Increase nonces to 8
- Add check for deposit addresses to be fetched for EVM chains and Solana, which proactively fetch amounts
- Add Solana swapping tests for regular swaps (no CCM and no contract swaps)
- Fixed nonce bug in the Allbatch transfer - it was using the same nonce for all transfers in the same Allbatch calls even though they are different transactions.
- Improve CFE to not do RPC calls if the vectors of accounts are empty
- Reduce maximum amount of fetches per batch to 5 to not hit the transaction length limit.
- Update initial durable hashes to the correct ones


ramizhasan111:
Would let someone else more familiar with the bouncer code review it. Otherwise LGTM

ramizhasan111 (approved):

PR #5031, opened 2024-07-09
fix: use rpath in engine dylib

dandanlen:
This fixes an issue with the engine dylib file that was preventing it from running on mac localnets. 

The dylib install path was previously set to a local path on my machine.
Now it's correctly set to a relative path. 

The engine lib build script already do this correctly on main, but that change hasn't been ported to the release branch. Hence the release version 1.4.5 libs were compiled with absolute paths instead of relative paths. 

Before:

```bash
otool -L old-engine-dylib/libchainflip_engine_v1_4_5.dylib | head -2
old-engine-dylib/libchainflip_engine_v1_4_5.dylib:
        /Users/dan/[...]/old-engine-dylib/libchainflip_engine_v1_4_5.dylib (compatibility version 0.0.0, current version 0.0.0)
```

After:

```bash
otool -L old-engine-dylib/libchainflip_engine_v1_4_5.dylib | head -2
old-engine-dylib/libchainflip_engine_v1_4_5.dylib:
        @rpath/old-engine-dylib/libchainflip_engine_v1_4_5.dylib (compatibility version 0.0.0, current version 0.0.0)
```


kylezs:
It should be `@rpath/libchainflip-engine...` and not `@rpath/old-engine-dylib/libchainflip-engine...`

kylezs:
FYI - this won't need to be set manually after 1.5 - as it's now part of the build. `rpath = true` in the Cargo.toml now.

PR #5035, opened 2024-07-11
chore: solana bouncer ccm

albert-llimos:
# Pull Request

Closes: PRO-1484

## Checklist

## Summary

After some changes in the SDK I've added:
- Support for regular CCM swaps in the tests
- Support for CCM swaps via contract call (EVM -> Solana)

Add CCM bouncer tests with CCM receiver contract event verification (Cf Tester)
Fixed source Chain encoding to match the Solana bytes encoding


state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -269,7 +278,7 @@ impl SolanaInstructionBuilder {
 			VaultProgram::with_id(vault_program)
 				.execute_ccm_native_call(
 					source_chain as u32,
-					source_address.encode(), // TODO: check if this is correct (scale encoding?)
+					Self::destructure_address(source_address),
```
albert-llimos, line 281: @dandanlen This encoding you added (I believe) doesn't work with Solana, it ends up with garbage bytes on chain

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -48,6 +46,17 @@ pub struct SolanaInstructionBuilder;
 const COMPUTE_LIMIT: SolComputeLimit = 300_000u32;
 
 impl SolanaInstructionBuilder {
+	// TODO: Shares code with EVM's ExecutexSwapAndCall or make it a function of ForeignChainAddress
+	fn destructure_address(source_address: Option<ForeignChainAddress>) -> Vec<u8> {
```
albert-llimos, line 50: This is also used in the EVM's ExecutexSwapAndCall. Should we have some common code or maybe just have it as a function of the ForeignChainAddress struct? @dandanlen @ramizhasan111

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1304,11 +1304,17 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 	fn do_egress_scheduled_ccm() {
+		let mut maybe_no_of_transfers_remaining =
+			T::FetchesTransfersLimitProvider::maybe_transfers_limit();
+
 		let ccms_to_send: Vec<CrossChainMessage<T::TargetChain>> =
 			ScheduledEgressCcm::<T, I>::mutate(|ccms: &mut Vec<_>| {
 				// Filter out disabled assets, and take up to batch_size requests to be sent.
-				ccms.extract_if(|ccm| !DisabledEgressAssets::<T, I>::contains_key(ccm.asset()))
-					.collect()
+				ccms.extract_if(|ccm| {
```
albert-llimos, line 1313: @ramizhasan111 please review this as it's in the lines of the other PR.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1304,11 +1304,17 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+					!DisabledEgressAssets::<T, I>::contains_key(ccm.asset()) &&
+						Self::should_fetch_or_transfer(&mut maybe_no_of_transfers_remaining)
+				})
```
ramizhasan111, line 1316: we can do better here by utilizing one more nonce since at this point we dont have to reserve one nonce for fetches. But I think we can leave it as is and leave this optimization (and potentially other optimizations) for when we actually do start hitting limits of parallel sol transactions, that is, we actually do have that much tx volume.

state-chain/chains/src/sol/instruction_builder.rs
ramizhasan111: We can already do this using the ```to_encoded_address``` function implemented for the ```ForeignChainAddress``` type. We should also probably also use that function for the ```destructure_address``` in evm's ExecutexSwapAndCall.

state-chain/pallets/cf-ingress-egress/src/lib.rs
albert-llimos: yes, technically I believe we could use `maybe_fetches_limit` and check if it's > 0 to get one extra nonce but it's very strange. I agree we can leave it as a future optimization.

ramizhasan111: we can do it in a better way by having a new function on the FetchesTransfersLimitProvider trait which would not affect the general code in ingress egress

state-chain/chains/src/sol/instruction_builder.rs
dandanlen: I don't think I added this - it was already there and I added the TODO comment about whether it should use `encode()` because it didn't make sense to me.

state-chain/chains/src/sol/instruction_builder.rs
dandanlen: I agree, adding this additional struct method seems like overkill.

state-chain/pallets/cf-ingress-egress/src/lib.rs
albert-llimos: Added a ticket as an optimisation: https://linear.app/chainflip/issue/PRO-1497/solana-optimize-nonce-consumption

state-chain/chains/src/sol/instruction_builder.rs
albert-llimos: Seems like unfortunately that doesn't work (Ramiz has also checked). So unless someone opposes to it I'll implement an extra struct method which we can reuse when egressing CCM calls for any chain.

ramizhasan111 (approved):

PR #5036, opened 2024-07-11
feat: sol min egress amount

ramizhasan111:
# Pull Request

Closes: PRO-1426

## Checklist

## Summary

If a sol egress is below a certain amount, we dont egress and swallow the amount.

#### Non-Breaking changes

### DIFF
--- a/foreign-chains/solana/sol-prim/src/consts.rs
+++ b/foreign-chains/solana/sol-prim/src/consts.rs
@@ -1,4 +1,4 @@
-use crate::{Address, Digest};
+use crate::{Address, Amount, Digest};
 use cf_utilities::bs58_array;
 
 pub const SOLANA_SIGNATURE_LEN: usize = 64;
@@ -38,3 +38,5 @@ pub const MAX_TRANSACTION_LENGTH: usize = 1_232;
 pub const NONCE_ACCOUNT_LENGTH: u64 = 80;
 
 pub const SOL_USDC_DECIMAL: u8 = 6u8;
+
+pub const MIN_EGRESS_AMOUNT: Amount = 0; // todo
--- a/foreign-chains/solana/sol-prim/src/consts.rs
+++ b/foreign-chains/solana/sol-prim/src/consts.rs
@@ -39,4 +39,4 @@ pub const NONCE_ACCOUNT_LENGTH: u64 = 80;
 
 pub const SOL_USDC_DECIMAL: u8 = 6u8;
 
-pub const MIN_EGRESS_AMOUNT: Amount = 1; // todo
+pub const MIN_EGRESS_AMOUNT: Amount = 890880;
--- a/foreign-chains/solana/sol-prim/src/consts.rs
+++ b/foreign-chains/solana/sol-prim/src/consts.rs
@@ -39,4 +39,4 @@ pub const NONCE_ACCOUNT_LENGTH: u64 = 80;
 
 pub const SOL_USDC_DECIMAL: u8 = 6u8;
 
-pub const MIN_EGRESS_AMOUNT: Amount = 0; // todo
+pub const MIN_EGRESS_AMOUNT: Amount = 1; // todo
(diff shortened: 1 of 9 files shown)


albert-llimos:
The same thing should be implemented for CCM transfers. 
I will update the `MIN_EGRESS_AMOUNT`

dandanlen (commented):
I think we should emit an event to notify that the egress was below a minimum amount, otherwise it will be impossible for the product to track/display when this happens.

ramizhasan111:
> I think we should emit an event to notify that the egress was below a minimum amount, otherwise it will be impossible for the product to track/display when this happens.

is that necessary to display on the product side? I think we also have this for polkadot (existential deposit) and if i understand correctly, we dont display this on product.

albert-llimos:
> > I think we should emit an event to notify that the egress was below a minimum amount, otherwise it will be impossible for the product to track/display when this happens.
> 
> is that necessary to display on the product side? I think we also have this for polkadot (existential deposit) and if i understand correctly, we dont display this on product.

I asked exactly that to Dan before he left on vacations and he said: "I can't remember precisely, but I think that for Polkadot we still send the egress, and it gets swallowed by Polkadot."
If it's swallowed by Polkadot then we are not swallowing it in Chainflip itself and then it doesn't feel mandatory to have one. However, in this case we are keeping it in the Chainflip and not egressing it and therefore we should have a way for the Product to know.
Fwiw I did some testing and now that we send transfers separately it could be an option to just not have a minimum egress and let them revert if the amount is very small. The issue I see there is that the nonce consumption is a bottleneck so trying to avoid unnecessary ceremonies seems worth it for amounts that are very small (~0.13$) and have a chance of failing. Let me know if you disagree.

ramizhasan111:
> > > I think we should emit an event to notify that the egress was below a minimum amount, otherwise it will be impossible for the product to track/display when this happens.
> > 
> > 
> > is that necessary to display on the product side? I think we also have this for polkadot (existential deposit) and if i understand correctly, we dont display this on product.
> 
> I asked exactly that to Dan before he left on vacations and he said: "I can't remember precisely, but I think that for Polkadot we still send the egress, and it gets swallowed by Polkadot." If it's swallowed by Polkadot then we are not swallowing it in Chainflip itself and then it doesn't feel mandatory to have one. However, in this case we are keeping it in the Chainflip and not egressing it and therefore we should have a way for the Product to know so I agree we should have an event. Fwiw I did some testing and now that we send transfers separately it could be an option to just not have a minimum egress and let them revert if the amount is very small. The issue I see there is that the nonce consumption is a bottleneck so trying to avoid unnecessary ceremonies seems worth it for amounts that are very small (~0.13$) and have a chance of failing. Let me know if you disagree.

I would tend towards letting the egress revert on chain for a few reasons. First, this would keep things symmetric. Since we already do that for polkadot, it makes sense that the protocol has the same behaviour for other chains. Secondly, i think it would be better to justify us egressing and it failing instead of swallowing it since from the protocol's point of view, it tried to egress which it should have done. Thirdly, we can already set a Minimum deposit for any asset via governance and we do it so that negligible value swaps dont flow through chainflip because of the reason you mentioned. I think the very small value sol egress case you mentioned would be covered there. Let me know if it makes sense.

albert-llimos:
> I would tend towards letting the egress revert on chain for a few reasons. First, this would keep things symmetric. Since we already do that for polkadot, it makes sense that the protocol has the same behaviour for other chains. Secondly, i think it would be better to justify us egressing and it failing instead of swallowing it since from the protocol's point of view, it tried to egress which it should have done. Thirdly, we can already set a Minimum deposit for any asset via governance and we do it so that negligible value swaps dont flow through Chainflip because of the reason you mentioned. I think the very small value sol egress case you mentioned would be covered there. Let me know if it makes sense.

That makes sense to me. The current approach was mainly thinking that we wanted to batch transfers and therefore they shouldn't fail, which is not true anymore, and that in Polkadot we do that.
The downside here, as I pointed out, is that we will do an extra signing ceremony and consume a nonce for an amount of time, which is quite annoying in Solana since we have limited nonces. But it's not even that this minimum egress solves a potential attack vector, just makes it a tiny bit more expensive.

I think I'm happy closing this, sorry for maybe not reevaluating this before.

PR #5045, opened 2024-07-16
fix: allow validators to deregister once they no longer have relevant key material

martin-chainflip:
# Pull Request

Closes: PRO-1503

## Checklist

## Summary

https://github.com/chainflip-io/chainflip-backend/blob/main/state-chain/pallets/cf-validator/src/lib.rs#L752
The corresponding code checks the range last_expired..=current but should check last_expired+1..=current

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -749,7 +749,7 @@ pub mod pallet {
 			>>::from_ref(&account_id);
 
 			ensure!(
-				(LastExpiredEpoch::<T>::get()..=CurrentEpoch::<T>::get())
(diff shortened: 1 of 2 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -749,7 +749,7 @@ pub mod pallet {
 			>>::from_ref(&account_id);
 
 			ensure!(
-				(LastExpiredEpoch::<T>::get()..=CurrentEpoch::<T>::get())
+				(LastExpiredEpoch::<T>::get() + 1..=CurrentEpoch::<T>::get())
```
kylezs, line 752: we should add a test for this too

martin-chainflip: done

kylezs (approved):

PR #5048, opened 2024-07-16
Feat/solana ccm gas limit

albert-llimos:
# Pull Request

Closes: PRO-1479 , PRO-1493

## Checklist

## Summary

Added gas limit calculation logic and tests in the bouncer. I had to refactor the `gas_limit_ccm` bouncer test, otherwise it became just too confusing to follow along the swaps for different chains.


state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -340,13 +343,38 @@ impl SolanaInstructionBuilder {
...
+		let budget_after_signature = gas_budget.saturating_sub(LAMPORTS_PER_SIGNATURE);
+		if compute_price == 0 {
+			return DEFAULT_COMPUTE_UNITS_PER_CCM_TRANSFER;
+		}
+		let compute_budget =
+			// Budget is in lamports, compute price is in CU/microlamport
+			sp_std::cmp::min(
+				MAX_COMPUTE_UNITS_PER_TRANSACTION as u128,
+				(budget_after_signature as u128 * MICROLAMPORTS_PER_LAMPORT as u128)
+					/ (compute_price as u128),
```
martin-chainflip, line 369: If compute_price really is "compute units per microlamport", then this should be a multiplication, not a divison

albert-llimos: The comment is what's wrong, `compute_price` is microlamports per compute unit. I'll fix it.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -340,13 +343,38 @@ impl SolanaInstructionBuilder {
...
+		if compute_price == 0 {
+			return DEFAULT_COMPUTE_UNITS_PER_CCM_TRANSFER;
+		}
+		let compute_budget =
+			// Budget is in lamports, compute price is in microlamports/CU
+			sp_std::cmp::min(
+				MAX_COMPUTE_UNITS_PER_TRANSACTION as u128,
+				(budget_after_signature as u128 * MICROLAMPORTS_PER_LAMPORT as u128)
+					/ (compute_price as u128),
+			) as SolComputeLimit;
```
ramizhasan111, line 370: what is the difference between the gas_budget provided while constructing the tx and the hardcoded limit of ```MAX_COMPUTE_UNITS_PER_TRANSACTION```? arent both limitation set by us? If yes, why do we need two different limtis?

albert-llimos: `gas_budget` is in a CCM the amount of native token on the egress chain (SOL in this case) that the user has paid for. In a CCM we should never spend more than that amount in fees.
`MAX_COMPUTE_UNITS_PER_TRANSACTION` is the maximum amount of CUs that a transaction can consume, which is a limit set by the SVM (Solana), not by us.

ramizhasan111: ok, so even if the user pays for a very a high budget, we cant have the budget implied compute units exceed the hardcoded limit? So in case that turns out to be the case, I guess we just swallow the extra budget provided by the user? is that correct?

albert-llimos: yes exactly that :+1:

ramizhasan111 (approved):

PR #5055, opened 2024-07-18
feat: CCM validity checker

syan095:
# Pull Request

Closes: PRO-1452
Closes: PRO-1502
Closes: PRO-1491

## Checklist

## Summary

Added a CCM message checker trait, and implemented Solana version of the CCM validity checker.
CCM messages lengths are checked when:
* opening deposit channel 
* On ccm deposit 

Accounts provided in CCM's cf_parameter are checked against a list of blacklisted accounts when building CcmTransfer API call.

On failure, a message is logged and an error is returned.
On CCM call build failure, the error is included in the event emitted


albert-llimos:
In general the approach looks good to me. There's just one thing that I'd say needs to be changed.
```
pub const MAX_CCM_MESSAGE_LENGTH_SOL: usize = 1_000usize;
pub const MAX_CCM_MESSAGE_LENGTH_USDC: usize = 1_000usize;
pub const MAX_CCM_EXTRA_ACCOUNTS: usize = 5usize;
```
I think that  instead of that we should have something like this (naming can be different):
```
pub const CCM_BYTES_PER_ACCOUNT: usize = 33usize;
pub const MAX_CCM_BYTES_USDC: usize = 1_000usize;
pub const MAX_CCM_BYTES_SOL: usize = 1_000usize;
```
So then instead of checking the lenght of the message and the number of extra accounts we need to check that the total amount of bytes doesn't exceed the limits. This is because the limit here is due to exceeding the transaction length and both of the parameters contribute to it.

 The pseudocode for the check should be this (dependant on the asset ofc):
```
if ccm.message.len() + ccm_accounts.remaining_accounts.len()  * CCM_BYTES_PER_ACCOUNT > MAX_CCM_BYTES_* {
	return Error
}
```
We could leave the separate check for each of them (ccm message length and remaining_accounts * bytes_per_account) against the MAX_CCM_BYTES_* if we thinkg that it can be useful to indicate if one of them by itself is already exceeding the limit.

foreign-chains/solana/sol-prim/src/consts.rs
```diff
@@ -34,11 +34,16 @@ pub const BPF_LOADER_UPGRADEABLE_ID: Address =
...
 
-pub const NONCE_ACCOUNT_LENGTH: u64 = 80;
+pub const NONCE_ACCOUNT_LENGTH: u64 = 80u64;
 
 pub const SOL_USDC_DECIMAL: u8 = 6u8;
+
+// TODO: Confirm the value
+pub const CCM_BYTES_PER_ACCOUNT: usize = 33usize;
+pub const MAX_CCM_BYTES_USDC: usize = 1_000usize;
+pub const MAX_CCM_BYTES_SOL: usize = 1_000usize;
```
albert-llimos, line 49: I'd move these three values to `state-chain/chains/src/sol.rs` as we do for `MAX_SOL_FETCHES_PER_TX` since these are not Solana specific values but rather our own parameters. Not technically true for the last one but let's move it there too because it's not obvious that it's not a parameter of ours.

state-chain/chains/src/sol/api.rs
```diff
@@ -251,6 +253,30 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
 		cf_parameters: Vec<u8>,
 	) -> Result<Self, SolanaTransactionBuildingError> {
+		// Verify the validity of the CCM message before building the call.
+		CcmValidityChecker::<Environment>::is_valid(
+			&CcmChannelMetadata {
+				message: message
+					.clone()
+					.try_into()
+					.expect("This is parsed from bounded vec, therefore the size must fit"),
+				gas_budget: 0,
```
albert-llimos, line 263: Having gas_budget set to zero here looks very strange. I understand that this is because it's very convenient to have the `CcmChannelMetadata` be the type passed to `is_valid`. And then my understanding is that we don't pass the gas_budget we have here because that is different from the `gas_budget` from when opening the deposit channel. If that's true, that's fine for me but maybe let's add a comment explaining it.

albert-llimos (commented):
We currently have this in `state-chain/chains/src/lib.rs`:
```
// TODO: PRO-1491
pub const MAX_CCM_MSG_LENGTH: u32 = 10_000;
pub const MAX_CCM_CF_PARAM_LENGTH: u32 = 1_000;
```
I presume that will enforce that the message and cf_params will always be within those limits for all chains, as that is bounding the type itself.
If that's right, I'm happy leaving those ones as is as general limits. We can then check them on a per-chain basis now that we have the ccm checker functionality.
If you agree with that then go ahead and remove that TODO comment.

albert-llimos:
Regarding my last commit, I just updated the limits and since the limits are easier to calculate with the `cf_receiver` included, I updated the check to only the remaining accounts.

foreign-chains/solana/sol-prim/src/consts.rs
syan095: OK. Moved to cf_chains/sols module.

state-chain/chains/src/sol/api.rs
syan095: Added a comment that clarified that the gas_budget is not used by Solana

state-chain/chains/src/ccm_checker.rs
```diff
@@ -0,0 +1,332 @@
...
+					return Err(CcmValidityError::CfParametersContainsInvalidAccounts)
+				}
+			};
+
+			Ok(())
+		} else {
+			Ok(())
+		}
+	}
+}
```
ramizhasan111, line 60: I dont think we need to make the ```CcmValidityChecker``` generic over the Environment. We inject it in the swapping pallet which is not instantiable and so doesnt need to be bound to a specific chain. If we need a checker for any other chain we have to add the logic of the checker in the same type (because swapping pallet is not instantiable). So basically, the same ```is_valid``` function will have the logic for the checker for other chains as well with a match statement.

ramizhasan111:
Also, I dont understand why we need to check the ccm validity (the same check) at 3 different stages of a ccm swap? checking it at any one of the 3 places would stop the swap. Is there a particular reason for this?

albert-llimos:
> Also, I dont understand why we need to check the ccm validity (the same check) at 3 different stages of a ccm swap? checking it at any one of the 3 places would stop the swap. Is there a particular reason for this?

We need to perform the check when building the transaction as that is the safest place to do it. Then, if we don't perform the checks when opening a deposit channel it means that we will accept opening deposit channels for transactions that will deterministically fail (too long, bad encoding, etc...). That seems like terrible user experience and a way for user's to lose funds, which we want to avoid.
Also, contract swaps (when implemented) bypass the opening deposit channels.
That's why the check is in multiple places.

ramizhasan111:
> > Also, I dont understand why we need to check the ccm validity (the same check) at 3 different stages of a ccm swap? checking it at any one of the 3 places would stop the swap. Is there a particular reason for this?
> 
> We need to perform the check when building the transaction as that is the safest place to do it. Then, if we don't perform the checks when opening a deposit channel it means that we will accept opening deposit channels for transactions that will deterministically fail (too long, bad encoding, etc...). That seems like terrible user experience and a way for user's to lose funds, which we want to avoid. Also, contract swaps (when implemented) bypass the opening deposit channels. That's why the check is in multiple places.

Ok, I think in that case we can split the checks and do them in parts because currently we are repeating the same checks 3 times in the lifetime of a swap. So, the stuff that can be checked on request_deposit_address (which is basically almost all of it except the aggkey check i guess) should be checked there and then while constructing the tx, we only check the rest of the stuff (the aggkey check). This could be done by having multiple functions in the ```CcmCheckValidity``` trait instead of one.

Regarding the aggkey check that is safer to do at the tx building level, I can think of a remote edge case that someone could initiate a ccm swap during the keygen verification but before successful keygen so that he knows the next aggkey but the key is still not confirmed on the state chain. But this would be super hard to execute. Is there any other reason the check while building the tx is safer?

ramizhasan111:
> It's only two times, isn't it? When requesting deposit address and when building the transaction.
> While I do agree that doing it two times is not ideal, I still think that we prevent users opening channels with bad parameters that will just fail when money is sent to the channel. That's why I'm in favor of doing the checks in both places (opening deposit channel and building the transaction).

Ah yes, in normal swaps its two times. The third time is actually for the contract swaps case. So basically, we check it two times for both normal swaps and contract swaps.

I agree with checking it on requesting deposit address. What I meant was maybe we check it only there and not check it while building the tx.

> Also because contract swaps (when implemented) bypass the request of deposit addresses. So having the checks when building the transaction ensures it always gets checked.

Currently, for the case of contract swaps as well we check it two times, one on witnessing the deposit and one while tx building. So in both kinds of swaps we do it twice. So basically, the check while building tx is redundant in both cases.

> The thing that could make sense to me is to not do the aggKey check on the request of deposit address, only on transaction building. This way we simplify the ccm validity checker as it then doesn't have to access the environment. That is ok for me because that is some potentious malicious use. However, the other checks (lengths mainly) are legit use cases that is might be difficult for users to get exactly right.

Yeah, I completely agree with the checks, just suggested to not do it multiple times.

albert-llimos:
> Currently, for the case of contract swaps as well we check it on deposit witness so we wouldnt bypass the check. So in both kinds of swaps we do it twice. So basically, the check while building tx is redundant in both cases.
>  I completely agree with the checks, just suggested to not do it multiple times.

Yes, I understand and if possible we'd definitely want to avoid that.

From my end I the solution should fulfill:
- All checks (if possible) should be done when requesting a deposit address to prevent users from losing money.
- In the Instruction builder we should check at least the aggKey and token vault (addresses in cf_parameters).
- Contract swaps should also be checked in the same way. In that case it's not so important to check it on ingress since at that point the funds have already been transferred so it doesn't matter if it fails straight away or it fails when building the transaction.
- The checks that are done more than once should be shared code to avoid issues.

> Currently, for the case of contract swaps as well we check it on deposit witness so we wouldnt bypass the check. So in both kinds of swaps we do it twice. So basically, the check while building tx is redundant in both cases.

If I understand you correctly, you are saying that on deposit witness we do that same checks as requesting a deposit address. If so, the solution could be:
- Check decoding of cf_parameters and message length there.
- Check addresses (aggKey) at the instruction builder
The thing there is that then we need to pass the decoded cf_parameters all along to not have to decode them again.

Does that make sense?

syan095:
> * Check decoding of cf_parameters and message length there.
> * Check addresses (aggKey) at the instruction builder
>   The thing there is that then we need to pass the decoded cf_parameters all along to not have to decode them again.
> 
> Does that make sense?

Ok I'll make the appropriate changes. Checking aggkey only when building the transaction will remove the need for the Generic Environment, as Ramiz pointed out.

albert-llimos:
> For this, It's very easy to decode data, but much harder to add in another chain-specific type to a pallet that is shared across all chains. so I think it would be easier to keep the cf_parameters as generic bytes until when we use it.

That's what I thought. We shall have decoding and length check when opening the deposit channel for UX. Therefore, let's have that one in both places. As you pointed out it shouldn't be expensive computationally.

Then the only question left is the `aggKey`and `token_vault_pda_account` check. It should definitely be checked when building the call for safety. Then, the problem there is that if we don't check when opening a deposit channel and it then fails to build, the UX is terrible (and money is lost). I honestly struggle to imagine a non-malicious scenario where anyone would pass any of those two accounts and that's why I'd be okay with putting that check only as part of the transaction building, especially if it simplifies the logic (the `CcmValidityChecker` not needing the environment) and/or significantly decreases the compute. Also, CCMs are not to be used by naive users but rather more sophisticated users, so it should be alright to have this.

After speaking to Roy we ended up agreeing on removing the key checks from requesting deposit channel so the final implementation will be:
- Decoding `cf_parameters` and length check in both deposit channel opening request and transaction builder.
- Check of `cf_parameters`'s accounts against the current `agg_key` and `token_vault_pda` in the instruction builder.

albert-llimos:
LGTM! I'll let Ramiz do the final review.

state-chain/chains/src/sol/api.rs
```diff
@@ -251,11 +253,36 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
+					.clone()
+					.try_into()
+					.expect("This is parsed from bounded vec, therefore the size must fit"),
+			},
+			transfer_param.asset.into(),
+		)
+		.map_err(SolanaTransactionBuildingError::InvalidCcm)?;
+
 		let ccm_accounts = SolCcmAccounts::decode(&mut &cf_parameters[..])
 			.map_err(|_| SolanaTransactionBuildingError::CannotDecodeCcmCfParam)?;
```
ramizhasan111, line 274: we are decoding the same ```cf_parameters``` two times here. one inside the ```is_valid()`` and one right after it. This is not ideal but I think deduplicating it might take an annoying refactor so I am also ok with leaving it as is.

state-chain/chains/src/sol/api.rs
```diff
@@ -251,11 +253,36 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
 		message: Vec<u8>,
 		cf_parameters: Vec<u8>,
 	) -> Result<Self, SolanaTransactionBuildingError> {
+		// Verify the validity of the CCM message before building the call.
+		CcmValidityChecker::is_valid(
```
ramizhasan111, line 257: I had a discussion with @albert-llimos regarding the ```is_valid()``` check here. This check is redundant here since we do the same check while doing request swap but I guess we want to have it as an extra safety layer. In this case I would add a comment mentioning that this check has already been done during the request swap but we do it again here as extra layer of safety. This would improve readability as someone might get confused why we do this check again.

ramizhasan111:
couple of comments, otherwise LGTM! will approve

ramizhasan111 (approved):

state-chain/chains/src/sol/api.rs
albert-llimos: And also we end up using a `ccm_accounts` that is decoded outside the `CcmValidityChecker` so we are not ensuring it's the same. This has very little room for going wrong right now but if we use this in the future for other chains the room for error increases.
We could deduplicate by having the `CcmValidityChecker` return the decoded cf_parameters with a different type per each chain or we could have a Validity check per each chain, but both options require some refactoring.
Another option could be to return an Enum. While it's not great, it adds some safety without much refactoring, something like this:
https://github.com/chainflip-io/chainflip-backend/compare/bc3923fe9362d1dcfc4ab15ba79041ba07307fd4..2f928397dec621e14ac13f75ba72a23f4406d3fe
Or we can just leave it as is and then in the future if/when we use this for other chains we can refactor it as it makes more sense depending on the requirements.

albert-llimos:
Let's discuss this comment before we merge: https://github.com/chainflip-io/chainflip-backend/pull/5055#discussion_r1701892245

state-chain/chains/src/sol/api.rs
syan095: Yeah good point. 
I also agree that this is the best approach here. I've made the appropriate changes.

state-chain/chains/src/sol/api.rs
syan095: Yep Updated the comments here

state-chain/chains/src/sol/api.rs
ramizhasan111: looks good now!

state-chain/chains/src/ccm_checker.rs
```diff
@@ -0,0 +1,317 @@
...
+	/// For Solana Chain: Performs decoding of the `cf_parameter`, and checks the expected length.
+	/// Returns the decoded `cf_parameter`.
+	fn check_and_decode(
+		ccm: &CcmChannelMetadata,
+		egress_asset: Asset,
+	) -> Result<DecodedCfParameters, CcmValidityError> {
+		if ForeignChain::from(egress_asset) == ForeignChain::Solana {
+			// Check if the cf_parameter can be decoded
+			let ccm_accounts = SolCcmAccounts::decode(&mut &ccm.cf_parameters.clone()[..])
+				.map_err(|_| CcmValidityError::CannotDecodeCfParameters)?;
```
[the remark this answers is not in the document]
albert-llimos: We can choose whichever encoding we want, as those will be arbitrary bytes regardless of where they come from (deposit channel or contract swap).
I assumed that scale-encoding was what is more convenient for the State Chain but I must say it was a bit annoying to do in typescript. I tried using the parity `scale-coded` package and hit some issues (can't recall what) and I ended up manually encoding it. I didn't spend much time tbh.
In any case, I'd say fine unless you have a better suggestion.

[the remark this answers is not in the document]
dandanlen: Scale encoding is convenient for us, but if we expect external parties to be able to support this and be able to pass arbitrary data, I would suggest a more universal format. Or for example we might decide that it's based on the native encoding of the external chain's accounts (does that make sense?) and it's the engine's responsibility to transcode this to scale. 

I don't it matters *that* much as long as we are clear about it and make a deliberate choice.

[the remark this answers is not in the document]
albert-llimos: The reason why maybe I wouldn't encode it according to the external chain is that it's still a bit unclear what we will use the `cf_parameters` for (besides this use) and it might have uses unrelated to the external destination chain.
Maybe some more universal format would make more sense, I'd just prefer something not very verbose since in contract swaps that will be on-chain data.
Whatever can work better and your end and that is easy to use for integrators/SDK. Let me know what would be a good option and we will have to also update the bouncer accordingly.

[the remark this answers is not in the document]
dandanlen: For general visibility - @albert-llimos and I discussed this and agreed scale encoding makes sense.

PR #5052, opened 2024-07-17
chore(bouncer): Install fok sdk version

niklasnatter:
# Pull Request

Closes: PRO-xxx

## Checklist

Please conduct a thorough self-review before opening the PR.

- [ ] I am confident that the code works.
- [ ] I have updated documentation where appropriate.

## Summary

*Please include a succinct description of the purpose and content of the PR. What problem does it solve, and how? Link issues, discussions, other PRs, and anything else that will help the reviewer.*

#### Non-Breaking changes

If this PR includes non-breaking changes, select the `non-breaking` label. On merge, CI will automatically cherry-pick the commit to a PR against the release branch.

PR #5056, opened 2024-07-18
test: Bouncer Fill or Kill test

j4m1ef0rd:
# Pull Request

Closes: PRO-1438, PRO-1500

## Checklist

## Summary

Might need #5052 to be merged first.

- changed broker to always use the `request_swap_deposit_address_with_affiliates` call.
- Added refund params to the bouncer new swap function
- new bouncer test that uses refund params with an unrealistic minimum price.

### DIFF
--- a/bouncer/package.json
+++ b/bouncer/package.json
@@ -6,7 +6,9 @@
     "prettier:write": "prettier --write ."
   },
   "dependencies": {
-    "@chainflip/cli": "1.3.0",
+    "@chainflip/cli": "1.5.0-fok.0",
+    "@chainflip/bitcoin": "0.2.2",
+    "@chainflip/rpc": "1.4.10",
     "@iarna/toml": "^2.2.5",
     "@polkadot/api": "^11.3.1",
     "@polkadot/keyring": "12.6.2",
--- a/bouncer/shared/fill_or_kill.ts
+++ b/bouncer/shared/fill_or_kill.ts
@@ -1,0 +1,127 @@
+import { InternalAsset as Asset, InternalAssets as Assets } from '@chainflip/cli';
+import assert from 'assert';
+import {
+  amountToFineAmount,
+  amountToFineAmountBigInt,
+  assetDecimals,
+  decodeDotAddressForContract,
+  isWithinOnePercent,
+  newAddress,
+  observeBalanceIncrease,
+} from './utils';
+import { requestNewSwap } from './perform_swap';
+import { send } from './send';
+import { getBalance } from './get_balance';
+import { getChainflipApi, observeEvent } from './utils/substrate';
+import { RefundParameters } from './new_swap';
+
+/// Do a swap with unrealistic minimum price so it gets retried and then refunded.
+async function testMinPriceRefund(asset: Asset, amount: number) {
+  const swapAsset = asset === Assets.Usdc ? Assets.Flip : Assets.Usdc;
+  const refundAddress = await newAddress(asset, 'FOK_REFUND');
+  const destAddress = await newAddress(swapAsset, 'FOK');
+  console.log(`Swap destination address: ${destAddress}`);
+
+  const refundBalanceBefore = await getBalance(asset, refundAddress);
+
+  const invalidMinPrice = amountToFineAmount(
+    '99999999999999999999999999999999999999999999999999999',
+    assetDecimals(asset),
+  );
+
+  await using chainflip = await getChainflipApi();
+  const currentRetryDuration = Number(await chainflip.query.swapping.swapRetryDelay());
+  console.log(`Current swap retry delay: ${currentRetryDuration} blocks`);
+
+  const refundParameters: RefundParameters = {
+    retryDuration: 1.6 * currentRetryDuration, // More than the swap retry delay, so it will be retried once.
+    refundAddress:
+      asset === Assets.Dot ? decodeDotAddressForContract(refundAddress) : refundAddress,
+    minPrice: invalidMinPrice,
+  };
+
+  console.log(`Requesting swap from ${asset} to ${swapAsset} with unrealistic min price`);
+  const swapRequest = await requestNewSwap(
+    asset,
+    swapAsset,
+    destAddress,
+    'FoK_Test',
+    undefined, // messageMetadata
+    0, // brokerCommissionBps
+    false, // log
+    0, // boostFeeBps
+    refundParameters,
+  );
+  const depositAddress = swapRequest.depositAddress;
+  const depositChannelId = swapRequest.channelId;
+
+  const observeSwapScheduled = observeEvent('swapping:SwapScheduled', {
+    test: (event) => event.data.origin.DepositChannel.channelId === depositChannelId.toString(),
+  }).event;
+
+  // Deposit the asset
+  await send(asset, depositAddress, amount.toString());
+  console.log(`Sent ${amount} ${asset} to ${depositAddress}`);
+
+  const swapId = Number((await observeSwapScheduled).data.swapId);
+  console.log(`Swap scheduled, swapId: ${swapId}`);
+
+  // TODO: Observing after the SwapScheduled event means its possible to miss the events, but we need to the swap id.
+  const observeRefundEgressScheduled = observeEvent(`swapping:RefundEgressScheduled`, {
+    test: (event) => Number(event.data.swapId) === swapId,
+  }).event;
+  const observeSwapExecuted = observeEvent(`swapping:SwapExecuted`, {
+    test: (event) => Number(event.data.swapId) === swapId,
+  }).event;
+  const observeSwapRescheduled = observeEvent(`swapping:SwapRescheduled`, {
+    test: (event) => Number(event.data.swapId) === swapId,
+  }).event;
+
+  const firstSwapEvent = await Promise.race([
+    observeSwapRescheduled,
+    observeSwapExecuted,
+    observeRefundEgressScheduled,
+  ]);
+  if (firstSwapEvent.name.method === 'SwapExecuted') {
+    throw new Error('Swap was executed instead of failing and being retried');
+  } else if (firstSwapEvent.name.method !== 'SwapRescheduled') {
+    throw new Error('Swap was not retried once as expected');
+  }
+  console.log('Swap was rescheduled');
+
+  const secondSwapEvent = await Promise.race([observeRefundEgressScheduled, observeSwapExecuted]);
+  if (secondSwapEvent.name.method === 'SwapExecuted') {
+    throw new Error('Swap was executed instead of being refunded');
+  }
+
+  const refundBalanceAfter = await observeBalanceIncrease(
+    asset,
+    refundAddress,
+    refundBalanceBefore,
+  );
+
+  console.log(`Refund balance before: ${refundBalanceBefore}, after: ${refundBalanceAfter}`);
+  // We expect the refund to be a little less due to ingress and egress fees.
+  assert(
+    isWithinOnePercent(
+      amountToFineAmountBigInt(refundBalanceAfter, asset),
+      amountToFineAmountBigInt(refundBalanceBefore, asset) +
+        amountToFineAmountBigInt(amount, asset),
+    ),
+    'Refund amount is incorrect',
+  );
+}
+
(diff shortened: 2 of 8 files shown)


bouncer/commands/go_bananas.ts
```diff
@@ -43,12 +43,12 @@ type AmountChange = null | {
 
 type LimitOrderResponse = {
   base_asset: {
-    chain: string;
-    asset: Asset;
+    chain: Chain;
```
niklasnatter, line 46: fixed the types to satisfy typescript

bouncer/package.json
```diff
@@ -46,5 +46,8 @@
     "tsx": "^4.7.2",
     "typescript": "^5.4.5"
   },
+  "engines": {
+    "pnpm": ">=9"
```
niklasnatter, line 50: throw an error if the installed pnpm version is too low

bouncer/shared/setup_boost_pools.ts
```diff
@@ -76,21 +76,22 @@ export async function setupBoostPools(): Promise<void> {
   console.log('=== Creating Boost Pools ===');
   const boostPoolCreationPromises: Promise<void>[] = [];
 
-  for (const chain in chainConstants) {
-    if (Object.prototype.hasOwnProperty.call(chainConstants, chain)) {
-      console.log(`Creating boost pools for all ${chain} assets`);
-      const newPools: BoostPoolId[] = [];
+  for (const chain of Object.values(Chains)) {
```
niklasnatter, line 79: iterating over `Object.values(Chains)` plays a bit nicer with typescript 🤷‍♂️

bouncer/package.json
j4m1ef0rd: My pnpm and node js were out of date. This may of been my issue.

bouncer/shared/contract_swap.ts
```diff
@@ -117,10 +118,12 @@ export async function performSwapViaContract(
...
-    console.log(`${tag} Old balance: ${oldBalance}`);
-    console.log(
-      `${tag} Executing (${sourceAsset}) contract swap to(${destAsset}) ${destAddress}. Current balance: ${oldBalance}`,
-    );
+    if (log) {
+      console.log(`${tag} Old balance: ${oldBalance}`);
+      console.log(
+        `${tag} Executing (${sourceAsset}) contract swap to(${destAsset}) ${destAddress}. Current balance: ${oldBalance}`,
+      );
+    }
```
kylezs, line 126: why would we not want to log stuff like this? Feel like we should always log and simplify the logic, remove the condition

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+} from './utils';
+import { requestNewSwap } from './perform_swap';
+import { send } from './send';
+import { getBalance } from './get_balance';
+import { getChainflipApi, observeEvent } from './utils/substrate';
+import { RefundParameters } from './new_swap';
+
+/// Do a swap with unrealistic minimum price so it gets retried and then refunded.
+async function testMinPriceRefund(asset: Asset, amount: number) {
+  const swapAsset = asset === Assets.Usdc ? Assets.Flip : Assets.Usdc;
```
kylezs, line 21: swapAsset is ambiguous
```suggestion
  const outputAsset = asset === Assets.Usdc ? Assets.Flip : Assets.Usdc;
```

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+  shortChainFromAsset,
+} from './utils';
+import { requestNewSwap } from './perform_swap';
+import { send } from './send';
+import { getBalance } from './get_balance';
+import { getChainflipApi, observeEvent } from './utils/substrate';
+import { RefundParameters } from './new_swap';
+
+/// Do a swap with unrealistic minimum price so it gets retried and then refunded.
+async function testMinPriceRefund(asset: Asset, amount: number) {
```
kylezs, line 20:
```suggestion
async function testMinPriceRefund(inputAsset: Asset, amount: number) {
```

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  const observeSwapScheduled = observeEvent('swapping:SwapScheduled', {
+    test: (event) =>
+      typeof event.data.origin === 'object' &&
+      'DepositChannel' in event.data.origin &&
+      event.data.origin.DepositChannel.channelId === depositChannelId.toString() &&
+      event.data.destinationAddress[shortChainFromAsset(swapAsset)]?.toLowerCase() ===
+        observeDestinationAddress.toLowerCase(),
+  }).event;
+
```
kylezs, line 70: These aren't necessarily unique right? 

I can have a bitcoin deposit channel address sending to ETH and a DOT address sending to ETH at the same address, and this code would match on both if the channel id is the same (which is possible since channel id is unique to chain and not global).

why not use the code that the evm deposits uses for SwapScheduled and factor this out? - though, see next comment first

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+    refundAddress,
+    refundBalanceBefore,
+  );
+
+  console.log(
+    `Refund balance before: ${refundBalanceBefore}, after: ${refundBalanceAfter} ${asset} `,
+  );
+  // We expect the refund to be a little less due to ingress and egress fees.
+  assert(
+    isWithinOnePercent(
```
kylezs, line 118: I think removing this 1% margin is going to reduce the chances of future flakiness a lot

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+import { requestNewSwap } from './perform_swap';
+import { send } from './send';
+import { getBalance } from './get_balance';
+import { getChainflipApi, observeEvent } from './utils/substrate';
+import { RefundParameters } from './new_swap';
+
+/// Do a swap with unrealistic minimum price so it gets retried and then refunded.
+async function testMinPriceRefund(asset: Asset, amount: number) {
+  const swapAsset = asset === Assets.Usdc ? Assets.Flip : Assets.Usdc;
+  const refundAddress = await newAddress(asset, 'FOK_REFUND');
```
kylezs, line 22: If we initiate two swaps with the same asset this breaks because it will generate the same address for each - which is not obvious (nor stated anywhere). We can and should use randomness here - that way we can be confident that no two tested FoK swaps will have the same refund address, so they won't interfere with each other (which would be a source of flakiness).

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+  const destAddress = await newAddress(swapAsset, 'FOK');
+  console.log(`Swap destination address: ${destAddress}`);
+
+  const refundBalanceBefore = await getBalance(asset, refundAddress);
+
+  const invalidMinPrice = amountToFineAmount(
```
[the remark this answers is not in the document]
msgmaxim: If we end up inlining, we should at least leave a comment to say that this price is "unrealistic" (which is key to understand the rest of the test).

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+  console.log(`Current swap retry delay: ${currentRetryDuration} blocks`);
+
+  const refundParameters: RefundParameters = {
+    retryDurationBlocks: 1.6 * currentRetryDuration, // More than the swap retry delay, so it will be retried once.
+    refundAddress:
+      asset === Assets.Dot ? decodeDotAddressForContract(refundAddress) : refundAddress,
+    minPrice: invalidMinPrice,
+  };
+
+  console.log(`Requesting swap from ${asset} to ${swapAsset} with unrealistic min price`);
```
msgmaxim, line 44: If we are going to run this test in parallel to others, we should tag all `console.log` statements (they way we do for other parallel tests) so we can link the message to the test.

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+    ),
+    `${asset} refund amount is incorrect (swapId: ${swapId})`,
+  );
+}
+
+export async function testFillOrKill() {
+  console.log('\x1b[36m%s\x1b[0m', '=== Running FoK test ===');
+
+  await Promise.all([
+    testMinPriceRefund(Assets.Flip, 500),
```
msgmaxim, line 131: I know this is going to be controversial, but I don't think we need to test this for multiple assets. I know it is easy to do, but I really think we should be more deliberate about what we are testing. What's important is that FoK works end-to-end. The rest is just a distraction (not to mention other, less important issues like scalability, maintenance etc). Testing multiple assets suggests that there are different code paths for FoK depending on the asset, but that's not the case. If someone adds a different code path for a some asset (which is unlikely), it will be their responsibility to add a test for that (and it will probably be a unit test anyway, not a bouncer test).

bouncer/shared/fill_or_kill.ts
msgmaxim: Actually, we should use `observeSwapRequested` (from utils) now. The fields you are looking for have been moved to the new `SwapRequested` event.

bouncer/shared/fill_or_kill.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  // Deposit the asset
+  await send(asset, depositAddress, amount.toString());
+  console.log(`Sent ${amount} ${asset} to ${depositAddress}`);
+
+  const swapScheduledEvent = await observeSwapScheduled;
+  const swapId = Number(swapScheduledEvent.data.swapId.replaceAll(',', ''));
+  console.log(`${asset} swap scheduled, swapId: ${swapId}`);
+
+  // TODO: Observing after the SwapScheduled event means its possible to miss the events, but we need to the swap id.
```
msgmaxim, line 79: Does this mean that this will be a flaky test?

bouncer/shared/contract_swap.ts
j4m1ef0rd: I needed to run the swapping test at the same time as my FoK test to trigger an error i was getting, but the logs were getting overwhelmed by the amount of logs from the swaps, making it had to see the logs from my FoK test, so i needed a way to disable the logs that i didn't care about. thought no harm in leaving it in, same as the non-contract swaps have this log condition that is on by default.

bouncer/shared/fill_or_kill.ts
j4m1ef0rd: Not a problem in practice because we start monitoring right after the swap scheduled event is seen, and the other events are always delayed by a few blocks. But yes, i agree a function that observes with a slight history check would be great.

bouncer/shared/fill_or_kill.ts
j4m1ef0rd: We don't do this anywhere else in the bouncer. Do you think its worth starting now? or should we come up with a better solution for logging in general? eg, something that can auto prefix.

j4m1ef0rd: There is a solution to this i just found in the gaslimit_ccm.ts test. Starts building a list of events before the deposit and then continuously checks the list once we know the `swapRequestId`. Maybe we could put this solution into a nicer utility of some sort.

bouncer/shared/fill_or_kill.ts
j4m1ef0rd: I assume you mean that just observing a balance increase at all is fine for this test.

kylezs: yeah exactly, especially if we use a random address, then it will be fine

bouncer/shared/fill_or_kill.ts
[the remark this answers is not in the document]
kylezs: yeah either or

bouncer/shared/contract_swap.ts
kylezs: Not a strong opinion so I'm ok with leaving it, but with the tags shouldn't it be relatively easy to trace the logs from a particular test?

kylezs (approved):

PR #5059, opened 2024-07-18
fix: account deletion logic to keep boost funds into consideration

marcellorigotti:
# Pull Request

Closes: PRO-1337

## Checklist

## Summary

Check boost balances of LP before allowing de-registration

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-lp/src/lib.rs
+++ b/state-chain/pallets/cf-lp/src/lib.rs
@@ -4,8 +4,8 @@
 use cf_chains::{address::AddressConverter, AnyChain, ForeignChainAddress};
 use cf_primitives::{AccountRole, Asset, AssetAmount, BasisPoints, ForeignChain};
 use cf_traits::{
-	impl_pallet_safe_mode, liquidity::LpBalanceApi, AccountRoleRegistry, Chainflip, DepositApi,
-	EgressApi, LpDepositHandler, PoolApi, ScheduledEgressDetails,
+	impl_pallet_safe_mode, liquidity::LpBalanceApi, AccountRoleRegistry, BoostApi, Chainflip,
+	DepositApi, EgressApi, LpDepositHandler, PoolApi, ScheduledEgressDetails,
 };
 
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -306,6 +314,12 @@ pub mod pallet {
 				Self::free_balances(&account_id)?.iter().all(|(_, amount)| *amount == 0),
 				Error::<T>::FundsRemaining
 			);
+			ensure!(
```
kylezs, line 317: we could add a test for this with a slightly smarter mock

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -294,6 +301,7 @@ pub mod pallet {
 		pub fn deregister_lp_account(who: OriginFor<T>) -> DispatchResult {
 			const STABLE_ASSET: Asset = Asset::Usdc;
 			let account_id = T::AccountRoleRegistry::ensure_liquidity_provider(who)?;
+			T::PoolApi::sweep(&account_id)?;
```
kylezs, line 304: why do we sweep here? I mean, i understand that we don't want to deregister if they have orders, but why not make them do this explicitly?

marcellorigotti: wdym? as far as I understood sweep doesn't close any order, it collects just the fees.
If they have open orders the extrinsic will fail on the next check (when the other PR updating `open_orders_count` to include zero sized orders will be merged)

state-chain/pallets/cf-lp/src/lib.rs
marcellorigotti: yeah sure! what did you have in mind?

kylezs: Just that this error is thrown when there are boost funds basically

kylezs (approved):

PR #5060, opened 2024-07-18
feat: generic configurable election-based witnessing pallet

AlastairHolmes:
Closes:  PRO-1429

This pallet facilitates the creation and deciding of elections as a model for deciding consensus of external off-chain state. Almost any aspect of how the elections behave including what is done when they reach consensus / finish can be configured via the `ElectoralSystem` associated type in the pallet's `Config` trait.


AlastairHolmes:
I'm guessing there is probably a few bugs due to so last minute changes, so I will go through this myself again in the coming days but it is done, excluding the code the calls to initialize (i.e. start the witnessing) for the new SolanaElections pallet, but I believe I will need to wait for the other Solana PRs before I can add that. But this should be very simple, as the code to do the initialization is implemented just there doesn't exist a reasonable place to call it in "main" at the moment.

albert-llimos:
> ... but I believe I will need to wait for the other Solana PRs before I can add that. 
All the Solana code has now been merged to main.

kylezs (commented):
just some small things as I'm going through it

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -0,0 +1,433 @@
...
+	<Sink::Chain as Chain>::DepositDetails: Default,
+{
+	pub fn open_channel<ElectoralAccess: ElectoralWriteAccess<ElectoralSystem = Self>>(
+		election_identifiers: Vec<
+			ElectionIdentifier<<Self as ElectoralSystem>::ElectionIdentifierExtra>,
+		>,
+		electoral_access: &mut ElectoralAccess,
+		channel: <Sink::Chain as Chain>::ChainAccount,
+		asset: <Sink::Chain as Chain>::ChainAsset,
+		close_block: <Sink::Chain as Chain>::ChainBlockNumber,
```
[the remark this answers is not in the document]
AlastairHolmes: That is correct. As a note, even if your tx is after the close_block it maybe processed. For example, if the protocol was offline/witnessing was paused as the external chain crossed the close_block. It is also possible for the tx to be processed in blocks just after the close_block (But not guaranteed).

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -0,0 +1,1955 @@
...
+
+				Ok(())
+			}
+			fn clear_votes(&mut self) {
+				let unique_monotonic_identifier = self.unique_monotonic_identifier();
+				ElectionBitmapComponents::<T, I>::clear(unique_monotonic_identifier);
+				for (_, (_, individual_component)) in
+					IndividualComponents::<T, I>::drain_prefix(unique_monotonic_identifier)
+				{
+					<<T::ElectoralSystem as ElectoralSystem>::Vote as VoteStorage>::visit_shared_data_references_in_individual_component(&individual_component, |shared_data_hash| {
```
dandanlen, line 568: Pedantic point - As far as I can tell, the `visit_shared_data_references` methods actually visit the hash, even thought they are always expected to either increment or decrement the reference count.

For example, it could be used to access the shared data itself (probably not very useful). My point is that maybe the closure should take a ReferenceDetails struct. Or even simpler, we add methods add/remove a reference.

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -0,0 +1,1955 @@
...
+
+			pub(super) fn add(
+				&mut self,
+				authority_index: AuthorityCount,
+				bitmap_component: BitmapComponentOf<T::ElectoralSystem>,
+				unique_monotonic_identifier: UniqueMonotonicIdentifier,
+				block_number: BlockNumberFor<T>,
+			) -> Result<(), CorruptStorageError> {
+				// We don't need to delete existing because we remove the existing in `take` which
+				// is called before this.
```
dandanlen, line 1011: This seems like quite a strong assumption... how important is it?

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -0,0 +1,1955 @@
...
+						{
+							// Update shared data reference counts
+							<<T::ElectoralSystem as ElectoralSystem>::Vote as VoteStorage>::visit_shared_data_references_in_individual_component(
+								&individual_component,
+								|shared_data_hash| Self::add_shared_data_reference(shared_data_hash, unique_monotonic_identifier, block_number),
+							);
+							// Store individual component
+							IndividualComponents::<T, I>::set(
+								unique_monotonic_identifier,
+								authority.clone(),
```
dandanlen, line 1212: nit: this should accept a reference. (might need to use insert instead of set)

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -0,0 +1,1955 @@
...
+			>(
+				current_epoch: EpochIndex,
+				unique_monotonic_identifier: UniqueMonotonicIdentifier,
+				f: F,
+			) -> Result<R, CorruptStorageError> {
+				Self::inner_with::<true, _, _>(current_epoch, unique_monotonic_identifier, f)
+			}
+
+			pub(super) fn add(
+				&mut self,
```
AlastairHolmes, line 1004: sdsd

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -0,0 +1,1955 @@
...
+		StorageMap<_, Identity, T::ValidatorId, (), OptionQuery>;
+
+	/// Votes will only be allowed if their `VoteSynchronisationBarrier` matches this value,
+	/// therefore authorities can invalidate in-flight votes by changing this value via an
+	/// extrinsic. Once they do that they can be confident that no old votes will be included, at
+	/// any block after the extrinsic is included. This allows validators to be able to ensure their
+	/// votes will not change, so they can delete/correct bad votes after detecting a problem for
+	/// example a reorg.
+	#[pallet::storage]
+	type AuthorityVoteSynchronisationBarriers<T: Config<I>, I: 'static = ()> =
```
kylezs, line 481:
```suggestion
	type AuthorityVoteSynchronisationBarrier<T: Config<I>, I: 'static = ()> =
```

kylezs: consistency, if it's a map we normally remove the s e.g. `NodeCFEVersion`

engine/src/witness/sol/sol_deposits.rs
```diff
@@ -1,204 +1,111 @@
...
 				}
-				Ok::<_, anyhow::Error>(())
-			}
-		})
-	}
+			}),
+	)
+	.chunks(MAX_MULTIPLE_ACCOUNTS_QUERY / 2)
+	.map(|deposit_channels_chunk| ingress_amounts(sol_rpc, deposit_channels_chunk, vault_address))
+	.buffered(MAXIMUM_CONCURRNET_RPCS)
```
albert-llimos, line 82: Typo, should be `MAXIMUM_CONCURRENT_RPCS`

engine/src/witness/sol/sol_deposits.rs
```diff
@@ -1,204 +1,111 @@
...
-												*deposit_channel.address_to_witness(),
-												value,
-											);
-										},
-									);
-								}
-							}
-						}
-					}
+const MAXIMUM_CONCURRNET_RPCS: usize = 16;
```
albert-llimos, line 33: Any particular reason why this value is set to 16?

engine/src/witness/sol/sol_deposits.rs
```diff
@@ -1,204 +1,111 @@
...
+		stream::iter(
+			ingress_amounts_chunk
+				.into_iter()
+				.filter_map(move |(deposit_channel, amount)| {
+					let amount: SolAmount = amount.saturated_into(); // TODO: Change the DeltaBasedIngress to not use the Chains Amount type but
+												 // instead use u256 or u128 or some generic.
+					let (_, current_total_ingressed) =
+						deposit_channels.get(deposit_channel.address()).unwrap();
+					if amount != current_total_ingressed.amount ||
+						slot < current_total_ingressed.block_number
```
albert-llimos, line 93: I imagine this statement is to avoid scenarios where nodes don't submit the amount because consensus has been reached (with other nodes' votes) and then they don't submit their own. However, I'm not sure that covers all cases.
Example in three nodes with an initial balance of 0 and deposit of 100 at slot 5.
Node A witnesses balance of 100 at slot 5.
Node B witnesses balance of 100 at slot 6.
I assume SC reaches consensus of balance 100 at slot 6.
Let's say node C was slow and didn't pull the deposit channel until after consensus is reached. Then node C witnesses balance of 100 at slot 7. It won't submit the balance since `amount == current_total_ingressed.amount` and `slot > current_total_ingressed.block_number`.

I don't think it's a problem since correct consensus has been reached but up to 1/3 of the nodes might end up not submitting the total_ingressed. I don't expect it to happen on normal behavior though, just wanted to double check that it makes sense.

AlastairHolmes: Not sure you totally understand the logic here, but yes the situation you describe technically can happen, i.e. someone doesn't vote as consensus was reached. The `if` statement's purpose to decrease the amount of vote data (by filtering out channels whose deposit amount hasn't changed). Authority's that vote, but don't specify a value for a given channel are assumed to be voting that the amount hasn't changed. This allows authority's to only vote if the amount changes, massively decrease the number of extrinsics/votes needed. Lets have a chat tomorrow to clear it up though.

albert-llimos: I should have specified, my comment was just about the `slot < current_total_ingressed.block_number`. The other part of the `if` statment makes total sense for the reason you're explaining.
If you agree that the situation that I explained can happen and is not problematic then it's all good :+1:

engine/src/witness/sol/sol_deposits.rs
kylezs: I don't believe there is - but seems a reasonable number.

kylezs (approved):

PR #5061, opened 2024-07-19
feat: upload `chainflip-engine` shared libraries to S3 💾

ahasna:


.github/workflows/_61_upload_engine_shared_libs.yml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_call:
+
+permissions:
+  id-token: write
+  contents: read
+jobs:
+  upload:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout 🛒
```
kylezs, line 11: why do we need to checkout here? we aren't doing anything with the repo as far as I can tell?

.github/workflows/_61_upload_engine_shared_libs.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Set Version 🔦
+        id: version
+        shell: bash
+        run: |
+          chmod +x chainflip-backend-bin/chainflip-node
+          VERSION=$(./ci/scripts/extract_version.sh "$(./chainflip-backend-bin/chainflip-node --version)" full)
+          echo "Version: $VERSION"
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "url_friendly_version=$(echo $VERSION | tr . _)" >> $GITHUB_OUTPUT
+          echo "full_version=$VERSION" >> $GITHUB_OUTPUT
```
kylezs, line 35: Only `url_friendly_version` is used as far as I can tell, where are the others used? Also `version` and `full_version` are the same?

.github/workflows/_61_upload_engine_shared_libs.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Move shared library files 📂
+        run: |
+          mkdir -p /tmp/chainflip-engine-shared-libs/
+          mv chainflip-backend-bin/libchainflip_engine_v${{ steps.version.outputs.url_friendly_version }}.so /tmp/chainflip-engine-shared-libs/
+          mv chainflip-backend-bin-mac-m2/libchainflip_engine_v${{ steps.version.outputs.url_friendly_version }}.dylib /tmp/chainflip-engine-shared-libs
```
kylezs, line 41: nit, consistency
```suggestion
          mv chainflip-backend-bin-mac-m2/libchainflip_engine_v${{ steps.version.outputs.url_friendly_version }}.dylib /tmp/chainflip-engine-shared-libs/
```

ci/scripts/extract_version.sh
```diff
@@ -15,6 +18,13 @@ fi
 
 # Extract major and minor version
 major_minor=$(echo "$version" | cut -d. -f1,2)
+full_version=$(echo "$version" | cut -d. -f1,2,3)
+
+if [[ "$version_type" == "major_minor" ]]; then
+    semver_version="$major_minor"
+else
+    semver_version="$full_version"
+fi
```
kylezs, line 27:
```suggestion
full_version=$(echo "$version" | cut -d. -f1,2,3)

if [[ "$version_type" == "major_minor" ]]; then
    echo $major_minor
else
    echo $full_version
fi
```

Then delete the echo `echo "$semver_version"` below.

No need to store it in a variable right? Also the variable name is misleading since it's possibly not a semver, it could be just a major_minor.

ci/scripts/extract_version.sh
```diff
@@ -15,6 +18,13 @@ fi
...
+
+if [[ "$version_type" == "major_minor" ]]; then
+    echo $major_minor
+else
+    echo $full_version
+fi
 
 # Output the major and minor version
-echo "$major_minor"
+echo "$semver_version"
```
kylezs, line 30: we don't need to echo this now

kylezs (approved):

PR #5063, opened 2024-07-19
feat: apicall stores signer, store current on chain key

ramizhasan111:
# Pull Request

Closes: PRO-1508
Closes: PRO-1509

## Checklist

## Summary

This PR introduces following two things:

- The Apicall now stores the key that the call was signed with
- The Threshold Signature pallet also stores the signer along with the signature which is then returned to the broadcast pallet to store in the apicall
- The current onchain key is now tracked in the broadcast pallet which is needed to determine if an apicall requires a signature refresh
- requires signature refresh for ethereum now checks the current on chain key against the key that the apicall was signed with.

todo: This will need migrations, fixing tests

#### Non-Breaking changes


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -1005,6 +1033,8 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
 				}
 		}
 
+		IncomingKeyAndBroadcastId::<T, I>::set(Some((new_key, broadcast_id)));
```
albert-llimos, line 1036: This approach to use `IncomingKeyAndBroadcastId` seems correct but I'm trying to think if there might be easier way to do it so we don't have to add yet another item to storage which requires another read to then delete it at `transaction_succeeded`.
I guess it could be possible to add a function for `ApiCall` to identify if the call is a key rotation and get the `new_key` from it if it is, right? The problem would then be that at `transaction_succeeded` we don't really have a way to know if it was a key rotation unless we store the `broadcastId`, which leads to the current approach. It just feels like simpler that we keep track of the key rotation through the system in the transaction itself rather than this.

The other solution I could come up with is that we implement logic in the engine to detect if transactions are key rotations and report that in `transaction_succeeded`. That would actually be fairly simple for Eth/Arb/Sol, which are the chains we need this logic for. But maybe it's just too cumbersome and the current approach is just simpler and therefore better?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -686,7 +712,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			if !pending_broadcasts.remove(broadcast_id) {
 				log::warn!("Expected broadcast with id {} to still be pending.", broadcast_id);
 			}
-			if let Some(broadcast_barrier_id) = BroadcastBarriers::<T, I>::get().first() {
+			while let Some(broadcast_barrier_id) = BroadcastBarriers::<T, I>::get().first() {
```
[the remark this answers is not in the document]
ramizhasan111: there is an edge case where if someone maliciously submits the rotation tx and there is a barrier on the tx before, when the barrier on the tx before is released, the barrier on the rotaton tx wont be released and so we need to release all barriers that are before the earliest pending tx.

That being said, I was still to do it. Somehow that while crept through unfinished. Should be correct in the latest commit.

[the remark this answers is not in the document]
albert-llimos: right, good catch. Also with the last commit it makes more sense.

state-chain/pallets/cf-broadcast/src/lib.rs
ramizhasan111: I dont think storage read/write is a problem because a rotation happens only once in 3 days and its just a few read/writes per rotation. I dont think I understand the second part of your comment.

Doing the recognition of rotation tx in the engine wont help us too much. The only thing we might, then, not need is the IncomingKeyAndBroadcastId. Other than that we still need all the functionality that enables us to store the signer inside the Apicall (to later compare the onchain key and the apicall signer). So not sure that little advantage is worth who knows how much effort in the engine.

albert-llimos: `IncomingKeyAndBroadcastId` is read in every `transaction_succeeded`  and the `CurrentOnChainKey` is read on every `start_next_broadcast_attempt`. If I'm not missing anything, that is for every transaction, not only in rotations.  My point was that those could be avoided if we did it in a clever way.
However, it's fair that a couple of extra reads for normal transactions and a couple more writes once in 3 days is fine to not have to overcomplicate it :+1:

ramizhasan111: ```IncomingKeyAndBroadcastId``` should not actually be read on every tx. The ```exists()``` function is supposed to be efficient in a way that it smartly/cheaply finds out if a storage item exists and only if it exists that we actually read the storage item. And because it would only exist around rotations, it would only be read for a few txs around that time. 

Its true that ```CurrentOnChainKey``` is read in ```start_next_broadcast_attempt``` but that is only called when a broadcast is retried and retry should only happen for a few txs. most of them should go through on first try. Also, at this point (```start_next_broadcast_attempt```) the information about what key is currently on the external chain has to be known and therefore has to be stored on the SC for the information to be available. I dont see how else would we retrieve this info (what the current onchain key is) on the fly. If we retrieve this info through some other storage item then that storage item has to be read so in any case we have to read storage.

albert-llimos: > The exists() function is supposed to be efficient in a way that it smartly/cheaply finds out if a storage item exists and only if it exists that we actually read the storage item
Alright, good to know!

Makes sense regarding the `CurrentOnChainKey` too, I was initially trying to not have to require the `IncomingKeyAndBroadcastId` but given what you said above it's not worth the effort. Therefore the current approach makes sense to me, thanks for providing extra info.

state-chain/pallets/cf-broadcast/src/lib.rs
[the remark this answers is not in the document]
kylezs: this is a good case to write a unit test for :)

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -309,6 +315,15 @@ pub mod pallet {
...
 	pub type AbortedBroadcasts<T, I = ()> = StorageValue<_, BTreeSet<BroadcastId>, ValueQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn incoming_key_broadcast_id)]
+	pub type IncomingKeyAndBroadcastId<T, I = ()> =
+		StorageValue<_, (AggKey<T, I>, BroadcastId), OptionQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn current_on_chain_key)]
+	pub type CurrentOnChainKey<T, I = ()> = StorageValue<_, AggKey<T, I>, OptionQuery>;
```
kylezs, line 325: I would leave a comment here explaining why this is necessary, in particular considering we have the key in the threshold signature pallet (which is not necessarily on the external chain since we optimistically rotate, but worth making that clear here).

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -438,8 +438,13 @@ pub mod pallet {
 	/// State of the threshold signature requested.
 	#[pallet::storage]
 	#[pallet::getter(fn signature)]
-	pub type Signature<T: Config<I>, I: 'static = ()> =
-		StorageMap<_, Twox64Concat, RequestId, AsyncResult<SignatureResultFor<T, I>>, ValueQuery>;
+	pub type Signature<T: Config<I>, I: 'static = ()> = StorageMap<
+		_,
+		Twox64Concat,
+		RequestId,
+		(AggKeyFor<T, I>, AsyncResult<SignatureResultFor<T, I>>),
```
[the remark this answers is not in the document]
ramizhasan111: I think the improvement in readability is pretty marginal since ```.0``` and ```.1``` was used in only in two places. I did the change anyways which was a bit of a pain to do. I couldnt get it to compile after changing the tuple to a struct in the storage item. couldnt figure out why but it was related to the ValueQuery. I therefore changed the ValueQuery to OptionQuery which meant some knock-on changes after which it compiled.

state-chain/pallets/cf-threshold-signature/src/migrations/migrate_signature_to_include_signer.rs
```diff
@@ -25,7 +25,9 @@ mod old {
 
 impl<T: crate::Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
 	fn on_runtime_upgrade() -> frame_support::weights::Weight {
-		let current_aggkey = Keys::<T, I>::get(CurrentKeyEpoch::<T, I>::get().unwrap()).unwrap();
+		let current_aggkey =
+			Keys::<T, I>::get(CurrentKeyEpoch::<T, I>::get().unwrap_or(Default::default()))
+				.unwrap_or(Default::default());
```
kylezs: Why do we default here (and elsewhere in this commit)? What are the cases when the key doesn't exist?

state-chain/pallets/cf-threshold-signature/src/migrations/migrate_signature_to_include_signer.rs
```diff
@@ -25,9 +25,8 @@ mod old {
 
 impl<T: crate::Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
 	fn on_runtime_upgrade() -> frame_support::weights::Weight {
-		let current_aggkey =
-			Keys::<T, I>::get(CurrentKeyEpoch::<T, I>::get().unwrap_or(Default::default()))
-				.unwrap_or(Default::default());
+		let current_aggkey = Keys::<T, I>::get(CurrentKeyEpoch::<T, I>::get().unwrap_or_default())
```
kylezs, line 28: Why are we defaulting here? When we do expect Keys for the current epoch to be empty?

ramizhasan111: So the idea is that there is no situation where the Keys would be empty on a running network except for the situation where a chain is not initialised in which case we wont have anything to migrate anyways so the default wont matter. Practically speaking, this situation can only occur with solana code being active on a chain (and solana not being initialized yet). To safeguard against that, the migration for the solana broadcast pallet is not even run. See the runtime lib file.

kylezs: I can maybe see the value in using default for the actual runtime upgrade just for very edge cases, we don't want it to panic for the non-solana chains.

However, for the pre and post checks we should not unwrap_or_default. There's current a hole in the test.

On runtime upgrade runs, we don't have a key, so we insert the default key.

Now on the post check we try to get a key and we get the default because it doesn't exist... now we're just comparing two default keys and the test *Passes*. This test should fail if we expect a key.

ramizhasan111: > I can maybe see the value in using default for the actual runtime upgrade just for very edge cases, we don't want it to panic for the non-solana chains.

As said before, for non-solana chains, there is no edge case on a running network since there is no scenario that Keys would be empty so not sure what you mean here.

> However, for the pre and post checks we should not unwrap_or_default. There's currently a hole in the test.
> 
> On runtime upgrade runs, we don't have a key, so we insert the default key.
> 
> Now on the post check we try to get a key and we get the default because it doesn't exist... now we're just comparing two default keys and the test _Passes_. This test should fail if we expect a key.

I don't agree with this. The case you are describing was already considered but as said before, is hypothetical. The tests wont run for the case you describe. The test for that case is different (see ```NoSolUpgrade```). Its no different than unwrapping a value giving a reason of why that is safe to do so. I would just ```unwrap``` here and not ```unwrap_or_default``` (which I did originally) but the ci upgrade test was panicing on the unwrap which, if anything, shows there is an issue with the ci upgrade test not the test for the migration.

ramizhasan111: discussed irl. The ```unwrap_or_default()``` should be changed back to ```unwrap()``` since the default case wont happen as mentioned above. The ci upgrade test should pass with the ```NoSolUpgrade``` which wasnt there when the test was failing.

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -399,7 +400,7 @@ fn re_request_threshold_signature_on_invalid_tx_params() {
 
 			assert_eq!(
 				MockThresholdSigner::<MockEthereumChainCrypto, RuntimeCall>::signature_result(0),
-				AsyncResult::Void
+				(Default::default(), AsyncResult::Void)
```
kylezs, line 403: Think we should use some non-default const for these instead of Default. Using Default will very easily hide potential bugs (e.g. if we unwrap_or_default() and hit a default, we could still pass the test)

state-chain/runtime/src/migrations/migrate_apicalls_to_store_signer.rs
kylezs, line 1: Nice tests here 👍

state-chain/runtime/src/lib.rs
```diff
@@ -1185,6 +1185,40 @@ type PalletMigrations = (
...
+		5,
+		6,
+	>,
+	VersionedMigration<
+		pallet_cf_broadcast::Pallet<Runtime, Instance4>,
+		migrations::migrate_apicalls_to_store_signer::ArbMigrateApicallsAndOnChainKey,
+		5,
+		6,
+	>,
+	VersionedMigration<
```
kylezs, line 1216: A comment here explaining why we do this would be helpful for future readers

.github/workflows/ci-development.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
```
kylezs, line 82: Reminder to remove this once we're happy it all passes

kylezs (commented):
Nice, just some small things

state-chain/chains/src/sol/api.rs
```diff
@@ -70,6 +71,10 @@ impl<Env: 'static> ApiCall<SolanaCrypto> for SolanaApi<Env> {
 	fn refresh_replay_protection(&mut self) {
 		todo!()
 	}
+
+	fn signer(&self) -> Option<<SolanaCrypto as ChainCrypto>::AggKey> {
+		todo!()
```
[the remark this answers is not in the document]
ramizhasan111: this is part of the skeleton for solana that is there on the current main. All of these todos are implemented in the solana integration PR.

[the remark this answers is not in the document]
kylezs: ah yeah got confused, thought this was targeting the solana branch, makes sense

.github/workflows/ci-development.yml
kylezs: Passes: https://github.com/chainflip-io/chainflip-backend/actions/runs/10161416189/job/28100173256?pr=5063 so yeah we can remove now

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1333,3 +1345,46 @@ fn threshold_sign_and_refresh_replay_protection() {
...
+		let tx_id_2 = [2; 4];
+		let broadcast_id_2 = initiate_and_sign_broadcast(
+			&api_call(2).1,
+			TxType::Rotation { new_key: Default::default() },
+		);
+
+		// the rotation tx should create barriers on both txs
+		let mut expected_barriers = BTreeSet::new();
+		expected_barriers.extend(vec![broadcast_id_1, broadcast_id_2]);
+		assert_eq!(BroadcastBarriers::<Test, Instance1>::get(), expected_barriers);
```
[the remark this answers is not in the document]
ramizhasan111: this test is for a specific case regarding the broadcast barrier where all the barriers upto the current pending txs are released. It directly relates to the change in this PR that deals with this storage. So I think the current test makes more sense. The fact that the broadcast gets rescheduled is basically testing that the barriers work as expected which is tested in other tests already there (we can always add more tests of course).

state-chain/pallets/cf-broadcast/src/tests.rs
ramizhasan111: We normally use Default for things that we dont test for in the test so I dont think its a big problem. In any case, I tried to do this but it is not so simple since signature_result is part of a mock pallet. It can be done but would take a bit of time. Dont know if we want this as part of this PR.

kylezs (approved):

PR #5064, opened 2024-07-19
Fix/workaround metadata bug for pallet config

martin-chainflip:
# Pull Request

Closes: PRO-1518

## Checklist

## Summary

Defining typeInfos for different enums with the same name causes PolkadotJS some confusion. For example, we have multiple Ingress-Egress pallets, one for each chain. Each of these pallets has an enum defined called "PalletConfigUpdate". Even though their type path and type ID are different, PolkadotJS seems to only look at the enum name when parsing the metadata. This leads to the situation where when trying to set a value for PalletConfigUpdate on Ethereum, the PolkadotJS interface uses the type for Solana (or whatever pallet was constructed last).

The solution is to not derive the typeInfo, but to explicitly construct it and assign different type names for each instance.

The main problem was that the name needs to be a string literal, but we can't construct it via concat!, because that will only work with string literals. Alastair suggested constcat::concat! instead, but that also doesn't work, because Rust doesn't allow to access static elements via generic types (wow).

In the end I just bruteforce write out all the chains. It sucks, but it is better than what we had before.

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -213,26 +213,17 @@ pub enum PalletConfigUpdate<T: Config<I>, I: 'static> {
 	SetMaxSwapRetryDurationBlocks { blocks: BlockNumber },
 }
 
-const fn get_name_for_set_minimum_deposit<T: Config<I>, I: 'static>() -> &'static str {
-	match T::TargetChain::NAME.as_bytes() {
-		b"Ethereum" => "SetMinimumDepositEthereum",
-		b"Polkadot" => "SetMinimumDepositPolkadot",
-		b"Bitcoin" => "SetMinimumDepositBitcoin",
-		b"Arbitrum" => "SetMinimumDepositArbitrum",
-		b"Solana" => "SetMinimumDepositSolana",
-		_ => panic!("Add the new chain to this list!"),
-	}
-}
-
-const fn get_name_for_pallet_config_update<T: Config<I>, I: 'static>() -> &'static str {
-	match T::TargetChain::NAME.as_bytes() {
-		b"Ethereum" => "PalletConfigUpdateEthereum",
-		b"Polkadot" => "PalletConfigUpdatePolkadot",
-		b"Bitcoin" => "PalletConfigUpdateBitcoin",
-		b"Arbitrum" => "PalletConfigUpdateArbitrum",
-		b"Solana" => "PalletConfigUpdateSolana",
-		_ => panic!("Add the new chain to this list!"),
-	}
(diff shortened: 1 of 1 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -217,6 +213,51 @@ pub enum PalletConfigUpdate<T: Config<I>, I: 'static = ()> {
...
+			"Polkadot" => concat!(stringify!($name), "Polkadot"),
+			"Bitcoin" => concat!(stringify!($name), "Bitcoin"),
+			"Arbitrum" => concat!(stringify!($name), "Arbitrum"),
+			"Solana" => concat!(stringify!($name), "Solana"),
+			_ => concat!(stringify!($name), "Other"),
+		}
+	};
+}
+
+impl<T, I> TypeInfo for PalletConfigUpdate<T, I>
```
kylezs, line 229: How can we know that this works? There are no compiler guarantees of anything here right?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -217,6 +213,51 @@ pub enum PalletConfigUpdate<T: Config<I>, I: 'static = ()> {
...
+	fn type_info() -> Type {
+		Type::builder()
+			.path(Path::new(append_chain_to_name!(PalletConfigUpdate), module_path!()))
+			.variant(
+				Variants::new()
+					.variant("ChannelOpeningFee", |v| {
+						v.index(0)
+							.fields(Fields::named().field(|f| f.ty::<T::Amount>().name("fee")))
+					})
+					.variant(append_chain_to_name!(SetMinimumDeposit), |v| {
```
kylezs, line 244: why is this one different? why do we append the chain name here but not the others?

kylezs (commented):
Is this fixed in a more recent version of substrate? or is there an open issue to fix it? Would be good to at least link the issue somewhere here to explain why this is required.

state-chain/pallets/cf-ingress-egress/src/lib.rs
martin-chainflip: this is the only one that has a different type across palettes. I tried to keep the renaming to the absolute minimum that still fixes the bug. This is the result of a lot of trial and error.

state-chain/pallets/cf-ingress-egress/src/lib.rs
martin-chainflip: no, the downside of implementing it manually is that we need to update the typeinfo together with the struct/enum itself.

kylezs (approved):

PR #5065, opened 2024-07-19
feat: RPC to return available pools

marcellorigotti:
# Pull Request

Closes: PRO-1506

## Checklist

## Summary

- Added an RPC `cf_available_pools` which rdated the code where we returns the available pools as a Vec<AssetPair>
- Added pools() to PoolApi to access these info internally, updated the code to use this new function instead of looping through all assets but usdc (error prone if we don't create the pool as soon as we release a new chain or if we decide to create pools with quote asset != USDC)

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-lp/src/lib.rs
+++ b/state-chain/pallets/cf-lp/src/lib.rs
@@ -304,8 +304,7 @@ pub mod pallet {
 
 			ensure!(
 				T::PoolApi::pools().iter().all(|asset_pair| {
-					T::PoolApi::open_order_count(&account_id, asset_pair.base, asset_pair.quote)
-						.unwrap_or_default() == 0
+					T::PoolApi::open_order_count(&account_id, asset_pair).unwrap_or_default() == 0
 				}),
 				Error::<T>::OpenOrdersRemaining
 			);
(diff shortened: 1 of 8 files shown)


state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -292,12 +292,11 @@ pub mod pallet {
...
 		#[pallet::weight(T::WeightInfo::deregister_lp_account())]
 		pub fn deregister_lp_account(who: OriginFor<T>) -> DispatchResult {
-			const STABLE_ASSET: Asset = Asset::Usdc;
 			let account_id = T::AccountRoleRegistry::ensure_liquidity_provider(who)?;
 
 			ensure!(
-				Asset::all().filter(|asset| *asset != STABLE_ASSET).all(|asset| {
-					T::PoolApi::open_order_count(&account_id, asset, STABLE_ASSET)
+				T::PoolApi::pools().iter().all(|asset_pair| {
+					T::PoolApi::open_order_count(&account_id, asset_pair.base, asset_pair.quote)
```
kylezs, line 299: why do we even separate base and quote here? out of scope for this PR, but maybe it  makes sense to pass asset_pair through a bit deeper - just reduces number of params, a little tidier, and means there are fewer places to mess up the order of passing in base/quote in the wrong place

kylezs (approved):

PR #5070, opened 2024-07-22
fix: grep correct spec_version

kylezs:
# Pull Request

## Checklist

## Summary

It was grepping two `spec_version` s from the file.

### DIFF
--- a/ci/scripts/check_runtime_spec_version.sh
+++ b/ci/scripts/check_runtime_spec_version.sh
@@ -5,7 +5,7 @@ network=$1
 runtime_spec_version_file="state-chain/runtime/src/lib.rs"
 
 # Extract the spec_version from the Rust file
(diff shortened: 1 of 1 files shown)


ci/scripts/check_runtime_spec_version.sh
```diff
@@ -5,7 +5,7 @@ network=$1
 runtime_spec_version_file="state-chain/runtime/src/lib.rs"
 
 # Extract the spec_version from the Rust file
-spec_version=$(grep 'spec_version:' $runtime_spec_version_file | awk '{print $2}' | tr -d ',')
+spec_version=$(grep '#[sp_version::runtime_version]' -A 10 $runtime_spec_version_file | grep 'spec_version' | awk '{print $2}' | tr -d ',')
```
martin-chainflip, line 8: I'd give it a few more lines than 10, just to be on the safe side.
Also, we might want to put the -F flag onto grep to turn off regex and search for the literal string instead

martin-chainflip (approved):

PR #5083, opened 2024-07-24
feat: Move asset balances into asset-balances pallet (PRO-1422)

Janislav:
# Pull Request

Closes: PRO-1422
Closes: PRO-1542

## Checklist

## Summary

Moved the storage management for the following Balances:

* FreeBalances
* HistoricalEarnedFees
* ~~CollectedRejectedFunds~~
* EarnedBrokerFees

Apart from that also refactored everything regarding the responsibilities around balance management in the pallets, added needed tests and migrations. Also merged FreeBalances and EarnedBrokerFees into one storage item.

#### Non-Breaking changes


state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -128,6 +146,25 @@ pub mod pallet {
...
+		StorageDoubleMap<_, Twox64Concat, T::AccountId, Identity, Asset, AssetAmount>;
+
+	#[pallet::storage]
+	/// Historical earned fees for an account.
+	pub type HistoricalEarnedFees<T: Config> =
+		StorageDoubleMap<_, Identity, T::AccountId, Twox64Concat, Asset, AssetAmount, ValueQuery>;
+
+	/// Network fees, in USDC terms, that have been collected and are ready to be converted to FLIP.
+	#[pallet::storage]
+	pub type CollectedNetworkFee<T: Config> = StorageValue<_, AssetAmount, ValueQuery>;
```
msgmaxim, line 162: This was recently (in https://github.com/chainflip-io/chainflip-backend/pull/5014) moved from LP to the Swapping pallet (as per [PRO-1404](https://linear.app/chainflip/issue/PRO-1404)). The motivation was to simplify code by avoiding indirection via traits. 

I'm surprised we want to move this again. Looking at https://linear.app/chainflip/issue/PRO-1422/move-asset-balances-into-dedicated-pallet, it seems like the plan was to move it to the swapping pallet rather than a new pallet (which has already been done)?

Also, doesn't look like this PR does not remove `CollectedNetworkFee` from its previous location. I'm guessing it is a merge error?

Janislav: I guess we don't then. I was not aware that this was the plan, but like it seems we gonna keep it in the swapping pallet in this case.

Yeah, that was due to merge error.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -128,6 +146,21 @@ pub mod pallet {
...
+
+	#[pallet::storage]
+	/// Historical earned fees for an account.
+	pub type HistoricalEarnedFees<T: Config> =
+		StorageDoubleMap<_, Identity, T::AccountId, Twox64Concat, Asset, AssetAmount, ValueQuery>;
+
+	/// Fund accrued from rejected swap and CCM calls.
+	#[pallet::storage]
+	pub type CollectedRejectedFunds<T: Config> =
+		StorageMap<_, Twox64Concat, Asset, AssetAmount, ValueQuery>;
```
dandanlen, line 163: I can see how it might make sense to track these funds here, but then we could track them in the WithheldAssets storage, for example. Not sure there's a real need for a distinct storage item.

state-chain/traits/src/liquidity.rs
```diff
@@ -12,20 +28,31 @@ pub trait LpDepositHandler {
...
+	/// Ensure that the given account has a refund address set for the given asset.
 	fn ensure_has_refund_address_for_pair(
 		who: &Self::AccountId,
 		base_asset: Asset,
 		quote_asset: Asset,
 	) -> DispatchResult;
+}
+
+/// API for interacting with the asset-balance pallet.
+pub trait BalanceApi {
```
dandanlen, line 51: If this is no longer lp-specific we should move it out of the liquidity module.

state-chain/traits/src/liquidity.rs
```diff
@@ -12,20 +28,31 @@ pub trait LpDepositHandler {
 	fn add_deposit(who: &Self::AccountId, asset: Asset, amount: AssetAmount) -> DispatchResult;
 }
 
-pub trait LpBalanceApi {
+/// API for interacting with the liquidity provider pallet.
+pub trait LpApi {
```
dandanlen, line 32: LpApi is a little too generic and clashes with the actual lp api. 

How about `LpRegistration`?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -423,8 +423,9 @@ pub mod pallet {
 		type AddressConverter: AddressConverter;
 
 		/// Pallet responsible for managing Liquidity Providers.
-		type LpBalance: LpBalanceApi<AccountId = Self::AccountId>
-			+ LpDepositHandler<AccountId = Self::AccountId>;
+		type LpBalance: BalanceApi<AccountId = Self::AccountId>;
```
dandanlen, line 426: The doc comment is out of date.

dandanlen (commented):
See my comment on the Linear issue - I think maybe the description wasn't clear enough but the main objective here is to track funds that are owned (and can be eg. withdrawn) by some account, in particular in order to unify the apis and storage for broker and lp asset balances.

So we need to add the brokers' / affiliates' earned fees, for example, but the HistoricalEarnedFees are not necessary (this doesn't track a balance, it's just a 'counter' used as a convenience for rpcs, it doesn't 'own' any funds). I'll add some other comments.

state-chain/pallets/cf-asset-balances/src/lib.rs
Janislav: I think we have to keep these two things separate, or we have to book keep that we don't refund these funds to validators. That storage item only tracks failed CCM swaps. We also emit an event in this case with all information about failed swap, I guess we do this, so the emitter of the CCM swap has some kind of receipt he can use for a later refund.

state-chain/pallets/cf-ingress-egress/src/mock_btc.rs
```diff
@@ -75,6 +78,18 @@ pub type MockEgressBroadcaster =
...
 
+impl LpDepositHandler for MockDepositHandler {
+	type AccountId = u64;
+
+	fn add_deposit(
+		_who: &Self::AccountId,
+		_asset: Asset,
+		_amount: AssetAmount,
+	) -> frame_support::pallet_prelude::DispatchResult {
+		todo!()
```
dandanlen, line 89: nit: use `unimplemented!()` unless we intend to implement this.

state-chain/pallets/cf-ingress-egress/src/mock_eth.rs
```diff
@@ -78,6 +78,18 @@ impl OnDeposit<Ethereum> for MockDepositHandler {}
...
 
+impl LpDepositHandler for MockDepositHandler {
+	type AccountId = u64;
+
+	fn add_deposit(
+		_who: &Self::AccountId,
+		_asset: Asset,
+		_amount: AssetAmount,
+	) -> frame_support::pallet_prelude::DispatchResult {
+		Ok(())
```
dandanlen, line 89: We can create a better mock than this... Something like:

```suggestion
		MockBalance::try_credit_account(account_id, asset, amount)?;
		Ok(())
```

(And then the tests can check the balance.)

dandanlen (commented):
Some initial comments, will continue reviewing tomorrow.

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -319,9 +311,11 @@ pub mod pallet {
 				Error::<T>::BoostedFundsRemaining
 			);
 
-			let _ = FreeBalances::<T>::clear_prefix(&account_id, u32::MAX, None);
+			// Clear all liquidity refund addresses for the account.
 			let _ = LiquidityRefundAddress::<T>::clear_prefix(&account_id, u32::MAX, None);
-			let _ = HistoricalEarnedFees::<T>::clear_prefix(&account_id, u32::MAX, None);
```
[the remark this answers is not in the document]
Janislav: Indeed, that's a bit problematic 😅. Ill check where it could make sense to add without reintroducing a dependency between pallets.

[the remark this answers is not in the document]
Janislav: I think there are two options and both include a cross dependency between pallets:

1) Keep it now in the pool's pallet and add a trait with a method to delete the storage
2) Move it back in the LP pallet with a trait to call from the pool pallet to record historical fees

A third option would be to put it to the `on_killed_account` method like you proposed, but that's a bit hacky in my opinion because the account is actually not killed, just unregistered. I am not 100% sure if we really want to remember the historical fees even if someone unregisters his account as LP. That could be seen as a bug or a feature, not sure though. What's your opinion?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -425,9 +425,10 @@ pub mod pallet {
 		/// representation.
 		type AddressConverter: AddressConverter;
 
-		/// Pallet responsible for managing Liquidity Providers.
-		type LpBalance: LpBalanceApi<AccountId = Self::AccountId>
-			+ LpDepositHandler<AccountId = Self::AccountId>;
+		/// Pallet responsible for managing account balances.
+		type Balance: BalanceApi<AccountId = Self::AccountId>;
+
+		type LpDepositHandler: LpDepositHandler<AccountId = Self::AccountId>;
```
Janislav, line 431: I think we can remove this and use the BalanceApi instead. The only problem is that we fire an event in the LP pallet which could be a breaking change for product.

Janislav: It doesn't seem they use it.

state-chain/pallets/cf-lp/src/lib.rs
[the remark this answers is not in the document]
dandanlen: Sorry, late reply. As discussed - I think using the OnKilledAccount hook is fine.

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -308,8 +292,11 @@ pub mod pallet {
 				}),
 				Error::<T>::OpenOrdersRemaining
 			);
+			T::PoolApi::sweep(&account_id)?;
```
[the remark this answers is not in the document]
Janislav: Oh, yeah. That must have been got in due to a merge. Good catch!

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1930,3 +1945,21 @@ impl<T: Config> Pallet<T> {
...
+		HistoricalEarnedFees::<T>::mutate(account_id, asset, |balance| {
+			*balance = balance.saturating_add(amount)
+		});
+	}
+	fn get_fee_amount(account_id: Self::AccountId, asset: Asset) -> AssetAmount {
+		HistoricalEarnedFees::<T>::get(account_id, asset)
+	}
+}
+
+impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
```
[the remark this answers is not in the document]
Janislav: Can not really find where this needs to go or where we are doing this for the other pallets.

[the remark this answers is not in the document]
dandanlen: I added it - it's in the System pallet config.

dandanlen:
I'll fix the conflicts.

PR #5086, opened 2024-07-25
feat: solana transaction length check

syan095:
# Pull Request

Closes: PRO-1427

## Checklist

## Summary
Added a final transaction length check to the SolanaInstructionBuilder. If the final transaction is too long, an error is returned.
For CCM and RotateAggKey, the DurableNonce is recovered (since these are not transactional, so no storage is reverted on error)
Added unit tests and integration tests accordingly.


engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -57,7 +57,7 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 			let state_chain_client = state_chain_client.clone();
 			async move {
 				let nonce_accounts: HashMap<SolAddress, SolHash> = state_chain_client
-					.storage_map::<pallet_cf_environment::SolanaUnAvailableNonceAccounts<
```
ramizhasan111, line 60: I guess this was a typo

syan095: Just double checked - the typo is now fixed

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -72,7 +75,25 @@ impl SolanaInstructionBuilder {
 
 		final_instructions.append(&mut instructions);
 
-		final_instructions
+		// Test serialize the final transaction to obtain its length.
+		let transaction = SolTransaction::new_unsigned(SolMessage::new_with_blockhash(
+			&final_instructions,
+			Some(&agg_key),
+			&durable_nonce.1.into(),
```
albert-llimos, line 82: You have changed the `nonce_account: SolPubkey` to `durable_nonce: DurableNonceAndAccount` in all these functions to then use that durable nonce here when mocking the final serialized transaction. I think that's unnecessary. We are already crafting this mocked transaction with an empty signature (using `new_unsigned` that uses `SolSignature::default()`) so we could do the same with the blockhash. Then we don't need to do those changes and pass the durable nonce value.

albert-llimos: you could either just add the default hash value or add the `new` function to the `Message`:
```
pub fn new(instructions: &[Instruction], payer: Option<&Pubkey>) -> Self {
Self::new_with_blockhash(instructions, payer, &Hash::default())
}
```

syan095: yeah you are right, using the default hash has no affect on the length of the final transaction.

albert-llimos:
@syan095  I think it's not right after the last commit.
If we want to check the length of the transaction crafting a mock transaction without a durable nonce (using `SolMessage::new`), then the Instruction builder should return `SolInstruction` as it was doing before.
If we want to check the length of the transaction by mocking the serialization of the valid transaction, then we should do as you did before and pass the durable nonce all the way to the transaction builder.
However, in this final implementation we craft the mock transaction with a Default durable nonce hash and then return that transaction. That should yield invalid transactions since it's mocked with the default durable nonce hash.
I'd suggest one of the two:
- Rolling back the last commit and just being OK with passing the durable nonces all the way down.
- Going back to the instruction builder returning instructions. It checks the length by crafting a mock transaction but it only returns the Instruction, and then the transaction is actually built in the APi as before.

syan095:
> @syan095 I think it's not right after the last commit. 

I woke up this morning and realized the change I made was incorrect. I was gonna fix it quietly but you were too quick on the review. I'll fix it now

albert-llimos:
> I think what we can do here is to revert the last change, so we only build the transaction once. Rename the InstructionBuilder to SolanaTransactionBuilder. The `finalize` function should be renamed to "build()". What do you think?

Agree with that 👍

albert-llimos (approved):

PR #5091, opened 2024-07-29
feat: protect against pending swaps during upgrade

kylezs:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

TODO: Test

Protects against pending swaps, as the migrations here will be quite difficult to get right, so better to just ensure there are no swaps. 

However, since it's possible we could always have swaps, we allow a force flag for emergency situations. This might also be useful to resolve an issue in a rotation too.


kylezs:
we have call_as_sudo, this isn't required

PR #5093, opened 2024-07-29
Feat: solana engine broadcast retries

albert-llimos:
# Pull Request

Closes: PRO-1499

## Checklist

## Summary

Add a check that the transactions is processed after it's send (with a delay). If not, it errors so it gets retried again by the retrier rpc. See an example of this in the Rust SDK client:
https://docs.rs/solana-rpc-client/2.0.2/src/solana_rpc_client/nonblocking/rpc_client.rs.html#592
This is because the `send_transaction` won't ensure that the transaction hasn't been dropped and in case of congestion there is no other mechanism than just manually checking if the transaction has landed and resending it otherwise.

### DIFF
--- a/engine/src/sol/retry_rpc.rs
+++ b/engine/src/sol/retry_rpc.rs
@@ -10,7 +10,7 @@ use cf_chains::{
 use core::time::Duration;
 use utilities::task_scope::Scope;
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use base64::Engine;
 
 use super::{
@@ -28,8 +28,9 @@ pub struct SolRetryRpcClient {
 const SOLANA_RPC_TIMEOUT: Duration = Duration::from_millis(1000);
 const MAX_CONCURRENT_SUBMISSIONS: u32 = 100;
 
-#[allow(dead_code)]
 const MAX_BROADCAST_RETRIES: Attempt = 10;
+// Retry every 3 seconds
+const RETRY_DELAY: u64 = 3000u64;
 
 impl SolRetryRpcClient {
 	pub async fn new(
@@ -208,9 +209,18 @@ impl SolRetryRpcApi for SolRetryRpcClient {
 				Box::pin(move |client| {
 					let encoded_transaction = encoded_transaction.clone();
 					#[allow(clippy::redundant_async_block)]
-					Box::pin(
-						async move { client.send_transaction(encoded_transaction, config).await },
-					)
+					Box::pin(async move {
+						let tx_signature =
+							client.send_transaction(encoded_transaction, config).await?;
+						tokio::time::sleep(Duration::from_millis(RETRY_DELAY)).await;
+						let signature_statuses =
+							client.get_signature_statuses(&[tx_signature], true).await?;
+						match signature_statuses.value.first().and_then(|status| status.as_ref()) {
+							Some(_) => Ok(tx_signature),
+							None =>
+								Err(anyhow!("Expected a Tx Status to be Some")),
+						}
+					})
 				}),
 				MAX_BROADCAST_RETRIES,
 			)
--- a/engine/src/sol/retry_rpc.rs
+++ b/engine/src/sol/retry_rpc.rs
@@ -216,7 +216,7 @@ impl SolRetryRpcApi for SolRetryRpcClient {
 						for status_retry in 0..GET_STATUS_BROADCAST_RETRIES {
 							let signature_statuses =
 								client.get_signature_statuses(&[tx_signature], true).await?;
-							match signature_statuses.value.iter().next() {
+							match signature_statuses.value.first() {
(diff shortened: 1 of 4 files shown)


engine/src/sol/retry_rpc.rs
```diff
@@ -208,9 +209,34 @@ impl SolRetryRpcApi for SolRetryRpcClient {
...
-					Box::pin(
-						async move { client.send_transaction(encoded_transaction, config).await },
-					)
+					Box::pin(async move {
+						let tx_signature =
+							client.send_transaction(encoded_transaction, config).await?;
+
+						for status_retry in 0..GET_STATUS_BROADCAST_RETRIES {
+							let signature_statuses =
+								client.get_signature_statuses(&[tx_signature], true).await?;
```
kylezs, line 218: Is there ever a case where we won't want to search the ledger cache? If not, we can just set it to true inside the inner RPC call and save passing through this boolean every time when it's called

engine/src/sol/retry_rpc.rs
```diff
@@ -208,9 +209,34 @@ impl SolRetryRpcApi for SolRetryRpcClient {
 				Box::pin(move |client| {
 					let encoded_transaction = encoded_transaction.clone();
 					#[allow(clippy::redundant_async_block)]
-					Box::pin(
-						async move { client.send_transaction(encoded_transaction, config).await },
-					)
+					Box::pin(async move {
```
kylezs, line 212: you could write a test for this by mocking the client to return error or Ok responses at particular times

engine/src/sol/retry_rpc.rs
kylezs, line 1: Noticed the tests in this file aren't actually testing anything, they are just calling external RPCs - would `#[ignore]` them

engine/src/sol/retry_rpc.rs
```diff
@@ -208,9 +209,34 @@ impl SolRetryRpcApi for SolRetryRpcClient {
...
+						let tx_signature =
+							client.send_transaction(encoded_transaction, config).await?;
+
+						for status_retry in 0..GET_STATUS_BROADCAST_RETRIES {
+							let signature_statuses =
+								client.get_signature_statuses(&[tx_signature], true).await?;
+							match signature_statuses
+								.value
+								.first()
+								.and_then(|status| status.as_ref())
```
kylezs, line 222: Doesn't make too much of a difference, but, slightly simpler
```suggestion
match signature_statuses.value.iter().next()
```

engine/src/sol/retry_rpc.rs
```diff
@@ -208,9 +209,34 @@ impl SolRetryRpcApi for SolRetryRpcClient {
...
+							{
+								Some(_) => return Ok(tx_signature),
+								None => {
+									// Ignore sleep at last step
+									if status_retry < GET_STATUS_BROADCAST_RETRIES {
+										// Retry twice a second
```
kylezs, line 228: this comment is only as accurate if the constant doesn't change - would just remove the comment tbh, it's pretty clear

engine/src/sol/retry_rpc.rs
albert-llimos: not fully clear. Just today I started with this:
https://github.com/chainflip-io/chainflip-backend/pull/5095
Where we might set that to false for at least some of those queries. So for now I want to leave it flexible as it might be needed in th efuture.

kylezs (approved):
Review from someone more solana aware is probably worthwhile too

engine/src/sol/retry_rpc.rs
albert-llimos: after a chat we decided that it's not worth it as it needs a bunch of logic for mocks for little benefit.

engine/src/sol/retry_rpc.rs
```diff
@@ -208,9 +209,28 @@ impl SolRetryRpcApi for SolRetryRpcClient {
...
+							client.send_transaction(encoded_transaction, config).await?;
+
+						for status_retry in 0..GET_STATUS_BROADCAST_RETRIES {
+							let signature_statuses =
+								client.get_signature_statuses(&[tx_signature], true).await?;
+							match signature_statuses.value.first() {
+								Some(_) => return Ok(tx_signature),
+								None => {
+									// Ignore sleep at last step
+									if status_retry < GET_STATUS_BROADCAST_RETRIES {
```
AlastairHolmes, line 223: This is never false.

AlastairHolmes: I'd suggest using a interval/make_periodic_tick and putting the sleep before instead of after, that way you avoid the need to count anything.

albert-llimos: I implemented it so it had the same behavior as the rust client but makes sense what you said. Just pushed an update using that.

AlastairHolmes (approved):

PR #5102, opened 2024-08-01
feat: emit swap request first

msgmaxim:
# Pull Request

## Checklist

## Summary

`SwaprRequested` is now emitted before any other event (e.g. SwapScheduled) that mentions swap_request_id, as was requested by the product. One difficulty was that we can reject a CMM request in some cases (the check is done in `init_swap_request` for various reasons), and originally I didn't want to emit `SwapRequested` event in that case, but now I'm thinking it is OK as long as we also emit `SwapRequestCompleted` immediately after `CcmFailed`, so that's what I did here. (The alternative was to split the match statement so we check check, then emit the event, and then later only assert on the fact that we have performed the check, but that would make the code a bit ugly).

### DIFF
--- a/bouncer/shared/utils.ts
+++ b/bouncer/shared/utils.ts
@@ -296,7 +296,7 @@ export async function observeSwapEvents(
   tag?: string,
   finalized = false,
 ): Promise<BroadcastChainAndId | undefined> {
-  let eventFound = false;
+  let broadcastEventFound = false;
   const subscribeMethod = finalized
     ? api.rpc.chain.subscribeFinalizedHeads
     : api.rpc.chain.subscribeNewHeads;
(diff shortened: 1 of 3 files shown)


bouncer/shared/utils.ts
```diff
@@ -318,81 +318,69 @@ export async function observeSwapEvents(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const events: any[] = await api.query.system.events.at(header.hash);
 
-    // Some events don't appear in the order one might expect (e.g. SwapRequested
-    // comes after SwapScheduled), so as a workaround, we add an other loop to be
-    // able to re-scan events that we already visited:
-    let keepSearching = true;
```
kylezs, line 324: nice 👍

kylezs (approved):

PR #5103, opened 2024-08-01
refactor: bouncer cmd for FoK swap

j4m1ef0rd:
# Pull Request

Closes: PRO-1537

## Checklist

## Summary

- Refactored the new swap command to use yargs to allow for more complex command line arguments.
- Added the 3 FoK refund parmas as command line options.
- Changed the `maxBoostFeeBps` to a command line option instead of an optional positional argument.

### DIFF
--- a/bouncer/commands/new_swap.ts
+++ b/bouncer/commands/new_swap.ts
@@ -1,35 +1,97 @@
 #!/usr/bin/env -S pnpm tsx
 // INSTRUCTIONS
 //
-// This command takes four arguments.
-// It will request a new swap with the provided parameters
-// Argument 1 is the source currency ("Btc", "Eth", "Dot" or "Usdc")
-// Argument 2 is the destination currency ("Btc", "Eth", "Dot" or "Usdc")
-// Argument 3 is the destination address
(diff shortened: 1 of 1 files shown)


bouncer/commands/new_swap.ts
```diff
@@ -1,35 +1,97 @@
...
-  console.log(`Requesting swap ${sourceAsset} -> ${destAsset}`);
+  if ((args.refundAddress === undefined) !== (args.minPrice === undefined)) {
+    throw new Error('Must specify both refundAddress and minimumPrice when using refund options');
+  }
+  const refundParameters: RefundParameters | undefined =
+    args.refundAddress !== undefined
+      ? {
+          retryDurationBlocks: args.refundDuration,
+          refundAddress: args.refundAddress,
+          minPrice: args.minPrice ? args.minPrice : '0',
```
kylezs, line 81: we could just use default as `0` instead of this?

kylezs (approved):

bouncer/commands/new_swap.ts
j4m1ef0rd: This '0' will never get hit because of the check above. Just needed this to make the linter happy. I can change it to something like 'error' so its more obvious.

PR #5104, opened 2024-08-01
fix: prewitness deposits without delaying for a block

kylezs:
# Pull Request

## Checklist

## Summary

Means we won't delay the prewitnessing of a particular. Previously we would wait for the *next* external chain block before prewitnessing.

This release would have to be done runtime upgrade first and then CFE upgrades.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -1752,7 +1752,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 	fn expiry_and_recycle_block_height(
 	) -> (TargetChainBlockNumber<T, I>, TargetChainBlockNumber<T, I>, TargetChainBlockNumber<T, I>)
 	{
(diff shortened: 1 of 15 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1752,7 +1752,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 	fn expiry_and_recycle_block_height(
 	) -> (TargetChainBlockNumber<T, I>, TargetChainBlockNumber<T, I>, TargetChainBlockNumber<T, I>)
 	{
-		let current_height = T::ChainTracking::get_block_height();
+		let current_height =
+			T::ChainTracking::get_block_height() + <T::TargetChain as Chain>::WITNESS_PERIOD;
```
martin-chainflip, line 1756: why this change?

AlastairHolmes (approved):

martin-chainflip (approved):

PR #5105, opened 2024-08-02
Chore: remove 1.5 migrations

msgmaxim:
# Pull Request

## Checklist

Please conduct a thorough self-review before opening the PR.

- [x] I am confident that the code works.
- [ ] I have updated documentation where appropriate.

## Summary

Removed migrations that are already in the 1.5 branch. Left `housekeeping` and `reap_old_accounts` as it looks like we want to apply them on every upgrade?

PR #5106, opened 2024-08-02
Feat: Core DCA

msgmaxim:
# Pull Request

Closes: PRO-1531, PRO-1539

## Checklist

## Summary

I'm hoping to merge this core functionality early so that @j4m1ef0rd could start working on this in parallel (we can/might also work on this branch ofc, but that's not ideal). This means some important features might come in separate PRs (see below).

- The commits are mostly self contained and can be reviewed one-by-one.
- Added `dca_parameters` to the `request_swap_deposit_address_with_affiliates` extrinsic and the the broker rpc method; the parameters are saved on the channel's actions (similarly to refund params). New migration adds None to any existing channels.
- DCA state is stored in `RequestState`. Now when a swap completes, we check DCA state to see what should happen next: if the swap/chunk is successful, we either schedule the next chunk (recording output) or egress if this was the final chunk; on failure, we refund any unexecuted amount, but still egress output from any previous chunks. No migration is needed since `RequestState`s don't exist in 1.5.

Out of scope:

- Configurable DCA Limits and parameter validation ([PRO-1538](https://linear.app/chainflip/issue/PRO-1538/dca-parameter-validation))
- Bouncer tests ([PRO-1532](https://linear.app/chainflip/issue/PRO-1532/bouncer-tests-for-dca))
- Broker fee charged per chunk ([PRO-1540](https://linear.app/chainflip/issue/PRO-1540/broker-fee-should-be-charged-per-dca-chunk))

**Update**: since this is early in review, I pushed another commit that now handles CCM too.


api/bin/chainflip-cli/src/main.rs
```diff
@@ -71,6 +71,7 @@ async fn run_cli() -> Result<()> {
 								params.boost_fee,
 								Default::default(),
 								None,
+								None,
```
kylezs, line 74: https://linear.app/chainflip/issue/PRO-1544/add-params-to-swaprequestparams-in-cli

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -242,27 +245,44 @@ pub struct CcmSwapAmounts {
...
-		principal_output_amount: AssetAmount,
-	},
+enum GasSwapState {
+	OutputReady { gas_budget: AssetAmount },
+	Scheduled { gas_swap_id: SwapId },
+	ToBeScheduled { gas_budget: AssetAmount, other_gas_asset: Asset },
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+struct DCAState {
```
dandanlen, line 255: Same nit as above: `DcaState` should be preferred over `DCAState`.

dandanlen (commented):
As mentioned in the previous PR, I think it's worth distinguishing more strongly between id types just for peace of mind. I've created a Linear issue PRO-1548.

dandanlen:
(Will continue reviewing tomorrow)

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1215,154 +1209,241 @@ pub mod pallet {
...
+			let Some(request) = SwapRequests::<T>::take(swap_request_id) else {
+				log_or_panic!("Swap request {swap_request_id} not found");
+				return;
+			};
 
-				let Some(output_amount) = swap.final_output else {
-					log_or_panic!("Swap {} is not completed yet!", swap.swap_id());
-					continue;
-				};
+			Self::deposit_event(Event::<T>::SwapRequestCompleted { swap_request_id: request.id });
```
dandanlen, line 1223: Any particular reason for sandwiching this between these two checks instead of emitting it after both?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1215,154 +1209,241 @@ pub mod pallet {
...
-									);
+			match request.state {
+				SwapRequestState::Regular {
+					dca_state: DCAState { remaining_input_amount, accumulated_output_amount, .. },
+					output_address,
+				} => {
+					// Refund the failed swap and any unused input amount:
+					Self::egress_for_swap(
+						request.id,
+						swap.input_amount + remaining_input_amount,
```
dandanlen, line 1238: I don't understand this - why refund the `input_amount + remaining_input_amount` and not just the remaining amount?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1215,154 +1209,241 @@ pub mod pallet {
...
+							output_amount + dca_state.accumulated_output_amount,
+							swap.output_asset(),
+							output_address.clone(),
+							None,  /* ccm */
+							false, /* refund */
+						);
+					}
+				},
+				SwapRequestState::NetworkFee => {
+					Self::deposit_event(Event::<T>::SwapRequestCompleted { swap_request_id });
```
dandanlen, line 1420: It feels like there ought to be a way to deduplicate this... It's not easy to see if this event is being emitted in all the right places (ie. that every request eventually completes).

As a step in that direction, we could require that this match statement returns some indication of whether the swap completed or not. Then at least we can be sure that each match arm handles the question of whether the swap is completed or not.

dandanlen (commented):
Actually a few more comments here. More to follow tomorrow.

state-chain/pallets/cf-ingress-egress/src/migrations/add_refund_params.rs
```diff
@@ -70,6 +70,7 @@ impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
 						destination_address,
 						broker_fees,
 						refund_params: None,
+						dca_params: None,
```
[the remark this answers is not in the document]
msgmaxim: The idea was/is to merge https://github.com/chainflip-io/chainflip-backend/pull/5105 first.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -242,27 +245,44 @@ pub struct CcmSwapAmounts {
...
+	ToBeScheduled { gas_budget: AssetAmount, other_gas_asset: Asset },
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+struct DCAState {
+	scheduled_chunk_swap_id: Option<SwapId>,
+	remaining_input_amount: AssetAmount,
+	remaining_chunks: u32,
+	accumulated_output_amount: AssetAmount,
+}
```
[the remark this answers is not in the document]
msgmaxim: What would this be useful for? Debugging by inspecting storage?

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: No reason, updating.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: `remaining_input_amount` is the amount to be used in future chunks. When a chunk fails, we still want to refund it, just like we are doing for non-DCA swaps in 1.5. You could argue that `remaining_input_amount` should include the  amount from the currently scheduled swap, but I don't know if one approach is clearly better than the other. Regular swaps didn't have `remaining_input_amount`, so we were refunding `swap.input_amount`. The current approach is just a natural progression of that code.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1540,6 +1542,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				destination_asset,
 				broker_fees,
 				refund_params,
+				dca_params,
```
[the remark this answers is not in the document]
msgmaxim: What is the exact plan to address this? A hotfix for 1.5 and a refactor here/separate PR?

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: > we could require that this match statement returns some indication of whether the swap completed or not.

Updated to do this. As a bonus, now putting the request back into storage is done in one place also.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: TBC...

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2021,13 +2028,14 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
 		channel_metadata: Option<CcmChannelMetadata>,
 		boost_fee: BasisPoints,
 		refund_params: Option<ChannelRefundParameters>,
+		dca_params: Option<DCAParameters>,
```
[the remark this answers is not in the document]
dandanlen: Ok, sorry, I do read the description, but this is long PR, it's easy to forget things.

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
dandanlen: I was thinking about things like rpc queries, but yes this too. 

Taking the idea a bit further, storing a history of chunk executions rather than just swap ids would make it easier to reason about correctness, and to simulate dca execution for testing: All of the values in this struct (remaining chunks, amount, total output) can be derived from the initial swap/dca parameters and the history of dca swap executions. Then we only need to mutate a single value (push a new chunk to a vec) rather than relying on consistently mutating multiple values.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
zoheb391: FWIW the PR description really helps rust illiterates like me so its not entirely in vain

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1246,154 +1220,248 @@ pub mod pallet {
...
+							Some(swap_id)
 						} else {
-							log_or_panic!(
-								"IngressEgressFee swap should not be to non-gas asset: {:?}",
-								swap.output_asset()
+							None
+						};
+
+						// See if we still need to schedule the gas swap
+						if let GasSwapState::ToBeScheduled { gas_budget, other_gas_asset } =
```
dandanlen, line 1369: Does this mean the gas swap is scheduled at the same block as the last chunk of the dca swap? Is this intentional?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1246,154 +1220,248 @@ pub mod pallet {
...
+
+						true
+					} else {
+						false
+					}
+				},
+				SwapRequestState::Regular { output_address, dca_state } => {
+					if dca_state.remaining_chunks > 0 {
+						let chunk_input_amount = dca_state.prepare_next_chunk();
+						dca_state.accumulated_output_amount += output_amount;
```
dandanlen, line 1403: We only add the output amount if the remaining chunks > 0. Is this intentional? In the ccm branch this line is outside the if clause. 

Seems like a good candidate for deduplication.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1246,154 +1220,248 @@ pub mod pallet {
...
+							request.id,
+						);
+
+						false
+					} else {
+						debug_assert!(dca_state.remaining_input_amount == 0);
+
+						Self::egress_for_swap(
+							swap_request_id,
+							output_amount + dca_state.accumulated_output_amount,
```
dandanlen, line 1421: Ok I see - it's because of this? 

Seems more consistent to add it above and then egress the `accumulated_output_amount`, no?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1816,17 +1942,37 @@ pub mod pallet {
...
 						};
 
-					// See if princial swap is needed, schedule it first if so:
+					// See if principal swap is needed, schedule it first if so:
 					if input_asset != output_asset && !principal_swap_amount.is_zero() {
+						let mut dca_state = DcaState {
+							scheduled_chunk_swap_id: None,
+							remaining_input_amount: principal_swap_amount,
+							// No DCA is equivalent to DCA with 1 chunk
+							remaining_chunks: dca_params.map(|p| p.number_of_chunks).unwrap_or(1),
```
dandanlen, line 1951: Just realised that the dca_params.chunk_interval isn't used anywhere... (and it's not mentioned in the PR description, I checked ;) )

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1816,17 +1942,37 @@ pub mod pallet {
...
+							accumulated_output_amount: 0,
+						};
+						let chunk_input_amount = dca_state.prepare_next_chunk();
+
+						let swap_refund_params = refund_params.as_ref().map(|params| {
+							utilities::calculate_swap_refund_parameters(
```
dandanlen, line 1957: Seems a bit odd to have a utility for this? 

Could it not be a method on one of the structs?

dandanlen (commented):
Some more comments...

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: No, it is scheduled at the same time as the second chunk (i.e. as soon as we have one successful chunk).

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: This part of the code uses if/else, which I think should be preferred over two separate if statements. The CCM case is more complex (two swaps can be executed concurrently), so a separate if statement was needed.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: See the answer above. It is a trade-off. We have to choose between consistency and rewriting if/esle with two separate ifs on the same condition, which seems ugly too.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: You are right, I missed this because the tests used the same value for chunk interval as the default SWAP_DELAY. Fixed now and updated the tests as well.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: @msgmaxim 

Update: There's an open PR to main - but it might be simpler to just fix it here. LMK.

Since the fix will be applied to 1.5, no need for a migration.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: But for example here, I can see that we don't update/assign the scheduled_chunk_swap_id. Maybe it's not significant, given that we only use it during ccm(?), but I can see how it will cause confusion. 

Surely when there is a swap execution, we *always* want to update the dca_state in the same way? So why can't we move this outside of the checks and do it consistently for every type of swap?

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: minor nit: an enum would have been nicer than a bool. Using bools, when reading the code and seeing that some match arm returns true/false, you need to scroll up again to remind yourself what the true/false means. Something like RequestStatus::Complete, for example, would be friendlier.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Ok - I thought the idea was to fully execute the principal swap and then do the gas swap. But I suppose this makes sense.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -242,27 +245,44 @@ pub struct CcmSwapAmounts {
...
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+struct DCAState {
+	scheduled_chunk_swap_id: Option<SwapId>,
+	remaining_input_amount: AssetAmount,
+	remaining_chunks: u32,
+	accumulated_output_amount: AssetAmount,
+}
+
+impl DCAState {
+	fn prepare_next_chunk(&mut self) -> AssetAmount {
```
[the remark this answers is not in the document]
msgmaxim: Yeah, I was thinking about it too. The only thing is that the very first chunk in init_swap_request always exists, but now we have to somewhat awkwardly check for that. Probably an improvement still though.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: > given that we only use it during ccm(?)

The intention was to also use it for regular swaps, just as a sanity check at least. Will update.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1201,199 +1226,261 @@ pub mod pallet {
...
+							accumulated_output_amount,
+							request.output_asset,
+							output_address,
+							None,  /* ccm */
+							false, /* refund */
+						);
+					}
+				},
+				SwapRequestState::Ccm { output_address: _, .. } => {
+					// TODO: add remaining DCA input to the refunded amount:
```
dandanlen, line 1292: This TODO should be included in this PR? Or a future one?

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
dandanlen: Not always - what about ccm swaps with zero principal?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -243,27 +247,59 @@ pub struct CcmSwapAmounts {
...
+	remaining_input_amount: AssetAmount,
+	remaining_chunks: u32,
+	chunk_interval: u32,
+	accumulated_output_amount: AssetAmount,
+}
+
+impl DcaState {
+	// Create initial DCA state; if no dca parameters provided (for non-DCA swaps),
+	// this creates state equivalent to 1 chunk DCA
+	fn new(input_amount: AssetAmount, params: Option<DcaParameters>) -> DcaState {
```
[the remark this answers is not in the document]
msgmaxim: It is used in 2 out of 3 places where DcaState is created. The third one is a special case where we want 0 dca chunks: in case of CCM where only gas swap (but not the principal) is required. I decided constructing the state by hand was justified for that case.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -243,27 +247,59 @@ pub struct CcmSwapAmounts {
...
+	}
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
 enum SwapRequestState {
 	Ccm {
-		state: CcmSwapState,
+		gas_swap_state: GasSwapState,
 		ccm_deposit_metadata: CcmDepositMetadata,
 		output_address: ForeignChainAddress,
```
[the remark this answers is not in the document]
msgmaxim: > we could easily make a requirement for every swap type - a NetworkFee's destination is the vault for example

I'm not sure it is a good idea. The CCM and Regular swaps let users specify output address and we actually use that value once the swap is executed. While we can think of some values that would seem not too "out of place", they aren't going to be used in the same way. We are not egressing network fees like we do for regular swaps, at least not directly, not in USDC. And even if we were, there is only one value (the vault) that makes sense, and it should be stored separately. Why would we store it individually for each swap and open us to the possiblity of accidentally putting the wrong value there?

I do see the how it might make sense to unify Ccm and Regular seeing that they have some fields in common, but I think we would be generalising prematurely. For example, we discussed how the outcome of a swap should be handled outside of the swapping pallet (and I think this is still the plan, just not the priority right now). Once `output_address` is out, I'm not sure it would be worth refactoring further just because of one duplicated field. I agree we we should keep these ideas in mind for the future though.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1603,6 +1691,59 @@ pub mod pallet {
...
+					if is_refund {
+						Self::deposit_event(Event::<T>::RefundEgressScheduled {
+							swap_request_id,
+							egress_id,
+							asset,
+							amount: egress_amount,
+							egress_fee: fee_withheld,
+						});
+					} else {
+						Self::deposit_event(Event::<T>::SwapEgressScheduled {
```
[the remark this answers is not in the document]
msgmaxim: We have discussed this before, and last I heard we were going to keep them separate... However, I started a thread on discord to revisit this.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1201,199 +1221,266 @@ pub mod pallet {
...
+					if dca_state.remaining_chunks > 0 {
+						let chunk_input_amount = dca_state.prepare_next_chunk();
+						dca_state.accumulated_output_amount += output_amount;
+
+						Self::schedule_swap(
+							request.input_asset,
+							request.output_asset,
+							chunk_input_amount,
+							// TODO: recompute swap params depending on the output amount we got
+							swap.swap.refund_params.clone(),
```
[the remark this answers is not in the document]
msgmaxim: I think we already have the distinction between "swap requests" and "swaps" (I chose to continue using "swap" in the context of amm to isloate the changes to the swapping pallet).

In this specific line, the two swaps are both actually at the "execution" level. I agree the naming isn't great, but the outer swap refers to `SwapState` (contains the result of the execution), while the inner is `Swap` the immutable parameters of the execution/swap. I will create a getter method for `refund_params` like I did for other fields to avoid this chaining.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: This is just a reminder for a future PR which will add support for FoK to CCM swaps.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: I have a slight bias towards functions over methods, but maybe a method would work better here. Updating.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: Cherry picked that PR. I also updated DCA migration code to include broker fee for ccm.

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: Then it doesn't get called. Chunks only exist on principal/regular swaps as a concept.

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
dandanlen: I was thinking output_address could become an enum, something like:

```rust
enum Destination {
	ExternalAccount(ForeignChainAddress),
	InternalAccount(AccountId),
	Vault(StorageItem), // not sure if/how this would work - might just be an inner enum
}
```

Or something like this. Admittedly not fully thought-through, but I feel like this would give us more flexibility - for example it would allow us to add features like 'swap deposits' almost trivially: it's just a swap request with an InternalAccount as destination.

With the current implementation we would need a new variant SwapDeposit with it's own DCA params, it's own (mostly duplicated) match arm in the process_outcome fn etc.. Anyway, not for this PR..

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -243,27 +247,93 @@ pub struct CcmSwapAmounts {
...
+	fn prepare_next_chunk(
+		&mut self,
+		prev_chunk_and_output: Option<(SwapId, AssetAmount)>,
+	) -> Option<AssetAmount> {
+		if let Some((prev_chunk_swap_id, prev_chunk_output_amount)) = prev_chunk_and_output {
+			if let Some(scheduled_swap_id) = self.scheduled_chunk_swap_id {
+				if scheduled_swap_id != prev_chunk_swap_id {
+					log_or_panic!(
+				"Invariant violation: the recorded chunk id {scheduled_swap_id} does not match executed {prev_chunk_swap_id}"
+			);
```
dandanlen, line 299: nit

```suggestion
					log_or_panic!(
						"Invariant violation: the recorded chunk id {scheduled_swap_id} does not match executed {prev_chunk_swap_id}"
					);
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -243,27 +247,93 @@ pub struct CcmSwapAmounts {
...
+
+		(state, first_chunk_amount)
+	}
+
+	fn prepare_next_chunk(
+		&mut self,
+		prev_chunk_and_output: Option<(SwapId, AssetAmount)>,
+	) -> Option<AssetAmount> {
+		if let Some((prev_chunk_swap_id, prev_chunk_output_amount)) = prev_chunk_and_output {
+			if let Some(scheduled_swap_id) = self.scheduled_chunk_swap_id {
```
dandanlen, line 295: Would it make sense to remove from the scheduled_chunk_swap_id? (Since it's no longer scheduled). 

```suggestion
			if let Some(scheduled_swap_id) = self.scheduled_chunk_swap_id.take() {
```

dandanlen (commented):
Thanks, looking good 🚀 . I'll take a closer look at tests on Monday.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Yes! Updated.

kylezs:
Migrations might be impacted, this PR was required because otherwise CI on main doesn't pass: https://github.com/chainflip-io/chainflip-backend/pull/5142

dandanlen (commented):
Test cases look fine to me. Although I feel we can do better to reduce duplication in the tests and make them more readable - they are quite verbose and repetitive (this applies to pre-existing tests too, not just the newly added ones). For example, the use of 'global' constants for test amounts etc makes things a little hard to follow. 

I think it's time we added some simple extension traits with common setup/update/check methods that are used across all the swapping tests. I'll raise an issue.

Apart from that I believe the only things that needs changing is what was discussed this morning re. expiry being implicit from the FoK retry delay and number of dca chunks - I think this implies simply updating the refund params for every chunk.

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -0,0 +1,582 @@
...
+				RuntimeEvent::Swapping(Event::SwapRequestCompleted {
+					swap_request_id: SWAP_REQUEST_ID
+				}),
+			);
+		});
+}
+
+/// Test that DCA with 1 chunk behaves like a regular swap
+#[test]
+fn dca_single_chunk() {
```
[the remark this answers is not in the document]
msgmaxim: Sure. Still, I feel like it is nicer to have an explicit test? Otherwise, it is not immediately clear that this case is (implicitly) covered by non-DCA tests. Can remove if you feel strongly about it though.

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -0,0 +1,582 @@
...
+					execute_at: CHUNK_2_BLOCK,
+					..
+				})
+			);
+		})
+		.then_execute_at_block(CHUNK_2_BLOCK, |_| {})
+		.then_execute_with(|_| {
+			assert_eq!(SwapRequests::<Test>::get(SWAP_REQUEST_ID), None);
+
+			assert_event_sequence!(
```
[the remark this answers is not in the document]
msgmaxim: OK, added extra checks for the dca state. Also updated all existing tests to use a swapping rate 1:2.

state-chain/pallets/cf-swapping/src/tests/dca.rs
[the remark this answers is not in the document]
dandanlen: Fine to leave it - as mentioned above, I think the tests are due a refactor anyway.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2181,11 +2171,12 @@ pub(crate) mod utilities {
...
 	pub(super) fn calculate_swap_refund_parameters(
 		params: &ChannelRefundParameters,
-		deposit_block: u32,
+		execute_at_block: u32,
 		input_amount: AssetAmount,
 	) -> SwapRefundParameters {
 		SwapRefundParameters {
-			refund_block: deposit_block.saturating_add(params.retry_duration),
+			// NOTE: We add 1 to make retry duration inclusive of the last block:
+			refund_block: execute_at_block.saturating_add(params.retry_duration).saturating_add(1),
```
dandanlen, line 2180: I don't understand why this was added - is it necessary/signficant?

dandanlen (approved):

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: You are right, it wasn't necessary, reverted. I think I forgot that we still execute one last time at `refund_block` before refunding as that block, so it already works as intended.

PR #5118, opened 2024-08-05
fix: qualify nodes based on reputation

dandanlen:
Fixes some issues I noticed while reviewing #5062. (Not strictly related to the PR itself, but still..)

1. The implementation of QualifyNode::filter_unqualified was wrong. Fortunately it was also unused (another mistake). 
2. We implemented the reputation-based qualification using is_qualified and the (buggy) default impl of filter_unqualified. The implementation using is_qualified is very ewxpensive (storage iteration over all validators for each validator, so n^2 storage reads).

I think the reason is wasn't noticed it probably that (a) it wasn't actually used (it should have been!) and (b) the naming was confusing.

I changed the name and made sure it's used when filtering backup validators.

I also created an explicit struct for the heartbeat-based qualification rather than using the Pallet.

Everything is squashed into a single commit, but it should be small enough to be reviewable as is. This is what the commit contains:

- Rename filter_unqualified to take_qualified
- Fix implementation (remove the `!`)
- Use take_qualified in the implementation
- Add clarifying documentation.
- Create HeartbeatQualification
- Move impls out of pallet module.
- Tests

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -1218,10 +1218,13 @@ impl<T: Config> Pallet<T> {
 	/// sorted by bids highest to lowest.
 	pub fn highest_funded_qualified_backup_node_bids(
 	) -> impl Iterator<Item = Bid<ValidatorIdOf<T>, <T as Chainflip>::Amount>> {
-		let mut backups: Vec<_> = Backups::<T>::get()
+		let backup_map = Backups::<T>::get();
+		let qualified_backups =
+			T::KeygenQualification::take_qualified(backup_map.keys().cloned().collect());
+		let mut backups = backup_map
 			.into_iter()
-			.filter(|(bidder_id, _)| T::KeygenQualification::is_qualified(bidder_id))
-			.collect();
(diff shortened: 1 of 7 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1218,10 +1218,13 @@ impl<T: Config> Pallet<T> {
 	/// sorted by bids highest to lowest.
 	pub fn highest_funded_qualified_backup_node_bids(
 	) -> impl Iterator<Item = Bid<ValidatorIdOf<T>, <T as Chainflip>::Amount>> {
-		let mut backups: Vec<_> = Backups::<T>::get()
+		let backup_map = Backups::<T>::get();
+		let qualified_backups =
+			T::KeygenQualification::take_qualified(backup_map.keys().cloned().collect());
```
kylezs, line 1223: both uses of `take_qualified` are then followed by a filter of something else, based on the `qualified` list, perhaps we can factor out further here

kylezs (approved):

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: Yeah I think I have an idea.

kylezs (approved):

PR #5119, opened 2024-08-05
fix: minor changes to asset-balances pallet

dandanlen:
These are just some minor changes/clarifications I added when reviewing #4962.

- I removed the distinction between refund limits for different chains, seems unnecessary (same for ARB and ETH, irrelevant for BTC, DOT, and soon to be removed). 
- I added some constants to clarify test numbers.


dandanlen:
Closing this, it's not important.

PR #5127, opened 2024-08-07
fix: ccm transfers should charge broker fee

dandanlen:
# Pull Request

Closes: PRO-1545

## Checklist

## Summary

Adds the broker fees to the ChannelAction::CcmTransfer variant and passes them to the swap request.

I will add the migration in the release branch since this is going straight to release/1.5.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/migrations/add_refund_params.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/migrations/add_refund_params.rs
@@ -18,6 +18,7 @@ pub(super) mod old {
 		CcmTransfer {
 			destination_asset: Asset,
 			destination_address: ForeignChainAddress,
+			broker_fees: Beneficiaries<AccountId>,
 			channel_metadata: CcmChannelMetadata,
 		},
 	}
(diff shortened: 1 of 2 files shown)


state-chain/pallets/cf-ingress-egress/src/migrations/add_refund_params.rs
```diff
@@ -76,10 +77,12 @@ impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
 					old::ChannelAction::CcmTransfer {
 						destination_asset,
 						destination_address,
+						broker_fees,
 						channel_metadata,
 					} => ChannelAction::CcmTransfer {
 						destination_asset,
 						destination_address,
+						broker_fees,
```
kylezs, line 85: Isn't the more correct thing to do here to delete the migration? It'll be applied on 1.5 so we don't need to update this

msgmaxim: This is a hotfix and should go directly into 1.5 I believe. My understanding (from chatting with @dandanlen) is that we will integrate this fix (in some form) directly into https://github.com/chainflip-io/chainflip-backend/pull/5106.

dandanlen:
Yes, migration should have been deleted. I'll delete this since I think it will be easier to add with/after #5106

PR #5128, opened 2024-08-07
Feat/migration sol chain state

ramizhasan111:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

#### Non-Breaking changes


state-chain/pallets/cf-chain-tracking/src/lib.rs
```diff
@@ -28,7 +28,7 @@ impl Get<FixedU128> for GetOne {
 	}
 }
 
-pub const PALLET_VERSION: StorageVersion = StorageVersion::new(3);
+pub const PALLET_VERSION: StorageVersion = StorageVersion::new(4);
```
kylezs, line 31: We don't need to update this, the storage hasn't changed, we're just inserting a value

Instead we can re-introduce the `SpecVersionedMigration`

```rust
use sp_std::{vec, vec::Vec};

use sp_runtime::DispatchError;

/// A runtime storage migration that will only be applied if the `SPEC_VERSION` matches the
/// post-upgrade runtime's spec version.
pub struct SpecVersionedMigration<U, const SPEC_VERSION: u32>(PhantomData<U>);

impl<U, const SPEC_VERSION: u32> OnRuntimeUpgrade for SpecVersionedMigration<U, SPEC_VERSION>
where
	U: OnRuntimeUpgrade,
{
	fn on_runtime_upgrade() -> Weight {
		if System::runtime_version().spec_version == SPEC_VERSION {
			U::on_runtime_upgrade()
		} else {
			log::info!(
				"⏭️ Skipping storage migration for version {:?} - consider removing this from the runtime.",
				SPEC_VERSION
			);
			Weight::zero()
		}
	}

	fn pre_upgrade() -> Result<Vec<u8>, DispatchError> {
		if System::runtime_version().spec_version == SPEC_VERSION {
			U::pre_upgrade()
		} else {
			Ok(vec![])
		}
	}

	fn post_upgrade(state: Vec<u8>) -> Result<(), DispatchError> {
		if System::runtime_version().spec_version == SPEC_VERSION {
			U::post_upgrade(state)
		} else {
			Ok(())
		}
	}
}
```

state-chain/runtime/src/lib.rs
```diff
@@ -1223,6 +1224,39 @@ type MigrateApicalls = (
...
+		NoopRuntimeUpgrade,
+		3,
+		4,
+	>,
+	VersionedMigration<
+		pallet_cf_chain_tracking::Pallet<Runtime, ArbitrumInstance>,
+		NoopRuntimeUpgrade,
+		3,
+		4,
+	>,
```
kylezs, line 1251: Above comment also means we don't have to do this

state-chain/pallets/cf-chain-tracking/src/lib.rs
ramizhasan111: yeah, it was either this or using spec-versioned migration but since we deleted it, I did it this way. I'll add the spec versioned migration back.

kylezs (approved):

state-chain/runtime/src/lib.rs
```diff
@@ -1186,6 +1186,10 @@ type PalletMigrations = (
 		11,
 	>,
 	MigrateApicalls,
+	migrations::spec_versioned_migration::SpecVersionedMigration<
+		migrations::solana_integration::SolanaChainState,
+		151,
```
dandanlen, line 1191: This is not good practice generally speaking since there's no guarantee we won't bump the spec version again before the release (ie. by the time we release to mainnet the spec version might be 152). But probably ok this time - we just need to remember to change this number if the version is bumped.

dandanlen (dismissed):

state-chain/runtime/src/lib.rs
dandanlen: Actually I think it's safer to just remove the spec version constraint.

ramizhasan111: will do

kylezs (dismissed):

dandanlen:
Please don't merge - I will consolidate all the changes in a single branch/PR.

PR #5129, opened 2024-08-07
fix: ccm transfers should charge broker fee

dandanlen:
Fixes ccm broker fees (see #5127).

It was a bit more work than on main because of some additional trait interfaces that needed to be updated. 

Also the migration is a little tricky because it updates an intermediate type from a previous migration (ChannelAction). 

There are still some issues with the migration when checking it with try-runtime, but I think these are unrelated.


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1618,6 +1622,14 @@ pub mod pallet {
 
 			let mut swap_output = CcmSwapOutput::default();
 
+			// Permill maxes out at 100% so this is safe.
+			let fee: u128 = Permill::from_parts(
```
kylezs, line 1626: Pull into function? - Ideally we can combine these two whole deposit functions as I think you were referring to yesterday - but that can be done on `main`

kylezs (approved):

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Yes the change is much smaller on main - this code path has already been consolidated with the other.

dandanlen:
Closing in favour of #5132

PR #5133, opened 2024-08-08
fix: api call migrations

kylezs:
# Pull Request

## Checklist

## Summary

Changes the order of the migrations, and how the migration is achieved to simplify it.

The issue before this fix was:
The 3->4 migration drains the ThresholdSignature data storage, and inserts the PendingApiCalls storage. When we are draining from ThresholdSignatureData, we are decoding into the new ApiCall type (note the lack of an old ApiCall) type. 

 This now passes try-runtime on mainnet and perseverance. Given the migration has already occurred on sisyphos, there's no need to worry about fixing it. After this is applied to persa and mainnet the storage will be at the same version and correct, so they will be consistent.

### DIFF
--- a/state-chain/runtime/src/migrations/migrate_apicalls_to_store_signer.rs
+++ b/state-chain/runtime/src/migrations/migrate_apicalls_to_store_signer.rs
@@ -11,7 +11,8 @@ use cf_chains::{
 	},
 };
 use dot::{api::PolkadotApi, PolkadotExtrinsicBuilder};
-use frame_support::traits::OnRuntimeUpgrade;
+use frame_support::{traits::OnRuntimeUpgrade, Twox64Concat};
+use pallet_cf_broadcast::ThresholdSignatureFor;
 use sp_std::marker::PhantomData;
 
 #[cfg(feature = "try-runtime")]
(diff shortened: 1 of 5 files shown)


dandanlen (approved):

state-chain/runtime/src/migrations/migrate_apicalls_to_store_signer.rs
```diff
@@ -499,37 +603,36 @@ macro_rules! eth_map_over_api_variants_old {
...
-				)
-				.unwrap();
+		old_threshold_sig_datas.into_iter().for_each(|(broadcast_id, old_apicall)| {
+			let (new_apicall, _sig) =
+				next::ThresholdSignatureData::<Runtime, $chain_pallet_instance>::get(broadcast_id)
+					.expect("next threshold sig data");
 
-			assert!(new_apicall.signer().unwrap() == current_key);
+			assert!(new_apicall.signer().expect("signer new apicall") == current_key);
 			assert!($assertion(old_apicall, new_apicall));
```
ramizhasan111, line 636: would be nice to also assert that ```sig``` is the same before and after migration.

ramizhasan111: I see that we remove the sig in the next migration so not useful to test sig.

PR #5136, opened 2024-08-09
feat: bouncer can observe past events

j4m1ef0rd:
# Pull Request

Closes: PRO-1550

## Checklist

## Summary

@acdibble or someone from product, can you please check `substrate.ts`.
@albert-llimos can you please check I didn't screw up your GasLimitCCM test.

- Added a utility that gets the past events from x blocks ago. 
- A class that caches events so its fast for later historical lookups.
	- cache can handle reorgs 
	- cache tracks the latest finalised block number and tracks pending requests so it will do the minimum rpc calls as possible (ie. No duplicate lookups).
- New `observeEvents` function that is able to find matches from past events and return a list.
	- if no event it found it will continue to look at new events as usual.
- Changed `observeEvent`  to use the new `observeEvents` function and just take the first match.
	- Also extended `observeEvents` to be able to look at past events. If no `historicCheckBlocks` is provided, then the behaviour is the same as before.
- Refactored the `FoK` test and the `GasLimitCCM` test to use the new functions.

Out of scope:
~~Caching the historic events. Each `observeEvents` call will retrieve the events for each block from the api each time. Could be a problem if running the test on a real network or if we start to use this a lot in tests.~~
~~Supporting reorgs on the event cache. Seems we are only using it on the StateChain for now.~~
who cares about scopes anyways.


martin-chainflip (approved):

martin-chainflip:
cool feature!

bouncer/shared/gaslimit_ccm.ts
```diff
@@ -119,130 +122,59 @@ async function trackGasLimitSwap(
...
+  ).data.transactionPayload;
 
-  if (chainGasAsset(destChain) === sourceAsset) {
-    egressBudgetAmount = messageMetadata.gasBudget;
-  } else {
-    const gasSwapId = reqIdToGasSwapId[swapRequestId];
-    egressBudgetAmount = Number(swapIdToEgressAmount[gasSwapId].replace(/,/g, ''));
-  }
+  const egressBudgetAmount =
+    chainGasAsset(destChain as Chain) === sourceAsset ? messageMetadata.gasBudget : outputAmount;
```
albert-llimos, line 176: Maybe I'm mistaken here, since swap events have recently changed, but I think this `outputAmount` is not correct. The test before the changes uses `swapIdToEgressAmount[gasSwapId]` to get the gas output amount. However, here the `outputAmount` comes from the SwapExecuted with swapRequestId that is the principal swap, isn't it? We should be getting that from the SwapExecuted that is the gas swap instead.

albert-llimos:
Very happy to get rid of the gas limit test's spaghetti code to handle that edge case, good stuff!

acdibble:
We could probably set up an indexer docker container and then you could run SQL queries against the blocks instead

msgmaxim:
> We could probably set up an indexer docker container and then you could run SQL queries against the blocks instead

I like the idea, but what will happen when the protocol makes changes to the evens (or adds new ones?). Presumably we won't be able to run bouncer tests until the indexer is updated too? I'm not 100% sure how this works, but I think currently it is maintained by the product team? Of course one solution would be to make it the protocol's team responsibility to update it 🤷. Not sure, but maybe something to consider.

bouncer/shared/fill_or_kill.ts
```diff
@@ -64,13 +64,14 @@ async function testMinPriceRefund(inputAsset: Asset, amount: number) {
   const swapRequestedEvent = await swapRequestedHandle;
   console.log(`Swap requested: ${JSON.stringify(swapRequestedEvent)}`);
   const swapRequestId = Number(swapRequestedEvent.data.swapRequestId.replaceAll(',', ''));
-  console.log(`${inputAsset} swap requested, swapId: ${swapRequestId}`);
+  console.log(`${inputAsset} swap requested, swapRequestId: ${swapRequestId}`);
 
-  // TODO: Observing after the SwapScheduled event means its possible to miss the events, but we need to the swap id.
   const observeSwapExecuted = observeEvent(`swapping:SwapExecuted`, {
     test: (event) => Number(event.data.swapRequestId.replaceAll(',', '')) === swapRequestId,
+    historicCheckBlocks: 10,
```
msgmaxim, line 71:
```suggestion
    historicalCheckBlocks: 10,
```

bouncer/shared/gaslimit_ccm.ts
```diff
@@ -119,130 +122,59 @@ async function trackGasLimitSwap(
...
-    `${destChain.toString().toLowerCase()}Broadcaster:TransactionBroadcastRequest`,
-    {
+  await send(sourceAsset, depositAddress);
+  const swapRequestId = Number((await swapRequestedHandle).data.swapRequestId.replaceAll(',', ''));
+
+  // Wait for the swap to complete
+  const outputAmount = (
+    await observeEvent('swapping:SwapExecuted', {
+      test: (event) => Number(event.data.swapRequestId.replaceAll(',', '')) === swapRequestId,
+      historicCheckBlocks: CHECK_PAST_BLOCKS_FOR_EVENTS,
```
msgmaxim, line 146: We could probably add a wrapper function that uses `CHECK_PAST_BLOCKS_FOR_EVENTS` (or whatever) by default, so it doesn't have to be repeated each time.

bouncer/shared/fill_or_kill.ts
msgmaxim: (same in other places)

bouncer/shared/utils/substrate.ts
```diff
@@ -217,12 +216,35 @@ const subscribeHeads = getCachedDisposable(
...
 );
 
+async function getPastEvents(chain: SubstrateChain, historicCheckBlocks: number): Promise<Event[]> {
+  const api = await apiMap[chain]();
+  const historicEvents: Event[] = [];
+  if (historicCheckBlocks > 0) {
+    const latestHeader = await api.rpc.chain.getHeader();
+    const latestBlockNumber = latestHeader.number.toNumber();
+    const startAtBlock =
+      latestBlockNumber - historicCheckBlocks > 0 ? latestBlockNumber - historicCheckBlocks : 0;
```
msgmaxim, line 226: Use can use `Math.max()` for this

bouncer/shared/gaslimit_ccm.ts
j4m1ef0rd: doesn't seem worth the extra code.

bouncer/shared/utils/substrate.ts
```diff
@@ -359,7 +361,7 @@ export function observeEvent<T = any>(
     chain = 'chainflip',
     test = () => true,
     finalized = false,
-    historicCheckBlocks = 0,
+    historicalCheckBlocks: historicCheckBlocks = 0,
```
msgmaxim, line 415: Nit: if `historicCheckBlocks` is renamed too, then the two should collapse.

bouncer/shared/utils/substrate.ts
```diff
@@ -200,7 +255,10 @@ const subscribeHeads = getCachedDisposable(
       const historicApi = await api.at(header.hash);
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const rawEvents = (await historicApi.query.system.events()) as unknown as any[];
-      subject.next(mapEvents(rawEvents, header.number.toNumber()));
+      const mappedEvents = mapEvents(rawEvents, header.number.toNumber());
+      const cache = eventCacheMap[chain];
+      cache.addEvents(header.number.toNumber(), mappedEvents);
```
msgmaxim, line 260: Might be more correct to cache by block hash rather than number? Shouldn't matter for localnet, but in case this is used on a real network where reorgs are possible, this might be problematic.

j4m1ef0rd: This was my original plan but it turned out to be harder and slower. I would have to fetch each block hash every time we do a block number lookup to check hashes match. We would also have to maintain an extra list that maps block numbers to hash's so we can delete the oldest blocks still. So i thought it would be fine to just do block numbers for version 1 and avoid the complexity and overhead. Happy to upgrade it later.

msgmaxim (approved):

PR #5135, opened 2024-08-08
fix: use encoded address for refund_address in events

dandanlen:
# Pull Request

Closes: PRO-1553

## Checklist

Please conduct a thorough self-review before opening the PR.

- [ ] I am confident that the code works. (Waiting for tests to pass)
- [x] I have written sufficient tests. (updated existing tests)
- [x] I have written and tested required migrations. (none require)
- [x] I have updated documentation where appropriate. (nothing required afaict)

## Summary

*Please include a succinct description of the purpose and content of the PR. What problem does it solve, and how? Link issues, discussions, other PRs, and anything else that will help the reviewer.*

See Linear description.

Changes are potentially engine-breaking (event update) which is why it would be good to include this in 1.5.1

PR #5138, opened 2024-08-09
fix: use encoded addresses in refund params for events

dandanlen:
Cherry pick of changes from #5135


kylezs (approved):

dandanlen:
Fixed compiler issues / clippy.

PR #5140, opened 2024-08-12
Feat: Solana gas tweaks

albert-llimos:
# Pull Request

Closes: PRO-1501

## Checklist

## Summary

- Set maximum CU per CCM transfer to 600k instead of 1.4M to avoid huge transactions which are less likely to be included by the leaders.
- Set a minimum of 1 priority fee (1 microlamport / CU), even if the median reported fee is zero. The cost of that priority fee will be negligible while it increases significantly the chances of being included.

### DIFF
--- a/state-chain/chains/src/sol.rs
+++ b/state-chain/chains/src/sol.rs
@@ -148,7 +148,7 @@ pub mod compute_units_costs {
 	pub const COMPUTE_UNITS_PER_ROTATION: SolComputeLimit = 8_000u32;
 	pub const COMPUTE_UNITS_PER_SET_GOV_KEY: SolComputeLimit = 15_000u32;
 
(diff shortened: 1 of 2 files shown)


syan095:
Sorry for hijacking the PR. I've renamed `calculate_ccm_gas_limit` function so it's clearer that it's only used for CCM functions.

syan095 (approved):
LGMT!

syan095:
1 question - we set the minimum "priority" fee to 1 - should we set this a bit higher?

albert-llimos:
> 1 question - we set the minimum "priority" fee to 1 - should we set this a bit higher?

While setting it to one (microlamports/CU) looks and is very low, in normal conditions (not high congestion) of the network a priority fee of 1 should be enough. However, since we're really interested in inclusion and that amount is still very small it's reasonable to raise it to 10 for now. I'll do that.

PR #5143, opened 2024-08-13
chore: benchmarking for Solana durable nonce recoveries

syan095:
# Pull Request

Closes: PRO-1556

## Checklist

## Summary

Added Benchmarking for the Solana Durable nonce recovery extrinsics


state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -483,7 +483,9 @@ pub mod pallet {
 		/// - [NonceAccountNotBeingUsed](Error::NonceAccountNotBeingUsedOrDoesNotExist)
 		/// - [NonceHashNotSuppliedWhereRequired](Error::NonceAccountNotBeingUsedOrDoesNotExist)
 		#[pallet::call_index(8)]
-		#[pallet::weight(Weight::zero())]
+		// This weight is not strictly correct but since it's a governance call, weight is
+		// irrelevant.
+		#[pallet::weight(T::WeightInfo::force_recover_sol_nonce())]
```
dandanlen, line 488:
```suggestion
		#[pallet::weight(T::WeightInfo::force_recover_sol_nonce())]
```

nit: I don't think this comment is necessary / applicable here. Whether the benchmark is correct or not depends on the implementation. If we are making any tradeoffs, they should be noted there.

dandanlen (approved):

PR #5144, opened 2024-08-13
feat: DCA parameter validation

j4m1ef0rd:
# Pull Request

Note: This PR is targeting the core DCA PR. Will rebase once its merged.

Closes: PRO-1538

## Checklist

## Summary

- Added a new setting `MaxSwapRequestDurationBlocks` (for DCA) to the swapping pallet.
- Moved `MaxSwapRetryDurationBlocks` (for FoK) to the swapping pallet.
	- Both can be set via the `update_pallet_config` rpc.
		- 1 new event `MaxSwapRequestDurationSet`
		- `MaxSwapRetryDurationSet` event is now emitted from the swapping pallet.
- Added both settings to the `SwappingEnvironment`
	- Updated the `SwappingEnvironment` snapshot
- New trait to send the swap request validation settings to the ingress egress pallet.
- Used the new `MaxSwapRequestDurationBlocks` to validate the swap request DCA params
- Moved and updated/Added tests
- Added a migration to remove the old `MaxSwapRetryDurationBlocks` storage item


state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -479,7 +480,48 @@ fn test_refund_parameter_validation() {
...
+		assert_ok!(request_dca_swap(max_chunks, max_chunk_interval));
+		assert_err!(
+			request_dca_swap(0, max_chunk_interval),
+			DispatchError::from(crate::Error::<Test, _>::InvalidNumberOfDcaChunks)
+		);
+		assert_err!(
+			request_dca_swap(max_chunks + 1, max_chunk_interval),
+			DispatchError::from(crate::Error::<Test, _>::InvalidNumberOfDcaChunks)
+		);
+		assert_err!(
```
msgmaxim, line 518: should this include swap interval = 1 (which is also invalid)?

state-chain/node/src/chain_spec.rs
```diff
@@ -787,6 +787,8 @@ fn testnet_genesis(
 			witness_safety_margin: Some(bitcoin_safety_margin),
 			dust_limits: vec![(btc::Asset::Btc, BITCOIN_DUST_LIMIT)],
 			max_swap_retry_duration_blocks: 600,
+			max_dca_chunks: 50,
```
msgmaxim, line 790: As discussed, doesn't the deafult value for the pallets config apply here (which sets these to reasonable values)? If so, we can just use `Default::default()` as below.

j4m1ef0rd:
As discussed, I will be refactoring this to move the params (including the Fok param) to the environment pallet so that we do not have separate limits for each chain that need to be updated separately.
I will also be changing the 2 DCA parmas to be 1 param: `maximum total request duration blocks`

dandanlen:
Some weirdness with the dca merge. Will rebase and fix.

dandanlen:
PR summary is no longer valid btw.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1003,6 +1039,19 @@ pub mod pallet {
...
+					PalletConfigUpdate::SetMaxSwapRetryDuration { blocks } => {
+						MaxSwapRetryDurationBlocks::<T>::set(blocks);
+						Self::deposit_event(Event::<T>::MaxSwapRetryDurationSet { blocks });
+					},
+					PalletConfigUpdate::SetMaxSwapRequestDuration { blocks } => {
+						// TODO JAMIE: check that this is correct.
+						ensure!(
+							blocks >= SWAP_DELAY_BLOCKS,
+							Error::<T>::MaxSwapRequestDurationTooShort
+						);
```
dandanlen, line 1051: I think we might also have to ensure that the max duration is greater than the max retry duration (?)

(Might depend on the answers to my other questions)

state-chain/primitives/src/lib.rs
```diff
@@ -151,6 +151,9 @@ pub struct TxId {
 /// The very first epoch number
 pub const GENESIS_EPOCH: u32 = 1;
 
+/// Number of blocks in the future a swap is scheduled for.
+pub const SWAP_DELAY_BLOCKS: u32 = 2;
```
dandanlen, line 155: I'm not a fan of putting configuration constants in the primitives...

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2092,10 +2072,28 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
...
+		}
+
+		if let Some(params) = &dca_params {
+			ensure!(
+				params.number_of_chunks > 0 && params.chunk_interval >= SWAP_DELAY_BLOCKS,
+				DispatchError::from(Error::<T, I>::InvalidDcaParameters)
+			);
+			let total_swap_request_duration = params
+				.number_of_chunks
+				.saturating_sub(1)
```
dandanlen, line 2090: What's the significance of this saturating_sub(1)?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2092,10 +2072,28 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
...
-				params.retry_duration <= MaxSwapRetryDurationBlocks::<T, I>::get(),
-				DispatchError::Other("Retry duration too long")
+				params.retry_duration <= swap_limits.max_swap_retry_duration_blocks,
+				DispatchError::from(Error::<T, I>::SwapRetryDurationTooLong)
+			);
+		}
+
+		if let Some(params) = &dca_params {
+			ensure!(
+				params.number_of_chunks > 0 && params.chunk_interval >= SWAP_DELAY_BLOCKS,
```
dandanlen, line 2085: To avoid checking SWAP_DELAY_BLOCKS here (and to keep this swapping-specific configuration isolated to the swapping pallet), we could just add the SWAP_DELAY_BLOCK to the chunk interval when we schedule the next chunk? Arguably this is more consistent with how we execute the first block: when we submit an initial swap, it is always scheduled for execution after three blocks. 

This also avoid having to communicate to users that there's a minimum chunk interval (ie. 0 just means reschedule the next chunk as soon as the first on is complete). Thoughts? (cc @msgmaxim )

state-chain/primitives/src/lib.rs
msgmaxim: We don't really have a place for sharing constants between crates/pallets. Is it worth creating a separate crate for just constants?

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: total_swap_request_duration = (number_of_chunks - 1) * interval + swap_delay. But i'm ignoring the initial swap delay for simplicity. So a request for 2 chunks with an interval of 5 blocks = (2-1) * 5 = 5 blocks, the gap between the first chunk and the second. I think that is correct.

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: `max retry duration`  is part of FoK which is independent from DCA, so i don't understand why we would have that restriction?

state-chain/primitives/src/lib.rs
msgmaxim: Alternatively, if the issue is just about the name, maybe we could rename "primitives" to "common"?

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: This is more of a product question, but personally I prefer keeping it the way it is. I think saying that there is a mimimum delay of 2 blocks is at least as (if not more) intuitive than having a hidden 2 block delay. 

> Arguably this is more consistent with how we execute the first block

I think we there are reasonable arbuments both ways. I think what's important is that we consistently notify LPs of a swap at least 2 blocks in advance. From the code's perspective I don't think we are too inconsistent: `schedule_swap` takes a block delay in all cases. It is just that for the first chunk there is a different delay since "interval" doesn't apply, and I think this is justified as the first chunk is literarly an edge case. Not sure if subtracting 2 from `chunk_interval` just so it can be added back at a later point makes things better. 

> to keep this swapping-specific configuration isolated to the swapping pallet

This is a good point, but I think the problem lies somewhere deeper. The two pallets are already tightly coupled: ingres-gress knows about channels which know about swaps. We should either fix this (not even sure if it is possible/desirable), or accept that there is strong coupling and not worry about workarounds to hide the symptoms.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -470,6 +457,8 @@ pub mod pallet {
 
 		/// Safe Mode access.
 		type SafeMode: Get<PalletSafeMode<I>>;
+
+		type SwapLimitsProvider: SwapRequestValidationProvider;
```
msgmaxim, line 461: As dicussed, we should keep the names consistent. Also, `SwapRequestValidationProvider` isn't very clear, I think I would prefer `SwapLimitsProvider` even if, as you say, it doesn't provide *all* the limits.

state-chain/primitives/src/lib.rs
dandanlen: Let's not rename the whole module just yet 😅

dandanlen (approved):

PR #5153, opened 2024-08-15
Chore: Asset-balances improvement

syan095:
## Checklist

## Summary
Minor improvement to the Asset-balances pallet
Added some new unit test cases.


state-chain/pallets/cf-asset-balances/src/tests.rs
```diff
@@ -180,16 +192,23 @@ pub fn refund_validators_polkadot() {
...
 			payed_gas(
 				ForeignChain::Ethereum,
 				100,
 				ForeignChainAddress::Eth(sp_core::H160([i as u8; 20])),
 			);
 		}
+		for i in 254u8..=255u8 {
+			payed_gas(
+				ForeignChain::Ethereum,
+				500,
```
dandanlen, line 205: I'm not sure what the benefit of this is? We don't test anything related to the amount.

state-chain/pallets/cf-asset-balances/src/tests.rs
```diff
@@ -231,3 +250,24 @@ pub fn do_not_refund_if_amount_is_too_low() {
 		assert_egress(0, None);
 	});
 }
+
+#[test]
+pub fn refund_validators_solana() {
+	new_test_ext().execute_with(|| {
+		const REFUND_AMOUNT: u128 = 1_000u128;
```
dandanlen, line 257: This test is testing the same case as refund_validators_btc - maybe we can write a generic test that just takes a `ForeignChain` as the argument?

state-chain/pallets/cf-asset-balances/src/tests.rs
syan095: Oh this is to test that the payouts are settled in reverse order (smallest to largest) so the 2 largest fees are left when we run out of available fund. I'll add a comment for this.

dandanlen (approved):
Thanks Roy

PR #5154, opened 2024-08-15
test: unsafe median electoral system

kylezs:
# Pull Request

## Checklist

## Summary

- Introduces mocks for the accessors and tests for unsafe median. These mocks are not quite complete, but for most tests they will be sufficient. The rest can be fleshed out as needed, the hard part is done.
- Tests the unsafe median electoral system
- Splits the unsafe median and monotonic median into their own files

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_system.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_system.rs
@@ -14,7 +14,6 @@ use crate::{
 ///
 /// Implementations of this trait should *NEVER* directly access the storage of the elections
 /// pallet, and only access it through the passed-in accessors.
-
 pub trait ElectoralSystem: 'static {
 	/// This is intended for storing any internal state of the ElectoralSystem. It is not
 	/// synchronised and therefore should only be used by the ElectoralSystem, and not be consumed
--- a/state-chain/pallets/cf-elections/src/electoral_system.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_system.rs
@@ -3,7 +3,6 @@ use frame_support::{
 	pallet_prelude::{MaybeSerializeDeserialize, Member},
 	Parameter,
 };
-use sp_std::vec::Vec;
 
 use crate::{
 	vote_storage::{AuthorityVote, VoteStorage},
@@ -15,6 +14,7 @@ use crate::{
 ///
 /// Implementations of this trait should *NEVER* directly access the storage of the elections
 /// pallet, and only access it through the passed-in accessors.
+
 pub trait ElectoralSystem: 'static {
 	/// This is intended for storing any internal state of the ElectoralSystem. It is not
 	/// synchronised and therefore should only be used by the ElectoralSystem, and not be consumed
@@ -214,6 +214,7 @@ mod access {
 
 	/// Represents the current consensus, and how it has changed since it was last checked (i.e.
 	/// 'check_consensus' was called).
+	#[cfg_attr(test, derive(Clone))]
 	pub enum ConsensusStatus<Consensus> {
 		/// You did not have consensus when previously checked, but now consensus has been gained.
 		Gained {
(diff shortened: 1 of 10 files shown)


state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -149,7 +151,18 @@ pub mod pallet {
 	#[derive(
 		PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug, Encode, Decode, TypeInfo, Default,
 	)]
-	struct UniqueMonotonicIdentifier(u64);
+	pub struct UniqueMonotonicIdentifier(u64);
+
+	#[cfg(test)]
+	impl UniqueMonotonicIdentifier {
+		pub(crate) fn new(value: u64) -> Self {
```
dandanlen, line 158: This combination pub and pub(crate) means anyone can access the type, but can't construct it. Is this intended? How do we expect the engine to construct a vote for an election?

state-chain/pallets/cf-elections/src/electoral_system.rs
```diff
@@ -397,3 +398,278 @@ pub use access::{
 	ConsensusStatus, ElectionReadAccess, ElectionWriteAccess, ElectoralReadAccess,
 	ElectoralWriteAccess,
 };
+
+#[cfg(test)]
+pub mod mocks {
```
dandanlen, line 403: Not sure I understand why we don't just the usual substrate pallet mock runtime?

kylezs: These mocks are used to test the electoral systems which do not require any runtime access.

state-chain/pallets/cf-elections/src/lib.rs
kylezs: The engine will get the identifier from the subscription that will be based on the `validator_election_data` function, it should never need to construct it

state-chain/pallets/cf-elections/src/electoral_system.rs
```diff
@@ -397,3 +397,278 @@ pub use access::{
...
+						.unwrap())
+				}
+			}
+		};
+	}
+
+	impl_election_read_access!(MockElectionReadAccess);
+	impl_election_read_access!(MockElectionWriteAccess);
+
+	// Do we even need this stuff?
```
AlastairHolmes, line 478: ?

state-chain/pallets/cf-elections/src/electoral_system.rs
```diff
@@ -397,3 +397,278 @@ pub use access::{
...
+				<Self::ElectoralSystem as ElectoralSystem>::ElectoralUnsynchronisedStateMapValue,
+			>,
+			CorruptStorageError,
+		> {
+			Ok(self
+				.unsynchronised_state_map
+				.iter()
+				.find(|(k, _)| k == key)
+				.map(|(_, v)| v.clone())
+				.expect("Key not found"))
```
AlastairHolmes, line 617: Should just return None

AlastairHolmes: Instead of expect

state-chain/pallets/cf-elections/src/electoral_system.rs
```diff
@@ -397,3 +397,278 @@ pub use access::{
...
+			impl<ES: ElectoralSystem> ElectionReadAccess for $struct_name<'_, ES> {
+				type ElectoralSystem = ES;
+
+				fn settings(
+					&self,
+				) -> Result<
+					<Self::ElectoralSystem as ElectoralSystem>::ElectoralSettings,
+					CorruptStorageError,
+				> {
+					Ok(self
```
AlastairHolmes, line 436: I see you are panicking, I'd suggest returning the CorruptStorage error instead in all thsese mocks as that is the behaviour of the pallet.

state-chain/pallets/cf-elections/src/electoral_system.rs
kylezs: stale

PR #5155, opened 2024-08-15
test: add Solana delta ingress witnessing test to bouncer

albert-llimos:
## Checklist

## Summary

Added test for Solana delta ingress witnessing. The test is failing at the moment, as expected:
https://github.com/chainflip-io/chainflip-backend/actions/runs/10401446608/job/28804557637
When this is fixed with the new pallet in PR #5060 we can add it as part of the bouncer tests.

### DIFF
--- a/bouncer/tests/delta_base_ingress.ts
+++ b/bouncer/tests/delta_base_ingress.ts
@@ -7,7 +7,7 @@ import { executeWithTimeout, observeFetch, sleep } from '../shared/utils';
 import { observeEvent } from '../shared/utils/substrate';
 import { killEngines, startEngines } from '../shared/upgrade_network';
 
-// To run:
+// To run locally:
 // ./tests/delta_base_ingress.ts prebuilt --bins ./../target/debug --localnet_init ./../localnet/init
 
 let swapsWitnessed: number = 0;
--- a/bouncer/tests/delta_base_ingress.ts
+++ b/bouncer/tests/delta_base_ingress.ts
@@ -9,6 +9,8 @@ import { killEngines, startEngines } from '../shared/upgrade_network';
 
 // To run locally:
 // ./tests/delta_base_ingress.ts prebuilt --bins ./../target/debug --localnet_init ./../localnet/init
+// To run in CI:
+// ./tests/delta_base_ingress.ts prebuilt --bins ./../ --localnet_init ./../localnet/init
 
 let swapsWitnessed: number = 0;
 
--- a/bouncer/tests/delta_base_ingress.ts
+++ b/bouncer/tests/delta_base_ingress.ts
@@ -1,16 +1,33 @@
 #!/usr/bin/env -S pnpm tsx
-import yargs from 'yargs';
-import { hideBin } from 'yargs/helpers';
-import { doPerformSwap } from '../shared/perform_swap';
-import { testSwap } from '../shared/swapping';
-import { executeWithTimeout, observeFetch, sleep } from '../shared/utils';
-import { observeEvent } from '../shared/utils/substrate';
-import { killEngines, startEngines } from '../shared/upgrade_network';
+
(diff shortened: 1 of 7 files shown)


bouncer/tests/delta_base_ingress.ts
```diff
@@ -0,0 +1,184 @@
...
+      }
+      return false;
+    },
+    abortable: true,
+  });
+
+  // Initiate swap from Solana
+  const swapParams = await testSwap(
+    sourceAsset,
+    'Eth',
```
kylezs, line 73: ArbEth? just so it's faster?

bouncer/tests/delta_base_ingress.ts
```diff
@@ -0,0 +1,184 @@
...
+  // Wait to ensure no new swap is being triggered after restart.
+  console.log('Waiting for 30 seconds to ensure no swap is being triggered after restart');
+  await sleep(30000);
+  swapScheduledHandle.stop();
+
+  if (swapsWitnessed !== 1) {
+    throw new Error('No swap was initiated. Swaps witnessed: ' + swapsWitnessed);
+  }
+
+  // Kill the engine
```
kylezs, line 95: think it's pretty clear what the function is doing 😆

bouncer/tests/delta_base_ingress.ts
```diff
@@ -0,0 +1,184 @@
...
+  });
+  await startEngines(localnetInitPath, binariesPath, numberOfNodes);
+
+  // Wait to ensure no additional new swap is being triggered after restart
+  // and check that swap completes.
+  await sleep(30000);
+  await swapHandle;
+  swapScheduledHandle.stop();
+
+  if (swapsWitnessed < 2) {
```
kylezs, line 137: Having this condition twice, here and inside SwapScheduled is redundant (and confusing imo).

bouncer/tests/delta_base_ingress.ts
```diff
@@ -0,0 +1,184 @@
...
+        inputAmount > Number(amountToFineAmount(amountSecondDeposit, assetDecimals(sourceAsset)))
+      ) {
+        throw new Error(
+          'Swap input amount is greater than the second deposit ' + inputAmount.toString(),
+        );
+      }
+      return false;
+    },
+    abortable: true,
+  });
```
kylezs, line 128: This is almost identical to the above call, we can make a local function to dedupe

bouncer/tests/delta_base_ingress.ts
```diff
@@ -0,0 +1,184 @@
...
+  amountToFineAmount,
+  assetDecimals,
+  executeWithTimeout,
+  observeFetch,
+  sleep,
+} from '../shared/utils';
+import { observeEvent } from '../shared/utils/substrate';
+import { killEngines, startEngines } from '../shared/upgrade_network';
+
+let swapsWitnessed: number = 0;
```
kylezs, line 33: move inside the function?

kylezs (commented):
Nice one 🙌  will be very handy soon, just some cleanliness comments

bouncer/tests/delta_base_ingress.ts
albert-llimos: It was to do it to two different destination chains but that will be tested in other tests so I'll do the change.

bouncer/tests/delta_base_ingress.ts
albert-llimos: it's not the same check. Inside the `SwapScheduled` we check that there are no more than two saps and here we check that there not less than two. We are making sure that one swap was seen (aka final number of swaps == 2)

bouncer/tests/delta_base_ingress.ts
albert-llimos: already done in the commit pushed almost at the same time  as the comment

kylezs (approved):

PR #5157, opened 2024-08-15
chore(ci): migrate fully to namespace runners 🏃‍♂️‍➡️

ahasna:


.github/actionlint.yaml
```diff
@@ -4,6 +4,7 @@ self-hosted-runner:
     [
       digitalocean,
       hetzner,
+      namespace-profile-default,
```
kylezs, line 7: the default is ubuntu 22?

kylezs (approved):

PR #5164, opened 2024-08-19
fix: allow unswept funds to be used in boost

kylezs:
# Pull Request

Closes: PRO-1547

## Checklist

- [x] I have written sufficient tests 

## Summary

Figured this was easy enough to get in before 1.6.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -39,7 +39,7 @@ use cf_traits::{
 	impl_pallet_safe_mode, AccountRoleRegistry, AdjustedFeeEstimationApi, AssetConverter,
 	AssetWithholding, BalanceApi, BoostApi, Broadcaster, Chainflip, DepositApi, EgressApi,
 	EpochInfo, FeePayment, FetchesTransfersLimitProvider, GetBlockHeight, IngressEgressFeeApi,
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1098,12 +1100,17 @@ pub mod pallet {
...
 			let booster_id = T::AccountRoleRegistry::ensure_liquidity_provider(origin)?;
+
 			ensure!(
 				T::SafeMode::get().add_boost_funds_enabled,
 				Error::<T, I>::AddBoostFundsDisabled
 			);
 			ensure!(amount > Zero::zero(), Error::<T, I>::AddBoostAmountMustBeNonZero);
 
+			// Free balances returns the unswept funds included, so if we want to be able to use
+			// those funds for boosting, we first need to sweep.
```
dandanlen, line 1111: nit: this comment is a little out of context - it's referring to the rpc? I think simply stating that try_debit *doesn't* implicitly sweep any open positions, and so we need to do it first, would be enough.

dandanlen (commented):
This will probably conflict with #5083 - would suggest adding it after that / branching off that PR.

msgmaxim (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: updated

dandanlen (approved):

dandanlen:
(I rebased against main)

PR #5166, opened 2024-08-19
feat: solana nonce tracking

kylezs:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

Implements Durable nonce witnessing for Solana.

### DIFF
--- a/engine/src/witness/sol.rs
+++ b/engine/src/witness/sol.rs
@@ -28,9 +28,11 @@ use pallet_cf_elections::{electoral_system::ElectoralSystem, vote_storage::VoteS
 use state_chain_runtime::{
 	chainflip::solana_elections::{
 		SolanaBlockHeightTracking, SolanaElectoralSystem, SolanaFeeTracking, SolanaIngressTracking,
+		SolanaNonceTracking,
 	},
 	SolanaInstance,
 };
+
 use std::sync::Arc;
 use utilities::{context, task_scope, task_scope::Scope};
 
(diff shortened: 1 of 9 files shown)


engine/src/witness/sol.rs
```diff
@@ -104,6 +106,36 @@ impl VoterApi<SolanaIngressTracking> for SolanaIngressTrackingVoter {
...
+impl VoterApi<SolanaNonceTracking> for SolanaNonceTrackingVoter {
+	async fn vote(
+		&self,
+		_settings: <SolanaNonceTracking as ElectoralSystem>::ElectoralSettings,
+		properties: <SolanaNonceTracking as ElectoralSystem>::ElectionProperties,
+	) -> Result<<<SolanaNonceTracking as ElectoralSystem>::Vote as VoteStorage>::Vote, anyhow::Error>
+	{
+		let (nonce_account, previous_nonce) = properties;
+		let (response_account, response_nonce) =
+			nonce_witnessing::get_durable_nonces(&self.client, vec![nonce_account])
```
albert-llimos, line 124: I'd really prefer to pass all nonce accounts instead of one by one. This way we can take advantage of the `getMultipleAccounts` and use 1 RPC call to query the state of all 10 nonces instead of 10 separate RPC calls.

engine/src/witness/sol.rs
```diff
@@ -104,6 +106,32 @@ impl VoterApi<SolanaIngressTracking> for SolanaIngressTrackingVoter {
...
+				.await?
+				.pop()
+				.expect("If the query succeeds, we expect a nonce for the account we queried for");
+
+		assert_eq!(response_account, nonce_account);
+
+		Ok(response_nonce)
```
ramizhasan111, line 131: In previous nonce witnessing functionality we only submitted the nonce witnesses to the SC if the witnessed nonce was different than the nonce on the SC. It looks like this change does not have that. I am not sure how the new witnessing works yet but it looks like it will always submit the nonce witness even if the existing nonce on the SC is the same as witnessed nonce.

kylezs: Yeah that's intentional, the SC will now handle this case

engine/src/witness/sol.rs
kylezs: https://linear.app/chainflip/issue/PRO-1588/witness-all-nonces-at-once

PR #5168, opened 2024-08-20
chore: benchmark `provide_shared_data` for the election pallet

syan095:
# Pull Request

Closes: PRO-1579

## Checklist

## Summary

Added benchmarking code for `provide_shared_data` for the election pallet.


state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -60,6 +57,30 @@ mod benchmarks {
...
+					.collect::<BTreeMap<_, _>>(),
+			)
+			.unwrap(),
+		));
+
+		#[extrinsic_call]
+		provide_shared_data(
+			RawOrigin::Signed(validator_id.into()),
+			T::ElectoralSystem::benchmark_shared_data(),
+		);
```
kylezs, line 81: Something like this to see that we actually inserted the shared data, which is the most expensive path.
```
		assert_eq!(
			SharedData::<T, I>::get(&SharedDataHash::of(&T::ElectoralSystem::benchmark_shared_data())),
			Some(T::ElectoralSystem::benchmark_shared_data())
		);
```

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -1298,7 +1298,7 @@ pub mod pallet {
 		}
 
 		#[pallet::call_index(1)]
-		#[pallet::weight(Weight::zero())] // TODO: Benchmarks
+		#[pallet::weight(T::WeightInfo::provide_shared_data())]
```
kylezs, line 1301: could we also add the weight to the vote extrinsic while we're here too

state-chain/pallets/cf-elections/src/benchmarking.rs
kylezs: I just pushed it since I tried it - and made the diagram text which I think was removed when the base branch was rebased

kylezs (approved):

PR #5178, opened 2024-08-21
feat: solana egress witnessing

kylezs:
# Pull Request

## Checklist

- [ ] I have written sufficient tests. - ticket created

## Summary

Implements Solana egress witnessing with election pallet

### DIFF
--- a/engine/src/witness/sol.rs
+++ b/engine/src/witness/sol.rs
@@ -149,7 +149,7 @@ impl VoterApi<SolanaEgressWitnessing> for SolanaEgressWitnessingVoter {
 			.next()
 			.ok_or(anyhow!("Success querying for {signature} but no items"))?;
 		assert_eq!(sig, signature, "signature we requested should be the same as in the response");
-		Ok(TransactionSuccessDetails { tx_fee })
+		Ok(TransactionSuccessDetails { tx_fee, signer: Default::default() })
 	}
 }
 
--- a/engine/src/witness/sol.rs
+++ b/engine/src/witness/sol.rs
@@ -14,21 +14,16 @@ use crate::{
 		chain_api::ChainApi, electoral_api::ElectoralApi,
 		extrinsic_api::signed::SignedExtrinsicApi, storage_api::StorageApi,
 	},
-	witness::{
-		common::{chain_source::Header, epoch_source::Vault},
-		sol::sol_deposits::get_channel_ingress_amounts,
-	},
+	witness::sol::sol_deposits::get_channel_ingress_amounts,
 };
-use anyhow::Result;
-use cf_chains::sol::{SolHash, SolSignature, LAMPORTS_PER_SIGNATURE};
-use cf_primitives::EpochIndex;
+use anyhow::{anyhow, Result};
+use cf_chains::sol::{SolSignature, LAMPORTS_PER_SIGNATURE};
 use futures::FutureExt;
-use futures_core::Future;
 use pallet_cf_elections::{electoral_system::ElectoralSystem, vote_storage::VoteStorage};
 use state_chain_runtime::{
 	chainflip::solana_elections::{
-		SolanaBlockHeightTracking, SolanaElectoralSystem, SolanaFeeTracking, SolanaIngressTracking,
-		SolanaNonceTracking,
+		SolanaBlockHeightTracking, SolanaEgressWitnessing, SolanaElectoralSystem,
+		SolanaFeeTracking, SolanaIngressTracking, SolanaNonceTracking, TransactionSuccessDetails,
 	},
 	SolanaInstance,
 };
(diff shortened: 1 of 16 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -575,3 +576,4 @@ generate_electoral_system_tuple_impls!(tuple_1_impls: ((A, A0)));
 generate_electoral_system_tuple_impls!(tuple_2_impls: ((A, A0), (B, B0)));
 generate_electoral_system_tuple_impls!(tuple_3_impls: ((A, A0), (B, B0), (C, C0)));
 generate_electoral_system_tuple_impls!(tuple_4_impls: ((A, A0), (B, B0), (C, C0), (D, D0)));
+generate_electoral_system_tuple_impls!(tuple_5_impls: ((A, A0), (B, B0), (C, C0), (D, D0), (EE, E0)));
```
kylezs, line 579: `EE` because `E` conflicts with the E generic as in Error... I might make them all double letters for consistency.... but not super important

dandanlen: Or since they aren't intended to be used, we could make that clear by prefixing them all? `_A, _B, _C` etc ?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -191,6 +192,10 @@ pub mod pallet {
 			AttemptCount = AttemptCount,
 		>;
 
+		type ElectionEgressWitnesser: WatchForEgressSuccess<
```
dandanlen, line 195: nit: might as well use the same names? ie. name the trait `ElectionEgressWitnesser`.

state-chain/pallets/cf-elections/src/electoral_systems/egress_success.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// This electoral system detects if something occurred or not. Voters simply vote if something
+/// happened, and if they haven't seen it happen, they don't vote.
+pub struct EgressSuccess<Identifier, Value, Settings, Hook> {
+	_phantom: core::marker::PhantomData<(Identifier, Value, Settings, Hook)>,
+}
+
+pub trait OnEgressSuccess<Identifier, Value> {
+	fn on_egress_success(id: Identifier, value: Value);
+}
```
dandanlen, line 25: This is basically identical to the `OnChangeHook` from the `change` elecoral system. 

Do we need another trait?

state-chain/pallets/cf-elections/src/electoral_systems/egress_success.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl<
+		Identifier: Member + Parameter + Ord,
+		Value: Member + Parameter + Eq,
+		Settings: Member + Parameter + MaybeSerializeDeserialize + Eq,
+		Hook: OnEgressSuccess<Identifier, Value> + 'static,
+	> EgressSuccess<Identifier, Value, Settings, Hook>
+{
+	pub fn watch_for_egress<ElectoralAccess: ElectoralWriteAccess<ElectoralSystem = Self>>(
```
dandanlen, line 34: This is also very similar to `watch_for_change`.

In general a lot of these impls are very similar...

state-chain/pallets/cf-elections/src/electoral_systems/egress_success.rs
```diff
@@ -0,0 +1,115 @@
...
+	fn check_consensus<ElectionAccess: ElectionReadAccess<ElectoralSystem = Self>>(
+		_election_identifier: ElectionIdentifierOf<Self>,
+		_election_access: &ElectionAccess,
+		_previous_consensus: Option<&Self::Consensus>,
+		votes: Vec<(VotePropertiesOf<Self>, <Self::Vote as VoteStorage>::Vote)>,
+		authorities: AuthorityCount,
+	) -> Result<Option<Self::Consensus>, CorruptStorageError> {
+		let votes_count = votes.len();
+		Ok(
+			if votes_count != 0 &&
```
dandanlen, line 106: What's the significance of this check? Just to avoid the call to `success_threshold_from_share_count`?

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -80,8 +124,14 @@ impl OnChangeHook<SolAddress, SolHash> for SolanaNonceTrackingHook {
 
 pub struct SolanaElectionHooks;
 
-impl Hooks<SolanaBlockHeightTracking, SolanaFeeTracking, SolanaIngressTracking, SolanaNonceTracking>
-	for SolanaElectionHooks
+impl
+	Hooks<
```
dandanlen, line 128: Not right now, but I think we should rename this trait to ElectionHooks or something... to distinguish it from the pallet hooks.

engine/src/witness/sol.rs
```diff
@@ -132,6 +127,32 @@ impl VoterApi<SolanaNonceTracking> for SolanaNonceTrackingVoter {
...
+impl VoterApi<SolanaEgressWitnessing> for SolanaEgressWitnessingVoter {
+	async fn vote(
+		&self,
+		_settings: <SolanaEgressWitnessing as ElectoralSystem>::ElectoralSettings,
+		signature: <SolanaEgressWitnessing as ElectoralSystem>::ElectionProperties,
+	) -> Result<
+		<<SolanaEgressWitnessing as ElectoralSystem>::Vote as VoteStorage>::Vote,
+		anyhow::Error,
+	> {
+		tracing::info!("Requesting success witnesses for {:?}", signature);
```
dandanlen, line 145: Does this need to be `info`? Maybe debug is enough? (I have no idea how frequently this might be polled.

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
kylezs: Yeah that probably works better

dandanlen (approved):

state-chain/pallets/cf-elections/src/electoral_systems/egress_success.rs
kylezs: I think it just makes it clear that this isn't called on a change, but only once.

kylezs: This electoral system is much more general than just egresses... but I couldn't think of a good general name for it. It's effectively "WatchForParticularThing"

engine/src/witness/sol.rs
kylezs: I was thinking of removing it entirely - the RPC is logging as we go

state-chain/pallets/cf-elections/src/electoral_systems/egress_success.rs
kylezs: yeah just defensive

engine/src/witness/sol.rs
kylezs: removed it

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
kylezs: I will do this once in main just to avoid more rounds of chained rebasing

state-chain/pallets/cf-elections/src/electoral_systems/egress_success.rs
dandanlen: Yeah, not saying we need to change this now, just noting that there seems to be a lot of repetition / boilerplate.

PR #5179, opened 2024-08-21
Benchmarking for ignore_my_votes and stop_ignoring_my_votes

Janislav:
# Pull Request

Closes: PRO-1565

## Checklist

## Summary

Added benchmarks for `ignore_my_votes` and `stop_ignoring_my_votes` for the simple case. I do benchmark the `recheck_contributed_to_consensuses` function independently, which is used by the other extrinsic to avoid the complexity of setting up a state of the system where we need to have multiple elections at the same time. We can just reuse the weight and multiply it by the length of the `ElectionConsensusHistoryUpToDate` storage item. I think that should be good enough for now.

#### Non-Breaking changes


dandanlen:
Closing in favour of #5180

PR #5181, opened 2024-08-21
Chore/add sol signer

albert-llimos:
## Checklist

## Summary

Submitting transaction signer for Solana egress witnessing.


albert-llimos:
We've decided to assign the fees to the current State chain aggregate key. It might not be strictly correct for some transaction around the rotation but that's fine and it will still get assigned to our keys.

PR #5182, opened 2024-08-22
Feat: Add DCA details to scheduled swaps subscription

j4m1ef0rd:
# Pull Request

Closes: PRO-1584

## Checklist

## Summary

- Added some DCA and FoK details to the scheduled swaps subscription.
- Added a unit test to cover it.

TODO: Update docs

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -12,7 +12,7 @@ use cf_chains::{
 };
 use cf_primitives::{
 	Affiliates, Asset, AssetAmount, Beneficiaries, Beneficiary, BlockNumber, ChannelId,
-	DcaParameters, ForeignChain, Price, SwapId, SwapLeg, SwapRequestId, TransactionHash,
+	DcaParameters, ForeignChain, SwapId, SwapLeg, SwapRequestId, TransactionHash,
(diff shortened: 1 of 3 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -164,12 +164,16 @@ pub struct Swap {
...
+	pub swap_request_id: SwapRequestId,
 	pub base_asset: Asset,
 	pub quote_asset: Asset,
 	pub side: Side,
 	pub amount: AssetAmount,
 	pub source_asset: Option<Asset>,
 	pub source_amount: Option<AssetAmount>,
+	pub remaining_chunks: u32,
+	pub chunk_interval: u32,
+	pub min_price: Option<Price>,
```
dandanlen, line 176: I think the min price might need some extra thought.

Since we include Buy/Sell sides here, I'm wondering if `min_price` is always applicable. 

Ie. for a sell it's a min price (I am only willing to sell above X price).

But for buys it's normally inverted (I am only willing to buy *below* X price). 

So maybe it makes sense to be explicit and call it the min_sell_price? Or simply price_limit? 

Also consider that the min_price will be stated for the full swap (eg. BTC-ETH) but this is the swap *leg* info, ie. BTC-USDC. So the min_price is a little meaningless in this context...

dandanlen: As discussed, we'll remove this for now.

dandanlen (approved):

PR #5185, opened 2024-08-22
chore: bouncer add check for solana nonces

albert-llimos:
## Checklist

## Summary

Add check at the end of the test that all nonces are available. It will especially be useful when we improve the way we witness the nonces to make sure nothing is broken.


bouncer/tests/all_concurrent_tests.ts
```diff
@@ -51,6 +51,24 @@ async function runAllConcurrentTests() {
...
+    if (availableNonces.length === solanaNumberOfNonces) {
+      break;
+    } else if (attempt === maxRetries - 1) {
+      throw new Error(
+        `Unexpected number of available nonces: ${availableNonces.length}, expected ${solanaNumberOfNonces}`,
+      );
+    } else {
+      await sleep(6000);
+    }
+  }
```
kylezs, line 71: could we move this into a function, just to keep `runAllConcurrentTests` readable at a high level

kylezs (approved):

PR #5186, opened 2024-08-22
feat: elections pallet tests

dandanlen:
# Pull Request

This includes a mock for the ElectoralSystem trait, and some initial tests of the consensus transitions when votes are added / ignored / deleted.

It's still incomplete - the next step would be to implement a mock of the VoteStorage trait and then `generate_vote_properties`.

### DIFF
--- a/state-chain/pallets/cf-elections/src/tests.rs
+++ b/state-chain/pallets/cf-elections/src/tests.rs
@@ -1,91 +1,251 @@
 #![cfg(test)]
-
-use super::*;
-use sp_std::collections::btree_set::BTreeSet;
 use std::collections::BTreeMap;
 
-use cf_traits::EpochInfo;
+use crate::{mock::*, *};
+use cf_primitives::AuthorityCount;
+use electoral_system::{
+	AuthorityVoteOf, ConsensusStatus, ElectionReadAccess, ElectoralReadAccess, ElectoralWriteAccess,
+};
+use electoral_systems::mock::MockElectoralSystem;
+use frame_support::traits::OriginTrait;
+use vote_storage::AuthorityVote;
 
-use electoral_system::{AuthorityVoteOf, ElectoralReadAccess};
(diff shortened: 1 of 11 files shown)


kylezs (approved):

state-chain/pallets/cf-elections/src/tests.rs
```diff
@@ -1,91 +1,251 @@
...
+
+				assert_eq!(Status::<Test, Instance1>::get(), Some(ElectoralSystemStatus::Running));
+
+				Pallet::<Test, Instance1>::with_electoral_access(|electoral_access| {
+					electoral_access
+						.election(ElectionIdentifier::new(unique_monotonic_identifier, ()))
+				})
+				.expect("Expected an initial election.");
+
+				ctx.umis.push(unique_monotonic_identifier);
```
kylezs, line 58: In the pallet logic itself, we use ElectionProperties as the source of truth about a particular UMI's existence. See `ensure_election_exists` - we could use ElectionProperties here instead - or at least assert that the UMIs are the same in each - `ElectionState` too - or perhaps we call `ensure_election_exixsts` here to avoid getting too close to implementation details

dandanlen: Yes, I wanted to avoid calling into any storage items directly. For example if we rely on ElectionProperties to get the list of active elections, then `ensure_election_exists` will always trivially pass.

Also, this `umis` item is not intended as a list of *live* elections - it could also be used to eg. assert that an election has been deleted. It's really just a record of what elections have been created during a particular test run.

I'm not sure there's much value in adding an assert here that the election exists: we just created it, and `ensure_election_exists` (which is a private method) is implicitly called in many places already.

dandanlen: As discussed - I added an integrity test to check the storage invariant.

kylezs (approved):

PR #5190, opened 2024-08-22
Tests elections pallet

Janislav:
# Pull Request

Closes: PRO-1575, PRO-1577

## Checklist

## Summary

WIP unit-tests for the elections pallet.

#### Non-Breaking changes


syan095:
I want to add some condition checks to the "delete_vote" test. should be done soon

dandanlen (commented):
Thanks - same as with the other PR, i'll reconcile this with the changes in #5186 and then get everything merged together.

dandanlen:
I've replaced one of the tests and commented out the others (for now) because they need to be updated to use the mock electoral system instead of the median system.

state-chain/pallets/cf-elections/src/mock.rs
```diff
@@ -64,6 +64,20 @@ impl TestSetup {
...
+			.into_iter()
+			.take(self.num_contributing_authorities as usize)
+			.collect()
+	}
+
+	pub fn non_contributing_authorities(&self) -> Vec<u64> {
+		self.all_authorities()
+			.into_iter()
+			.skip(self.num_contributing_authorities as usize)
+			.collect()
```
kylezs, line 79: could combine them and `split`? - means they're not going to go out of sync

state-chain/pallets/cf-elections/src/tests.rs
```diff
@@ -1,35 +1,138 @@
...
-		validator_ids: &[u64],
-		call: impl Fn(&u64) -> Call<Test, I>,
-	) -> Self
-	where
-		Test: Config<I>;
 	fn submit_votes<I: 'static>(
 		self,
 		validator_ids: &[u64],
 		vote: AuthorityVoteOf<MockElectoralSystem>,
 		expected_error: Option<Error<Test, I>>,
```
kylezs, line 131: does feel a little strange to use Option here - why not expect Result<(), Error> .... we end up converting it to result anyway.

state-chain/test-utilities/src/rich_test_externalities.rs
```diff
@@ -261,11 +267,64 @@ where
 		)
 	}
 
+	/// Keeps executing pallet hooks until the given predicate returns true.
+	///
+	/// Preserves the context.
+	#[track_caller]
+	pub fn then_process_blocks_until<F: Fn(Ctx) -> bool>(mut self, predicate: F) -> Self {
```
kylezs, line 274: nit, predicates usually get `P` 
```suggestion
	pub fn then_process_blocks_until<P: Fn(Ctx) -> bool>(mut self, predicate: P) -> Self {
```

state-chain/pallets/cf-elections/src/tests.rs
```diff
@@ -1,35 +1,138 @@
...
+		.update_settings(&[BehaviourUpdate::DeleteOnFinalizeConsensus(true)])
+		.snapshot();
+
+	// Case 1: Provide Shared Data through provide_shared_data extrinsic:
+	let case_1 = TestRunner::from_snapshot(initial_test_state.clone())
+		.assert_calls_ok(&authorities[..1], |_| Call::<Test, Instance1>::provide_shared_data {
+			shared_data: (),
+		});
+
+	// Case 2: Provide Shared Data through vote extrinsic:
```
kylezs, line 61: 👍

kylezs (approved):

state-chain/pallets/cf-elections/src/mock.rs
dandanlen: The intention is that this is just the test setup and therefore is immutable - it doesn't return the *current* contributing/non-contributing authorities.

state-chain/pallets/cf-elections/src/tests.rs
dandanlen: Yeah I think I agree - the idea was that we usually don't expect an error so we could just use Default::default, but agree it's bit unexpected..

PR #5192, opened 2024-08-23
feat: use encoded address in ccm deposit metadata in events

j4m1ef0rd:
# Pull Request

Closes: PRO-1597

## Checklist

## Summary

Made `CcmDepositMetadata` generic so we can encode the address inside of it when outputting it in the swap requested event and the ccm failed event.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -1966,7 +1966,14 @@ pub mod pallet {
 							output_address: T::AddressConverter::to_encoded_address(
 								output_address.clone(),
 							),
-							ccm_deposit_metadata: ccm_deposit_metadata.clone(),
+							ccm_deposit_metadata: cf_chains::CcmDepositMetadataEncoded {
+								source_chain: ccm_deposit_metadata.source_chain,
+								source_address: ccm_deposit_metadata
+									.source_address
+									.clone()
+									.map(T::AddressConverter::to_encoded_address),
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1966,7 +1966,14 @@ pub mod pallet {
 							output_address: T::AddressConverter::to_encoded_address(
 								output_address.clone(),
 							),
-							ccm_deposit_metadata: ccm_deposit_metadata.clone(),
+							ccm_deposit_metadata: cf_chains::CcmDepositMetadataEncoded {
```
msgmaxim, line 1969: I think `CcmDepositMetadata` is also exposed via the CcmFailed event, so that one should be updated to use the encoded version.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -688,6 +688,14 @@ mod ccm {
 	#[track_caller]
 	fn init_ccm_swap_request(input_asset: Asset, output_asset: Asset, input_amount: AssetAmount) {
 		let ccm_deposit_metadata = generate_ccm_deposit();
+		let ccm_deposit_metadata_encoded = cf_chains::CcmDepositMetadataEncoded {
```
msgmaxim, line 691: We could have a method on `CcmDepositMetadata` to convert it into the encoded verson (taking AddressConverter as a generic parameter?). I think we already do this for some of the other structs.

dandanlen (approved):

PR #5201, opened 2024-08-26
tests/electoral-systems

Janislav:
# Pull Request

Closes: PRO-1582, PRO-1580

## Checklist

## Summary

Tests the specific assumptions around consensus and the finalization of an election for the Change and the MonotonicMedian electoral systems. 

#### Non-Breaking changes


state-chain/pallets/cf-elections/src/electoral_systems/change.rs
```diff
@@ -143,3 +143,70 @@ impl<
...
+	fn finalize_election() {
+		let mut electoral_system =
+			MockElectoralSystem::<Change<(), u64, (), MockHook>>::new((), (), ());
+		electoral_system
+			.new_election((), ((), 1), ())
+			.unwrap()
+			.set_consensus_status(ConsensusStatus::Changed { previous: 1, new: 2 });
+		electoral_system.finalize_elections(&()).unwrap();
+		// Hook should have been called
+		assert!(HOOK_HAS_BEEN_CALLED.with(|hook_called| hook_called.get()));
```
dandanlen, line 210: This test can be more thorough:

- Create election with initial value
- Set consensus to the same as the initial value
- on_finalize => hook should not be called
- Set consensus to a different value
- on_finalize => hook should be called, election deleted.

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_median.rs
```diff
@@ -114,3 +114,127 @@ impl<
...
+			(),
+		);
+		let mut election = electoral_system.new_election((), (), ()).unwrap();
+		election.set_consensus_status(ConsensusStatus::Changed {
+			previous: INIT_UNSYNCHRONISED_STATE,
+			new: NEXT_UNSYNCHRONISED_STATE,
+		});
+		electoral_system.finalize_elections(&()).unwrap();
+		// Hock has been called
+		assert!(HOOK_HAS_BEEN_CALLED.with(|hook_called| hook_called.get()));
```
dandanlen, line 236: Let's encapsulate the raw static variable accesses behind a method in the Mock definition.

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_median.rs
```diff
@@ -114,3 +114,127 @@ impl<
...
+		// Unsynchronised state has been updated
+		assert_eq!(electoral_system.unsynchronised_state().unwrap(), NEXT_UNSYNCHRONISED_STATE);
```
dandanlen, line 238: We should check it has the correct value? We could check more properties (i'll add another comment).

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_median.rs
```diff
@@ -114,3 +114,127 @@ impl<
...
+	const INIT_UNSYNCHRONISED_STATE: u64 = 22;
+
+	#[test]
+	fn check_consensus_correctly_calculates_median_when_all_authorities_vote() {
+		let mut electoral_system = MockElectoralSystem::<MonotonicMedian<u64, (), MockHook>>::new(
+			INIT_UNSYNCHRONISED_STATE,
+			(),
+			(),
+		);
+		let mut votes = vec![((), 1), ((), 5), ((), 3), ((), 2), ((), 8), ((), 6), ((), 9)];
```
dandanlen, line 146: We can make this easier to read - the order of the values doesn't need to be randomised (it gets shuffled anyway). Also we could be more deliberate with the values we choose (see my other comment).

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_median.rs
dandanlen, line 1: There are some key properties of the 'monotonic median' are not explicitly tested. For example:

- The unsynchronised state (ie. the value passed to the hook) can only *increase*.
- It should not be possible for a minority of nodes (failure_threshold_from_share_count(authority_count)) to influence the outcome of the median calculation. For example, if the 'real' value is `x`, then *even if `failure_threshold` authorities vote for some other value*, we should only be able to reach consensus on `x`. 

I think this is what is meant by this comment in the code:

```rust
// Calculating the median this way means atleast 2/3 of validators would be needed to
// increase the calculated median.
```

The comment is not super clear, and clearly you have written a test that comes to consensus on a *lower* value than the initial state (initial state `22` but consensus returns `Some(5)`), so we should make sure we understand what the assumptions are and that they do in fact hold.

dandanlen:
Same as the other PRs - I'll merge combine everything into one and merge.

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
```diff
@@ -0,0 +1,321 @@
...
+			$(
+				pub fn $check_name() -> Self {
+					Self::new(#[track_caller] |$arg_1, $arg_2| $check_body)
+				}
+			)+
+		}
+	};
+	(
+		$(
+			#[ extra_constraints: $( $t:ty : $tc:path ),+ ]#
```
kylezs, line 264: nit:
(and in the comment above too)
```suggestion
			#[extra_constraints: $( $t:ty : $tc:path ),+]#
```

state-chain/pallets/cf-elections/src/electoral_systems/tests/change.rs
```diff
@@ -0,0 +1,117 @@
...
+
+type Vote = u64;
+type SimpleChange = Change<(), Vote, (), MockHook>;
+
+register_checks! {
+	SimpleChange {
+		hook_has_been_called(_pre, _post) {
+			assert!(MockHook::called(), "Hook should have been called!");
+		},
+		hook_not_been_called(_pre, _post) {
```
kylezs, line 31: nit
```suggestion
		hook_called(_pre, _post) {
			assert!(MockHook::called(), "Hook should have been called!");
		},
		hook_not_called(_pre, _post) {
```

state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_median.rs
```diff
@@ -0,0 +1,207 @@
...
+fn check_consensus_correctly_calculates_median_when_exactly_super_majority_authorities_vote() {
+	const AUTHORITY_COUNT: AuthorityCount = 10;
+	let vote_count = cf_utilities::success_threshold_from_share_count(AUTHORITY_COUNT);
+	let votes = (0..vote_count).map(|v| ((), v as u64)).collect::<Vec<_>>();
+
+	with_default_context().expect_consensus(AUTHORITY_COUNT, votes, Some(3));
+}
+
+#[test]
+fn to_few_votes_consensus_not_possible() {
```
kylezs, line 81:
```suggestion
fn too_few_votes_consensus_not_possible() {
```

state-chain/pallets/cf-elections/src/electoral_systems/tests/change.rs
```diff
@@ -0,0 +1,117 @@
...
+		(0..SUCCESS_THRESHOLD)
+			.map(|_| ((), CORRECT_VALUE))
+			.chain((SUCCESS_THRESHOLD..AUTHORITY_COUNT).map(|_| ((), INCORRECT_VALUE)))
+			.collect(),
+		Some(CORRECT_VALUE),
+	);
+}
+
+#[test]
+fn finalization_only_on_consensus_change() {
```
kylezs, line 91:
```suggestion
fn hook_only called_on_consensus_change() {
```

state-chain/pallets/cf-elections/src/electoral_systems/tests/change.rs
```diff
@@ -0,0 +1,117 @@
...
+			AUTHORITY_COUNT,
+			vec![((), Vote::default() + 1); AUTHORITY_COUNT as usize],
+			Some(Vote::default() + 1),
+		)
+		.test_on_finalize(
+			&(),
+			|_| {
+				assert!(!MockHook::called());
+			},
+			vec![Check::<SimpleChange>::hook_has_been_called()],
```
kylezs, line 115: could also check last_election_deleted here, which would imply that we can't have two change updates for the same election - which I think is a worthwhile check here

kylezs (approved):

PR #5203, opened 2024-08-26
test: add new LP to execute new tests in parallel

marcellorigotti:
# Pull Request

Closes: PRO-1600

## Checklist

## Summary

Added //LP_3 to the default lp accounts of localnet so that we can run other tests in parallel without interfering with existing tests

### DIFF
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: true
+      run-job: false
   publish:
     needs: [package]
     uses: ./.github/workflows/_30_publish.yml
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
   publish:
     needs: [package]
     uses: ./.github/workflows/_30_publish.yml
--- a/bouncer/shared/create_and_delete_multiple_orders.ts
+++ b/bouncer/shared/create_and_delete_multiple_orders.ts
@@ -76,8 +76,8 @@ export async function createAndDeleteMultipleOrders(numberOfLimitOrders: number)
   await Promise.all(promises);
   console.log('Orders successfully submitted');
 
-  let openOrders = await countOpenOrders('BTC', 'USDC', lp.address, chainflip);
-  openOrders += await countOpenOrders('ETH', 'USDC', lp.address, chainflip);
+  let openOrders = await countOpenOrders('BTC', 'USDC', lp.address);
+  openOrders += await countOpenOrders('ETH', 'USDC', lp.address);
   console.log(`Number of open orders: ${openOrders}`);
 
   console.log('Deleting opened orders...');
@@ -92,8 +92,8 @@ export async function createAndDeleteMultipleOrders(numberOfLimitOrders: number)
   await orderDeleteEvent;
   console.log('All orders successfully deleted');
 
-  openOrders = await countOpenOrders('BTC', 'USDC', lp.address, chainflip);
-  openOrders += await countOpenOrders('ETH', 'USDC', lp.address, chainflip);
+  openOrders = await countOpenOrders('BTC', 'USDC', lp.address);
+  openOrders += await countOpenOrders('ETH', 'USDC', lp.address);
   console.log(`Number of open orders: ${openOrders}`);
 
   assert.strictEqual(openOrders, 0, 'Number of open orders should be 0');
--- a/bouncer/shared/create_and_delete_multiple_orders.ts
+++ b/bouncer/shared/create_and_delete_multiple_orders.ts
@@ -90,12 +90,6 @@ export async function createAndDeleteMultipleOrders(numberOfLimitOrders: number)
   let openOrders = await countOpenOrders('BTC', 'USDC', lp.address, chainflip);
   openOrders += await countOpenOrders('ETH', 'USDC', lp.address, chainflip);
   console.log(`Number of open orders: ${openOrders}`);
-  // number of open orders should be equal to: numberOfLimitOrders * 2 (we create numberOfLimitOrders for both ETH and BTC) + 2 (we create 2 extra range orders)
-  assert.strictEqual(
-    openOrders,
-    numberOfLimitOrders * 2 + 2,
-    'Number of open orders do not match the correct amount, were there existing orders already?',
-  );
 
   console.log('Deleting opened orders...');
   const orderDeleteEvent = observeEvent('liquidityPools:LimitOrderUpdated', {
--- a/bouncer/shared/create_and_delete_multiple_orders.ts
+++ b/bouncer/shared/create_and_delete_multiple_orders.ts
@@ -81,7 +81,7 @@ export async function createAndDeleteMultipleOrders(numberOfLimitOrders: number,
   console.log(`Number of open orders: ${openOrders}`);
 
   console.log('Deleting opened orders...');
-  const orderDeleteEvent = observeEvent('liquidityPools:LimitOrderUpdated', {
+  const orderDeleteEvent = observeEvent('liquidityPools:RangeOrderUpdated', {
     test: (event) => event.data.lp === lp.address && event.data.baseAsset === 'Btc',
   }).event;
   await lpMutex.runExclusive(async () => {
--- a/bouncer/shared/create_and_delete_multiple_orders.ts
+++ b/bouncer/shared/create_and_delete_multiple_orders.ts
@@ -1,34 +1,17 @@
+import { ApiPromise } from '@polkadot/api';
 import { Keyring } from '../polkadot/keyring';
 import { handleSubstrateError, lpMutex } from '../shared/utils';
 import { getChainflipApi, observeEvent } from './utils/substrate';
 import { limitOrder } from './limit_order';
+import { rangeOrder } from './range_order';
 
-export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
-  console.log(`=== cancel_orders_batch test ===`);
-  await using chainflip = await getChainflipApi();
-
-  const keyring = new Keyring({ type: 'sr25519' });
-  keyring.setSS58Format(2112);
-  const lpUri = process.env.LP_URI || '//LP_1';
-  const lp = keyring.createFromUri(lpUri);
-
-  // create a series of limit_order and save their info to delete them later on
-  const promises = [];
-  const orderToDelete: {
-    Limit?: { base_asset: string; quote_asset: string; side: string; id: number };
-    Range?: { base_asset: string; quote_asset: string; id: number };
-  }[] = [];
-  let i = 0;
-  while (i < numberOfOrders) {
-    promises.push(limitOrder('Btc', 0.00000001, i, i));
-    orderToDelete.push({ Limit: { base_asset: 'BTC', quote_asset: 'USDC', side: 'sell', id: i } });
-    i++;
-  }
-  console.log('Submitting orders');
-  await Promise.all(promises);
-  console.log('Orders successfully submitted');
-
-  let orders = await chainflip.rpc('cf_pool_orders', 'BTC', 'USDC', lp.address);
+async function countOpenOrders(
+  baseAsset: string,
+  quoteAsset: string,
+  lp: string,
+  chainflip: ApiPromise,
+) {
+  const orders = await chainflip.rpc('cf_pool_orders', baseAsset, quoteAsset, lp);
   if (!orders) {
     throw Error('Rpc cf_pool_orders returned undefined');
   }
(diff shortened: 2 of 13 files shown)


state-chain/node/src/chain_spec/testnet.rs
```diff
@@ -139,6 +139,12 @@ pub fn extra_accounts() -> Vec<(AccountId, AccountRole, FlipBalance, Option<Vec<
 			100 * FLIPPERINOS_PER_FLIP,
 			Some(b"Chainflip Testnet LP 2".to_vec()),
 		),
+		(
+			get_account_id_from_seed::<sr25519::Public>("LP_3"),
+			AccountRole::LiquidityProvider,
+			100 * FLIPPERINOS_PER_FLIP,
+			Some(b"Chainflip Testnet LP 1".to_vec()),
```
j4m1ef0rd, line 146:
```suggestion
			Some(b"Chainflip Testnet LP 3".to_vec()),
```

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,15 +1,39 @@
...
+}
 
 export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
   console.log(`=== cancel_orders_batch test ===`);
   await using chainflip = await getChainflipApi();
 
   const keyring = new Keyring({ type: 'sr25519' });
   keyring.setSS58Format(2112);
-  const lpUri = process.env.LP_URI || '//LP_1';
+  const lpUri = process.env.LP_URI || '//LP_3';
```
j4m1ef0rd, line 36: nit: would be nice to define '//LP_3' in one place at the top of the file. Makes it easy to later see what account this test is using and change it.

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -59,18 +84,11 @@ export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
...
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.limit_orders.asks.length || 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.limit_orders.bids.length || 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.range_orders.length || 0;
+
+  openOrders = await countOpenOrders('BTC', 'USDC', lp.address, chainflip);
+  openOrders += await countOpenOrders('ETH', 'USDC', lp.address, chainflip);
+  console.log(`Number of open orders: ${openOrders}`);
```
j4m1ef0rd, line 90: log is repeated below

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -59,18 +84,11 @@ export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
...
-  openOrders = 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.limit_orders.asks.length || 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.limit_orders.bids.length || 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.range_orders.length || 0;
+
+  openOrders = await countOpenOrders('BTC', 'USDC', lp.address, chainflip);
+  openOrders += await countOpenOrders('ETH', 'USDC', lp.address, chainflip);
```
j4m1ef0rd, line 89: might as well assert openOrders == 0 here instead of just printing it.

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -20,34 +44,35 @@ export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
...
-
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.limit_orders.asks.length || 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.limit_orders.bids.length || 0;
-  // @ts-expect-error limit_orders does not exist on type AnyJson
-  openOrders += orders?.range_orders.length || 0;
+  let openOrders = await countOpenOrders('BTC', 'USDC', lp.address, chainflip);
+  openOrders += await countOpenOrders('ETH', 'USDC', lp.address, chainflip);
   console.log(`Number of open orders: ${openOrders}`);
```
j4m1ef0rd, line 74: would be nice to check openOrders == the correct amount here?.

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -20,34 +44,35 @@ export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
   }[] = [];
   let i = 0;
   while (i < numberOfOrders) {
-    promises.push(limitOrder('Btc', 0.00000001, i, i));
+    promises.push(limitOrder('Btc', 0.00000001, i, i, '//LP_3'));
     orderToDelete.push({ Limit: { base_asset: 'BTC', quote_asset: 'USDC', side: 'sell', id: i } });
     i++;
   }
+  i = 0;
+  while (i < numberOfOrders) {
```
j4m1ef0rd, line 52: Its misleading now that numberOfOrders does not include the range orders. Maybe rename it `numberOfLimitOrders`.

j4m1ef0rd (commented):
I think this will fail the upgrade test because the LP_3 account will not exist on the old chainspec that was upgraded. We hit this problem before when we added the 'LP_BOOST' account. @kylezs do you remember how we got around it?

bouncer/shared/create_and_delete_multiple_orders.ts
j4m1ef0rd: I like to use `assert.strictEqual()`

kylezs:
We solved it by running `setup_swaps()` in the upgrade_network, can see on the 1.4 tag: https://github.com/chainflip-io/chainflip-backend/blob/1250bd9014e727280d11fc276be24ea505c72ce1/bouncer/shared/upgrade_network.ts#L195

Here it'll be a little more involved - setting up account etc. But having the utilities to assign account roles within bouncer will surely come in handy at some point anyway - so we might as well develop those pieces as utilities and then do this.

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,97 @@
...
+    depositLiquidity('Usdc', 10000, false, LP),
+    depositLiquidity('Eth', deposits.get('Eth')!, false, LP),
+    depositLiquidity('Dot', deposits.get('Dot')!, false, LP),
+    depositLiquidity('Btc', deposits.get('Btc')!, false, LP),
+    depositLiquidity('Flip', deposits.get('Flip')!, false, LP),
+    depositLiquidity('Usdt', deposits.get('Usdt')!, false, LP),
+    depositLiquidity('ArbEth', deposits.get('ArbEth')!, false, LP),
+    depositLiquidity('ArbUsdc', deposits.get('ArbUsdc')!, false, LP),
+    depositLiquidity('Sol', deposits.get('Sol')!, false, LP),
+    depositLiquidity('SolUsdc', deposits.get('SolUsdc')!, false, LP),
```
kylezs, line 55: These should all be taking `lpUri` instead of `LP` else this test breaks if someone supplies an `LP_URI`
Would also rename `LP` to `DEFAULT_LP_URI` or something.

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,97 @@
...
+  if (!orders) {
+    throw Error('Rpc cf_pool_orders returned undefined');
+  }
+  let openOrders = 0;
+
+  // @ts-expect-error limit_orders does not exist on type AnyJson
+  openOrders += orders?.limit_orders.asks.length || 0;
+  // @ts-expect-error limit_orders does not exist on type AnyJson
+  openOrders += orders?.limit_orders.bids.length || 0;
+  // @ts-expect-error limit_orders does not exist on type AnyJson
```
kylezs, line 29:
```suggestion
  // @ts-expect-error range_orders does not exist on type AnyJson
```

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,97 @@
...
     Range?: { base_asset: string; quote_asset: string; id: number };
   }[] = [];
   let i = 0;
-  while (i < numberOfOrders) {
-    promises.push(limitOrder('Btc', 0.00000001, i, i));
+  while (i < numberOfLimitOrders) {
+    promises.push(limitOrder('Btc', 0.00000001, i, i, LP));
     orderToDelete.push({ Limit: { base_asset: 'BTC', quote_asset: 'USDC', side: 'sell', id: i } });
     i++;
   }
```
kylezs, line 69: nit
```suggestion
  for (let i = 0; i < numberOfLimitOrders; i++) {
    promises.push(limitOrder('Btc', 0.00000001, i, i, LP));
    orderToDelete.push({ Limit: { base_asset: 'BTC', quote_asset: 'USDC', side: 'sell', id: i } });
  }
```

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,97 @@
...
+    promises.push(limitOrder('Eth', 0.000000000000000001, i, i, LP));
+    orderToDelete.push({ Limit: { base_asset: 'ETH', quote_asset: 'USDC', side: 'sell', id: i } });
+    i++;
+  }
+
+  promises.push(rangeOrder('Btc', 0.1, LP));
+  orderToDelete.push({
+    Range: { base_asset: 'BTC', quote_asset: 'USDC', id: 0 },
+  });
+  promises.push(rangeOrder('Eth', 0.01, LP));
```
kylezs, line 81: same here `lpUri`

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,97 @@
...
+  i = 0;
+  while (i < numberOfLimitOrders) {
+    promises.push(limitOrder('Eth', 0.000000000000000001, i, i, LP));
+    orderToDelete.push({ Limit: { base_asset: 'ETH', quote_asset: 'USDC', side: 'sell', id: i } });
+    i++;
+  }
+
+  promises.push(rangeOrder('Btc', 0.1, LP));
+  orderToDelete.push({
+    Range: { base_asset: 'BTC', quote_asset: 'USDC', id: 0 },
```
kylezs, line 79: We could instead get the await on the order creations and get the OrderId, rather than rely on it always being 0. If we run this test again it'll fail because of this.

We can likely do it such that we can still Promise.all() on all the orders at the same time, and then we can get the order ids of the respective orders to place in ordersToDelete

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,97 @@
...
+import { deposits } from './setup_swaps';
 
-export async function createAndDeleteMultipleOrders(numberOfOrders: number) {
+const LP: string = '//LP_3';
+
+async function countOpenOrders(
+  baseAsset: string,
+  quoteAsset: string,
+  lp: string,
+  chainflip: ApiPromise,
```
kylezs, line 17: There's no need to pass in the API here, can just initialise it with `await using api = await getChainflipApi();`  - the initialiser of that function ensures there's only one underlying client and we effectively just get a reference to it.

bouncer/shared/setup_lp_account.ts
```diff
@@ -0,0 +1,96 @@
...
+  amountToFineAmount,
+} from './utils';
+import { approveErc20 } from './approve_erc20';
+import { getChainflipApi, observeEvent } from './utils/substrate';
+
+export async function setupLpAccount(lpKey: string) {
+  const keyring = new Keyring({ type: 'sr25519' });
+  keyring.setSS58Format(2112);
+  const lpUri = lpKey;
+  const lp = keyring.createFromUri(lpUri);
```
kylezs, line 24: this should probably be a function now, repeated a few times

bouncer/shared/setup_lp_account.ts
```diff
@@ -0,0 +1,96 @@
...
+  const txOptions = {
+    nonce: await getNextEvmNonce('Ethereum'),
+  } as const;
+
+  console.log('Funding 1000 Flip to ' + lp.address);
+  let pubkey = lp.address;
+  try {
+    pubkey = decodeFlipAddressForContract(lp.address);
+  } catch {
+    // ignore error
```
kylezs, line 56: why? wouldn't it be bad if decoding failed?

bouncer/shared/setup_lp_account.ts
```diff
@@ -0,0 +1,96 @@
...
+  getWhaleKey,
+  assetDecimals,
+  lpMutex,
+  handleSubstrateError,
+  amountToFineAmount,
+} from './utils';
+import { approveErc20 } from './approve_erc20';
+import { getChainflipApi, observeEvent } from './utils/substrate';
+
+export async function setupLpAccount(lpKey: string) {
```
kylezs, line 20: This whole function is *very* similar to fund_flip - we should be able to share a lot more code here

bouncer/shared/upgrade_network.ts
```diff
@@ -350,5 +351,9 @@ export async function upgradeNetworkPrebuilt(
     await incompatibleUpgradeNoBuild(localnetInitPath, binariesPath, runtimePath, numberOfNodes);
   }
 
+  if (cleanOldVersion === '1.6.1') {
```
kylezs, line 354: We might bump the versions... as we have, we're at 1.6.4 now

I believe you can do something like
```suggestion
  if (cleanOldVersion.startsWith("1.6")) {
```

bouncer/shared/setup_lp_account.ts
marcellorigotti: Actually my bad here I don't know why I copy pasted the code of fund_flip instead of simply calling it :/
Anyway fixed the comments in fund_flip directly

bouncer/shared/create_and_delete_multiple_orders.ts
marcellorigotti: This is not true, the test pass every time since rangeOrder always create an order with id = 0;
The test won't pass only if the LP is used to other tests, in which it is used to create more orders with an id that doesn't get deleted by this test, but the whole point was to create this new lp specifically for this test so that we could run it in parallel hence I don't think this is a problem.

.github/workflows/ci-development.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
```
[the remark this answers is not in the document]
marcellorigotti: I'll remove it as soon as it pass all the CI checks after the latest changes 👍

bouncer/shared/create_and_delete_multiple_orders.ts
kylezs: My thinking is beyond just this test - if we wanted to right something like a "chaos mode" that runs a bunch of bouncer tests randomly one after the other, then we would want to be able to run it twice without assuming an order id (applies to limit orders too ofc not just here)

marcellorigotti: yeah but what I am saying is that this test always create orders with id starting from 0 to numLimitOrders (for limit orders) and id = 0 for range orders and deletes them no matter what.
It works as many times in a row you want. 
For limit orders this is not required since we are passing the orderId to the function, hence we are sure it will always delete the order it creates, I could do the same for the rangeOrder function if we want to be sure to delete the same orders we create (I just used 0 since it was hardcoded in the rangeOrder function).

bouncer/shared/create_and_delete_multiple_orders.ts
```diff
@@ -1,54 +1,86 @@
...
 
-  const keyring = new Keyring({ type: 'sr25519' });
-  keyring.setSS58Format(2112);
-  const lpUri = process.env.LP_URI || '//LP_1';
-  const lp = keyring.createFromUri(lpUri);
+  const lpUri = process.env.LP_URI || DEFAULT_LP;
+  const lp = createLpKeypair(lpUri);
+
+  await Promise.all([
+    // provide liquidity to LP_3
```
j4m1ef0rd, line 37: LP_3 will be funded each time we run the test, this is slow and once we are using the new chainspec, it will be unnecessary. Shouldn't we just put this in the `setup_lp_account` function? Makes sense that we would always want to fund an LP account when making one.

j4m1ef0rd (approved):
Will be easier if you can merge this PR before my bouncer PR.

bouncer/shared/create_and_delete_multiple_orders.ts
marcellorigotti: Feel free to merge it if you're ok with it Jaime

j4m1ef0rd: oh i see. I thought the account was pre-funded in the chainspec. Yeah thats fine, we can move it later.

kylezs:
Looks like failure could be related to this test @marcellorigotti

marcellorigotti:
> Looks like failure could be related to this test @marcellorigotti

Yeah it can fail since the limit_orders can be filled and not be deleted, hence not emitting the correct event.
Change it to check the RangeOrders instead since those can never be filled!

PR #5210, opened 2024-08-29
chore: tidy up solana vote impls

dandanlen:
This is a refactor / tidy up I did as I was debugging other issues. 

Some minor changes to behaviour, specifically these two case:
- when the rpc returns no priority fees -> now we return a default value MIN_PRIORITIZATION_FEE
- when the rpc returns no new durable nonce -> now we return the *old* nonce instead of erroring, and the impl of the `Change` electoral system handles the rest.

Details from the commit message:

- move implementation details into their own modules.
- make impls return Option instead of Vec and take single arg where appropriate
- remove unused data and assertions (slot, signature comparison etc)
- Handle case when there are no Prio Fees (add MIN_PRIORITIZATION_FEE)
- Durable Nonce default to old nonce if the new one isn't witnessed (election logic handles this).

### DIFF
--- a/engine/src/witness/sol.rs
+++ b/engine/src/witness/sol.rs
@@ -1,3 +1,5 @@
+mod egress_witnessing;
+mod fee_tracking;
 mod nonce_witnessing;
 mod sol_deposits;
 
@@ -6,18 +8,13 @@ use crate::{
 	sol::{
 		commitment_config::CommitmentConfig,
 		retry_rpc::{SolRetryRpcApi, SolRetryRpcClient},
-		rpc_client_api::{
-			RpcTransactionConfig, TransactionConfirmationStatus, UiTransactionEncoding,
-		},
 	},
 	state_chain_observer::client::{
 		chain_api::ChainApi, electoral_api::ElectoralApi,
 		extrinsic_api::signed::SignedExtrinsicApi, storage_api::StorageApi,
 	},
-	witness::sol::sol_deposits::get_channel_ingress_amounts,
 };
-use anyhow::{anyhow, Result};
-use cf_chains::sol::{SolSignature, LAMPORTS_PER_SIGNATURE};
+use anyhow::Result;
 use futures::FutureExt;
 use pallet_cf_elections::{electoral_system::ElectoralSystem, vote_storage::VoteStorage};
 use state_chain_runtime::{
@@ -29,7 +26,7 @@ use state_chain_runtime::{
 };
 
 use std::sync::Arc;
-use utilities::{context, task_scope, task_scope::Scope};
+use utilities::{task_scope, task_scope::Scope};
 
 #[derive(Clone)]
 struct SolanaBlockHeightTrackingVoter {
@@ -55,6 +52,9 @@ struct SolanaFeeTrackingVoter {
 	client: SolRetryRpcClient,
 }
 
+// TODO: decide on a reasonable value for this.
+const MIN_PRIORITIZATION_FEE: u64 = 0;
+
 #[async_trait::async_trait]
 impl VoterApi<SolanaFeeTracking> for SolanaFeeTrackingVoter {
 	async fn vote(
@@ -63,13 +63,9 @@ impl VoterApi<SolanaFeeTracking> for SolanaFeeTrackingVoter {
 		_properties: <SolanaFeeTracking as ElectoralSystem>::ElectionProperties,
 	) -> Result<<<SolanaFeeTracking as ElectoralSystem>::Vote as VoteStorage>::Vote, anyhow::Error>
 	{
-		let priorization_fees = self.client.get_recent_prioritization_fees().await;
-
-		let mut priority_fees: Vec<u64> =
-			priorization_fees.iter().map(|f| f.prioritization_fee).collect();
-		priority_fees.sort();
-
-		Ok(context!(priority_fees.get(priority_fees.len().saturating_sub(1) / 2).cloned())?)
+		Ok(fee_tracking::get_median_prioritization_fee(&self.client)
+			.await
+			.unwrap_or(MIN_PRIORITIZATION_FEE))
 	}
 }
 
@@ -88,7 +84,7 @@ impl VoterApi<SolanaIngressTracking> for SolanaIngressTrackingVoter {
 		<<SolanaIngressTracking as ElectoralSystem>::Vote as VoteStorage>::Vote,
 		anyhow::Error,
 	> {
-		get_channel_ingress_amounts(
+		sol_deposits::get_channel_ingress_amounts(
 			&self.client,
 			settings.vault_program,
 			settings.usdc_token_mint_pubkey,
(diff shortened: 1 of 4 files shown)


engine/src/witness/sol.rs
```diff
@@ -55,6 +52,9 @@ struct SolanaFeeTrackingVoter {
 	client: SolRetryRpcClient,
 }
 
+// TODO: decide on a reasonable value for this.
+const MIN_PRIORITIZATION_FEE: u64 = 0;
```
dandanlen, line 56: @albert-llimos would be good to get your input on whether there's a sensible value to use for this.

engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -49,20 +52,21 @@ where
...
 
-			result.push((*nonce_account, hash));
-		} else {
-			return Err(anyhow!("Expected UiAccountData::Json(ParsedAccount)"));
-		}
+			Ok(Some(hash))
+		},
+		Some(_) => Err(anyhow!("Expected UiAccountData::Json(ParsedAccount)")),
+		None => Ok(None),
 	}
```
dandanlen, line 69: This is the main change to the nonce witnessing. 

Previously, if the rpc returned `None` we would error. Now we return `Ok(None)` to indicate that the rpc succeeded but that the account wasn't found. This should cause the vote to be filtered out, and retried later.

engine/src/witness/sol.rs
albert-llimos: yes this is fine. Shouldn't it be named `DEFAULT_PRIORITIZATION_FEE` though?

engine/src/witness/sol.rs
```diff
@@ -114,16 +110,12 @@ impl VoterApi<SolanaNonceTracking> for SolanaNonceTrackingVoter {
...
-
-		assert_eq!(response_account, nonce_account);
-
-		Ok(response_nonce)
+		let (nonce_account, previous_nonce) = properties;
+		Ok(nonce_witnessing::get_durable_nonce(&self.client, nonce_account)
+			.await?
+			// If the nonce is not found, we default to the previous nonce.
+			// The `Change` electoral system ensure this vote is filtered.
+			.unwrap_or(previous_nonce))
```
albert-llimos, line 118: Wouldn't be rather not vote? We'll get like 150*10 extrinics every 6 seconds for no reason.

engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -11,16 +12,16 @@ use anyhow::{anyhow, Result};
 use serde_json::Value;
 use std::str::FromStr;
 
-pub async fn get_durable_nonces<SolRetryRpcClient>(
+pub async fn get_durable_nonce<SolRetryRpcClient>(
```
albert-llimos, line 15: I don't think I'd do these changes necessarily. We have an open ticket to get all durable nonces in one call so we don't have a separate `get_multiple_accounts` rpc call for each nonce but rather do them together. @kylezs

engine/src/witness/sol.rs
dandanlen: As per the comment - this is dealt with by the implementation of the 'change' electoral system. We vote for the 'latest' value and if it is unchanged, the vote is filtered out and effectively ignored. This is not super obvious, I agree. I would prefer if it was more explicit and we returned, say, `None`. But I think returning the current nonce is the correct thing to do here, given the expected behaviour of the electoral system. And changing it to Option<Vote> would be a much deeper change, requiring migrations. I would rather avoid this until we have a better understanding of the elections pallet.

engine/src/witness/sol/nonce_witnessing.rs
dandanlen: It's effectively just a refactor in order to better reflect what we actually do currently, removing the need to eg. assert on the size of the returned vec. If we change it later, we can update the implementation to reflect that, but IMO the correct and most consistent way to structure this is to have one set of interfaces that reflect what we want to do from the voter's POV (in this case: get the nonce for a single account), and these should hide the details of the solana rpcs (use get_multiple with a single-value array).

engine/src/witness/sol.rs
dandanlen: True - will rename.

engine/src/witness/sol/nonce_witnessing.rs
albert-llimos: I agree that it's better with the current implementation, it's just that it will change when we implement the improved approach. But it's a fairly easy change so it can be done if/when we do that as part of the ticket :+1:

engine/src/witness/sol/fee_tracking.rs
```diff
@@ -0,0 +1,24 @@
...
+	rpc_client_api::RpcPrioritizationFee,
+};
+
+pub async fn get_median_prioritization_fee(sol_client: &SolRetryRpcClient) -> Option<u64> {
+	let mut fees = sol_client
+		.get_recent_prioritization_fees()
+		.await
+		.into_iter()
+		.map(|RpcPrioritizationFee { prioritization_fee, .. }| prioritization_fee)
+		.collect::<Vec<_>>();
```
kylezs, line 12: We can do this after checking if empty

engine/src/witness/sol/egress_witnessing.rs
```diff
@@ -0,0 +1,114 @@
...
+			let transaction_meta = sol_client
+				.get_transaction(
+					&signature,
+					RpcTransactionConfig {
+						encoding: Some(UiTransactionEncoding::Json),
+						// Using finalized there could be a race condition where this
+						// doesn't get the tx. But "Processed" is timing out so we better
+						// retry with finalized.
+						commitment: Some(CommitmentConfig::finalized()),
+						// Getting also type 0 even if we don't use them atm
```
kylezs, line 38: this comment doesn't make sense to me

engine/src/witness/sol/egress_witnessing.rs
```diff
@@ -0,0 +1,114 @@
...
+				// witness something because no meta is returned.
+				None => LAMPORTS_PER_SIGNATURE,
+			})
+		},
+		Some(TransactionStatus { confirmation_status: other_status, .. }) => Err(anyhow::anyhow!(
+			"Transaction status is {:?}, waiting for {:?}.",
+			other_status,
+			TransactionConfirmationStatus::Finalized
+		)),
+		None => Err(anyhow::anyhow!("Unknown Transaction.")),
```
kylezs, line 60: Wouldn't this be reasonably likely? This is what we discussed yesterday, that it'll log error for now but we can improve it later?

engine/src/witness/sol/egress_witnessing.rs
dandanlen: Me neither - is copy-pasted from the previous impl. @albert-llimos ?

engine/src/witness/sol/egress_witnessing.rs
dandanlen: Yes - will add a comment to clarify.

engine/src/witness/sol/fee_tracking.rs
dandanlen: (And IMO reads a little better like this)

engine/src/witness/sol/egress_witnessing.rs
albert-llimos: no, this comment is correct.
There are legacy transactions and type 0 transactions. Type 0 being one version after legacy, it's just the naming that is confusing.
By default legacy transactions are queried, which are also the transactions we craft. However, we are querying max to type 0 just to query any possible transaction regardless of the type. As explained in the comment we could use None to just query legacy, which would be fine, but I thought I'd just extend it to all types just in case.

kylezs: Yeah not doubting it's correctness... just the phrasing before I can comment on the correctness

engine/src/witness/sol/egress_witnessing.rs
```diff
@@ -0,0 +1,116 @@
...
+			let transaction_meta = sol_client
+				.get_transaction(
+					&signature,
+					RpcTransactionConfig {
+						encoding: Some(UiTransactionEncoding::Json),
+						// Using finalized there could be a race condition where this
+						// doesn't get the tx. But "Processed" is timing out so we better
+						// retry with finalized.
+						commitment: Some(CommitmentConfig::finalized()),
+						// Getting also type 0 even if we don't use them atm
```
albert-llimos, line 38:
```suggestion
						// Getting both type 0 and legacy even if we our broadcasts are type legacy
```

engine/src/witness/sol/egress_witnessing.rs
albert-llimos: gotcha. Added a suggestion.

dandanlen: I updated the comment.

PR #5213, opened 2024-08-29
refactor: download dylibs from runner build script

kylezs:
# Pull Request

Closes: PRO-1447
Closes: PRO-1425
Closes: PRO-1606
Closse: PRO-1486

## Checklist

## Summary

- Downloads shared libraries from s3 for previous version. This removes the need to store large build artifacts in the repo. 
- Uploads both old and new dylibs as part of build artifact upload
- Utilises s3 for downloading the old version of the binaries as part of the upgrade-test.
   - From https://github.com/chainflip-io/chainflip-backend/pull/5225 onwards, the binaries will be zipped and put on s3, and this zip will include *both* dylibs required for that version, because both are uploaded in build.


.github/workflows/_25_package.yml
```diff
@@ -35,109 +35,6 @@ jobs:
...
-        with:
-          workflow: release-sisyphos.yml
-          name: chainflip-backend-bin
-          github_token: ${{ secrets.CF_BACKEND_GITHUB_TOKEN }}
-          path: latest-release
-
-      # See: engine-runner-bin/Cargo.toml for more information
-      - name: Move latest release dylib into the target/release ready to be packaged by cargo-deb.
-        run: |
-          mv latest-release/*.so target/release/
```
kylezs, line 139: We can delete this and all of the above because the `engine-runner` build script now downloads it the old dylib.

.github/workflows/upgrade-test.yml
```diff
@@ -129,75 +85,51 @@ jobs:
...
-          workflow_conclusion: ""
-          name: chainflip-backend-bin
-          github_token: ${{ secrets.CF_BACKEND_GITHUB_TOKEN }}
-          path: latest-release-bins
-          search_artifacts: true
+        run: |
+          mkdir -p upgrade-from-bins && cd upgrade-from-bins
+          curl https://artifacts.chainflip.io/${{ env.OLD_VERSION }}/chainflip-backend-bin.zip --output chainflip-backend-bin.zip
+          unzip chainflip-backend-bin.zip -d .
+          cd .. && ls -l upgrade-from-bins
```
anton-chain, line 101:
```suggestion
          mkdir -p upgrade-from-bins; ( cd upgrade-from-bins
          curl https://artifacts.chainflip.io/${{ env.OLD_VERSION }}/chainflip-backend-bin.zip --output chainflip-backend-bin.zip
          unzip chainflip-backend-bin.zip -d .
          ); ls -l upgrade-from-bins
```

.github/workflows/upgrade-test.yml
```diff
@@ -129,75 +85,51 @@ jobs:
...
         run: |
           set -x
-          RELEASE_VERSION=$(./latest-release-bins/chainflip-node --version)
-          echo $RELEASE_VERSION
-          echo "RELEASE_VERSION=${RELEASE_VERSION}" >> $GITHUB_ENV
-
-      - name: Download upgrade-to binaries for push or pull request 📥
+          OLD_BIN_VERSION=$(./upgrade-from-bins/chainflip-node --version)
+          echo $OLD_BIN_VERSION
+          echo "OLD_BIN_VERSION=${OLD_BIN_VERSION}" >> $GITHUB_ENV
```
anton-chain, line 115: can we store this value to `GITHUB_OUTPUT` too? would be handy if we'll need this info across other jobs in future

engine-runner-bin/build.rs
```diff
@@ -1,8 +1,56 @@
...
+		let response = get(format!("https://artifacts.chainflip.io/{OLD_VERSION}/{dylib_name}"))?;
+
+		if response.status().is_success() {
+			std::fs::create_dir_all(dest_folder)?;
+			let mut dest: BufWriter<File> = BufWriter::new(File::create(dylib_location)?);
+			let content = response.bytes()?;
+			copy(&mut content.as_ref(), &mut dest)?;
+			Ok(())
+		} else {
+			Err(Box::from(format!("Failed to download file: {}", response.status())))
```
dandanlen, line 42: Could a bit more informative - what file?

engine-runner-bin/build.rs
```diff
@@ -1,8 +1,56 @@
...
+	// or added another commit on top then we get the latest build artifacts for a particular
+	// version.
+	if prebuilt_supported {
+		let response = get(format!("https://artifacts.chainflip.io/{OLD_VERSION}/{dylib_name}"))?;
+
+		if response.status().is_success() {
+			std::fs::create_dir_all(dest_folder)?;
+			let mut dest: BufWriter<File> = BufWriter::new(File::create(dylib_location)?);
+			let content = response.bytes()?;
+			copy(&mut content.as_ref(), &mut dest)?;
```
dandanlen, line 39: Looks like this can be written directly using response.copy_to(..) to avoid allocating the additional `content` Vec (which is quite big?).

dandanlen (approved):
LGTM - as discussed, would be good to somehow ensure the build artifacts are immutable, then we don't need to download them on every build.

PR #5214, opened 2024-08-30
feat: pool fills rpc subscription

msgmaxim:
# Pull Request

Closes: PRO-1543

## Checklist

## Summary

- Moved most of the logic from `order_fills` lp-api command under `custom-rpc` so that it can be used/imported in both places.
- Added the ability to `new_subscription` to pass state from one call of the closure to the next (needed to pass previous pool state in this case, but this could be useful in other subscriptions too).
- Added new api methods to runtime: `cf_pools` and `cf_lp_events`; limited events to LP only to minimise any compatibility issues between runtime and node. The order updates are computed outside of runtime since they require querying (or caching) previous state of the pools storage item (not sure if this can be done from within runtime directly).

- Tested on localnet with `wscat -c ws://localhost:9944` followed by `{"id":1, "jsonrpc":"2.0", "method": "cf_subscribe_lp_order_fills"}` and performing some swaps.


state-chain/custom-rpc/src/order_fills.rs
```diff
@@ -0,0 +1,179 @@
...
+		quote_asset: Asset,
+		side: Side,
+		id: U256,
+		tick: Tick,
+		sold: U256,
+		bought: U256,
+		fees: U256,
+		remaining: U256,
+	},
+	RangeOrder {
```
niklasnatter, line 29: it would be great if we can add the amount that was filled by this order

state-chain/runtime/src/lib.rs
```diff
@@ -2009,9 +2022,15 @@ impl_runtime_apis! {
...
 
-		fn cf_pools() -> Vec<PoolPairsMap<Asset>> {
+		fn cf_pool_pairs() -> Vec<PoolPairsMap<Asset>> {
 			LiquidityPools::pools()
 		}
+
+		fn cf_pools() -> BTreeMap<AssetPair, Pool<Runtime>> {
+
+			pallet_cf_pools::Pools::<Runtime>::iter().collect()
+
```
dandanlen, line 2032: This is a (potentially quite confusing) breaking change. 

It looks like this is just a workaround to access the pools storage - we can access it directly by adding a reference to the state backend to the rpc provider, and avoid adding another api method we will need to maintain.

state-chain/custom-rpc/src/order_fills.rs
dandanlen: I raised an issue for this, unfortunately it doesn't look so simple..

dandanlen:
@msgmaxim i've added the capability to query storage directly and reverted the change to cf-pools. Let me know what you think.

dandanlen:
Cool - think I'll rebase it (39 commits behind main) and let CI run once more before merging.

msgmaxim:
I'm asking @j4m1ef0rd to approve this so it can be merged.

j4m1ef0rd (approved):

PR #5226, opened 2024-09-03
tests: Swapping unit tests cleanup

j4m1ef0rd:
# Pull Request

Closes: PRO-1603

## Checklist

## Summary

- Experimented with factoring out common code in the DCA tests but settled with just having the common code in simple functions.
- moved some tests into files to help clean things up.
- Cleaned up a few swapping tests to make them consistent with the others.
- Added one more DCA test.


state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -2,6 +2,102 @@ use super::*;
...
+			remaining_input_amount: INPUT_AMOUNT - chunk_amount,
+			remaining_chunks: number_of_chunks - 1,
+			chunk_interval,
+			accumulated_output_amount: 0,
+		}
+	);
+}
+
+#[track_caller]
+fn assert_chunk_1_executed(
```
msgmaxim, line 60: This function is always called with the same parameters. I know that some of these may be useful to change in the future for other tests, but I also feel like the tests would be more clear if we removed either some or all of them (e.g. by highlighting that this isn't the focus of the tests). No strong opinion though.

I do think it is a little strage that we have to pass both `chunk_amount_after_fee` and `chunk_amount`. Once you have one you should be able to derive the other (using `BROKER_FEE_BPS`). As another option, you can assume `CHUNK_AMOUNT` and `CHUNK_AMOUNT_AFTER_FEE` are used in all tests and remove the amount parameters.

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -484,6 +484,137 @@ fn dca_with_fok_fully_executed() {
...
+					status: DcaStatus::ChunkScheduled(2),
+					remaining_input_amount: INPUT_AMOUNT,
+					remaining_chunks: NUMBER_OF_CHUNKS - 2,
+					chunk_interval: CHUNK_INTERVAL,
+					// Should still be 0
+					accumulated_output_amount: 0,
+				}
+			);
+		})
+		.then_execute_at_block(INIT_BLOCK + (SWAP_DELAY_BLOCKS + CHUNK_INTERVAL) as u64, |_| {})
```
msgmaxim, line 590: Nit: I would extract these (here and below) into constatns `CHUNK_1_BLOCK` etc (as we did in other tests), otherwise this math is a distraction.

msgmaxim (approved):

state-chain/pallets/cf-swapping/src/tests/fill_or_kill.rs
```diff
@@ -3,11 +3,11 @@ use super::*;
 const BROKER_FEE: AssetAmount = INPUT_AMOUNT * BROKER_FEE_BPS as u128 / 10_000;
 
 fn fok_swap(refund_params: Option<TestRefundParams>) -> TestSwapParams {
-	params(None, refund_params, false)
+	TestSwapParams::new(None, refund_params, false)
```
dandanlen, line 6: Why not create a constructor/builder instead of using a free function?

```
// Fok without ccm
TestSwapParams::fok(refund_params);
// or
TestSwapParams::new().with_fok(refund_params);

// Fok with ccm
TestSwapParams::fok_with_ccm(refund_params);
// or
TestSwapParams::new().with_fok(refund_params).with_ccm();
```


This can be extended to dca: we need fok params for dca, and we also want to test ccm with dca.

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -623,6 +472,138 @@ fn dca_with_fok_fully_executed() {
...
+			assert_has_matching_event!(
+				Test,
+				RuntimeEvent::Swapping(Event::SwapRequestCompleted {
+					swap_request_id: SWAP_REQUEST_ID
+				}),
+			);
+		});
+}
+
 mod ccm_tests {
```
dandanlen, line 607: nit: I think it makes more sense for these to live in the same module as the other ccm tests.

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -2,6 +2,100 @@ use super::*;
 
 const CHUNK_INTERVAL: u32 = 3;
 
+#[track_caller]
+fn setup_dca_swap(
+	number_of_chunks: u32,
+	chunk_interval: u32,
+	refund_params: Option<TestRefundParams>,
+) {
```
dandanlen, line 10: It seems like a function like this could have been factored out more generally. 

There are checks here that are repeated eg. for FOK swaps (the sanity check, asserting on SwapRequested and SwapScheduled. It's not clear why we have a special-case function only for dca. Could it not just be `setup_swaps(&[TestSwapParams])`?

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -645,7 +626,17 @@ mod ccm_tests {
 	const GAS_SWAP_ID: SwapId = 3;
 
 	#[track_caller]
-	fn check_ccm_dca_request_received() {
+	fn setup_ccm_dca_swap(
+		number_of_chunks: u32,
+		chunk_interval: u32,
+		refund_params: Option<TestRefundParams>,
+	) {
```
dandanlen, line 633: Similar to the other comment - why create two almost-but-not-quite identical functions setup_ccm_dca_swap and setup_dca_swap? Surely we want to do almost all the same checks? I can see differences between the two functions, but it's unclear if these differences are significant in any way.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1402,15 +856,16 @@ fn can_handle_ccm_with_zero_swap_outputs() {
...
+
 			swap_with_custom_broker_fee(Asset::Usdc, Asset::Eth, 100, bounded_vec![]);
 			swap_with_custom_broker_fee(Asset::Usdc, Asset::Eth, 1, bounded_vec![]);
 
 			// Change the swap rate so swap output will be 0
 			SwapRate::set(0.01f64);
-			System::reset_events();
 		})
-		.then_process_blocks_until(|_| System::block_number() == 4)
+		.then_execute_at_block(INIT_BLOCK + SWAP_DELAY_BLOCKS as u64, |_| {})
```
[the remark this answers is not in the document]
dandanlen: (added it here [e4a5138](https://github.com/chainflip-io/chainflip-backend/pull/5226/commits/e4a5138b0480a24251c22ca92599986fab2c99fc))

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -2,6 +2,100 @@ use super::*;
...
+		refund_params,
+		false,
+	)]);
+
+	// Check that the swap request was received;
+	assert_has_matching_event!(
+		Test,
+		RuntimeEvent::Swapping(Event::SwapRequested {
+			swap_request_id: SWAP_REQUEST_ID,
+			input_amount: INPUT_AMOUNT,
```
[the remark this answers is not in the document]
dandanlen: In this case, it looks like we just want to extract the input amount so we can calculate the chunk amount (the input amount doesn't matter: we want to know that the input, chunks and chunk amount are consistent).

dandanlen (commented):
This is definitely a big step in the right direction, but I feel like it go a bit further still, particularly wrt. test setup where I feel like there's still a bit of excess duplication. 

I'm also not a fan of inheriting constants like this ('like this' = implicitly inheriting values from parent modules), although I can see how it's convenient in many places...

state-chain/pallets/cf-swapping/src/tests/dca.rs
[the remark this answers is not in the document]
msgmaxim: I think we want tests to highlight what's important (i.e. how a given test case is different from others) and the input amount, like you said, is not important here. If we need a specific/different amount for some reason, we can generalise this or simply write custom setup code. I don't think we need to generalise prematurely.

state-chain/pallets/cf-swapping/src/tests/dca.rs
msgmaxim: What's the difference? These tests are also DCA, so they are not exactly out of place here.

state-chain/pallets/cf-swapping/src/tests/fill_or_kill.rs
j4m1ef0rd: didn't seem worth it. It is already quite easy and clean to just provide None if not using it.

state-chain/pallets/cf-swapping/src/tests/dca.rs
j4m1ef0rd: There are a few other ccm test that are not in the ccm module because i wanted to keep them with other test of the same type. eg, `reject_invalid_ccm_deposit`, `swap_excess_are_confiscated_ccm_via_deposit` ...
So i think its fine to leave them here. The ccm module is for ccm specific behaviour test.

state-chain/pallets/cf-swapping/src/tests/dca.rs
j4m1ef0rd: It got messy when accounting for the differences between ccm and not ccm, I would have to add logic to handle the gas and do extra checks. But I do not like putting logic in test code. It's better to have explicit hard coded asserts without any logic.

state-chain/pallets/cf-swapping/src/tests/dca.rs
j4m1ef0rd: It got messy when accounting for the differences between ccm and not ccm, I would have to add logic to handle the gas and do extra checks. But I do not like putting logic in test code. It's better to have explicit hard coded asserts without any logic.

state-chain/pallets/cf-swapping/src/tests/dca.rs
j4m1ef0rd: The dca test are quite different and only ever do one swap. The FoK tests use many swaps at a time and the setup code for them is already factored out and simple. Combining the setups makes both more complicated for no clear benefit.

state-chain/pallets/cf-swapping/src/tests/dca.rs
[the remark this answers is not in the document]
dandanlen: I'm not talking about generalising for the sake of it. If the value is not important, so why are we checking it? For a reader of the tests, seeing this check suggests that the INPUT_AMOUNT is somehow relevant, and automatically then thinks, but why? And where is it defined? What if it's different? I think proper scoping, or using default test setups where it's clear that we're using, effectively, a placeholder value, would make this clearer.

state-chain/pallets/cf-swapping/src/tests/dca.rs
dandanlen: I disagree - the clear benefit is that the checks are consistently applied across all tests. I can see there some differences between the two and I don't know if this is deliberate and significant, or just because one copy of the code has been updated, but not the other.

state-chain/pallets/cf-swapping/src/tests/dca.rs
dandanlen: My view is basically this:
- ccm can be fok and/or dca. 
- Non-ccm can also be fok and/or dca. 
- CCM is not really a sub-category of dca swaps. 

Therefore a module hierarchy that reflects this would make more sense IMO.

dandanlen:
As mentioned in the comments, I disagree with some of the opinions. 
I can live with this being merged though.

state-chain/pallets/cf-swapping/src/tests/dca.rs
msgmaxim: I think it would also be incorrect to say that DCA is a subcategory of CCM 🤷. Conceptually, I don't think there is a big difference, either way a test combining them will look out of place. FWIW fill-or-kill tests also include CCM tests already. I think we should take the pragmatic approach and decide where to put a test based on where it can benefit more from feature-specific helpers (which might very well be the CCM module in this case, I don't know, but I don't think the difference would be significant).

state-chain/pallets/cf-swapping/src/tests/fill_or_kill.rs
msgmaxim: I don't mind the builder pattern. I somewhat agree with Jamie that it might not be worth it at this stage, but we will likely have more paramters in the future, and I think it is a change we will eventually want to make at some point.

state-chain/pallets/cf-swapping/src/tests/dca.rs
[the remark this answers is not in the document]
msgmaxim: I'm not sure I understand your concerns. It is still important to use *some* value and check against it *and* other values derived from it. You can let rust-analyzer show where it is defined. It is defined externally, because other pieces of the same test rely on it. It is global so we don't have to pass it around, which make tests more clear. If it needs to be differnet for some reason in some test, then we can write a more specialized test just for that case, and we don't have to generalize everything just for that one case.

> it looks like we just want to extract the input amount so we can calculate the chunk amount

I see what you mean, but still, the same value is used (or derived from) in other places, so it is not good enough to check that something is "locally correct".

j4m1ef0rd:
I will merge now to avoid delaying other work both I and Max are doing involving swapping tests. But can re-visit later to explore Dan's options.

PR #5228, opened 2024-09-03
fix: decrease broadcast timeout constants for sol and arb

dandanlen:
I decreased it to 2 minutes.

### DIFF
--- a/state-chain/runtime/src/lib.rs
+++ b/state-chain/runtime/src/lib.rs
@@ -924,7 +924,7 @@ impl pallet_cf_broadcast::Config<Instance4> for Runtime {
 	type EnsureThresholdSigned =
 		pallet_cf_threshold_signature::EnsureThresholdSigned<Self, EvmInstance>;
 	type BroadcastReadyProvider = BroadcastReadyProvider;
-	type BroadcastTimeout = ConstU32<{ 90 * MINUTES }>;
+	type BroadcastTimeout = ConstU32<{ 2 * MINUTES }>;
 	type WeightInfo = pallet_cf_broadcast::weights::PalletWeight<Runtime>;
 	type SafeMode = RuntimeSafeMode;
 	type SafeModeBlockMargin = ConstU32<10>;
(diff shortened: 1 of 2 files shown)


kylezs (approved):

kylezs:
As discussed, we should increase this to 4 minutes, and then decrease the number of retries in the sol `retry_rpc.rs` to be 5 - which makes the max just over 2 minutes (currently it's just over an hour)

state-chain/runtime/src/lib.rs
```diff
@@ -924,7 +924,7 @@ impl pallet_cf_broadcast::Config<Instance4> for Runtime {
 	type EnsureThresholdSigned =
 		pallet_cf_threshold_signature::EnsureThresholdSigned<Self, EvmInstance>;
 	type BroadcastReadyProvider = BroadcastReadyProvider;
-	type BroadcastTimeout = ConstU32<{ 90 * MINUTES }>;
+	type BroadcastTimeout = ConstU32<{ 2 * MINUTES }>;
```
dandanlen, line 927:
```suggestion
	type BroadcastTimeout = ConstU32<{ 4 * MINUTES }>;
```

state-chain/runtime/src/lib.rs
```diff
@@ -957,7 +957,7 @@ impl pallet_cf_broadcast::Config<Instance5> for Runtime {
 	type EnsureThresholdSigned =
 		pallet_cf_threshold_signature::EnsureThresholdSigned<Self, SolanaInstance>;
 	type BroadcastReadyProvider = BroadcastReadyProvider;
-	type BroadcastTimeout = ConstU32<{ 90 * MINUTES }>;
+	type BroadcastTimeout = ConstU32<{ 2 * MINUTES }>;
```
dandanlen, line 960:
```suggestion
	type BroadcastTimeout = ConstU32<{ 4 * MINUTES }>;
```

albert-llimos:
I see that the EVM timeout. Could we reduce that also to 2 or 4 minutes?

state-chain/runtime/src/lib.rs
dandanlen, line 927: Not sure about this - the EVM retrier is the same for ETH and ARB chains. Should the protocol timeout be the same too? 🤔 

Thoughts?

state-chain/runtime/src/lib.rs
```diff
@@ -872,7 +872,7 @@ impl pallet_cf_broadcast::Config<Instance2> for Runtime {
 	type EnsureThresholdSigned =
 		pallet_cf_threshold_signature::EnsureThresholdSigned<Self, PolkadotInstance>;
 	type BroadcastReadyProvider = BroadcastReadyProvider;
-	type BroadcastTimeout = ConstU32<{ 10 * MINUTES }>;
+	type BroadcastTimeout = ConstU32<{ 4 * MINUTES }>;
```
kylezs, line 875: Polkadot? I assume is supposed to instance 4 for arb?

state-chain/runtime/src/lib.rs
kylezs: Protocol timeout should be longer for ETH, since we have to wait for safety margin

state-chain/runtime/src/lib.rs
dandanlen: I figured there was no reason not to reduce it for polkadot too - 4 is a bit of arbitrary choice, but 10 minutes for finality seemed excessive.

PR #5230, opened 2024-09-03
chore: fix compute units

albert-llimos:
## Checklist

## Summary

Accomodate for extra compute units depending on the amount of derivation required on-chain.


state-chain/chains/src/sol/transaction_builder.rs
```diff
@@ -113,18 +114,23 @@ impl SolanaTransactionBuilder {
...
 						compute_limit += COMPUTE_UNITS_PER_FETCH_NATIVE;
+						let fetch_pda_and_bump = derive_fetch_account(
+							param.deposit_fetch_id.address,
+							sol_api_environment.vault_program,
+						)
+						.map_err(SolanaTransactionBuildingError::FailedToDeriveAddress)?;
+
+						// Add extra compute units for each bump derivation required on-chain
+						compute_limit += (AccountBump::MAX - fetch_pda_and_bump.bump) as u32 *
+							COMPUTE_UNITS_PER_BUMP_DERIVATION;
```
kylezs, line 125: Seems we could dedup this

kylezs, line 125: Why is it `(AccountBump::MAX - fetch_pda_and_bump.bump)`, intuitively I'd have thought it to be just `fetch_pda_and_bump.bump * COMPUTE_UNITS_PER_BUMP_DERIVATION`

dandanlen: ... for example make it a method on the PdaAndBump struct. `impl PdaAndBump { fn compute_units(&self) -> ComputeUnits }`

albert-llimos: because bumps are started in reverse. So from the max value towards zero.

dandanlen: I think this will be easier to understand if we move the implementation into `PdaAndBump` impl as mentioned above (maybe with a comment to explain).

state-chain/chains/src/sol/transaction_builder.rs
```diff
@@ -52,15 +49,14 @@ fn token_program_id() -> SolAddress {
...
-	fn derivation_compute_units(&self) -> SolComputeLimit {
-		(AccountBump::MAX - self.bump) as u32 * COMPUTE_UNITS_PER_BUMP_DERIVATION;
-	}
-}
 pub struct SolanaTransactionBuilder;
 
 impl SolanaTransactionBuilder {
+	// Compute extra compute units for each bump derivation required on-chain
+	fn derivation_compute_units(bump: AccountBump) -> SolComputeLimit {
+		(AccountBump::MAX - bump) as u32 * COMPUTE_UNITS_PER_BUMP_DERIVATION
```
kylezs, line 57: a comment saying that bumps start from max and go towards 0 would be helpful

state-chain/chains/src/sol/transaction_builder.rs
```diff
@@ -137,6 +148,13 @@ impl SolanaTransactionBuilder {
 
 						compute_limit += COMPUTE_UNITS_PER_FETCH_TOKEN;
 
+						let fetch_pda_and_bump =
+							derive_fetch_account(ata.address, sol_api_environment.vault_program)
+								.map_err(SolanaTransactionBuildingError::FailedToDeriveAddress)?;
+
+						// Add extra compute units for on-chain derivation
+						compute_limit += Self::derivation_compute_units(fetch_pda_and_bump.bump);
```
kylezs, line 156: think I slightly preferred the last way you had it tbh, having the function on the fetch_pda_and_bump - since it doesn't require passing anything in, and it's specific to the "Pda and bumping" and not so much to the whole builder. was there a reason you changed it? 
Not a strong opinion either way

state-chain/chains/src/sol/transaction_builder.rs
albert-llimos: The `PdaAndBump` is part of the Solana primitives while this logic is more of a compute unit estimation (gas estimation) when crafting a transaction to be broadcasted. It might not be as straight forward to understand but putting it as part of an address primitive is conceptually wrong. Therefore, I've added it as a function to the TransactionBuilder.

state-chain/chains/src/sol/transaction_builder.rs
```diff
@@ -137,6 +152,13 @@ impl SolanaTransactionBuilder {
 
 						compute_limit += COMPUTE_UNITS_PER_FETCH_TOKEN;
 
+						let fetch_pda_and_bump =
```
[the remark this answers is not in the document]
albert-llimos: it's a PdaAndBump type but it's actually an ata (associated token account). Naming it that way makes it easier to identify the token accounts.

state-chain/chains/src/sol/transaction_builder.rs
albert-llimos: address it here (hopefully): https://github.com/chainflip-io/chainflip-backend/pull/5230#discussion_r1743405658

kylezs: So because this derivation shouldn't be done on PdaAndBump types directly (because it might actually be an ATA for example, we want to put it on transactionbuilder? am I understanding that correctly?

albert-llimos: no. Because the PdaAndBump is a primitive of Solana in a sense that it follows the Solana standard of derivation of addresses. ATA is a particular type of address derivation.
However, the compute units that we add per derivation (`derivation_compute_units`) is something we are adding as an estimation of the cost that we will incurr when deriving an address on-chain. It's acting here as a gas estimation more than anything else and it's not according to any standard/primitive of Solana.

dandanlen:
I cherry picked a squashed version of this to #5234

dandanlen (approved):

PR #5231, opened 2024-09-03
chore: Cleanup DispatchError that contains strings

syan095:
# Pull Request

Closes: PRO-1617

## Checklist

## Summary

Cleaned up strings used in DispatchError::Others. Refactored them into named errors if appropriate.
Removed deprecated extrinsics.

### DIFF
--- a/state-chain/chains/src/btc/api.rs
+++ b/state-chain/chains/src/btc/api.rs
@@ -148,8 +148,10 @@ impl<E: ReplayProtectionProvider<Bitcoin>> ExecutexSwapAndCall<Bitcoin> for Bitc
 
 // transfer_fallback is unsupported for Bitcoin.
 impl<E: ReplayProtectionProvider<Bitcoin>> TransferFallback<Bitcoin> for BitcoinApi<E> {
-	fn new_unsigned(_transfer_param: TransferAssetParams<Bitcoin>) -> Result<Self, DispatchError> {
-		Err(DispatchError::Other("Bitcoin's TransferFallback is not supported."))
+	fn new_unsigned(
+		_transfer_param: TransferAssetParams<Bitcoin>,
+	) -> Result<Self, TransferFallbackError> {
+		Err(TransferFallbackError::Unsupported)
 	}
 }
 
--- a/state-chain/chains/src/lib.rs
+++ b/state-chain/chains/src/lib.rs
@@ -500,6 +500,9 @@ pub enum AllBatchError {
 	/// Unable to select Utxos.
 	UtxoSelectionFailed,
 
+	/// Failed to build Solana transaction.
+	FailedToBuildSolanaTransaction(SolanaTransactionBuildingError),
+
 	/// Some other DispatchError occurred.
 	DispatchError(DispatchError),
 }
@@ -548,8 +551,15 @@ pub trait ExecutexSwapAndCall<C: Chain>: ApiCall<C::ChainCrypto> {
 	) -> Result<Self, ExecutexSwapAndCallError>;
 }
 
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo)]
+pub enum TransferFallbackError {
+	/// The chain does not support this functionality.
+	Unsupported,
+	/// Failed to lookup the given token address, so the asset is invalid.
+	CannotLookupTokenAddress,
+}
 pub trait TransferFallback<C: Chain>: ApiCall<C::ChainCrypto> {
-	fn new_unsigned(transfer_param: TransferAssetParams<C>) -> Result<Self, DispatchError>;
+	fn new_unsigned(transfer_param: TransferAssetParams<C>) -> Result<Self, TransferFallbackError>;
 }
 
 pub trait FeeRefundCalculator<C: Chain> {
--- a/state-chain/pallets/cf-environment/src/lib.rs
+++ b/state-chain/pallets/cf-environment/src/lib.rs
@@ -110,6 +110,8 @@ pub mod pallet {
 		EthAddressNotUpdateable,
 		/// The nonce account is currently not being used or does not exist.
 		NonceAccountNotBeingUsedOrDoesNotExist,
+		/// The given UTXO parameters are invalid.
+		InvalidUtxoParameters,
 	}
 
 	#[pallet::pallet]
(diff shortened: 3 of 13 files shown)


syan095:
I know I shouldn't mix tickets, but adding an extra field in the event was trivial, and this PR also touches on the same area of the code. I thought it was easier to just do it together than create a new PR for it.

state-chain/chains/src/btc/api.rs
```diff
@@ -149,7 +149,8 @@ impl<E: ReplayProtectionProvider<Bitcoin>> ExecutexSwapAndCall<Bitcoin> for Bitc
 // transfer_fallback is unsupported for Bitcoin.
 impl<E: ReplayProtectionProvider<Bitcoin>> TransferFallback<Bitcoin> for BitcoinApi<E> {
 	fn new_unsigned(_transfer_param: TransferAssetParams<Bitcoin>) -> Result<Self, DispatchError> {
-		Err(DispatchError::Other("Bitcoin's TransferFallback is not supported."))
+		// Bitcoin's TransferFallback is not supported.
+		Err(DispatchError::Other(""))
```
kylezs, line 153: Maybe I'm missing something, but what does this improve exactly?

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -379,7 +381,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			T::EnsureGovernance::ensure_origin(origin)?;
 
-			ensure!(params.are_valid(), DispatchError::Other("Invalid parameters"));
+			ensure!(params.are_valid(), Error::<T>::InvalidConsolidationParameters);
```
kylezs, line 384: The changes like this make sense where we're actually changing the type, but I don't really understand the above change (and the similar ones where we're still using DispatchError::Other).

state-chain/traits/src/mocks/account_role_registry.rs
```diff
@@ -11,6 +11,8 @@ impl MockPallet for MockAccountRoleRegistry {
 }
 
 const ACCOUNT_ROLES: &[u8] = b"AccountRoles";
+pub const ALREADY_REGISTERED_ERROR: DispatchError =
+	DispatchError::Other("Account already registered");
```
kylezs, line 15: So in this case we don't want to take it out of the Other("string") and into a comment?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -615,6 +615,7 @@ pub mod pallet {
 			amount: AssetAmount,
 		},
 		CcmFailed {
+			request_id: SwapRequestId,
```
kylezs, line 618: Personally would've made this change in a separate PR as it's not related to the error stuff at all

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -992,7 +993,11 @@ pub mod pallet {
 				// NOTE: DCA not yet supported for swaps from the contract
 				None,
 				SwapOrigin::Vault { tx_hash },
-			)?;
+			)
+			.is_err()
+			{
+				log_or_panic!("Regular swap request should never fail");
```
kylezs, line 999: CCM is the only case that the function can fail in

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: Also this issue was in triage, not in TODO so strictly speaking wasn't ready to pick up / hadn't be decided that we should do it (though in this case I do think we should do it).

dandanlen (commented):
Agree with Kyle - not quite sure what the benefit/aim of these changes is. We seem to be deleting information (but not always). 

Also agree with keeping PRs isolated, especially where there are no conflicts/dependencies between the PRs. It doesn't take much effort to open a new branch.

state-chain/chains/src/btc/api.rs
syan095: Changed this to a proper typed error enum

state-chain/pallets/cf-environment/src/lib.rs
syan095: Yeah because if we just use DispatchError::Other, you will get an error with no information. Here with a pallet error, you can get information about the error type from the name

state-chain/traits/src/mocks/account_role_registry.rs
syan095: As discussed, we decided that we can leave the cases where the string are used in tests only as is.

state-chain/pallets/cf-swapping/src/lib.rs
syan095: This is removed. WIll be done in a separate PR after it is triaged

state-chain/pallets/cf-swapping/src/lib.rs
syan095: Updated the comments

state-chain/chains/src/lib.rs
```diff
@@ -548,8 +551,15 @@ pub trait ExecutexSwapAndCall<C: Chain>: ApiCall<C::ChainCrypto> {
 	) -> Result<Self, ExecutexSwapAndCallError>;
 }
 
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo)]
+pub enum TransferFallbackError {
```
kylezs, line 555: 👍

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: Not really sure what you mean here... it looks like you removed this check and "log_or_panic" which I think is correct, but doesn't match what you're saying here... "Updated the comments. Also changed to log::error so it is testable." just want to double check

syan095: Oh Sorry! i got confused while working on 2 separate PRs. 

This PR here:
https://github.com/chainflip-io/chainflip-backend/pull/5242
removed the `log_or_panic` to log::error, so this code path is testable.

kylezs (approved):

dandanlen (approved):
Much better, thanks.

PR #5240, opened 2024-09-05
Feat: Bouncer test execution overhaul

j4m1ef0rd:
# Pull Request

Closes: PRO-1551

## Checklist

## Summary

**ExecutableTest**
- New class called `ExecutableTest`
- All test now have exactly 1 instance of this class
- It has details about the test, ie. name and timeout.
- It can store information about the test, ie. swap context
```ts
export const testFillOrKill = new ExecutableTest('FoK', main, 600);

async function main() {
	//  ...test code
}
```
Then to run the test in code:
```ts
import { testFillOrKill } from '../tests/fill_or_kill';

// Executes the test with the inbuilt timeout and exit the process
await testFillOrKill.runAndExit();

// Or if you don't want it to exit the process so you can run in parallel with something else:
const tests = [
  testFillOrKill.run(),
  testDCASwaps.run(),
]; 
await Promise.all(tests);
```

**How to run a test from the CLI**
No need to create a file just to be able to run a test from the command line. Just use the new "run_test" command
```sh
./commands/run_test.ts <test_name> <optional_arg1> <optional_arg2> ...

# Run a single test (not case sensitive and can be provided with or without spaces and hyphens)
./commands/run_test.ts "dca swaps"

# You can also provide the file path or file name of the test (so you can tab complete, you lazy boy)
./commands/run_test.ts DCA_test.ts

# Get a list of all available tests
./commands/run_test.ts -h
```
If you have created a new test and it is not on the list, simply add it to the list in `run_test.ts`.

**Folder Changes**
- All files in the `test/` are now **not** runnable via command line. They are the actual test code. Instead of being put in the `shared` folder with random utility code, they are all in one place here and can only be ran via executing it with a command file.
- All files in the `test_commands` are runnable via command line. But these are mostly legacy, and I want to remove all of them at some point in the future. 

**Other Changes**
- `console.log` has been replaced with `myExecutableTest.log` in all test. This allows auto adding of the test name as a prefix.
- You can use `myExecutableTest.logDebug` to log things that only show up of the test is being ran by its self, ie. via the `run_test.ts` command. Helpful for debugging a single test.
- All concurrent tests now uses the tests inbuilt timeout for each test. PRO-1551
- If a test errors or times out, it will log the name of the test that failed along with the error.
	- The swap context only prints a report if the test that failed was using the swap context
- Made a few tests use the swap context

**Out of scope / Future Plans**
- Get more tests to use the swap context
- Store and then output to a file the top level logs of the test that failed to help with debugging.
- Make the test know details about the arguments that it can take so the `run_test` command can let you know
- Make the test know what it is waiting for at any given time, so if it times out, it can tell you why.
- Remove the `test_commands` folder, or at least delete all of the test files that are replaced by the `run_test` command.
- Generalize the executable class code so we can use it for non-test commands and tasks.


bouncer/shared/executable_test.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (this.status === TestStatus.Running) {
+      throw new Error(`${this.name} Test is already running`);
+    }
+    console.log('\x1b[36m%s\x1b[0m', `=== Running ${this.name} test ===`);
+    this.status = TestStatus.Running;
+    await runWithTimeout(this.runFunction(...args), this.timeoutSeconds * 1000).catch((error) => {
+      // Print the swap report if the swap context was used
+      if (this.swapContext.allSwaps.size > 0) {
+        this.swapContext.print_report();
+      }
```
kylezs, line 45: This seems out of place to me - it seems like something that should be the responsibility of the swapping test - not of the general executable test function. Here it would probably look like something that the test being called can provide back to the caller like "callOnTimeout()" or "callOnFailure()", which would be `this.swapContext.print_report()` for the swapping test for example

bouncer/shared/executable_test.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.timeoutSeconds = timeoutSeconds;
+    this.swapContext = new SwapContext();
+  }
+
+  /// Run the test with the pre-defined timeout and any given arguments
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async run(...args: any[]) {
+    if (this.status === TestStatus.Running) {
+      throw new Error(`${this.name} Test is already running`);
+    }
```
kylezs, line 38: So it can be called again after it's already completed? what's the intention with this condition specifically? Why couldn't we kick off two instances of the same test?

bouncer/shared/executable_test.ts
```diff
@@ -0,0 +1,94 @@
...
+    console.log('\x1b[36m%s\x1b[0m', `=== Running ${this.name} test ===`);
```
kylezs, line 39: `\x1b[36m%s\x1b[0m` could we make this thing a constant of some kind () - it's not clear what this is by reading it

I understand the others are slightly different (it's not clear to me why they're different either), but we should be able to factor something out to make it more readable regardless.

kylezs (commented):
I like the general approach 👍  some comments that I think could improve things a bit further.

> All files in the test_commands are runnable via command line. But these are mostly legacy, and I want to remove all of them at some point in the future.

I think we should do this next so everything's consistent.

Followed by:

> A few tests need timeout values to be set. They were not using timeouts before.

The rest of the future plans, I like the ideas but we should evaluate priorities after these ones^ before continuing onto those.

bouncer/shared/executable_test.ts
j4m1ef0rd: Because I create the instance for the test in the files "global" context, we will only every have one instance, and we can guarantee it exists at all times. I went with this approach because:
- Way simpler than anything else i could think of. 
- Allows us to define the details of the test in one place and in the same place as the test itself.
- Allows using of the test instance for logging and other things without passing it to every function the test has.
- At the moment we never run 2 instances of a test.

If you want to still run 2 instances of a test, you can create a test that does that by using the inner function of the test instead.
```js
export const mySingleTest = new ExecutableTest('single', main, 600);
async function main(){
  // normal test 
}

export const myDoubleTest = new ExecutableTest('double', doubleTest, 600);
async function doubleTest(){
  await await Promise.all([main, main]);
}
```

bouncer/shared/executable_test.ts
```diff
@@ -0,0 +1,94 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  log(message: string, ...optionalParams: any[]) {
+    console.log('\x1b[1m%s\x1b[0m', `[${this.name}] ${message}`, ...optionalParams);
+  }
+
+  /// Only logs a message if the debug flag is enabled for this test
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  debugLog(message: string, ...optionalParams: any[]) {
+    if (this.debug) {
+      console.log('\x1b[30m%s\x1b[0m', `[${this.name}] ${message}`, ...optionalParams);
```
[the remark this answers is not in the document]
j4m1ef0rd: not worth it

bouncer/tests/all_concurrent_tests.ts
```diff
@@ -31,22 +29,22 @@ async function runAllConcurrentTests() {
...
+    swapLessThanED.run(),
+    testAllSwaps.run(),
+    testEvmDeposits.run(numberOfNodes),
+    testFundRedeem.run('redeem'),
+    testMultipleMembersGovernance.run(),
+    testLpApi.run(),
+    testBrokerFeeCollection.run(),
+    testBoostingSwap.run(),
+    testFillOrKill.run(),
+    testDCASwaps.run(),
```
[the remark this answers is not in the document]
j4m1ef0rd: We can make a map of `ExecutableTest` and `arguments[]`, but it ends up being messier than what we have already.

bouncer/shared/executable_test.ts
j4m1ef0rd: I disagree. Having it auto print the report is better than having to remember to add code to every test that will use the swap context to output the report on failure.
The `callOnTimeout` and `callOnFailure` is a good idea for when a test has custom stuff it needs to do, like cleaning code, but so far no tests need this, so I will add it when required.

bouncer/commands/execute_test.ts
```diff
@@ -0,0 +1,89 @@
...
+import { testRotatesThroughBtcSwap } from '../tests/rotates_through_btc_swap';
+import { testRotateAndSwap } from '../tests/rotation_barrier';
+import { testGasLimitCcmSwaps } from '../tests/gaslimit_ccm';
+import { testAllSwaps } from '../tests/all_swaps';
+import { testSwapAfterDisconnection } from '../tests/swap_after_temp_disconnecting_chains';
+
+async function main() {
+  const testName = process.argv[2];
+
+  // Every time we add a new test, we need to add it to this list.
```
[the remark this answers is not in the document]
j4m1ef0rd: > This doesn't have to be the case - why not ls tests and get the list that way?

I gave it a try. I can get a list of all of the names of instances of the class in all test files, but turning them back into actual instances that i can call functions on is really hard.

bouncer/commands/execute_test.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  // Match the test and run it
+  const additionalArgs = process.argv.slice(3);
+  for (const test of allTests) {
+    if (
+      testName.toLowerCase().replace(/-/g, '').replace(/ /g, '') ===
+      test.name.toLowerCase().replace(/-/g, '').replace(/ /g, '')
+    ) {
+      // This will exit the process when the test is done.
```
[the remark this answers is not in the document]
j4m1ef0rd: because if we want it to exit the process with the correct 1 or -1 (pass or error) we would have to call .run().catch and catch the error and exit with -1. Seemed easier just to let the execute function handle it and not repeat code.

bouncer/shared/boost.ts
```diff
@@ -20,6 +20,10 @@ import { requestNewSwap } from './perform_swap';
 import { createBoostPools } from './setup_boost_pools';
 import { jsonRpc } from './json_rpc';
 import { observeEvent, Event, getChainflipApi } from './utils/substrate';
+import { ExecutableTest } from './executable_test';
+
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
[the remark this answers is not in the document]
j4m1ef0rd: Haven't had a problem with it yet. But no idea if it has some hidden issue. @acdibble might be able to help?

bouncer/shared/evm_deposits.ts
```diff
@@ -170,9 +174,9 @@ async function testDoubleDeposit(sourceAsset: Asset, destAsset: Asset) {
   }
 }
 
-export async function testEvmDeposits(numberOfNodes: number) {
-  console.log('=== Testing EVM Deposits ===');
-
+async function main(numberOfNodes = 1) {
+  console.log('Running EVM deposit tests', numberOfNodes);
+  return;
```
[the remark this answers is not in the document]
j4m1ef0rd: was just for testing. fixed in the next commit already.

bouncer/shared/executable_test.ts
kylezs: Not every test is a swapping test is my point though, it's not relevant for every test, and therefore shouldn't be here

bouncer/commands/execute_test.ts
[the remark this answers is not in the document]
kylezs: Perhaps by using the exact model you have now it's hard, but does seem like it would be possible

bouncer/shared/executable_test.ts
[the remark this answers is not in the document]
kylezs: It ensures consistency between log message formats - and consistency is something we want to achieve with this change. It's also not a difficult change to make.

bouncer/shared/executable_test.ts
kylezs: makes sense

bouncer/commands/execute_test.ts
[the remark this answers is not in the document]
j4m1ef0rd: Yeah can be an upgrade later. It's still easier to add it to the list then it was to create a command file to run the test.

bouncer/shared/executable_test.ts
j4m1ef0rd: The swap context is a property of the executable test, so it seems fine to make printing the report it's responsibility.  
I agree that it's not perfect, but doing it another way will add complexity for no clear benefit.

bouncer/commands/execute_test.ts
j4m1ef0rd, line 1: We could call this `run_test.ts` instead? anyone prefer that?

kylezs: yeah, rolls off the tongue a bit better

kylezs (approved):

martin-chainflip:
I have to say, I'm not a fan of this redesign. IMO the ticket was meant to change the timeout logic to apply to individual tests, which could have been implemented in a much easier way. Adding a full refactor (and changes to how tests are written, logged, executed etc.) into such a ticket without prior discussions is not good style. Having said that, if nobody objects, then I'm fine with merging this.

j4m1ef0rd:
Any idea why the CI is failing? @ahasna

PR #5244, opened 2024-09-06
[RC 1.6.4]

dandanlen:
Replaces #5243

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -42,6 +42,7 @@ tree --no-default-features --depth 1 --edges=features,normal
 # - RUSTSEC-2022-0061: This advisory is related to the deprecated `parity-wasm`, not a security issue. It's a dependency of `substrate`.
 # - RUSTSEC-2024-0336: This adivsory comes from rustls, which is a dependency of the `try-runtime-cli` crate.
 # - RUSTSEC-2024-0320: Unmaintained transitive `yaml-rust` dependency of `insta` crate. We only use insta for testing.
+# - RUSTSEC-2024-0370: Unmaintained transitive dependency. Only affects macro generation efficiency.
 cf-audit = '''
 audit -D unmaintained -D unsound
     --ignore RUSTSEC-2022-0093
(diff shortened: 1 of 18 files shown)


albert-llimos:
While my initial reaction was thinking that we should do this, I'm not so convinced anymore.
I think the reason Alastair implemented it with `all_same` is because the nonce mechanism is critical to the system and we really want to be totally sure that a nonce has changed before updating in the State Chain. Updating it with a wrong or old value would lead to an invalid transaction, which is extremely problematic in, for instance, a rotation transaction.
The downside of course is that one node could then keep reporting the wrong node and prevent the node from being updated. While I don't like this downside, when it comes to nonces I'd always prefer to be on the side of safety and compromise "liveliness" over signing invalid transactions.

dandanlen:
@albert-llimos 

If we require unanimity then a single validator can prevent any nonce from being updated...

albert-llimos:
> If we require unanimity then a single validator can prevent any nonce from being updated...

I acknowledged that in my comment. My point was that there was most certainly a reason why Alastair implement it that way (safety) and it's worth considering if there is some edge case that he considered in where we would want unanimity.
If we can't think of one or we just don't think we require more safety, then that's fine.

dandanlen:
> > If we require unanimity then a single validator can prevent any nonce from being updated...
> 
> I acknowledged that in my comment. My point was that there was most certainly a reason why Alastair implement it that way (safety) and it's worth considering if there is some edge case that he thought about in which we would want unanimity. If we can't think of one or we just don't think we require more safety, then that's fine.

Maybe Alastair made a mistake? I don't think it adds more safety - it introduces a DoS vector.

I think the concern is that the validators might agree on some 'old' nonce, but requiring unanimity doesn't fix that: once we accept that a majority of validators might be wrong, we have to accept that they might all be wrong. I think the ultimate solution would be something like requiring them to witness (slot, nonce) and to require both a super-majority consensus on the nonce *and* that the consensus slot number increases.

albert-llimos:
> I think the concern is that the validators might agree on some 'old' nonce, but requiring unanimity doesn't fix that: once we accept that a majority of validators might be wrong, we have to accept that they might all be wrong. I think the ultimate solution would be something like requiring them to witness (slot, nonce) and to require both a super-majority consensus on the nonce _and_ that the consensus slot number increases.
Makes sense. That was probably the concern, and I'd say it makes sense because in Solana it's not uncommon for nodes to fall behind, which is what your ultimate solution would address. But anyway yes, you're right, let's do this PR for now.

kylezs (approved):

PR #5263, opened 2024-09-12
feat: move contract swap extrinsics into ingress-egress pallet

msgmaxim:
# Pull Request

Closes: PRO-1417

## Checklist

## Summary

- Moved `schedule_swap_from_contract` and `ccm_deposit` from cf-swapping to cf-ingress-egress as per the linear ticket (now called `contract_swap_request` and `contract_ccm_swap_request`).
- Because ingress-egress has different instances, the now engine needs to call the correct instance based on chain. I added `IngressCallBuilder` trait to the witnesser for this. Also it seems like `vault_transfer_failed` was incorrectly calling Ethereum's instance even for Arbitrum contracts. This fixes that too.
- Extended `AddressConverter` with `decode_and_validate_address_for_asset` so that both address and asset can be check in a single call from both pallets.
- There is now a second `CcmFailed` event in the ingress-egress pallet... (but see the discussion below.)
- Now if witness extrinsic fails due to the `decode_and_validate_address_for_asset` check, we print an erorr instead of returning DispatchError.
- Was able to simplify/remove some tests. Notable changes are:

    - `swap_by_witnesser_happy_path` -> `can_request_swap_via_extrinsic`
    - `can_process_ccms_via_extrinsic` -> `can_request_ccm_swap_via_extrinsic`
    - `swap_excess_are_confiscated_ccm_via_extrinsic` and `swap_excess_are_confiscated_for_swap_via_extrinsic` have been removed (confiscating happens in the swapping pallet, and the fact that it is initiated from contract by ingress-egress pallet is irrelevant)
 
  - `swap_excess_are_confiscated_for_swap_via_deposit` -> `swap_excess_are_confiscated` (renamed only)
  - `swap_excess_are_confiscated_ccm_via_deposit` -> removed (it doesn't test anything that swap_excess_are_confiscated doesn't already cover)
  - `can_swap_ccm_below_max_amount` -> removed (it doesn't test anything that can_swap_below_max_amount doesn't already cover) 
  - `reject_invalid_ccm_deposit`: removed unnecesary swaps initiated via contract which don't improve test coverage 

---
### Discussion:
- In the new benchmarks (which I mostly moved from the other pallet), I'm not sure how to verify the the swap request is created, since this is done in the other pallet. Since the mock isn't used, I'm not sure if we can read the other pallet's storage in that case. For now we just check that the extrinsic returns Ok().
- After resolving conficts with a PR that added `swap_request_id` to `CcmFailed` I realised that `swap_request_id` isn't always available when that event is emitted. `CcmFailed` is only emitted when some parameters are invalid, and we probably shouldn't even create a swap request when this is the case (it was done the way it is now to simplify the code, but maybe thsi wasn't the best desicion). I think it might make sense to check the parameters outside of the swapping pallet and so `CcmFailed` will only be emitted in one pallet, and no swap request will be created (this also means no `swap_request_id` will be included fyi @acdibble). This would also mean that init_swap_request will be infallible. Is it worth making that change here?


dandanlen:
@msgmaxim thanks, this looks good.

Regarding the discussion points: 

- Benchmarks: look fine. You can see from the weight files that the storage is in fact accessed (I ran it locally), so I think simply asserting OK is enough here. The alternative would be to add a `get_swap_request` method to the `SwapRequestHandler` trait (gated with runtime-benchmarks) and use that to check.
- Regarding the question of whether or not to assign a request_id, I feel like it makes sense to *not* assign one, to be consistent with normal swaps. However in that case we should make the check *before* calling `init_swap_request` (like we do with all other checks) and making this method infallible (which would be nice). We can do this by refactoring the `principal_and_gas_amounts` check: we can move it into the chains crate where all the 'ccm metadata' structs are defined. This would allow us to remove the CcmFailed event from the swapping pallet. *<-- I had a go at this I think it's an improvement. See branch feat/move-contract-swap-witnessing-ccm-swap-metadata: 388e932a3*

dandanlen (approved):

PR #5268, opened 2024-09-13
Handle CFE version submission sometimes silently failing.

MxmUrw:
# Pull Request

Closes: PRO-1608

## Checklist

## Summary

Currently, submitting a new CFE version might silently fail. We could not reproduce the exact failure conditions, but to safeguard against such an accidental failure we now do the following:

After submitting a new version, we get the actual (hopefully new) version number from storage to check whether our submission was successful. If it was not, we retry up to 10 times. We also retry if we get explicit errors during submission.

#### Non-Breaking changes

### DIFF
--- a/engine/src/state_chain_observer/client.rs
+++ b/engine/src/state_chain_observer/client.rs
@@ -813,7 +813,7 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 				// Here we call the submission code multiple times before giving up with an error.
 				const MAX_UPDATE_VERSION_RETRIES: usize = 10;
 				let mut update_successful = false;
-				for retry in 0..MAX_UPDATE_VERSION_RETRIES {
+				for retry in 1..=MAX_UPDATE_VERSION_RETRIES {
 					let block_hash = finalized_block_stream.cache().hash;
 					let block_number = finalized_block_stream.cache().number;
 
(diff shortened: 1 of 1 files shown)


engine/src/state_chain_observer/client.rs
```diff
@@ -795,52 +806,84 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
...
+					.map_err(|_| anyhow::anyhow!("Timed out trying to submit CFE version"))
+					.map(|x| x.map_err(|e| e.into())); // map inner subxt::Error to anyhow::Error
+
+					match result.flatten() {
+						Ok(_) => {
+							// Submitting did not return an error, but this does not mean it was
+							// successful. Thus we check whether the version was actually updated:
+							let new_version =
+								get_recorded_cfe_version(&subxt_client, &subxt_signer).await?;
+							if new_version == *CFE_VERSION {
```
kylezs, line 871: nit:
it might not actually be the "new version" in the case of it not updating, can just be inlined to avoid naming it
```suggestion
							if get_recorded_cfe_version(&subxt_client, &subxt_signer).await? == *CFE_VERSION {
```

engine/src/state_chain_observer/client.rs
```diff
@@ -795,52 +806,84 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
...
-						)
-						.await?;
+				// Submitting the CFE version has three possible results:
+				//   - Successful
+				//   - Returns no error, but not successful
+				//   - Returns error
+				// Here we call the submission code multiple times before giving up with an error.
+				const MAX_UPDATE_VERSION_RETRIES: usize = 10;
+				let mut update_successful = false;
+				for retry in 0..MAX_UPDATE_VERSION_RETRIES {
```
kylezs, line 816: We can avoid the retry+1s below :)
```suggestion
				for retry in 1..=MAX_UPDATE_VERSION_RETRIES {
```

kylezs (approved):

PR #5269, opened 2024-09-16
tests for ensure governance checks

j4m1ef0rd:
# Pull Request

Closes: PRO-1638

## Checklist

## Summary

- Removed the `NeverFailingOriginCheck` and added `FailOnNoneOrigin` & `OnlyAllowRootOrigin` instead. These allow us to test that the origin is being checked.
	- Mocks defaultly use `OnlyAllowRootOrigin` for governance checks and `FailOnNoneOrigin` for all others.
- Added tests for every governance extrinsic except for the ones in the elections pallet.
- Added a few test for non-governance origin checks.
- Changed the governance pallet to use the actual ensure governance check instead of a mock.
- Added several missing tests for extrinsics.
- Changed 2 extrinsics from a root origin check to a governance check.
- Looked over all other extrinsics to try and find any missing/incorrect origin checks. all good :)

### DIFF
--- a/state-chain/pallets/cf-governance/src/mock.rs
+++ b/state-chain/pallets/cf-governance/src/mock.rs
@@ -3,8 +3,11 @@ use std::cell::RefCell;
 use crate::{self as pallet_cf_governance};
 use cf_primitives::SemVer;
 use cf_traits::{
-	impl_mock_chainflip, mocks::time_source, AuthoritiesCfeVersions, CompatibleCfeVersions,
-	ExecutionCondition, RuntimeUpgrade,
+	mocks::{
+		account_role_registry::MockAccountRoleRegistry, ensure_origin_mock::FailOnNoneOrigin,
+		funding_info::MockFundingInfo, time_source,
+	},
+	AuthoritiesCfeVersions, CompatibleCfeVersions, ExecutionCondition, RuntimeUpgrade,
 };
 use frame_support::{derive_impl, dispatch::DispatchResultWithPostInfo, ensure, parameter_types};
 use frame_system as system;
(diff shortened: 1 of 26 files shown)


state-chain/pallets/cf-governance/src/mock.rs
```diff
@@ -63,7 +66,26 @@ impl system::Config for Test {
...
+
+impl cf_traits::Chainflip for Test {
+	type Amount = u128;
+	type ValidatorId = <Test as frame_system::Config>::AccountId;
+	type RuntimeCall = RuntimeCall;
+	type EnsureWitnessed = FailOnNoneOrigin<Self>;
+	type EnsurePrewitnessed = FailOnNoneOrigin<Self>;
+	type EnsureWitnessedAtCurrentEpoch = FailOnNoneOrigin<Self>;
+	// Using actual EnsureGovernance instead of mock
+	type EnsureGovernance = crate::EnsureGovernance;
```
dandanlen, line 84: 👌

dandanlen (commented):
I replaced one trait with EnsureRoot as mentioned. I also noticed there was an unused EnsureWitnessed mock macro. I removed it too. 

As a general remark, I think the 'negative' tests (asserting that the governance calls can't be called from an arbitrary signed origin) feel a little superfluous. I feel like asserting the origin(s) that *are* valid is sufficient, and we don't much extra by asserting which ones *aren't* valid. 

The only additional regression we might catch would be where we for some reason accidentally modified a governance call to accept the governance origin *or* the signed origin. 

Also the naming of these tests as `only_governance_can_do_x` is not strictly accurate: it only tests that the call cannot be dispatch by the signed origin. 

What do you think?

j4m1ef0rd:
> What do you think?

I agree it's not perfect, but it's better than nothing and it was easy to implement.

dandanlen (approved):

dandanlen:
> > What do you think?
> 
> I agree it's not perfect, but it's better than nothing and it was easy to implement.

I feel like those are not good reasons tbh.

PR #5258, opened 2024-09-11
feat: change broadcast timeout to be a storage value.

MxmUrw:
# Pull Request

Closes: PRO-1625

## Checklist

Please conduct a thorough self-review before opening the PR.

- [x] I am confident that the code works.
- [x] I have written sufficient tests.
- [x] I have written and tested required migrations.
- [ ] I have updated documentation where appropriate.

## Summary

Previously, the broadcast timeout was given by a compile-time constant (`#[pallet::constant]`). In order to make this value updateable at runtime, this PR does the following:
 - Define broadcast timeout as a storage value.
 - Add a generic extrinsic `update_pallet_config()` to the `broadcast` pallet, allowing to update the broadcast timeout when called by governance.
 - Add tests and benchmarking.

#### Non-Breaking changes

If this PR includes non-breaking changes, select the `non-breaking` label. On merge, CI will automatically cherry-pick the commit to a PR against the release branch.

PR #5270, opened 2024-09-16
Denote broadcast timeout in target chain blocks.

MxmUrw:
# Pull Request

Closes: PRO-989

## Checklist

## Summary

This change involves the following (in the broadcast pallet):
 - `BroadcastTimeout` is now specified in `ChainBlockNumberFor<T,I>`
 - `Timeouts` is now specified in `ChainBlockNumberFor<T,I>`
 - A new constant `SafeModeBlockMarginForTargetChain`, also in `ChainBlockNumberFor<T,I>` is introduced since the previously used `SafeModeBlockMargin` stays denominated in `BlockNumberFor<T>` for other purposes.
 - For convenience, introduction of block time constants for all target chains.
 - Update the migration introduced in #5258 since it hasn't been released yet.

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-broadcast/src/lib.rs
+++ b/state-chain/pallets/cf-broadcast/src/lib.rs
@@ -281,11 +281,9 @@ pub mod pallet {
 	/// A mapping from block number to a list of broadcasts that expire at that
 	/// block number.
 	#[pallet::storage]
-	pub type Timeouts<T: Config<I>, I: 'static = ()> = StorageMap<
+	pub type Timeouts<T: Config<I>, I: 'static = ()> = StorageValue<
 		_,
-		Twox64Concat,
-		ChainBlockNumberFor<T, I>,
-		BTreeSet<(BroadcastId, T::ValidatorId)>,
+		Vec<(ChainBlockNumberFor<T, I>, BTreeSet<(BroadcastId, T::ValidatorId)>)>,
 		ValueQuery,
 	>;
 
@@ -410,22 +408,24 @@ pub mod pallet {
 			// nominated. If there are no more broadcaster available, then the broadcast is aborted.
 
 			let current_chain_block = T::ChainTracking::get_block_height();
+			let mut expiries = BTreeSet::new();
 
 			// take all broadcasts which have timed out. Since iterators break if we update
 			// the map during iteration, we do this in two steps: first collect all expired_keys
 			// and separately the expired values, after that iterate again to delete all collected
 			// keys from storage.
-			let mut expired_keys = Vec::new();
-			let mut expiries = BTreeSet::new();
-			for (ix, val) in Timeouts::<T, I>::iter() {
-				if ix <= current_chain_block {
-					expiries.append(&mut val.clone());
-					expired_keys.push(ix);
-				}
-			}
-			for key in expired_keys {
-				Timeouts::<T, I>::remove(key);
-			}
+			Timeouts::<T, I>::mutate(|timeouts| {
+				// NB: As per documentation, retain visits every element exactly once,
+				// and can thus be used for iterations with state change.
+				timeouts.retain(|(ix, val)| {
+					if *ix <= current_chain_block {
+						expiries.append(&mut val.clone());
+						false
+					} else {
+						true
+					}
+				});
+			});
 
 			let pending_broadcasts = PendingBroadcasts::<T, I>::get();
 			let mut delayed_retries = DelayedBroadcastRetryQueue::<T, I>::take(block_number);
@@ -462,10 +462,10 @@ pub mod pallet {
 					});
 				}
 			} else {
-				Timeouts::<T, I>::mutate(
+				Timeouts::<T, I>::append((
 					current_chain_block.saturating_add(T::SafeModeBlockMarginForTargetChain::get()),
-					|current| current.append(&mut expiries),
-				);
+					expiries,
+				));
 				DelayedBroadcastRetryQueue::<T, I>::mutate(
 					block_number.saturating_add(T::SafeModeBlockMargin::get()),
 					|current| current.append(&mut delayed_retries),
@@ -909,10 +909,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			broadcast_data.nominee = Some(nominated_signer.clone());
 			AwaitingBroadcast::<T, I>::insert(broadcast_id, broadcast_data.clone());
 
-			Timeouts::<T, I>::append(
+			Timeouts::<T, I>::append((
 				T::ChainTracking::get_block_height() + BroadcastTimeout::<T, I>::get(),
-				(broadcast_id, nominated_signer.clone()),
-			);
+				BTreeSet::from([(broadcast_id, nominated_signer.clone())]),
+			));
 
 			T::CfeBroadcastRequest::tx_broadcast_request(TxBroadcastRequest {
 				broadcast_id,
--- a/state-chain/pallets/cf-broadcast/src/lib.rs
+++ b/state-chain/pallets/cf-broadcast/src/lib.rs
@@ -278,17 +278,10 @@ pub mod pallet {
 	pub type DelayedBroadcastRetryQueue<T: Config<I>, I: 'static = ()> =
 		StorageMap<_, Twox64Concat, BlockNumberFor<T>, BTreeSet<BroadcastId>, ValueQuery>;
 
-	/// A vector containing external chainblock numbers, together with the set of broadcasts
-	/// that expire at this block number.
-	///
-	/// Note that there might be multiple entries for the same timeout block, we allow this
-	/// because it is more efficient to append something to a vector than mutating its contents.
+	/// A vector containing broadcast_ids, together with the chain block numbers they time out at.
 	#[pallet::storage]
-	pub type Timeouts<T: Config<I>, I: 'static = ()> = StorageValue<
-		_,
-		Vec<(ChainBlockNumberFor<T, I>, BTreeSet<(BroadcastId, T::ValidatorId)>)>,
-		ValueQuery,
-	>;
+	pub type Timeouts<T: Config<I>, I: 'static = ()> =
+		StorageValue<_, Vec<(ChainBlockNumberFor<T, I>, BroadcastId, T::ValidatorId)>, ValueQuery>;
 
 	/// Stores the signed external API Call for a broadcast.
 	#[pallet::storage]
@@ -418,9 +411,9 @@ pub mod pallet {
 			// and separately the expired values, after that iterate again to delete all collected
 			// keys from storage.
 			Timeouts::<T, I>::mutate(|timeouts| {
-				timeouts.retain(|(expiry_block, broadcast_ids)| {
+				timeouts.retain(|(expiry_block, broadcast_id, nominee)| {
 					if *expiry_block <= current_chain_block {
-						expiries.append(&mut broadcast_ids.clone());
+						expiries.insert((broadcast_id.clone(), nominee.clone()));
 						false
 					} else {
 						true
@@ -463,10 +456,17 @@ pub mod pallet {
 					});
 				}
 			} else {
-				Timeouts::<T, I>::append((
-					current_chain_block.saturating_add(T::SafeModeChainBlockMargin::get()),
-					expiries,
-				));
+				for (broadcast_id, nominee) in expiries {
+					Timeouts::<T, I>::append((
+						current_chain_block.saturating_add(T::SafeModeChainBlockMargin::get()),
+						broadcast_id,
+						nominee,
+					))
+				}
+				// Timeouts::<T, I>::append((
+				// 	current_chain_block.saturating_add(T::SafeModeChainBlockMargin::get()),
+				// 	expiries,
+				// ));
 				DelayedBroadcastRetryQueue::<T, I>::mutate(
 					block_number.saturating_add(T::SafeModeBlockMargin::get()),
 					|current| current.append(&mut delayed_retries),
@@ -912,7 +912,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 			Timeouts::<T, I>::append((
 				T::ChainTracking::get_block_height() + BroadcastTimeout::<T, I>::get(),
-				BTreeSet::from([(broadcast_id, nominated_signer.clone())]),
+				broadcast_id,
+				nominated_signer.clone(),
 			));
 
 			T::CfeBroadcastRequest::tx_broadcast_request(TxBroadcastRequest {
--- a/state-chain/pallets/cf-broadcast/src/lib.rs
+++ b/state-chain/pallets/cf-broadcast/src/lib.rs
@@ -284,7 +284,6 @@ pub mod pallet {
 	pub type Timeouts<T: Config<I>, I: 'static = ()> = StorageMap<
 		_,
 		Twox64Concat,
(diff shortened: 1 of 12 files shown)


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -405,7 +408,25 @@ pub mod pallet {
...
+
+			let current_chain_block = T::ChainTracking::get_block_height();
+
+			// take all broadcasts which have timed out. Since iterators break if we update
+			// the map during iteration, we do this in two steps: first collect all expired_keys
+			// and separately the expired values, after that iterate again to delete all collected
+			// keys from storage.
+			let mut expired_keys = Vec::new();
+			let mut expiries = BTreeSet::new();
+			for (ix, val) in Timeouts::<T, I>::iter() {
```
dandanlen, line 420: I don't like having to iterate over a map of all timeouts on every block... maybe it would be easier to just store this as a single `Vec<(BlockNumber, BTreeMap<..>)>` value?

state-chain/pallets/cf-broadcast/src/migrations/initialize_broadcast_timeout_storage.rs
```diff
@@ -15,20 +22,21 @@ mod old {
 
 	// Same timeout values as previously defined in `#[pallet::constant]`s
 	// and same as currently used in `node/src/chain_spec.rs`
-	pub const ETHEREUM_BROADCAST_TIMEOUT: BlockNumber = 5 * MINUTES;
-	pub const POLKADOT_BROADCAST_TIMEOUT: BlockNumber = 4 * MINUTES;
-	pub const BITCOIN_BROADCAST_TIMEOUT: BlockNumber = 90 * MINUTES;
-	pub const ARBITRUM_BROADCAST_TIMEOUT: BlockNumber = 2 * MINUTES;
-	pub const SOLANA_BROADCAST_TIMEOUT: BlockNumber = 4 * MINUTES;
+	pub const ETHEREUM_BROADCAST_TIMEOUT: BlockNumber = 5 * BLOCKS_PER_MINUTE_ETHEREUM;
```
dandanlen, line 25: FWIW due to rounding this works out as more like 4.5 minutes. I don't think it matters but worth keeping in mind in case we try to coordinate this with anything external.

state-chain/pallets/cf-broadcast/src/lib.rs
kylezs: by BlockNumber you mean ChainBlockNumber?

kylezs: if yes - then yeah, looking now we only append to timeouts so writing is still cheap

state-chain/traits/src/mocks/block_height_provider.rs
```diff
@@ -19,6 +19,13 @@ impl<C: Chain> BlockHeightProvider<C> {
 	pub fn set_block_height(height: C::ChainBlockNumber) {
 		Self::put_value(BLOCK_HEIGHT_KEY, height);
 	}
+
+	pub fn mutate_block_height<F>(f: F)
```
kylezs, line 23: we only ever use this to increment - we could just simplify and make this "increment" instead

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -965,22 +1006,69 @@ fn broadcast_can_be_aborted_due_to_time_out() {
...
+			let (broadcast_id, _) = start_mock_broadcast();
+			broadcast_ids.push(broadcast_id);
+			ready_to_abort_broadcast(broadcast_id);
+			(broadcast_ids, expiry)
+		})
+		.then_process_next_block()
+		.then_process_next_block()
+		// move external block height into the future past both timeouts
+		.then_execute_with_keep_context(|(_, expiry)| {
+			BlockHeightProvider::<MockEthereum>::set_block_height(expiry + 5)
```
kylezs, line 1037: how do we know 5 is past both timeouts? It's a magic number - we could tie to the values defined more closely so it's clearer

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: Yeah - To allow appending it might have to be a `Vec<(ChainBlockNumber, BroadcastId, ValidatorId)>`.

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -1191,25 +1287,30 @@ fn broadcast_retries_will_not_be_overwritten_during_safe_mode() {
 			assert_ok!(Broadcaster::transaction_failed(RuntimeOrigin::signed(0u64), broadcast_id));
 
 			// On safe mode next block, storage will be re-added to this target block.
-			let next_block = System::block_number() + 1u64;
-			target = next_block +
+			let statechain_next_block = System::block_number() + 1u64;
```
kylezs, line 1290: generally if there's a context where state chain and external chain blocks are used then we use "block" and "chain block" for state chain and external chain blocks respectively, so we should be consistent here

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -965,22 +1006,69 @@ fn broadcast_can_be_aborted_due_to_time_out() {
...
+		.execute_with(|| {
+			let mut broadcast_ids = Vec::new();
+			let (broadcast_id, _) = start_mock_broadcast();
+			broadcast_ids.push(broadcast_id);
+			ready_to_abort_broadcast(broadcast_id);
+
+			let expiry = BlockHeightProvider::<MockEthereum>::get_block_height()
+				.saturating_add(crate::BroadcastTimeout::<Test, Instance1>::get());
+
+			(broadcast_ids, expiry)
```
kylezs, line 1021: nit - no need to mut initialise and push here
```suggestion
			(vec![broadcast_id], expiry)
```

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -917,21 +953,23 @@ fn witness_broadcast(tx_out_id: [u8; 4]) {
 
 #[test]
 fn timed_out_broadcaster_are_reported() {
```
kylezs, line 955: while we're here...
```suggestion
fn timed_out_broadcasters_are_reported() {
```

kylezs: not strictly part of this PR so feel free to leave out (and maybe include in a separate PR) - but this test says that we report them... but then stops short of actually testing that - which we can fix, we do:
```
MockOffenceReporter::assert_reported(
			PalletOffence::FailedToBroadcastTransaction,
			failed_authorities,
		);
``` 
elsewhere in these tests - no reason we couldn't here.

state-chain/pallets/cf-broadcast/src/lib.rs
MxmUrw: Ok, it seems that @kylezs agrees with you on this, so I'm gonna try your suggestion.

state-chain/pallets/cf-broadcast/src/tests.rs
MxmUrw: Speaking of magic numbers, what do you think about these?

https://github.com/chainflip-io/chainflip-backend/blob/6a9eb233cedf771bef4c65f1242f0f3e356ea776/state-chain/pallets/cf-broadcast/src/tests.rs#L516-L526

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: The reason to not use a BTreeSet is that you can use StorageAppend to just `::append(..)` the value efficiently to the end of the Vec (ie. just write a single element). Otherwise you need to read, decode, update, encode, write the whole Vec.

state-chain/pallets/cf-broadcast/src/tests.rs
MxmUrw: Ok, changing. (The convention is quite confusing though, isn't it...)

kylezs: only when you're not used to it ;) - but yeah i agree, the options are a little limited without having huge type names etc.

state-chain/pallets/cf-broadcast/src/tests.rs
MxmUrw: Simply adding this into the last `then_execute_at_next_block` fails though. The failed broadcasters are appearently only reported once the transaction itself has been successful (in `egress_success`). So if you want to add this I guess there should be some steps more happening, and would be better in a new PR.

state-chain/pallets/cf-broadcast/src/migrations/initialize_broadcast_timeout_storage.rs
MxmUrw: Added comment noting this.

state-chain/traits/src/mocks/block_height_provider.rs
MxmUrw: Changed to `increment_block_height()`.

state-chain/pallets/cf-broadcast/src/tests.rs
kylezs: yeah makes sense to do separately

state-chain/pallets/cf-broadcast/src/tests.rs
MxmUrw: I computed the expiry of all timeouts with `std::cmp::max` now.

state-chain/pallets/cf-broadcast/src/tests.rs
MxmUrw: I guess then the new constant I added should be called `SafeModeChainBlockMargin` instead of `SafeModeBlockMarginForTargetChain`, right?

kylezs (approved):
nice 👌 just some minor comments

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -405,7 +409,27 @@ pub mod pallet {
...
+
+			let current_chain_block = T::ChainTracking::get_block_height();
+			let mut expiries = BTreeSet::new();
+
+			// take all broadcasts which have timed out. Since iterators break if we update
+			// the map during iteration, we do this in two steps: first collect all expired_keys
+			// and separately the expired values, after that iterate again to delete all collected
+			// keys from storage.
+			Timeouts::<T, I>::mutate(|timeouts| {
+				// NB: As per documentation, retain visits every element exactly once,
```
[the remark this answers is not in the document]
MxmUrw: Yes, sensible assumption I guess.

state-chain/pallets/cf-broadcast/src/tests.rs
```diff
@@ -965,22 +1026,69 @@ fn broadcast_can_be_aborted_due_to_time_out() {
...
+		.then_execute_with(|(mut broadcast_ids, expiry1)| {
+			let (broadcast_id, _) = start_mock_broadcast();
+			broadcast_ids.push(broadcast_id);
+			ready_to_abort_broadcast(broadcast_id);
+			let expiry2 = BlockHeightProvider::<MockEthereum>::get_block_height()
+				.saturating_add(crate::BroadcastTimeout::<Test, Instance1>::get());
+			(broadcast_ids, max(expiry1, expiry2))
+		})
+		.then_process_next_block()
+		.then_process_next_block()
```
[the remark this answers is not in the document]
MxmUrw: Historical/incidental. I removed one of them, test works fine.

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: Ok i see what you mean. tbh. I don't see a great reason to use a BTreeSet in storage at all. We need to iterate all the values anyway most of the time, and when we append we almost always append a single value (the only exception to this is during safe mode).

dandanlen (approved):
Apart from the somewhat needless use of a BTreeSet (https://github.com/chainflip-io/chainflip-backend/pull/5270#discussion_r1766491468) LGTM.

kylezs:
> Apart from the somewhat needless use of a BTreeSet (https://github.com/chainflip-io/chainflip-backend/pull/5270#discussion_r1766491468) LGTM.

@MxmUrw and I discussed this - it does compress storage in some cases, with the tradeoff being that it's not strictly necessary so could be confusing - and neither of us had a super strong opinion either way.

kylezs (commented):
Ah, we're also going to need migrations for Timeouts

dandanlen:
> > Apart from the somewhat needless use of a BTreeSet ([#5270 (comment)](https://github.com/chainflip-io/chainflip-backend/pull/5270#discussion_r1766491468)) LGTM.
> 
> @MxmUrw and I discussed this - it does compress storage in some cases, with the tradeoff being that it's not strictly necessary so could be confusing - and neither of us had a super strong opinion either way.

AFAICT The only case where it compresses storage is when we're in safe mode. Under normal operation it's always worse. 

In general I would rather we stick to using the simplest possible solutiions (in this case `Vec<(A,B,C)>`) unless there's a good reason to use a more complex option (`Vec<(A, BTreeSet<(B,C)>)>`) and in this case I don't see a good reason to add the extra complexity. But I don't mind merging this as is.

MxmUrw:
> AFAICT The only case where it compresses storage is when we're in safe mode. Under normal operation it's always worse.

Yes that's true. I didn't consider the fact that constructing (and accessing) a singleton `BTreeSet` is gonna have some overhead, but of course it would have. Changing.

state-chain/pallets/cf-broadcast/src/migrations/migrate_timeouts.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct Migration<T: Config<I>, I: 'static>(PhantomData<(T, I)>);
+
+impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
+	fn on_runtime_upgrade() -> Weight {
+		// Instead of trying to translate the previous timeout into external chain blocks,
+		// we simply reset the remaining timeout duration to the new `BroadcastTimeout` value.
+		let new_timeout = BlockHeightProvider::<T::TargetChain>::get_block_height() +
+			BroadcastTimeout::<T, I>::get();
+		let mut old_keys = Vec::new();
+		for (key, timeouts) in old::Timeouts::<T, I>::iter() {
```
kylezs, line 28: can use drain here

MxmUrw: Yes, nice.

dandanlen (approved):

state-chain/pallets/cf-broadcast/src/migrations/migrate_timeouts.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[cfg(test)]
+mod migration_tests {
+	#[test]
+	fn test_migration() {
+		use super::*;
+		use crate::mock::*;
+
+		new_test_ext().execute_with(|| {
```
kylezs, line 81: there's currently nothing to migrate here so this test isn't doing anything really

MxmUrw: Added some timeout values before the migration runs.

kylezs (approved):

PR #5275, opened 2024-09-17
chore: rename `inspect_storage` to `then_execute_with_keep_context`.

MxmUrw:
# Pull Request

Part of: PRO-989

## Checklist

## Summary

During #5270 I noticed that it would be nice to have a test utility function which works like `then_execute_with`, but instead of requiring the user to forward or construct a new context it simply retains the old one.

#### Non-Breaking changes


kylezs:
And as discussed "inspect" suggests strictly not mutating anything, but here, we're just saying that we're not mutating the context - but we may mutate something else

MxmUrw:
Renamed to `then_execute_with_keep_context` (removing the "and") to make it slightly easier to read, since the name is already long as is. Fine?

dandanlen (approved):
Still doen't quite roll off the tongue but ok :)

PR #5277, opened 2024-09-17
feat: add solana monitoring data

marcellorigotti:
# Pull Request

Closes: PRO-1658

## Checklist

## Summary

Add solana monitoring data to monitoring rpc:
- Available nonces/Unavailable nonces
- Broadcaster.incomingKeyAndBroadcastId() -> This is missing for all the chains
   - Plus the signature of the transaction for solana -> we need to check it didn't revert  
- Solana aggKey
- SolanaBroadcaster.currentOnChainKey -> Sol funds held here

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -8,6 +8,7 @@ use cf_chains::{
 	address::{ForeignChainAddressHumanreadable, ToHumanreadableAddress},
 	dot::PolkadotAccountId,
 	eth::Address as EthereumAddress,
+	sol::SolAddress,
 	Chain,
 };
 use cf_primitives::{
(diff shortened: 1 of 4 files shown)


dandanlen (commented):
Looks fine to me, i'd just like to understand better how this will be used. Let's chat tomorrow.

marcellorigotti:
> Looks fine to me, i'd just like to understand better how this will be used. Let's chat tomorrow.

These changes are also breaking, bringing the same problems as https://github.com/chainflip-io/chainflip-backend/pull/5279 does, correct?

dandanlen (approved):
To summarize: 

- New node with old runtime will be broken.
- Old node with new runtime will be ok.

So we can update the monitoring rpc node *after* the runtime upgrade and we should be ok. However @marcellorigotti is looking into how best to manage this so that we can update the node ahead of time.

PR #5278, opened 2024-09-18
feat: liveness electoral system

kylezs:
# Pull Request

Closes: PRO-1645

## Checklist

- [x] I am confident that the code works - tested on 3 node localnet, took down a node and waited for a liveness completion and saw the node be punished :)

## Summary

Generic liveness electoral system that could be used for all chains. It works by asking the engines to return the block hash for a particular block number in a range from the current block height to 100 blocks below the current block height - this is to provide some variation on slower chains such as bitcoin. If we were to use the chain height and we're doing liveness every minute, then we would be returning the same value which is not ideal.

We punish any nodes who:
a) voted for the non-consensus value
b) did not vote when there was a consensus value

We don't punish nodes in any other scenarios such as:
- no one voted
- votes but no consensus

The offence contains the `ForeignChain` to identify it against other instances of the liveness system in the future to help with alerts and general observability.

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
@@ -1,0 +1,183 @@
+use crate::{
+	electoral_system::{
+		AuthorityVoteOf, ConsensusVote, ConsensusVotes, ElectionIdentifierOf, ElectionReadAccess,
+		ElectionWriteAccess, ElectoralSystem, ElectoralWriteAccess, VotePropertiesOf,
+	},
+	vote_storage::{self, VoteStorage},
+	CorruptStorageError,
+};
+use cf_primitives::AuthorityCount;
+use cf_utilities::success_threshold_from_share_count;
+use frame_support::{
+	pallet_prelude::{MaybeSerializeDeserialize, Member},
+	Parameter,
+};
+
+use itertools::Itertools;
+use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
+
+pub struct Liveness<ChainBlockNumber, ChainBlockHash, BlockNumber, Hook, ValidatorId> {
+	_phantom: core::marker::PhantomData<(
+		ChainBlockNumber,
+		ChainBlockHash,
+		BlockNumber,
+		Hook,
+		ValidatorId,
+	)>,
+}
+
+pub trait OnCheckComplete<ValidatorId> {
+	fn on_check_complete(validator_ids: Vec<ValidatorId>);
+}
+
+impl<
+		ChainBlockNumber: Member + Parameter + Eq + From<u64> + Into<u64> + Copy,
+		ChainBlockHash: Member + Parameter + Eq + Ord,
+		BlockNumber: Member
+			+ Parameter
+			+ Eq
+			+ MaybeSerializeDeserialize
+			+ frame_support::sp_runtime::Saturating
+			+ Ord
+			+ Copy,
+		Hook: OnCheckComplete<ValidatorId> + 'static,
+		ValidatorId: Member + Parameter + Ord + MaybeSerializeDeserialize,
+	> ElectoralSystem for Liveness<ChainBlockNumber, ChainBlockHash, BlockNumber, Hook, ValidatorId>
+{
+	type ValidatorId = ValidatorId;
+	type ElectoralUnsynchronisedState = ();
+	type ElectoralUnsynchronisedStateMapKey = ();
+	type ElectoralUnsynchronisedStateMapValue = ();
+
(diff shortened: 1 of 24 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
```diff
@@ -0,0 +1,183 @@
...
+
+		let (consensus_validators, non_consensus_validators): (Vec<_>, Vec<_>) =
+			active_votes.into_iter().partition(|(_, validator_ids)| {
+				validator_ids.len() as AuthorityCount >= success_threshold
+			});
+
+		debug_assert!(
+			consensus_validators.len() <= 1,
+			"there should only be one group that gained consensus"
+		);
```
dandanlen, line 167: We know this is true because success_threshold is a majority if votes. So since BTreeMap is sorted, we could just take the last entry and check if it contains more than success_threshold votes.

state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
```diff
@@ -0,0 +1,183 @@
...
+		let success_threshold = success_threshold_from_share_count(num_authorities);
+
+		let mut active_votes = BTreeMap::new();
+		let mut non_voters = Vec::new();
+		for ConsensusVote { vote, validator_id } in consensus_votes.votes {
+			match vote {
+				Some(vote) => {
+					active_votes.entry(vote.1).or_insert_with(Vec::new).push(validator_id);
+				},
+				None => non_voters.push(validator_id),
```
dandanlen, line 155: Instead of maintaining a separate vec we could accumulate everything into a Map<Vote, Option<Voter>>?

state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
```diff
@@ -0,0 +1,183 @@
...
+	type ElectoralSettings = BlockNumber;
+	type ElectionIdentifierExtra = ();
+
+	// The external chain block number that the engines will get the hash for.
+	type ElectionProperties = ChainBlockNumber;
+
+	// The SC block number that we started the election at.
+	type ElectionState = BlockNumber;
+	type Vote = vote_storage::bitmap::Bitmap<ChainBlockHash>;
+	type Consensus = Vec<Self::ValidatorId>;
```
dandanlen, line 63: Does it make sense to use a BTreeSet to make it easier to convince ourselves there are no duplicate report? (And would allow us to simply assert equality in the tests instead of having to sort etc.)

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
```diff
@@ -96,16 +97,45 @@ where
...
+				assert_eq!(new_consensus, expected_consensus);
+			},
+		)
+	}
+
+	pub fn expect_consensus_by<C: FnOnce(Option<ES::Consensus>, Option<ES::Consensus>)>(
 		self,
 		mut consensus_votes: ConsensusVotes<ES>,
 		expected_consensus: Option<ES::Consensus>,
+		check: C,
```
dandanlen, line 138: This function signature is a bit odd: we pass in the expected_consensus, but we also pass in a closure that receives the expected consensus. 

We could instead just capture the expected consensus in the closure?

dandanlen (commented):
Some minor suggestions but LGTM 👌

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
kylezs: Just going to delete it since using a BTreeSet makes this redundant

state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
kylezs: Removed as part of making it a single map

dandanlen (approved):

state-chain/runtime/src/lib.rs
```diff
@@ -1264,6 +1265,7 @@ type MigrationsForV1_7 = (
 	VersionedMigration<pallet_cf_broadcast::Pallet<Runtime, PolkadotInstance>, NoopUpgrade, 8, 9>,
 	VersionedMigration<pallet_cf_broadcast::Pallet<Runtime, BitcoinInstance>, NoopUpgrade, 8, 9>,
 	VersionedMigration<pallet_cf_broadcast::Pallet<Runtime, ArbitrumInstance>, NoopUpgrade, 8, 9>,
+	LivenessSettingsMigration,
```
dandanlen, line 1268: I think this should be VersionedMigration otherwise it's not idempotent.

PR #5279, opened 2024-09-19
fix: expire all previous epochs

j4m1ef0rd:
# Pull Request

Closes: PRO-1613

## Checklist

## Summary

- Changed the `expire_epoch` function to recursively expire all previous epochs that had not expired yet in chronological order. This should avoid the bug stated in the issue.
	- Added a test to make sure it expires as expected.
- Renamed `BlocksPerEpoch` -> `EpochDuration`
	- Added a migration and migration test for the rename.

#### Non-Breaking changes

I think this is non breaking, i'll let you tag it @dandanlen

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -1013,7 +1013,14 @@ impl<T: Config> Pallet<T> {
 	}
 
 	fn expire_epoch(epoch: EpochIndex) -> Weight {
-		LastExpiredEpoch::<T>::set(epoch);
+		// Recursively expire all previous epochs that have not been expired yet. This is possible
+		// due to epoch duration being configurable.
+		let previous_epoch = epoch.saturating_sub(1);
+		if epoch > 0 && LastExpiredEpoch::<T>::get() != previous_epoch {
+			Self::expire_epoch(previous_epoch);
+		}
+
+		// Expire given epoch
 		let mut num_expired_authorities = 0;
 		for authority in EpochHistory::<T>::epoch_authorities(epoch).iter() {
 			num_expired_authorities += 1;
@@ -1023,6 +1030,7 @@ impl<T: Config> Pallet<T> {
 			}
 			T::Bonder::update_bond(authority, EpochHistory::<T>::active_bond(authority));
 		}
+		LastExpiredEpoch::<T>::set(epoch);
 		T::EpochTransitionHandler::on_expired_epoch(epoch);
 
 		let validators = HistoricalAuthorities::<T>::take(epoch);
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -379,8 +379,8 @@ pub mod pallet {
 			let mut weight = Weight::zero();
 
 			// Check expiry of epoch and store last expired.
-			if let Some(epoch_index) = EpochExpiries::<T>::take(block_number) {
-				weight.saturating_accrue(Self::expire_epoch(epoch_index));
+			if let Some(epoch_to_expire) = EpochExpiries::<T>::take(block_number) {
+				Self::expire_epochs_up_to(epoch_to_expire);
 			}
 
 			weight.saturating_accrue(Self::punish_missed_authorship_slots());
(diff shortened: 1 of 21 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1018,7 +1018,14 @@ impl<T: Config> Pallet<T> {
 	}
 
 	fn expire_epoch(epoch: EpochIndex) -> Weight {
-		LastExpiredEpoch::<T>::set(epoch);
+		// Recursively expire all previous epochs that have not been expired yet. This is possible
+		// due to epoch duration being configurable.
+		let previous_epoch = epoch.saturating_sub(1);
+		if epoch > 0 && LastExpiredEpoch::<T>::get() != previous_epoch {
+			Self::expire_epoch(previous_epoch);
+		}
```
dandanlen, line 1026: I don't see a good reason to use recursion here, much clearer to just iterate. It also breaks the semantics of the function, it's no longer 'expire epoch', it's 'expire this and any previous unexpired epoch'. So at the very least we should rename it. 

I think a better option is to take the responsibility of updating LastExpiredEpoch outside of this function to the caller.

state-chain/pallets/cf-validator/src/migrations/rename_blocks_per_epoch.rs
```diff
@@ -0,0 +1,65 @@
...
+
+// Rename BlocksPerEpoch -> EpochDuration
+impl<T: Config> OnRuntimeUpgrade for BlocksPerEpochMigration<T> {
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, DispatchError> {
+		Ok(old::BlocksPerEpoch::<T>::get().encode())
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		EpochDuration::<T>::put(old::BlocksPerEpoch::<T>::get());
```
dandanlen, line 26:
```suggestion
		EpochDuration::<T>::put(old::BlocksPerEpoch::<T>::take());
```

state-chain/pallets/cf-validator/src/migrations/rename_blocks_per_epoch.rs
```diff
@@ -0,0 +1,65 @@
...
+		Weight::zero()
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(state: Vec<u8>) -> Result<(), DispatchError> {
+		assert_eq!(
+			EpochDuration::<T>::get(),
+			BlockNumberFor::<T>::decode(&mut &state[..]).unwrap()
+		);
+		Ok(())
```
dandanlen, line 36:
```suggestion
			EpochDuration::<T>::get(),
			BlockNumberFor::<T>::decode(&mut &state[..]).unwrap()
		);
		assert!(!old::BlocksPerEpoch::<T>::exists());
		Ok(())
```

state-chain/runtime/src/monitoring_apis.rs
```diff
@@ -27,7 +27,7 @@ pub struct BtcUtxos {
 
 #[derive(Serialize, Deserialize, Encode, Decode, Eq, PartialEq, TypeInfo, Debug)]
 pub struct EpochState {
-	pub blocks_per_epoch: u32,
+	pub epoch_duration: u32,
```
dandanlen, line 30: Also a breaking change here, fyi @marcellorigotti

dandanlen (commented):
@j4m1ef0rd I'll make the below changes. 

Will wait for @acdibble and @marcellorigotti 's inputs re. api breakage before merging.

state-chain/runtime/src/monitoring_apis.rs
marcellorigotti: Does this mean that after the next binary release until we run the runtime-upgrade this rpc won't be available?
Or will it still work right away but with the new field name (blocks_per_epoch -> epoch_duration)?

dandanlen: Yeah it would mean that the binary release needs to be coordinated with the runtime upgrade. I think a better option would be to add a new field with the new name, and remove it in the following release (1.7 or 1.8).

marcellorigotti: I agree with the other option. If this RPC doesn't work then monitoring doesn't work as well (the majority of it at least)

kylezs:
What's the status on this one?

dandanlen (approved):
Conflicts need to be resolved, then it should be good.

kylezs:
Looks like it's hitting a migration issue @j4m1ef0rd

state-chain/pallets/cf-validator/src/migrations.rs
```diff
@@ -2,8 +2,10 @@ use crate::Pallet;
 use cf_runtime_upgrade_utilities::{PlaceholderMigration, VersionedMigration};
 
 mod delete_old_epoch_data;
+mod rename_blocks_per_epoch;
 
 pub type PalletMigration<T> = (
 	VersionedMigration<Pallet<T>, delete_old_epoch_data::Migration<T>, 3, 4>,
-	PlaceholderMigration<Pallet<T>, 4>,
+	VersionedMigration<Pallet<T>, rename_blocks_per_epoch::BlocksPerEpochMigration<T>, 3, 5>,
```
dandanlen, line 9: I think it's this (my fault):

```suggestion
	VersionedMigration<Pallet<T>, rename_blocks_per_epoch::BlocksPerEpochMigration<T>, 4, 5>,
```

PR #5281, opened 2024-09-19
PRO-1620: Ensure that default port is used if none is given in configuration

MxmUrw:
# Pull Request

Closes: PRO-1620

## Checklist

## Summary

Currently the polkadot http rpc client uses jsonrpsee's HttpClient. Unfortunately, since we are on version 0.16 of jsonrpsee, this client has the rule that a port must always be given, even if it should be the default port. Later versions of jsonrpsee forgo this requirement.

To make sure that the user configures a valid http endpoint for polkadot, we currently check that the url contains a port. Unfortunately the checking logic uses a regex and accepts urls such as `https://endpoint/path:port` as valid, even though it shouldn't.

It also seems that the Solana configuration logic mistakenly was copied from Polkadot since it calls the same validation logic, without there being a need for that.

This PR does multiple things:
 - [x] Instead of requiring the user to always give a port for polkadot endpoints, we automatically insert the default port in case none was given.
 - [x] Remove tests which check that polkadot endpoints always contain ports.
 - [x] Remove the port validation logic for solana.
 
#### Non-Breaking changes

### DIFF
--- a/engine/src/dot/http_rpc.rs
+++ b/engine/src/dot/http_rpc.rs
@@ -82,7 +82,7 @@ impl RpcClientT for PolkadotHttpClient {
 fn ensure_port(url: SecretUrl) -> Result<SecretUrl> {
 	// we use url::Url to get the default port for our scheme
 	let targetport = Url::parse(url.as_ref())
-		.unwrap() // We know that `url` is valid
+		.expect("SecretUrl was validated by being passed into `Url::parse`.")
 		.port_or_known_default()
 		.ok_or(anyhow::anyhow!("Unknown scheme and no port given."))?;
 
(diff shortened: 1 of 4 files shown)


MxmUrw:
Currently open question: How about solana port validation?

engine/src/dot/http_rpc.rs
```diff
@@ -70,6 +72,46 @@ impl RpcClientT for PolkadotHttpClient {
...
+	// We use http::uri::Uri to ensure that the default port is added if none exists
+	// We split the uri into components, insert the port, and reconstruct.
+	let mut parts = url
+		.as_ref()
+		.parse::<Uri>()
+		.unwrap() // We know that `url` is valid
+		.into_parts();
+
+	// Update the authority part of the uri by mapping over it.
+	let authority = parts.authority.unwrap();
```
kylezs, line 98: Should throw error instead of panicking here, or expect() with a reason why it's safe
```suggestion
	let authority = parts.authority.unwrap();
```

engine/src/dot/http_rpc.rs
```diff
@@ -70,6 +72,46 @@ impl RpcClientT for PolkadotHttpClient {
...
+			})?)
+		} else {
+			Some(authority)
+		};
+
+	// Reconstruct uri.
+	let uri = Uri::from_parts(parts)
+		.map_err(|err| anyhow!("Unexpected error when trying to append port to url: {err}"))?;
+
+	Ok(format!("{uri}").into())
```
kylezs, line 112:
```suggestion
	Uri::from_parts(parts)
		.map_err(|err| anyhow!("Unexpected error when trying to append port to url: {err}"))
		.map(|uri| uri.to_string().into())
```

engine/src/dot/http_rpc.rs
```diff
@@ -70,6 +72,46 @@ impl RpcClientT for PolkadotHttpClient {
...
+	// Update the authority part of the uri by mapping over it.
+	let authority = parts.authority.unwrap();
+	parts.authority =
+		if authority.port().is_none() {
+			Some(format!("{}:{}", authority.as_str(), targetport).parse().map_err(|err| {
+				anyhow!("Unexpected error when trying to append port to url: {err}")
+			})?)
+		} else {
+			Some(authority)
+		};
```
kylezs, line 106: can dedup the Some()
```suggestion
	parts.authority = Some(if authority.port().is_none() {
		format!("{}:{}", authority.as_str(), targetport)
			.parse()
			.map_err(|err| anyhow!("Unexpected error when trying to append port to url: {err}"))?
	} else {
		authority
	});
```

engine/src/dot/http_rpc.rs
MxmUrw: Yes, `expect` makes sense. Updated all occurences of unwrap.

engine/src/dot/http_rpc.rs
MxmUrw: Nice, done.

engine/src/dot/http_rpc.rs
MxmUrw: Rewrote execution flow.

kylezs (approved):

PR #5282, opened 2024-09-19
PRO-1594: Add healthcheck endpoints to broker and lp apis

MxmUrw:
# Pull Request

Closes: PRO-1594

## Checklist

## Summary

Currently, only the engine binary itself has a healthcheck endpoint. This PR adds the same functionality to the broker and lp api binaries. In particular, the `healthcheck.rs` file is moved into the `utilities` crate to allow all binaries to reuse it.

#### Non-Breaking changes


api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -148,6 +168,17 @@ async fn main() -> anyhow::Result<()> {
 
 	task_scope(|scope| {
 		async move {
+			// initialize healthcheck endpoint
+			let has_completed_initialising = Arc::new(AtomicBool::new(false));
+			if opts.health_check_hostname.is_some() || opts.health_check_port.is_some() {
+				let error_msg = "Clap enforces that both health_check.hostname and health_check.port are present.";
+				let h = HealthCheck {
```
kylezs, line 175: can just inline h, as a variable name doesn't add much

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -619,6 +625,20 @@ pub struct LPOptions {
...
+		requires("health_check.port")
+	)]
+	pub health_check_hostname: Option<String>,
+	#[clap(
+		id = "health_check.port",
+		long = "health_check.port",
+		help = "Port for this LP server's healthcheck. Requires --health_check.hostname to be present as well.",
+		requires("health_check.hostname")
+	)]
+	pub health_check_port: Option<u16>,
```
kylezs, line 641: Seems like we could dedupe this (would just require a more general log message): https://docs.rs/clap/3.2.16/clap/trait.Args.html#example

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -638,12 +658,26 @@ async fn main() -> anyhow::Result<()> {
 
 	task_scope(|scope| {
 		async move {
+			// initialize healthcheck endpoint
+			let has_completed_initialising = Arc::new(AtomicBool::new(false));
+			if opts.health_check_hostname.is_some() || opts.health_check_port.is_some() {
```
kylezs, line 663: we could also deduplicate too this by moving it into a function that wraps the healthcheck

api/bin/chainflip-lp-api/src/main.rs
MxmUrw: Done.

api/bin/chainflip-lp-api/src/main.rs
MxmUrw: Also done.

utilities/src/with_std/health.rs
```diff
@@ -6,10 +6,29 @@
...
 use serde::Deserialize;
 use tracing::info;
 use warp::Filter;
 
+#[derive(Args, Debug, Clone, Default)]
+pub struct HealthCheckOptions {
+	#[clap(
+		id = "HEALTH_CHECK_HOSTNAME",
+		long = "health_check.hostname",
+		help = "Hostname for this server's healthcheck. Requires HEALTH_CHECK_PORT to be set as well.",
```
kylezs, line 19: `health_check.port` instead of HEALTH_CHECK_PORT - since the user won't see this id - they need the id they can use, same below for hostname

kylezs (approved):
👌

CumpsD:
Yay, no more TCP healthcheck :)

MxmUrw:
> Yay, no more TCP healthcheck :)

Yes!

PR #5285, opened 2024-09-20
feat: remove swap and retry batch on price impact

msgmaxim:
# Pull Request

Closes: PRO-1440

## Checklist

## Summary

Moved the retry loop from `on_finalize` (where we only retried on FoK failures) to `execute_batch` (where now we retry on both price impact *and* FoK failures). In order to determine which swap to remove in case of "price impact" we compare `swap_amount` (which may be input amount, or may be the result of the first leg) and remove the largest; for this I added swap_states to `SwapLegFailed`.

One thing to note is that now `BatchSwapFailed` may be emitted more than once per block due to retrying. I think it makes sense: each error would refer to a different batch, potentially to different pools even, but it also might not be ideal if the same error is repeated many times with consequent errors not providing much additional information.


dandanlen (commented):
Looks good - I want to take another look on Monday but I don't expect any major comments.

dandanlen (commented):
I was wrong - I have some comments :)

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1217,31 +1196,95 @@ pub mod pallet {
...
+
+						if violating.is_empty() {
+							break Ok(BatchExecutionOutcomes {
+								successful_swaps: non_violating,
+								failed_swaps: removed_swaps,
+							});
+						} else {
+							swaps = non_violating.into_iter().map(|ctx| ctx.swap).collect();
+							removed_swaps.extend(violating.into_iter().map(|ctx| ctx.swap));
+						}
```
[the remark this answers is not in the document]
msgmaxim: Indeed. I did have a test that we revert storage on price impact, but not on FoK failures... I ended up moving FoK checks inside `try_execute_without_violations` which can now return both types of failures. I think the code looks cleaner now too. I also added `storage_state_rolls_back_on_fok_violation` to test the desired behaviour.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1217,31 +1196,95 @@ pub mod pallet {
...
+						// empty, abort; otherwise, remove the swap and try again with the
+						// remaining swaps:
+						match swap_to_remove {
+							Some(swap_to_remove) if !remaining_swaps.is_empty() => {
+								removed_swaps.push(swap_to_remove);
+								swaps = remaining_swaps;
+							},
+							_ => {
+								if let Some(swap_to_remove) = swap_to_remove {
+									removed_swaps.push(swap_to_remove);
```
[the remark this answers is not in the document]
msgmaxim: I moved `if remaining_swaps.is_empty()` right after `loop {` so that it is always checked (and prevents us from calling functions unnecessarily when we have 0 swaps to begin with). Looks much cleaner now.

state-chain/pallets/cf-swapping/src/mock.rs
```diff
@@ -75,6 +76,7 @@ parameter_types! {
 	pub static NetworkFee: Permill = Permill::from_perthousand(0);
 	pub static Swaps: Vec<(Asset, Asset, AssetAmount)> = vec![];
 	pub static SwapRate: f64 = DEFAULT_SWAP_RATE as f64;
+	pub static Liquidity: HashMap<Asset, AssetAmount> = HashMap::default();
```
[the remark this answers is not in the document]
msgmaxim: Updated to use storage (also now check that it is rolled back in one of the tests).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1638,9 +1681,14 @@ pub mod pallet {
...
-			for (asset, swaps) in swap_groups {
-				Self::execute_group_of_swaps(swaps, asset, direction).map_err(|amount| {
-					BatchExecutionError::SwapLegFailed { asset, direction, amount }
+			for (asset, mut swaps) in swap_groups {
+				Self::execute_group_of_swaps(&mut swaps, asset, direction).map_err(|amount| {
+					BatchExecutionError::SwapLegFailed {
+						asset,
+						direction,
+						amount,
+						swaps: swaps.into_iter().map(|swap| swap.clone()).collect(),
```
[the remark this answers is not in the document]
msgmaxim: updated

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2266,4 +2314,28 @@ pub(crate) mod utilities {
 			.unwrap_or(u128::MAX),
 		}
 	}
+
+	pub(super) fn split_off_highest_impact_swap<T: Config>(
+		mut swaps: Vec<Swap<T>>,
+		swap_states: &[SwapState<T>],
+		direction: SwapLeg,
+	) -> (Option<Swap<T>>, Vec<Swap<T>>) {
```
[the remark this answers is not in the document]
msgmaxim: > mutate the Vec in-place

updated

> In this function, there are some key invariants that we can enforce/assert on

I will add debug asserts for these. However, I'm not sure whether we should change the signature: that would mean asserting in release too, and I thought we aim to avoid runtime panics. It is one thing to assert on something locally obvious, but in this case there is enough "action at a distance" that I'd rather have a fallback way of handling a scenario that we think is impossible now.

Note that if this invariant is violated doesn't (necessarily) mean that something has gone so wrong that panicking is the lesser evil. Far more likely would be that we made a mistake in `do_group_and_swap` (or some other similar code that we will write in the future) where `SwapLegFailed` is incorrectly reported and a totally reasonable/desirable way to handle this would be to simply give up on the entire batch (which is what I'm doing here).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1238,13 +1222,73 @@ pub mod pallet {
...
+						asset,
+						direction,
+						amount,
+						failed_swap_group,
+					}) => {
+						Self::deposit_event(Event::<T>::BatchSwapFailed {
+							asset,
+							direction,
+							amount,
+						});
```
dandanlen, line 1259: It might we worth adjusting how we communicate this: the swaps have not longer actually failed, they might be retried. Maybe we don't need this event at all any more?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1238,13 +1222,73 @@ pub mod pallet {
...
+
+						// Find the largest swap from the failing pool/direction:
+						let swap_to_remove = utilities::split_off_highest_impact_swap(
+							&mut swaps_to_execute,
+							&failed_swap_group,
+							direction,
+						);
+
+						// Abort if we couldn't find a swap to remove; otherwise, remove the
+						// swap and try again with the remaining swaps:
```
dandanlen, line 1269: This comment is not quite accurate any more / a bit confusing.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1238,13 +1222,73 @@ pub mod pallet {
...
+						}
+					},
+					Err(BatchExecutionError::PriceViolation {
+						violating_swaps,
+						non_violating_swaps,
+					}) => {
+						failed_swaps.extend(violating_swaps);
+						swaps_to_execute = non_violating_swaps;
+					},
+					Err(BatchExecutionError::DispatchError { error }) => return Err(error),
```
dandanlen, line 1287: This is the only error condition in this loop, and it can only happen if we exceed the the transaction depth limit, right? So am I right in thinking that this function (execute_batch) doesn't need to be transactional (since it can never really error, and never needs to roll back)? 

ie. instead of passing this error back up the chain, we could just log it at this level.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2266,4 +2315,40 @@ pub(crate) mod utilities {
 			.unwrap_or(u128::MAX),
 		}
 	}
+
+	// Note:
```
dandanlen, line 2319: Missing note :)

state-chain/pallets/cf-swapping/src/mock.rs
```diff
@@ -91,6 +92,20 @@ impl MockSwappingApi {
 	fn swaps_should_fail() -> bool {
 		SWAPS_SHOULD_FAIL.with(|cell| cell.get())
 	}
+
+	pub fn add_liquidity(asset: Asset, amount: AssetAmount) {
+		let liquidity = Liquidity::get()
```
dandanlen, line 97: I was thinking there might be a simpler way to simulate this. We don't need to add amounts etc, we could just configure the mock to behave *as if* there is no liquidity in a certain pool (ie. instead of a liquidity amount we just have a bool).

This would avoid having to calculate/reason about how much input is required at current prices to make the test fail etc.

dandanlen (commented):
Looking good now. I'll give it another pass and try to get it merged tomorrow.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Right, I changed this to never return BatchExecutionOutcomes directly now.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Don't remember what I wanted to say, so I just deleted it now :)

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Combined with the expression above and updated the comment.

state-chain/pallets/cf-swapping/src/mock.rs
msgmaxim: I almost ended up making this change, but then ran into the `price_impact_removes_one_swap` test where we actually want low liquidity, but enough for *one* swap to test the retrying works (and even the swaps from the failing pool/side can be retired). Because retrying is done entirely within a single call to on_finalize there is no way for us to change liquidity between iteration as we remove swaps, so we need a bit more control in the mock's liquidity than to just say "all swaps in the pool should fail".

Also, I thought it was nice that we could check the resulting liquidity in presence of rollbacks (e.g. in `storage_state_rolls_back_on_fok_violation`), which I thought was the whole point of marking `Liquidity` as `storage`.

dandanlen (commented):
I thought we wanted to remove the BatchSwapFailed event?

state-chain/cf-integration-tests/src/swapping.rs
```diff
@@ -660,64 +666,31 @@ fn failed_swaps_are_rolled_back() {
...
-		assert_eq!(
-			Some(eth_price),
-			LiquidityPools::current_price(Asset::Eth, STABLE_ASSET)
-				.map(|pool_price| pool_price.price)
-		);
 		assert_eq!(
-			Some(btc_price),
-			LiquidityPools::current_price(Asset::Btc, STABLE_ASSET)
+			Some(flip_price),
+			LiquidityPools::current_price(Asset::Flip, STABLE_ASSET)
```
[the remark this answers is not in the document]
msgmaxim: This is just what the existing test was checking (and presumably the reason why they used a range order). Now replaced this with checking the full state of pools.

msgmaxim:
> I thought we wanted to remove the BatchSwapFailed event?

Maybe. Though it also sounded like we wanted to at least replace it with another way to tell that a swap failed due to low liquidity (for example, by adding a field to `SwapRescheduled`). It seemed more appropriate to leave it for another PR.

dandanlen (approved):

dandanlen:
Ok I'll open an issue to follow this up.

PR #5286, opened 2024-09-20
fix: scale types for pending ceremonies

martin-chainflip:
# Pull Request

Closes: PRO-1676

## Checklist

## Summary

This change renames the types returned by the ThresholdSignature -> pendingCeremonies call so that different types have different names.


state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -364,6 +363,8 @@ pub mod pallet {
 		/// A marker trait identifying the chain that we are signing for.
 		type TargetChainCrypto: ChainCrypto;
 
+		type TargetChain: Chain;
+
```
dandanlen, line 367:
```suggestion
		/// A marker trait identifying the chain that we are signing for.
		type TargetChainCrypto: ChainCrypto;

		type TargetChain: Chain<ChainCrypto = Self::TargetChainCrypto>;

```

Just to make sure can't do something stupid...

state-chain/runtime/src/lib.rs
```diff
@@ -788,6 +788,7 @@ impl pallet_cf_threshold_signature::Config<Instance16> for Runtime {
 	type ThresholdCallable = RuntimeCall;
 	type ThresholdSignerNomination = chainflip::RandomSignerNomination;
 	type TargetChainCrypto = EvmCrypto;
+	type TargetChain = Arbitrum;
```
dandanlen, line 791: What about Ehthereum?

state-chain/pallets/cf-threshold-signature/src/lib.rs
dandanlen: Actually I'm not sure we need (or *can* provide) the target chain here. Multiple chains can use the same instance of this pallet.

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -1199,6 +1200,66 @@ pub mod pallet {
...
+macro_rules! append_chain_to_name {
+	($name:ident) => {
+		match T::TargetChain::NAME {
+			"Ethereum" => concat!(stringify!($name), "Ethereum"),
+			"Polkadot" => concat!(stringify!($name), "Polkadot"),
+			"Bitcoin" => concat!(stringify!($name), "Bitcoin"),
+			"Arbitrum" => concat!(stringify!($name), "Arbitrum"),
+			"Solana" => concat!(stringify!($name), "Solana"),
+			_ => concat!(stringify!($name), "Other"),
+		}
```
dandanlen, line 1212:
```suggestion
		concat!(stringify!($name), T::TargetChain::NAME)
```

But as mentioned elsewhere it will have to be ChainCrypto not Chain.

state-chain/pallets/cf-threshold-signature/src/lib.rs
martin-chainflip: that's fine, though. If it is the same pallet, then it will also have the same types and the same serialization. All we need is for different instances of this generic pallet to have different names. The most straightforward way would be to use the NAME parameter of the 'Chain' type (ChainCrypto does not provide a name)

state-chain/runtime/src/lib.rs
martin-chainflip: We don't instantiate `pallet_cf_threshold_signature::Config` for `Instance1` anywhere.

dandanlen: Exactly: it's instantiated for the EVM, it's neither Arbitrum nor Ethereum.

state-chain/pallets/cf-threshold-signature/src/lib.rs
dandanlen: Then we should add a name? Otherwise it's just confusing. We will see 'Arbitrum' even when it's Ethereum. If we're going to add names, we might as well make them accurate?

dandanlen (commented):
TLDR - I think we should add a name to the ChainCrypto trait instead of just using Arbitrum.

dandanlen:
In fact it doesn't even have to be a NAME string, it just has to be something we can match on.

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -1199,6 +1197,65 @@ pub mod pallet {
...
+{
+	type Identity = Self;
+	fn type_info() -> Type {
+		Type::builder()
+			.path(Path::new(append_chain_to_name!(RequestContext), module_path!()))
+			.composite(
+				Fields::named()
+					.field(|f| f.ty::<RequestId>().name("request_id"))
+					.field(|f| f.ty::<AttemptCount>().name("attempt_count"))
+					.field(|f| f.ty::<PayloadFor<T, I>>().name(append_chain_to_name!(PayloadFor))),
```
dandanlen, line 1254: Sorry, another comment... 

I don't understand why we need to use this `append_chain_to_name` on the field names... surely it's only needed on the path? 

If we want to resolve the type `PayloadFor<T, I>` then we need to apply the same treatment to the `PayloadFor<T, I>` *types* and somehow direct the type_info to resolve that type.

martin-chainflip: The only way to direct type_info on how to resolve a type is via its name. If we don't append the chain name to PayloadFor, then there will be multiple types called "PayloadFor" and the recursive type resolution will fail again.

martin-chainflip: my previous message was not correct. I added separate "field names" and "type names" now

dandanlen (approved):

PR #5294, opened 2024-09-25
refactor: Strongly-typed identifiers for SwapId/SwapRequestId

MxmUrw:
# Pull Request

Closes: PRO-1548

## Checklist

## Summary

Previously, `SwapId` and `SwapRequestId` were both type aliases for `u64`, making them easy to accidentally misuse. This PR changes them to be struct wrappers around `u64`, generated with a new macro, `define_wrapper_type`.

The macro implements `From<$inner>` for the wrapper struct, such that strongly-typed ids can now be created by, e.g., `my_u64.into()`. It also implements the `Deref` and `DerefMut` traits to access the underlying `u64` value.

One particular problem during implementation was that multiple test files in the swapping pallet use macros such as `assert_event_sequence` which expect patterns to be passed in. Previously, literal values for both SwapId and SwapRequestId were used in such calls. In order to preserve such a usage pattern, the **constructor for the new struct wrappers have to be public**. This is because calls such as `.into()` are not valid in patterns.

#### Non-Breaking changes

### DIFF
--- a/state-chain/primitives/src/lib.rs
+++ b/state-chain/primitives/src/lib.rs
@@ -21,7 +21,7 @@ pub mod chains;
 
 #[macro_export]
 macro_rules! define_wrapper_type {
-	($name: ident, $inner: ty, extra_derives: $( $extra_derive: ident ),* ) => {
+	($name: ident, $inner: ty $(, extra_derives: $( $extra_derive: ident ),*)? ) => {
 
 		#[derive(
 			Clone,
@@ -34,7 +34,7 @@ macro_rules! define_wrapper_type {
 			TypeInfo,
 			MaxEncodedLen,
 			Default,
-			$( $extra_derive ),*
+			$($( $extra_derive ),*)?
 		)]
 		pub struct $name(pub $inner);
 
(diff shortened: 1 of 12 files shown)


state-chain/primitives/src/lib.rs
```diff
@@ -13,10 +13,58 @@ use serde::{Deserialize, Serialize};
...
+			RuntimeDebug,
+			PartialEq,
+			Eq,
+			Encode,
+			Decode,
+			TypeInfo,
+			MaxEncodedLen,
+			Default,
+			serde::Serialize,
+			serde::Deserialize,
```
dandanlen, line 36: FWIW this might cause issues in pallets/crates that don't include serde. Maybe we can let users add their own?

Not sure this works?

```rust
#[derive(X)]
define_wrapper_type!(SwapId, u64);
```

Might have to be:

```rust
define_wrapper_type!(SwapId, u64, extra_derives: X, Y, Z);
```

state-chain/primitives/src/lib.rs
```diff
@@ -63,6 +64,11 @@ macro_rules! define_wrapper_type {
 			}
 		}
 	};
+
+	($name: ident, $inner: ty) => {
+		$crate::define_wrapper_type!($name, $inner, extra_derives:)
```
dandanlen, line 69: This needs a trailing `;`.

Alternatively you can just make `extra_derives` optional and remove this second case:

```
	( $name: ident, $inner: ty $(, extra_derives: $(, $extra_derive: ident ),* )? )` 
```
(etc.)

dandanlen (approved):

PR #5297, opened 2024-09-25
feat: Submit a slot number alongside nonce

marcellorigotti:
# Pull Request

Closes: PRO-1635

## Checklist

## Summary

Updated the Change Electoral system to Nonce Electoral system, which takes both nonce value and slot as votes and update the nonce to the new value only if the new value is different than the previous one and the slot is higher than the previous one.


engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -9,17 +9,19 @@ use crate::sol::{
...
+use pallet_cf_elections::vote_storage::nonce::NonceVote;
 use serde_json::Value;
+use sol_prim::SlotNumber;
 use std::str::FromStr;
-
 pub async fn get_durable_nonce<SolRetryRpcClient>(
 	sol_client: &SolRetryRpcClient,
 	nonce_account: SolAddress,
-) -> Result<Option<SolHash>>
+	previous_nonce: SlotNumber,
```
albert-llimos, line 19: This should be `previous_slot`

engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -29,10 +31,11 @@ where
 				encoding: Some(UiAccountEncoding::JsonParsed),
 				data_slice: None,
 				commitment: Some(CommitmentConfig::finalized()),
-				min_context_slot: None,
+				min_context_slot: Some(previous_nonce),
```
albert-llimos, line 34: Same here -> `Some(previous_slot)`

dandanlen (commented):
Looks good to me, just some minor comments mostly around naming. I'll take a closer look at tests on Monday.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -758,3 +758,11 @@ impl<T: Config> NetworkEnvironmentProvider for Pallet<T> {
 		Self::network_environment()
 	}
 }
+
+pub struct NoUsedNonce<T: Config>(PhantomData<T>);
+
+impl<T: Config> ExecutionCondition for NoUsedNonce<T> {
+	fn is_satisfied() -> bool {
+		SolanaAvailableNonceAccounts::<T>::get().len() == 10
```
dandanlen, line 766: I thought it was 7 😅 

Could we not use `SolanaUnavailableNonceAccounts::<T>::iter().next().is_none()`?

albert-llimos: It was originally 7 indeed. However, I did some optimizations to be able to bump it up to ten, as that is the bottleneck.

engine/src/witness/sol.rs
```diff
@@ -110,12 +111,13 @@ impl VoterApi<SolanaNonceTracking> for SolanaNonceTrackingVoter {
...
-			.await?
-			// If the nonce is not found, we default to the previous nonce.
-			// The `Change` electoral system ensure this vote is filtered.
-			.unwrap_or(previous_nonce))
+		let (nonce_account, previous_nonce, previous_slot) = properties;
+
+		let nonce_and_slot =
+			nonce_witnessing::get_durable_nonce(&self.client, nonce_account, previous_slot).await?;
+		// If the nonce is not found, we default to the previous nonce and slot.
+		// The `Nonce` electoral system ensure this vote is filtered.
```
kylezs, line 119:
```suggestion
		// The `NonceWitnessing` electoral system ensure this vote is filtered.
```

state-chain/pallets/cf-elections/src/electoral_systems/nonce_wintessing.rs
```diff
@@ -34,51 +34,46 @@ pub trait OnChangeHook<Identifier, Value> {
 impl<
 		Identifier: Member + Parameter + Ord,
 		Value: Member + Parameter + Eq + Ord,
+		Slot: Member + Parameter + Eq + Ord + Copy + Default,
 		Settings: Member + Parameter + MaybeSerializeDeserialize + Eq,
 		Hook: OnChangeHook<Identifier, Value> + 'static,
 		ValidatorId: Member + Parameter + Ord + MaybeSerializeDeserialize,
-	> Change<Identifier, Value, Settings, Hook, ValidatorId>
+	> NonceWitnessing<Identifier, Value, Slot, Settings, Hook, ValidatorId>
```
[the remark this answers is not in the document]
kylezs: imo, making the names more specific is better in this case

state-chain/pallets/cf-elections/src/electoral_systems/nonce_wintessing.rs
```diff
@@ -107,18 +102,35 @@ impl<
...
 		_context: &Self::OnFinalizeContext,
 	) -> Result<Self::OnFinalizeReturn, CorruptStorageError> {
 		for election_identifier in election_identifiers {
 			let mut election_access = electoral_access.election_mut(election_identifier)?;
-			if let Some(value) = election_access.check_consensus()?.has_consensus() {
-				let (identifier, previous_value) = election_access.properties()?;
-				if previous_value != value {
+			if let Some((value, slot)) = election_access.check_consensus()?.has_consensus() {
+				let (identifier, previous_value, previous_slot) = election_access.properties()?;
+				if previous_value != value && slot > previous_slot {
```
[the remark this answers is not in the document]
kylezs: Here we have the on-chain logic, which should prevent it on its own, without having to rely on correct behaviour from the engines. So I don't mind having this here as a way of ensuring the chain logic is valid independent of engine behaviour - though I do think there should be a comment here that we don't expect this to be false because in general the engines should behave.

state-chain/pallets/cf-elections/src/electoral_systems/tests/nonce_witnessing.rs
```diff
@@ -0,0 +1,214 @@
...
+	}
+}
+
+type Vote = NonceVote<Value, Slot>;
+type Value = u64;
+type Slot = u32;
+type SimpleChange = NonceWitnessing<(), Value, Slot, (), MockHook, ()>;
+
+register_checks! {
+	SimpleChange {
```
kylezs, line 33:
```suggestion
	SimpleNonceWitnessing {
```

state-chain/pallets/cf-elections/src/electoral_systems/tests/nonce_witnessing.rs
```diff
@@ -0,0 +1,214 @@
...
+			}))
+			.chain(
+				(0..AUTHORITY_COUNT - correct_voters - incorrect_voters)
+					.map(|_| ConsensusVote { vote: None, validator_id: () }),
+			)
+			.collect(),
+	}
+}
+
+fn generate_votes_with_differen_slots(
```
kylezs, line 76:
```suggestion
fn generate_votes_with_different_slots(
```

state-chain/pallets/cf-elections/src/vote_storage/nonce.rs
```diff
@@ -0,0 +1,82 @@
...
+	pallet_prelude::{Member, TypeInfo},
+	Parameter,
+};
+
+use super::{AuthorityVote, VoteComponents, VoteStorage};
+
+use crate::{CorruptStorageError, SharedDataHash};
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode, TypeInfo)]
+pub struct NonceVote<Value, Slot> {
```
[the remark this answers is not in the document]
kylezs: I think we can generalise it when it necessary? else it's a little confusing it has a general name and is used for one (quite specific) thing

state-chain/pallets/cf-elections/src/electoral_systems/nonce_wintessing.rs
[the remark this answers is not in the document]
dandanlen: Why?

state-chain/pallets/cf-elections/src/vote_storage/nonce.rs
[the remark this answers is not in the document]
dandanlen: My view is that if we implement something generically, then it should have generic names. Reading generic code with specifically-named values is confusing because you always have the specific use-case in mind.

state-chain/pallets/cf-elections/src/electoral_systems/nonce_wintessing.rs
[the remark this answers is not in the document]
dandanlen: Yes I'm saying we remove it, just that add something so that we don't silently fail if the assumption is wrong.

state-chain/pallets/cf-environment/src/lib.rs
dandanlen: Sorry, should been less indirect: we can't assume that it will always be 10. We've already change it once, we might change it again. So the implementation should not assume the number.

marcellorigotti:
I kept everything generic in the end!

state-chain/pallets/cf-elections/src/electoral_systems/nonce_wintessing.rs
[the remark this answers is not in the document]
kylezs: yeah i don't think it matters much, but unless it has a general use better to name it specific. Like we don't name every parameter "variable", the name becomes more general as what it is actually doing becomes more general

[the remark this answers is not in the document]
dandanlen: Yes, but in this case, we're implementing a type with a bunch of generics, so it's pretty clear that it's supposed to work generically. And of course I'm not saying we should name everything `variable` and `thing`. What I mean is that naming should be as specific as possible within its own context. In this particular case, for example, there's nothing in the functionality of the implementation that is nonce-specific.

state-chain/pallets/cf-elections/src/vote_storage/change.rs
```diff
@@ -0,0 +1,85 @@
...
+	type PartialVote = MonotonicChangeVote<T, S>;
+	type IndividualComponent = S;
+	type BitmapComponent = T;
+	type SharedData = ();
+
+	fn vote_into_partial_vote<H: FnMut(Self::SharedData) -> SharedDataHash>(
+		vote: &Self::Vote,
+		mut _h: H,
+	) -> Self::PartialVote {
+		(*vote).clone()
```
kylezs, line 31: This means we don't have a partial vote, and we always vote the full vote. The plan is to extend this electoral system for the contract witnessing, and so there the value would be much larger in size - and we don't want every validator submitting all that data - we should probably store a hash of the value in the bitmap, and then construct the full vote from the shared data

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
```diff
@@ -137,13 +161,23 @@ impl<
...
-				if *count >= success_threshold {
-					Some(vote.clone())
+
+			counts.iter().find_map(|(vote, blocks_height)| {
+				let num_votes = blocks_height.len() as u32;
+				if num_votes >= success_threshold {
+					let mut blocks_height = blocks_height.clone();
+					let (_, consensus_block_height, _) = {
+						blocks_height
+							.select_nth_unstable(threshold_from_share_count(num_votes) as usize)
```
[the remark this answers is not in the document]
marcellorigotti: My idea was to keep only the valid votes into consideration for the slot.
Cause if someone voted for a wrong value their slot could be way off and it doesn't makes much sense to keep that into consideration

engine/src/witness/sol/nonce_witnessing.rs
```diff
@@ -9,17 +9,19 @@ use crate::sol::{
...
 use serde_json::Value;
+use sol_prim::SlotNumber;
 use std::str::FromStr;
-
 pub async fn get_durable_nonce<SolRetryRpcClient>(
 	sol_client: &SolRetryRpcClient,
 	nonce_account: SolAddress,
-) -> Result<Option<SolHash>>
+	previous_slot: SlotNumber,
+) -> Result<Option<MonotonicChangeVote<SolHash, SlotNumber>>>
```
dandanlen, line 20: nit: To keep the idea of a vote within the VoterApi, this fn should not return a vote, it can just return `(SolHash, SlotNumber)`.

state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs
dandanlen, line 1: I don't see a test that checks that we can only form consensus if the block increases. (ie. everyone votes for a new value, but at a lower block).

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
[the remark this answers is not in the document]
dandanlen: Yes, but if we are in this branch of the code, then we know that there are at least `success_threshold` valid votes. Our thresholds should always be based on the total number of authorities.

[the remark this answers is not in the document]
marcellorigotti: Ok I'll switch back to use the authority_count

state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs
marcellorigotti: Actually this is not true, we can form consensus even if the block decrease.
Votes are filtered at the moment of voting only, so if we manually create votes that have a lower block we can still form consensus.

marcellorigotti: I am implementing `is_vote_valid()` for the mock so that we can test that these bad votes are correctly rejected!

state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs
```diff
@@ -152,6 +156,37 @@ fn consensus_when_all_votes_the_same_but_different_slot() {
 	);
 }
 
+#[test]
+fn votes_with_old_value_or_lower_block_are_rejected() {
```
dandanlen, line 160: 👌

dandanlen (approved):

dandanlen:
As discussed: the upgrade condition check doesn't work, we can delete it. 

We need to migrate by deleting all the old elections and then requesting nonce witness elections for any missing nonces.

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -51,6 +52,338 @@ pub type SolanaElectoralSystem = Composite<
...
+				.map(|(key, _)| {
+					key
+				})
+				.collect::<Vec<ElectionIdentifierOf<super::SolanaElectoralSystem>>>();
+			log::info!("Post upgrade number of elections: {:?}", election_identifiers.len() as u32);
+			assert!(previous_number_election == election_identifiers.len() as u32);
+
+			Ok(())
+		}
+		fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
kylezs, line 273: can we make sure we find some selection of blocks on mainnet that have elections of this type and run try-runtime against them

marcellorigotti: I tested it against some specific blocks with some unavailable nonces as well as during bouncer (to be sure to have a lot of unavailable nonces) and it passed everything.
I am unable to test it against mainnet though, the command always times out.. Here is a block with many unavailable nonces:
`0x254f6d74ae97f08681b448169b43cdd1363c576f0a713e6f8e2ef06371c4c0ca` -> 9 nonces unavailable

dandanlen (approved):

PR #5299, opened 2024-09-26
refactor: Collect settings extrinsic in threshold signer pallet into a single extrinsic

MxmUrw:
# Pull Request

Closes: PRO-1181

## Checklist

## Summary

Previously, there had been three separate extrinsics for configuring storage values for the `threashold-signature` pallet:
    - `set_threshold_signature_timeout()`
    - `set_keygen_response_timeout()`
    - `set_keygen_slash_amount()`

This PR **removes these extrinsics** and consolidates them into a single `update_pallet_config()` extrinsic. Additionally, the event types being emitted have been consolidated into a single `PalletConfigUpdated` event. There are the following changes in behaviour:
 - setting the threshold signature response timeout now **always emits an event**, even if the new value is the same as the old one.
 - setting the keygen response time out now **always emits an event**, even if the new value is the same as the old one.
 - setting the keygen slash amount now **always emits an event**; previously there was no event associated with this extrinsic.

#### Non-Breaking changes


state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -72,6 +72,17 @@ pub type KeygenResponseStatus<T, I> =
 pub type KeyHandoverResponseStatus<T, I> =
 	ResponseStatus<T, KeyHandoverSuccessVoters<T, I>, KeyHandoverFailureVoters<T, I>, I>;
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub enum PalletConfigUpdate {
+	/// Set the maximum duration (in blocks) of a threshold signing ceremony before it is timed out
+	/// and retried
+	ThresholdSignatureResponseTimeout { new_timeout: u32 },
+	/// TODO: Comment required
```
dandanlen, line 80: 👀

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -1168,31 +1156,41 @@ pub mod pallet {
 			)
 		}
 
-		#[pallet::call_index(7)]
-		#[pallet::weight(T::Weights::set_keygen_response_timeout())]
-		pub fn set_keygen_response_timeout(
+		/// Apply a list of configuration updates to the pallet.
```
dandanlen, line 1159: This is a good idea but it's not how it's implemented ;)

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -1168,31 +1156,41 @@ pub mod pallet {
...
-				KeygenResponseTimeout::<T, I>::put(new_timeout);
-				Pallet::<T, I>::deposit_event(Event::KeygenResponseTimeoutUpdated { new_timeout });
-			}
-
-			Ok(())
-		}
+			match update {
+				PalletConfigUpdate::ThresholdSignatureResponseTimeout { new_timeout } => {
+					let new_timeout: BlockNumberFor<T> = new_timeout.into();
+					if new_timeout != ThresholdSignatureResponseTimeout::<T, I>::get() {
```
dandanlen, line 1173: I'm not sure we need this check: it just adds additional noise (and possibly confusion: 'why was my update not applied?'). We can assume that governance are competent enough not to make this kind of mistake.

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -1168,31 +1156,41 @@ pub mod pallet {
...
-		#[pallet::call_index(8)]
-		#[pallet::weight(T::Weights::set_keygen_response_timeout())]
-		pub fn set_keygen_slash_amount(
-			origin: OriginFor<T>,
-			amount_to_slash: FlipBalance,
-		) -> DispatchResult {
-			T::EnsureGovernance::ensure_origin(origin)?;
+				PalletConfigUpdate::KeygenResponseTimeout { new_timeout } => {
+					let new_timeout: BlockNumberFor<T> = new_timeout.into();
+					if new_timeout != KeygenResponseTimeout::<T, I>::get() {
```
dandanlen, line 1181: Same here (and below).

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -1168,31 +1156,41 @@ pub mod pallet {
...
+						KeygenResponseTimeout::<T, I>::put(new_timeout);
+						Self::deposit_event(Event::<T, I>::PalletConfigUpdated { update });
+					}
+				},
 
-			KeygenSlashAmount::<T, I>::put(amount_to_slash);
+				PalletConfigUpdate::KeygenSlashAmount { amount_to_slash } => {
+					if amount_to_slash != KeygenSlashAmount::<T, I>::get() {
+						KeygenSlashAmount::<T, I>::put(amount_to_slash);
+						Self::deposit_event(Event::<T, I>::PalletConfigUpdated { update });
```
dandanlen, line 1190: Then we can pull the event outside of the match statement too (since it can't fail).

state-chain/pallets/cf-threshold-signature/src/lib.rs
MxmUrw: This was the previous behaviour, I would be happy to make it work like you suggest.

state-chain/pallets/cf-threshold-signature/src/lib.rs
MxmUrw: Sure thing.

state-chain/pallets/cf-threshold-signature/src/lib.rs
MxmUrw: Updated.

dandanlen (approved):
Good work, thanks.

PR #5301, opened 2024-09-27
feat: cleanup aborted broadcasts

j4m1ef0rd:
# Pull Request

Closes: PRO-1609

## Checklist

## Summary

- Now removing an aborted broadcast when it gets witnessed successful.
	- Added a unit test to cover cleanup after a manual broadcast.
- Added a housekeeping migration that removes the stale aborted broadcasts on mainnet and perseverance.
	- Tested this with a try-runtime against mainnet & perseverance.

### DIFF
--- a/state-chain/pallets/cf-broadcast/src/migrations/remove_aborted_broadcasts.rs
+++ b/state-chain/pallets/cf-broadcast/src/migrations/remove_aborted_broadcasts.rs
@@ -9,9 +9,6 @@ pub const ETHEREUM_MAX_ABORTED_BROADCAST_PERSEVERANCE: BroadcastId = 1609;
 pub const ARBITRUM_MAX_ABORTED_BROADCAST_PERSEVERANCE: BroadcastId = 665;
 pub const POLKADOT_MAX_ABORTED_BROADCAST_PERSEVERANCE: BroadcastId = 634;
 
-pub struct EthereumMigration;
-pub struct ArbitrumMigration;
-
 pub fn remove_stale_and_all_older<T: Config<I>, I: 'static>(latest_stale_broadcast: BroadcastId) {
 	AbortedBroadcasts::<T, I>::mutate(|aborted| {
 		aborted.retain(|id| id > &latest_stale_broadcast);


state-chain/runtime/src/migrations/remove_aborted_broadcasts.rs
```diff
@@ -0,0 +1,85 @@
...
+		10,
+	>,
+	VersionedMigration<pallet_cf_broadcast::Pallet<Runtime, SolanaInstance>, NoopUpgrade, 9, 10>,
+	VersionedMigration<pallet_cf_broadcast::Pallet<Runtime, PolkadotInstance>, NoopUpgrade, 9, 10>,
+	VersionedMigration<pallet_cf_broadcast::Pallet<Runtime, BitcoinInstance>, NoopUpgrade, 9, 10>,
+);
+
+// Stale aborted broadcasts on mainnet as of 23/09/2024
+const ETHEREUM_ABORTED_BROADCASTS: [BroadcastId; 5] = [3026, 3684, 3686, 11350, 11592];
+const ARBITRUM_ABORTED_BROADCASTS: [BroadcastId; 5] = [238, 239, 345, 423, 426];
```
dandanlen, line 32: These values are only valid on mainnet. It's unlikely but possible that we unintentionally delete something on the other networks.

I think this would make more sense in the housekeeping migration:

- the operation is inherently idempotent: we can run it multiple times and the end-effect will always be the same. 
- we didn't make any changes to the storage format
- we can also delete sisyphos and perseverance values 

Then we can avoid bumping the pallet storage version. 

I would also suggest just writing it in terms of the maximum id that we want to delete: ie. `aborted.retain(|id| id > highest_value_to_delete)`.

j4m1ef0rd: ah, didnt know "housekeeping" was a thing. yeah makes sense.

kylezs:
There are Polkadot ones on Perseverance now, FYI @j4m1ef0rd

state-chain/pallets/cf-broadcast/src/migrations/remove_aborted_broadcasts.rs
```diff
@@ -0,0 +1,27 @@
...
+// Mainnet
+pub const ETHEREUM_MAX_ABORTED_BROADCAST_BERGHAIN: BroadcastId = 11592;
+pub const ARBITRUM_MAX_ABORTED_BROADCAST_BERGHAIN: BroadcastId = 426;
+// Perseverance testnet
+pub const ETHEREUM_MAX_ABORTED_BROADCAST_PERSEVERANCE: BroadcastId = 1609;
+pub const ARBITRUM_MAX_ABORTED_BROADCAST_PERSEVERANCE: BroadcastId = 665;
+pub const POLKADOT_MAX_ABORTED_BROADCAST_PERSEVERANCE: BroadcastId = 634;
+
+pub struct EthereumMigration;
+pub struct ArbitrumMigration;
```
dandanlen, line 13: These can be deleted.

dandanlen (approved):

PR #5302, opened 2024-09-27
Chore/debug 3 nodes

albert-llimos:
# Pull Request

Closes: PRO-1591

## Checklist

## Summary

The failure was due to missing emitted events, which is fixed after we improved the way we witness events in the bouncer.
Then a couple of timeouts needed to be slightly increased so it wouldn't time out for 3-nodes.
The bouncer passes here with three nodes: https://github.com/chainflip-io/chainflip-backend/actions/runs/11069328012/job/30756962348


dandanlen:
Updated the branch to pull in the audit changes to allow the PR to pass.

kylezs (approved):

PR #5305, opened 2024-10-01
chore: bump asset-balances to match release

kylezs:
## Summary

Version bump had not yet been pulled across to main from release - meaning try-runtime/migrations would fail due to version mismatch.

See: ea83d71bf70b37ce0224160255b89ee710d3c359


dandanlen (approved):

kylezs:
1.5.1 .so file needs to be added to the bin.zip - Anton will do after his lunch

PR #5309, opened 2024-10-02
feat: Solana Vault swaps (base PR)

albert-llimos:
# Pull Request

Closes: PRO-1630, PRO-1631

## Checklist

## Summary

~~Missing the migration to add the new variables into the environment.~~ Done


martin-chainflip (approved):

albert-llimos:
We will create PRs on top of this one for the Solana Program Swap feature. Therefore I'd not merge this to main and merge the other PRs onto this one. Once the feature is complete we can then merge it to main.

state-chain/pallets/cf-broadcast/src/migrations.rs
```diff
@@ -7,5 +7,5 @@ mod migrate_timeouts;
 pub type PalletMigration<T, I> = (
 	VersionedMigration<Pallet<T, I>, initialize_broadcast_timeout_storage::Migration<T, I>, 6, 7>,
 	VersionedMigration<Pallet<T, I>, migrate_timeouts::Migration<T, I>, 7, 8>,
-	PlaceholderMigration<Pallet<T, I>, 8>,
```
ramizhasan111, line 10: why this change?

state-chain/chains/src/sol/transaction_builder.rs
```diff
@@ -437,6 +442,41 @@ impl SolanaTransactionBuilder {
...
+		agg_key: SolAddress,
+		durable_nonce: DurableNonceAndAccount,
+		compute_price: SolAmount,
+	) -> Result<SolTransaction, SolanaTransactionBuildingError> {
+		let number_of_accounts = event_accounts.len();
+		let event_and_sender_vec: Vec<AccountMeta> = event_accounts
+			.into_iter()
+			.flat_map(|(event_account, payee)| vec![event_account, payee])
+			// Both event account and payee should be writable and non-signers
+			.map(|address| AccountMeta::new(address.into(), false))
```
ramizhasan111, line 462: i think you dont need the extra map. you could do this inside the ```flat_map```:
```rust
.flat_map(|(event_account, payee)| vec![AccountMeta::new(event_account.into(), false), AccountMeta::new(payee.into(), false)])
```

state-chain/runtime/src/migrations/sol_api_environment.rs
```diff
@@ -0,0 +1,96 @@
+use frame_support::traits::OnRuntimeUpgrade;
```
ramizhasan111, line 1: we dont need this migration to be implemented here at the top level of the runtime. We can move it into the pallet level migration.

state-chain/pallets/cf-broadcast/src/migrations.rs
albert-llimos: @syan095

state-chain/runtime/src/migrations/sol_api_environment.rs
albert-llimos: @syan095

state-chain/pallets/cf-broadcast/src/migrations.rs
syan095: We have `SerializeSolanaBroadcastMigration` in the Runtime crate that migrates broadcaster pallets from version 8 -> 9, so I've updated the "placeholder" migration (which should always points to the current version).

I've updated this with a comment to make it clearer

syan095: I see. I was under the impression that the placeholder migration should point to the latest version after update, but seems like we don't need to. 

This change was overridden by a merge from main.

state-chain/runtime/src/migrations/sol_api_environment.rs
syan095: Got it. Moved it into the Environment pallet.

state-chain/pallets/cf-broadcast/src/migrations.rs
syan095: UPDATE: the placeholder version should match the latest pallet's storage version, or it will give out a warning on try-runtime.
I've updated this to version 10 (as 10 is the latest version after merge from main)

syan095: Update again: Looks like with version 10, we get a warning when doing the pallet migrations, but we are OK in the post-migration checks. 
If we leave the version at 8, we are OK when doing pallet migrations, but will get a warning in post-migration checks.

this is due to how we do the migrations - we do pallet migrations first, then the Runtime ones.

It's just warnings, so I guess I'll just leave it at version 10, since that's the version the pallets ended up on after all the migrations are done.

syan095 (approved):

PR #5310, opened 2024-10-02
feat: tainted transaction reporting

Janislav:
# Pull Request

Closes: PRO-1664

## Checklist

## Summary

- Added the functionality to mark a transaction as tainted as a broker or LP.
- Check transactions when we process the ingress and stop if the channel is marked as tainted.
- Saving tainted transaction details to refund it later.

#### Non-Breaking changes


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -122,6 +123,15 @@ pub enum DepositIgnoredReason {
 	NotEnoughToPayFees,
 }
 
+/// Holds information about a tainted transaction.
+#[derive(RuntimeDebug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo)]
+pub struct TaintedTransactionDetails<AccountId> {
+	/// The broker that created the tainted transaction.
+	pub broker: AccountId,
+	/// The address we use for refunding.
+	pub refund_address: Option<ForeignChainAddress>,
```
dandanlen, line 132: I would have thought that having the refund address here was unnecessary since we already have channel refund params.

dandanlen (commented):
Some early comments.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1776,8 +1860,21 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		);
+
 		let channel_id = deposit_channel_details.deposit_channel.channel_id;
 
+		if let Err(tainted_transaction_details) = Self::check_if_tx_is_tainted(
+			TaintedTransactions::<T, I>::take(&deposit_details),
+			&deposit_channel_details,
+		) {
+			TaintedTransactions::<T, I>::insert(deposit_details, tainted_transaction_details);
+			return Err(Error::<T, I>::TransactionTainted.into())
```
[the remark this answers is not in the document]
dandanlen: Also - I think we will need more information than what is stored here. We should probably just store the full DepositWitness and possibly the channel details too. Basically, all the info we need in order to construct the refund tx .

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: We take the refund address from the deposit channel. But at the same point when we later want to refund we need this address, right? The other option would be saving the channel_id, but saving the refund address directly seemed a bit more convenient to me.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -524,6 +538,16 @@ pub mod pallet {
 	pub type PrewitnessedDepositIdCounter<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, PrewitnessedDepositId, ValueQuery>;
 
+	/// Stores the details of tainted transactions against the deposit details.
+	#[pallet::storage]
+	pub type TaintedTransactions<T: Config<I>, I: 'static = ()> = StorageMap<
+		_,
+		Twox64Concat,
+		<T::TargetChain as Chain>::DepositDetails,
+		TaintedTransactionDetails<T::AccountId>,
```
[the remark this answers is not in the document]
Janislav: Good point

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: See my other comment - we actually need more than just the refund address so that we can construct a refund tx. I agree we should save it though, rather than just store a reference to the channel (unless we can guarantee that the channel won't be deleted).

state-chain/traits/src/liquidity.rs
```diff
@@ -23,6 +23,12 @@ pub trait LpRegistration {
 		address: cf_chains::ForeignChainAddress,
 	);
 
+	/// Get the liquidity refund address for the given account and asset.
+	fn get_liquidity_refund_address(
```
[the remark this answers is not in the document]
Janislav: So you are saying that if an LP is requesting a deposit address, he should also provide the refund address directly here instead of setting it over the extrinsic in the LP pallet?

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
Janislav: I store now the DepositWitness + refund address.

Janislav:
Still working on the benchmark. It's written, but there is something odd with the way we generate the weights.

state-chain/traits/src/lib.rs
```diff
@@ -724,6 +724,7 @@ pub trait DepositApi<C: Chain> {
 		lp_account: Self::AccountId,
 		source_asset: C::ChainAsset,
 		boost_fee: BasisPoints,
+		refund_address: Option<ForeignChainAddress>,
```
dandanlen, line 727: This should not be optional. 

If the LP hasn't registered a refund address, they can't open a channel, so we should always have this.

state-chain/pallets/cf-lp/src/lib.rs
```diff
@@ -199,11 +199,15 @@ pub mod pallet {
 				Error::<T>::NoLiquidityRefundAddressRegistered
 			);
 
+			let refund_address =
+				LiquidityRefundAddress::<T>::get(&account_id, ForeignChain::from(asset));
```
dandanlen, line 203: See the `ensure!` just above here. We always have this. We combine these and then we don't need the option (See other comment in the trait definition):

```suggestion
			let refund_address =
				LiquidityRefundAddress::<T>::get(&account_id, ForeignChain::from(asset))
					.ok_or(Error::<T>::NoLiquidityRefundAddressRegistered)?;
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -120,6 +126,19 @@ pub enum DepositIgnoredReason {
...
 	NotEnoughToPayFees,
+	TransactionTainted,
+}
+
+/// Holds information about a tainted transaction.
+#[derive(RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo, CloneNoBound)]
+#[scale_info(skip_type_params(T, I))]
+pub struct TaintedTransactionDetails<T: Config<I>, I: 'static> {
+	/// The broker that created the tainted transaction.
+	pub broker: T::AccountId,
```
dandanlen, line 137: nit: it's not always going to be a broker.

```suggestion
	/// The account that reported the transaction.
	pub reporter: T::AccountId,
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -252,14 +271,15 @@ pub mod pallet {
 	#[derive(CloneNoBound, RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo)]
 	#[scale_info(skip_type_params(T, I))]
 	pub struct DepositChannelDetails<T: Config<I>, I: 'static> {
+		/// The owner of the deposit channel.
+		pub owner: T::AccountId,
```
dandanlen, line 275: This will need a migration. In the migration, we won't know the owner, so will just have to use some default value.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -701,12 +738,22 @@ pub mod pallet {
...
+		/// The deposit channel does not exist.
+		DepositChannelDoesNotExist,
+		/// Unauthorized call
+		Unauthorized,
+		/// The deposit channel has been tainted and is no longer usable.
+		DepositChannelTainted,
+		/// Transaction tainted
+		TransactionTainted,
+		/// Unsupported chain
+		UnsupportedChain,
```
dandanlen, line 750: Most of these appear to be unused...

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1251,6 +1327,26 @@ impl<T: Config<I>, I: 'static> IngressSink for Pallet<T, I> {
...
 impl<T: Config<I>, I: 'static> Pallet<T, I> {
+	fn mark_transaction_as_tainted_inner(
+		origin: OriginFor<T>,
+		tx_id: <T::TargetChain as Chain>::DepositDetails,
+	) -> DispatchResult {
+		let account_id = ensure_signed(origin)?;
+		let is_broker = T::AccountRoleRegistry::has_account_role(&account_id, AccountRole::Broker);
+		let is_lp =
+			T::AccountRoleRegistry::has_account_role(&account_id, AccountRole::LiquidityProvider);
+		ensure!(is_broker || is_lp, BadOrigin);
```
dandanlen, line 1338: Actually I'm not sure we want/need LPs to mark transactions as tainted. We need their deposit address *in case* a tx is marked, but in general they won't be marking their own.

We will probably add another account role or similar for the 'reporter' role, but that can be in a future PR. For now we can remove the LP here.

dandanlen: I'd also suggest unwrapping the origin and doing permission checks in the extrinsic rather than in this `inner` method. Might make testing/benchmarking easier.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1251,6 +1327,26 @@ impl<T: Config<I>, I: 'static> IngressSink for Pallet<T, I> {
...
+		let tainted_transaction = TaintedTransactionDetails {
+			broker: account_id.clone(),
+			refund_address: None,
+			deposit_witness: None,
+			expires_at: <frame_system::Pallet<T>>::block_number()
+				.saturating_add(BlockNumberFor::<T>::from(TAINTED_TX_EXPIRATION_BLOCKS)),
+			marked_for_refund: false,
```
dandanlen, line 1345: I think it would be clearer/safer/more efficient if we split this into two storage items: 

```rust
// in a map of AccountId -> TxId -> TaintedTransactionDetails
struct TaintedTransactionDetails {
	// I don't think we need this, it's already in the key of the map:
	// reporter: AccountId,
	expires_at: BlockNumber,
}

// in a value of Vec<RejectionDetails>
struct RejectionDetails {
	refund_address: Address,
	deposit_witness: DepositWitness,
}
```

This avoids having to deal with the `Option`s and filter against `marked_for_refund` etc.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1776,7 +1872,57 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		let deposit_channel_details = DepositChannelLookup::<T, I>::get(&deposit_address)
 			.ok_or(Error::<T, I>::InvalidDepositAddress)?;
 
+		ensure!(
+			deposit_channel_details.deposit_channel.asset == asset,
+			Error::<T, I>::AssetMismatch
+		);
+
 		let channel_id = deposit_channel_details.deposit_channel.channel_id;
+		let channel_owner = deposit_channel_details.clone().owner;
```
dandanlen, line 1881: We don't need to clone the full details struct just to get the owner.

```suggestion
		let channel_owner = deposit_channel_details.owner.clone();
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1776,7 +1872,57 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
+		if let Some(tainted_tx) =
+			TaintedTransactions::<T, I>::take(channel_owner.clone(), deposit_details.clone())
+		{
+			if tainted_tx.broker == channel_owner {
```
dandanlen, line 1886: Since we are reading this from a storage item with a key of `channel_owner`, this condition should always be true (see my comment about not really needed the `broker` field).

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1776,7 +1872,57 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		ensure!(
+			deposit_channel_details.deposit_channel.asset == asset,
+			Error::<T, I>::AssetMismatch
+		);
+
 		let channel_id = deposit_channel_details.deposit_channel.channel_id;
+		let channel_owner = deposit_channel_details.clone().owner;
+
+		if let Some(tainted_tx) =
+			TaintedTransactions::<T, I>::take(channel_owner.clone(), deposit_details.clone())
```
dandanlen, line 1884: nit: I think we should consider moving this check below the minimum deposit check, simply because the min deposit check is cheaper. Also, if the deposit is below the minimum, we drop it anyway (ie. we don't need to refund it) .

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: (But also, i think we can remove this field altogether? See: https://github.com/chainflip-io/chainflip-backend/pull/5310#discussion_r1795051064

dandanlen:
Also please remember to rebase against main to pick up the new compiler version.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: "Inner" was a more temporary solution since we only wanted to support BTC for now and that would open a whole new number of problems with tests and benchmarks. Since I think it could make sense to support at least ETH as well, we can do this properly and the inner function warp can disappear. In general, I think we should try to benchmark as close as possible to be on the safe side.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: I don't think we need DepositWitness tbh. Definitely not for BTC (tx_id is enough here) and for ETH we need other stuff I (refund_address, asset, amount). Because we need to pass this information to the API call, I think we're going to end up to define a new type for that anyway. I would go for now with something static that works for ETH/BTC and extend the type as we go or make generic for the chain.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: Good point. This is redundant since we changed from `TxId->TaintedDetails` to `AccountId->TxId->TaintedDetails`

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -120,6 +123,17 @@ pub enum DepositIgnoredReason {
...
 	/// required to process the requisite transactions.
 	NotEnoughToPayFees,
+	TransactionTainted,
+}
+
+/// Holds information about a tainted transaction.
+#[derive(RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo, CloneNoBound)]
+#[scale_info(skip_type_params(T, I))]
+pub struct TaintedTransactionDetails<T: Config<I>, I: 'static> {
+	pub refund_address: Option<ForeignChainAddress>,
```
Janislav, line 133: Until refund parameters are optional, this has to stay optional. We can change it after this is done.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -748,6 +789,23 @@ pub mod pallet {
 					Self::recycle_channel(&mut used_weight, address);
 				}
 			}
+
+			let mut expired_transactions = Vec::new();
+
+			for (account, tx_id, expire_block) in TaintedTransactions::<T, I>::iter() {
+				if expire_block <= now {
```
dandanlen, line 796: I don't like having to iterate through all of the TaintedTransactions just to see if any have expired (99% of the time this will return nothing). It would be good to think of a better way to do this.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1810,6 +1895,39 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			return Ok(())
 		}
 
+		let channel_owner = deposit_channel_details.owner.clone();
+
+		if TaintedTransactions::<T, I>::take(channel_owner.clone(), deposit_details.clone())
```
dandanlen, line 1900:
```suggestion
		if TaintedTransactions::<T, I>::take(&channel_owner, &deposit_details)
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: But we can't refund anything if there is no refund address? 

Doesn't it make more sense to make it required here, and handle the case where we don't have it before storing this value?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -524,6 +540,23 @@ pub mod pallet {
 	pub type PrewitnessedDepositIdCounter<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, PrewitnessedDepositId, ValueQuery>;
 
+	/// Stores the reporter and the tx_id against the BlockNumber when the report expires.
+	#[pallet::storage]
+	pub(crate) type TaintedTransactions<T: Config<I>, I: 'static = ()> = StorageDoubleMap<
+		_,
+		Identity,
+		T::AccountId,
+		Twox64Concat,
```
dandanlen, line 549:
```suggestion
		Blake2_128Concat,
```

DepositDetails can be freely set by the caller, so we need to be defensive here.

state-chain/pallets/cf-ingress-egress/src/migrations/add_owner_to_channel_details.rs
```diff
@@ -0,0 +1,128 @@
...
+		OptionQuery,
+	>;
+}
+
+pub struct Migration<T: Config<I>, I: 'static>(PhantomData<(T, I)>);
+
+impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for Migration<T, I> {
+	fn on_runtime_upgrade() -> Weight {
+		for (account, channel_details) in old::DepositChannelLookup::<T, I>::drain() {
+			let dummy_account = T::AccountId::decode(&mut &[0u8; 32][..]).unwrap();
```
dandanlen, line 33: Prefer to use translate rather than drain+insert. According to the docs, inserting during iteration can cause unwanted side effects.

dandanlen (commented):
Some minor nits but I think it's nearly ready.

Don't worry about the version bump CI check - should go away when #5327 is merged.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: I don't believe it doesn't make any difference other than having to deal with less complex code if we go with the option variant. In the point of the swap flow where we have (or have not) the refund address, it's already too late to do anything else than saving what we have and exit the swap logic. What do you want to save if, instead, if there is just no address? The solution is to make the refund parameters a requirement, so this can not happen, but since this is not done, yet we go with an option here and replace it after @msgmaxim has solved this in his PR.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: We could introduce another storage item that also saves `ExpiryBlock => Vec<(account, tx_id)>` and then we change the TaintedTransactionMap to `AcountId, TxID => ()`. With this, we can remove the entry directly and don't need to iterate over all TaintedTransactions. As long no one is spamming, I don't see this would be a problem, but it's true that this could open up some kind of attack vector where any broker account could add a million tainted TX. I think we should avoid introducing any parameter that could affect the execution of a hook and can be manipulated from the outside.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: Yeah, my preference is that we keep the option in this PR. Then once I change the interface to always require refund address, it can update this code too (and the code won't get compiled if I forget to do this).

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -64,6 +64,9 @@ use sp_std::{
 };
 pub use weights::WeightInfo;
 
+// 1 hour.
+const TAINTED_TX_EXPIRATION_BLOCKS: u32 = 600;
```
msgmaxim, line 68: Should we tie this up to `SECONDS_PER_BLOCK` in case we do end up changin it? 😄

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1810,6 +1895,39 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+			};
+
+			let tainted_transaction_details = TaintedTransactionDetails {
+				refund_address,
+				amount: deposit_amount,
+				asset,
+				tx_id: deposit_details.clone(),
+			};
+
+			ScheduledTxForReject::<T, I>::append(tainted_transaction_details);
```
msgmaxim, line 1918: How will this work if the deposit has already been boosted? Right now it looks like you would be able to drain the boost pool if your deposit is rejected.

dandanlen: @Janislav ☝️ 

We forgot to fully consider boost in this.

Janislav: Mhhh... I would not call it "forgot" as long it is not live 😅.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1328,12 +1331,14 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		tx_id: <T::TargetChain as Chain>::DepositDetails,
 	) -> DispatchResult {
 		let account_id = T::AccountRoleRegistry::ensure_broker(origin)?;
-		TaintedTransactions::<T, I>::insert(
-			account_id,
-			tx_id,
+		TxExpiresAt::<T, I>::mutate(
 			<frame_system::Pallet<T>>::block_number()
```
dandanlen, line 1335: I'd suggest making it such that block numbers are grouped into 'hourly' buckets. The exact time is not important. 

Also we can use append:

```rust
TxExpiresAt::<T, I>::append(
	// Expiry scheduled between 1 and 2 hours in the future:
	((<frame_system::Pallet<T>>::block_number() % 600) + 2) * 600,
	(account_id.clone(), tx_id.clone())
);
```

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -2120,7 +2120,8 @@ fn tainted_transactions_expire_if_not_witnessed() {
 		let tx_id = DepositDetails::default();
 		let expiry_at = System::block_number() + TAINTED_TX_EXPIRATION_BLOCKS as u64;
 
-		TaintedTransactions::<Test>::insert(BROKER, tx_id.clone(), expiry_at);
+		TaintedTransactions::<Test>::insert(BROKER, tx_id.clone(), ());
+		TxExpiresAt::<Test>::insert(expiry_at, vec![(BROKER, tx_id.clone())]);
```
dandanlen, line 2124: We should test using the real public methods, or at least with the `_inner` method.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: Why should we do this? 🤔 Is it not also bad for brokers (for e.x) to not reliable know when something expires?

dandanlen: Actually yeah I can't remember why I thought this would be a good idea - never mind. (Although using append is still a better than mutate.)

Janislav: Agree 👍

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: We will not change it :D. But yes, using `SECONDS_PER_BLOCK` is better 👍

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1854,6 +1869,48 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+	/// Checks if the transaction is tainted and schedules it for rejection if it is.
+	fn check_if_tx_is_tainted_and_schedule_reject(
+		channel_owner: T::AccountId,
+		deposit_details: <T::TargetChain as Chain>::DepositDetails,
+		deposit_amount: TargetChainAmount<T, I>,
+		asset: TargetChainAsset<T, I>,
+		deposit_address: TargetChainAccount<T, I>,
+		action: ChannelAction<T::AccountId>,
+	) -> bool {
+		if TaintedTransactions::<T, I>::take(channel_owner.clone(), deposit_details.clone())
```
msgmaxim: Nit, but it seems like we are doing way to many clones here. For example, when you call `check_if_tx_is_tainted_and_schedule_reject` you preemptively clone `deposit_address` but in the common case of "transaction not tainted" that value isn't even used. Also, I believe you can pass references to `take` in this line. I haven't tested this, but conceptually, we shouldn't clone anything (including at check_if_tx_is_tainted_and_schedule_reject`'s callers side) unless the transaction is in fact tainted and you need to create new objects like `TaintedTransactionDetails` and `DepositIgnored`.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1901,34 +1958,15 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-				amount: deposit_amount,
-				asset,
-				tx_id: deposit_details.clone(),
-			};
-
-			ScheduledTxForReject::<T, I>::append(tainted_transaction_details);
-
-			Self::deposit_event(Event::<T, I>::DepositIgnored {
-				deposit_address: deposit_address.clone(),
+		if deposit_channel_details.boost_status == BoostStatus::NotBoosted &&
```
msgmaxim: Would it be useful to log a warning if a deposit is tainted, but we had to accept it since we already boosted it?

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -2120,14 +2122,175 @@ fn tainted_transactions_expire_if_not_witnessed() {
...
+		eth::Asset::Eth,
+		1000,
+		10
+	));
+
+	address
+}
+
+#[test]
+fn ignore_boosted_channels_if_tainted_after_prewitness() {
```
msgmaxim, line 2191: This name seems misleading. What seems to be ingnored is the "tainted" status, but the name suggests that the channel (which I would read as "deposit") will be ignored, which is the opposite of what should happen.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -790,16 +811,12 @@ pub mod pallet {
 				}
 			}
 
-			let mut expired_transactions = Vec::new();
-
-			for (account, tx_id, expire_block) in TaintedTransactions::<T, I>::iter() {
-				if expire_block <= now {
-					expired_transactions.push((account.clone(), tx_id.clone()));
+			for (account, tx_id) in ReportExpiresAt::<T, I>::take(now) {
```
msgmaxim, line 814: Can we add a code comment to explain what this is for? It took me a bit of time to understand the motivation here. Also, are we ever removing `TaintedTransactions` record if we prewitnessed a deposit? I couldn't find it. I think we could remove it once the channel expires (which is aslo when we forget the "boosted" status).

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -76,6 +79,13 @@ pub enum BoostStatus<ChainAmount> {
 	NotBoosted,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, Default)]
+pub enum Prewitnessed {
```
msgmaxim, line 83: As a suggestion, maybe we could rename this to `BoostRejected`? It helped me understand the code when thinking about it this way. Right now it is possible that you prewitnessed a tx, then marked it as tainted, but the status will still be `Prewitnessed::No`, correct?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1814,6 +1927,44 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			return Ok(())
 		}
 
+		let channel_owner = deposit_channel_details.owner.clone();
+
+		if let Some(prewitness_state) =
+			TaintedTransactions::<T, I>::take(&channel_owner, &deposit_details)
+		{
+			if deposit_channel_details.boost_status == BoostStatus::NotBoosted ||
```
msgmaxim, line 1935: I think this would benefit from a comment to explain the logic. Something like: "We refund if (a) we rejected a boost of this deposit earlier, or (b) if there has been no attempt to boost in the first place".

Alternatively, it may be simpler to understand the code if we recorded deposits that *have* boosted, then you could have a single condition here. I'm a little worried about relying on `boost_status` since it is about a channel rather than a deposit (we may have to consider cases where there are multiple desposits into the same channel). However, I realise that this isn't a trivial change, so may not be worth it.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: We remove it when it's finally witnessed. The idea behind the code is that for some reason the timespan between pre witness and witness is so long that it expires in the meantime, we have to add it back since otherwise the boosted swap would go through. Eventually it will get removed from `TaintedTransactions ` though.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: I agree that this is not very intuitive. All the information if a deposit was boosted or not is there, though. I think what could help to make it more understandable is first adding a comment that explains the code (like you suggested) and break the code down a bit with proper naming of the single conditions like that:
```
let is_not_boosted = deposit_channel_details.boost_status == BoostStatus::NotBoosted;
let was_tainted_before_boosting = prewitness_state == Prewitnessed::Yes;

if is_not_boosted || was_tainted_before_boosting {
    // ... (rest of the code)
}
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: > Eventually it will get removed from TaintedTransactions  though.

Does this include the case where a a tx is prewitnessed, but is never fully witnessed? (This is the case when boosters lose money essentially.)

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: > was_tainted_before_boosting

This is still misleading though... If it is tainted before it is prewitnessed, then we *don't* boost it. IMO "boost rejected" (like I suggested above) would be more accurate. And sorry for being nitpicky about naming/comments 🙂, it is just that it is not easy to reason about both boosting and tainting at the same time, so I think in this particular case it is worth being more accurate about names to make reasoning about the code a little easier.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -752,6 +810,25 @@ pub mod pallet {
...
 			}
+
+			// A report get's cleaned up after approx 1 hour and needs to be re-reported by the
+			// broker if necessary. This is need as some kind of garbage collection mechanism
+			// for tainted deposits.
+			for (account, tx_id) in ReportExpiresAt::<T, I>::take(now) {
+				let marked_status = TaintedTransactions::<T, I>::take(&account, &tx_id);
+				// A special case when a deposit was boosted and rejected in the prewitness step but
+				// expired between the prewitness and witness step. In this case we put it back into
+				// the list of expired transactions so it get's witness and finally rejected.
```
msgmaxim, line 821:
```suggestion
				// the list of tainted transactions so it is rejected when it is finally witnessed.
```

msgmaxim:
I took the liberty to push a commit with some updates to comments (a few things weren't quite right, and I thought it would help if I did it myself). Feel free to revert @Janislav.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -752,6 +810,26 @@ pub mod pallet {
 					Self::recycle_channel(&mut used_weight, address);
 				}
 			}
+
+			// A report gets cleaned up after approx 1 hour and needs to be re-reported by the
```
msgmaxim, line 814: What if the broker re-reports? Would it then overwrite the `BoostRejected` record? (This would be problematic.)

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1352,6 +1354,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		tx_id: <T::TargetChain as Chain>::DepositDetails,
 	) -> DispatchResult {
 		let account_id = T::AccountRoleRegistry::ensure_broker(origin)?;
+		ensure!(
+			!TaintedTransactions::<T, I>::contains_key(&account_id, &tx_id),
```
dandanlen, line 1358: I don't think this is enough? We don't need to know if it has been reported, we need to know if it has been boosted.

For example, what about a case where a transaction has been boosted (without being reported)?

dandanlen:
I added some changes (From the commit description):

- mark boosted transactions as boosted instead of using channel status
- allow pallet config instead of relying on chain
- add event for tx reports
- only allow reporting of unseen transactions
- add doc comments
- renaming of types/events
- remove unused error

The main one is that I added a Boosted variant (and renamed) to the transaction report enum. I think this makes everything a bit clearer, in particular avoids having to rely on the channel status to know if a deposit has been boosted: now we just look up the tx_id.

dandanlen (approved):

Janislav:
Somehow, the storage of the old DepositChannelDetails is not decodable during migration. No idea why yet.

dandanlen (approved):

PR #5311, opened 2024-10-03
feat: BTC contract swap encoding

msgmaxim:
# Pull Request

Closes: PRO-1673

## Checklist

## Summary

- Defines `UtxoEncodedData` which effectively determines how swap parameters are layed out/encoded. For encoding/decoding we use Scale to keep things smiple. It still results in a reasonably compact encoding: the `check_utxo_encoding` test demonstrates which bytes are used for what (and that we don't exceed the limit of 80 bytes). I chose not to use "insta" in this case to make things a bit more explicit.
- `SharedCfParameters` defines a subset of `UtxoEncodedData`'s fields which we expect to use in other smart contracts (this is what you intended @albert-llimos?)
- As discussed on discord, min price is encoded as min output amount to save space. The encoding of other fields is more or less straightforward.


albert-llimos:
> SharedCfParameters defines a subset of UtxoEncodedData's fields which we expect to use in other smart contracts (this is what you intended @albert-llimos?)
Yes exactly, that looks good to me.

albert-llimos:
The only thing left to "match" the contract swaps is the CCM parameters (`message` and `gasAmount`). I understand that we have this 80bytes limitation so it definitely doesn't fit there. I guess there might be ways around it (e.g. add another utxo for that, although still limited in length) but I assume we don't care about supporting CCM swaps, especially right now. I wouldn't care but just double checking that we're all aligned. @martin-chainflip @msgmaxim

engine/src/witness/btc/smart_contract.rs
```diff
@@ -0,0 +1,338 @@
...
+		return None;
+	};
+
+	let deposit_amount = utxo_to_vault.value.to_sat();
+
+	// Derive min price (encoded as min output amount to save space):
+	let min_price = sqrt_price_to_price(bounded_sqrt_price(
+		data.parameters.min_output_amount.into(),
+		deposit_amount.into(),
+	));
```
dandanlen, line 157: We should make it possible to pass the minimum_amount instead of price when witnessing the contract swaps. The price was used as a workaround for channels, because channels are opened before the swap is made. That's not the case with contract swaps. All we are doing here is converting to a price, which is then converted back to a min output on-chain.

engine/src/witness/btc/smart_contract.rs
```diff
@@ -0,0 +1,338 @@
...
+#[allow(dead_code)]
+fn encode_data_in_nulldata_utxo(data: &[u8]) -> Option<ScriptBuf> {
+	const MAX_LEN: u8 = 80;
+
+	let len = data.len() as u8;
+	if len > MAX_LEN {
+		return None
+	}
+
+	let script_bytes: Vec<_> = [OP_RETURN.to_u8(), OP_PUSHDATA1.to_u8(), len]
```
martin-chainflip, line 59: unfortunately its not that easy. have a look at `impl SerializeBtc for BitcoinOp`. If `len` is below 76, then we cant use OP_PUSHDATA1. instead we just push `len` followed by `data`. If len == 1 (will probably not happen), then we need special opcodes, too. At least we should chekc that len > 1 above.

engine/src/witness/btc/smart_contract.rs
msgmaxim: It is not clear that we should prefer the minimum amount. If we use DCA, for example, we want to know the price rather than the amount since we have to derive the min output for each chunk individually. Same for CCM. I'd rather we use Price consistently when requesting a swap from the swapping pallet, and think of the conversion to/from min output here simply as a way to compress this parameter in contract calls.

engine/src/witness/btc/smart_contract.rs
msgmaxim: > yes. we will always push at least the destination address, so OP_PUSHDATA1 should work for all cases

This is what you told me in DM when I asked about it 🥲. Or did you mean that we would have to add padding for this to work? Why wouldn't `OP_PUSHDATA1` work for arbitrary length? I thought the special codes were just an optimisation... Seems silly of bitcoin not to support this.

> If len == 1 (will probably not happen), then we need special opcodes

Are you sure? `OP_PUSHBYTES_1` does exist and seems to work same as all other `OP_PUSHBYTES_x` opcodes.

engine/src/witness/btc/smart_contract.rs
dandanlen: Ok, makes sense 👌

engine/src/witness/btc/smart_contract.rs
msgmaxim: Updated now @martin-chainflip

dandanlen: I would suggest we move the *encoding* of params (and associated types) into the chains crate. We already have utilities for pushing bytes that use the correct opcode depending on the size of the vec (see BitcoinOp::PushBytes). And the encoding will be used by the apis, which should not need to add a dependency on the engine witnessing code.

If we add the OP_RETURN variant to `BitcoinOp`, building the script will be something like `BitcoinScript::new([BitcoinOp::Return, BitcoinOp::PushBytes { params.encode() }])`.

msgmaxim: Moved encodeing to cf-chains @dandanlen (and now use `BitcoinScript`). The decoding is still in the engine since it should only be used by the witnesser. We can consider moving it too, but I think it might be worth waiting to see what the other components of btc smart contract will look like.

engine/src/witness/btc/smart_contract.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// Currently unused, but will be used by the deposit wintesser:
+#[allow(dead_code)]
+pub fn try_extract_contract_call(
+	tx: &VerboseTransaction,
+	vault_address: ScriptPubkey,
+) -> Option<BtcContractCall> {
+	// A correctly constructed transaction carrying CF swap parameters must have at least 3 outputs:
+	let [utxo_to_vault, nulldata_utxo, change_utxo, ..] = &tx.vout[..] else {
```
dandanlen, line 92: At least 3 or exactly 3? Maybe we can be strict?

state-chain/chains/src/btc.rs
```diff
@@ -965,6 +974,7 @@ pub enum BitcoinOp {
 	PushArray20 { bytes: [u8; 20] },
 	PushArray32 { bytes: [u8; 32] },
 	PushVersion { version: u8 },
+	Return,
```
dandanlen, line 977: nit: this should go above the comment saying `// Not part of the bitcoin spec`.

state-chain/chains/src/btc.rs
```diff
@@ -954,7 +963,7 @@ impl<T: SerializeBtc> SerializeBtc for &[T] {
 #[derive(Encode, Decode, TypeInfo, MaxEncodedLen, Clone, RuntimeDebug, PartialEq, Eq)]
 pub enum BitcoinOp {
 	PushUint { value: u32 },
-	PushBytes { bytes: BoundedVec<u8, ConstU32<MAX_SEGWIT_PROGRAM_BYTES>> },
+	PushBytes { bytes: BoundedVec<u8, ConstU32<MAX_PUSHABLE_BYTES>> },
```
dandanlen, line 966: This seems a bit unnecessary? `MAX_SEGWIT_PROGRAM_BYTES` was already more than enough.

state-chain/chains/src/btc/smart_contract_encoding.rs
```diff
@@ -0,0 +1,95 @@
+use crate::ForeignChainAddress;
+use cf_primitives::{Asset, AssetAmount};
+use codec::{Decode, Encode};
+
+use super::{BitcoinOp, BitcoinScript};
+
+// The maximum length of data that can be encoded in a nulldata utxo
+const MAX_NULLDATA_UTXO_LENGTH: usize = 80;
```
dandanlen, line 8:
```suggestion
const MAX_NULLDATA_LENGTH: usize = 80;
```

Strictly speaking the utxo is longer than that (because it includes the opcodes).

state-chain/chains/src/btc/smart_contract_encoding.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[derive(Encode, Decode, Clone)]
+pub struct UtxoEncodedData {
+	pub output_asset: Asset,
+	pub output_address: ForeignChainAddress,
```
dandanlen, line 13: ForeignChainAddress is an internal type, not really for external use. It also duplicates information already held in the asset field (the asset implies the chain so we don't need to encode this again in the address). 

How about we move the address to the end and assume that any trailing bytes are part of the address? 

```rust
pub struct UtxoEncodedData {
	pub parameters: SharedCfParameters,
	pub output_asset: Asset,
	// Assumed to be the raw bytes, 20 bytes for evm, 32 for dot/sol
	// btc should be rejected
	pub output_address: Vec<u8>,
}
```

Alternatively we decode the bytes based on the output asset:

```rust
pub struct UtxoEncodedData {
	pub parameters: SharedCfParameters,
	pub output_asset: Asset,
	pub output_address: EncodedAddress,
}

impl Decode for UtxoEncodedData {
	// Custom impl that uses try_into_encoded_address
	// See implementation in evm::vault.rs
}
```

state-chain/chains/src/btc/smart_contract_encoding.rs
```diff
@@ -0,0 +1,95 @@
...
+	pub parameters: SharedCfParameters,
+}
+
+// The encoding of these parameters is the same across chains
+#[derive(Encode, Decode, Clone)]
```
dandanlen, line 18:
```suggestion
#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone)]
```

We will need these too.

state-chain/chains/src/btc/smart_contract_encoding.rs
```diff
@@ -0,0 +1,95 @@
...
+	pub chunk_interval: u16,
+	// Boost fields:
+	pub boost_fee: u8,
+}
+
+#[allow(dead_code)]
+pub fn encode_data_in_nulldata_utxo(data: &[u8]) -> Option<BitcoinScript> {
+	if data.len() > MAX_NULLDATA_UTXO_LENGTH {
+		return None;
+	}
```
dandanlen, line 34: This shouldn't be possible. We should debug_assert this and/or fail loudly instead of silently dropping it.

engine/src/witness/btc/smart_contract.rs
```diff
@@ -0,0 +1,295 @@
...
+	};
+
+	// First output must be a deposit into our vault:
+	if utxo_to_vault.script_pubkey.as_bytes() != vault_address.bytes() {
+		return None;
+	}
+
+	// Second output must be a nulldata UTXO (with 0 amount):
+	if nulldata_utxo.value.to_sat() != 0 {
+		return None;
```
dandanlen, line 103: A warning or debug statement might be appropriate here too.

engine/src/witness/btc/smart_contract.rs
msgmaxim: What is the benefit? I did decide to be strict about the order to make parsing easier, but allowing additional UTXOs doesn't make things harder for us. I think it is better to keep this more flexible.

state-chain/chains/src/btc.rs
msgmaxim: What do you mean? `MAX_SEGWIT_PROGRAM_BYTES` is only 40 bytes, but we want to allow up to 80 bytes for our use case. It also makes more sense. `MAX_SEGWIT_PROGRAM_BYTES` was only used because that happened to be what was being pushed, however bitcoin script allows pushing up to `MAX_PUSHABLE_BYTES`, so that's more correct.

state-chain/chains/src/btc/smart_contract_encoding.rs
msgmaxim: Yeah, I thought this was confusing, but didn't think we could do better without making the name even longer. `MAX_NULLDATA_LENGTH` also isn't great, as to me it sounds like we are talking about an array of 80 zeros, rather than nulldata being simply the type of the utxo... No strong opinion though, I updated to use your suggestion.

state-chain/chains/src/btc/smart_contract_encoding.rs
msgmaxim: The funciton accepts `&[u8]` from whoever is calling it, so it is very much possible. Also, how is possible to "silently drop" the result of encoding? Note that return `None` which should be handled by the caller to get the result. It is the caller's job to assert (with debug_asert or otherwise) that encoding worked.

Of course we can assert when passing the result of `params.encode()` that the result of encoding `UtxoEncodedData` with Scale always fits within 80 bytes. I was a little hesitant to do this since, it is easy to add another field or a variant to an existing type that would result in exceeding the max length. (We have tests for this of course, but it is hard to test exhaustively). Seems better to return Error to the consumer of this for a particular set of parameters than to crash the broker? Anyway, I assert internally that `params.encode()` fits, but I think a lot of this fine-tuning of the interfaces would make more sense when we add components that use them.

engine/src/witness/btc/smart_contract.rs
msgmaxim: I'm adding a warning, but are we not expecting other transactions into our vault's change address? Though I guess if we are sending change to ourselves it will never be the first UTXO.

engine/src/witness/btc/smart_contract.rs
```diff
@@ -0,0 +1,295 @@
...
+fn try_extract_utxo_encoded_data(script: &bitcoin::ScriptBuf) -> Option<&[u8]> {
+	let bytes = script.as_script().as_bytes();
+
+	// First opcode must be OP_RETURN
+	if bytes[0] != OP_RETURN.to_u8() {
+		return None;
+	}
+
+	// Second opcode must be either OP_PUSHBYTES_X (1..=75) or OP_PUSHDATA1 (76):
+	let (data_len, data_bytes) = match bytes[1] {
```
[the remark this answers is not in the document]
msgmaxim: Indeed. Added more test cases too.

engine/src/witness/btc/smart_contract.rs
dandanlen: Yeah I was just thinking in terms of debugging that this it might help. No big deal though.

state-chain/chains/src/btc.rs
dandanlen: Sorry I thought I saw 128, must have misread.

engine/src/witness/btc/smart_contract.rs
dandanlen: I wasn't sure, it was a question. I think i agree we can leave it as is.

state-chain/chains/src/btc/smart_contract_encoding.rs
```diff
@@ -8,15 +8,57 @@ use super::{BitcoinOp, BitcoinScript};
...
+		r
+	}
+}
+
+impl Decode for UtxoEncodedData {
+	fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
+		let output_asset = Asset::decode(input)?;
+
+		let output_address = match output_asset {
+			Asset::Eth | Asset::Flip | Asset::Usdc | Asset::Usdt =>
```
dandanlen, line 45: You can do this: 

```rust
let output_address = match ForeignChain::from(output_asset) {
			ForeignChain::Ethereum => EncodedAddress::Eth(<[u8; 20]>::decode(input)?),
	// ...
```

state-chain/chains/src/btc/smart_contract_encoding.rs
msgmaxim: Forgot to mention that this has also been updated @dandanlen

state-chain/chains/src/btc/smart_contract_encoding.rs
msgmaxim: updated

dandanlen (approved):

state-chain/chains/src/btc.rs
```diff
@@ -958,22 +967,32 @@ impl<T: SerializeBtc> SerializeBtc for &[T] {
...
-#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, RuntimeDebug, PartialEq, Eq)]
+#[derive(Encode, Decode, TypeInfo, Clone, RuntimeDebug, PartialEq, Eq)]
 pub struct BitcoinScript {
-	bytes: BoundedVec<u8, ConstU32<MAX_BITCOIN_SCRIPT_LENGTH>>,
+	bytes: Vec<u8>,
+}
+
+// We only store (spendable) BitcoinScripts that we authored ourselves.
+// For these, the max encoded length is MAX_BITCOIN_SCRIPT_LENGTH.
+// For anything else, we don't need MaxEncodedLen.
```
dandanlen, line 991: I noticed that PushBytes allows MAX_PUSHABLE_BYTES, but we were storing it in a bounded vec of size MAX_BITCOIN_SCRIPT_LENGTH. 

The only reason we defined this was to derive MaxEncodedLen, so I moved the definition of this here.

dandanlen (commented):
LGTM - I added one small change (see comment). Will let you double-check my commit and then push the button if you're happy.

PR #5314, opened 2024-10-07
feat: minimum chunk size setting

j4m1ef0rd:
# Pull Request

Closes: PRO-1703

## Checklist

## Summary

- Added a storage item for `MinimumChunkSize`
	- Added it to the pallet config so we can update the setting for each asset.
		- Added the new setting to the pallet config unit test.
- Added logic in the `init_swap_request` function that lowers the number of chunks so it will always be >= to the minimum.
	- Added a unit test to cover this new logic
- Added  a `minimum_chunk_size` `AssetMap` to the `SwappingEnvironment`
	- Updated the test snapshot


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -505,6 +508,14 @@ pub mod pallet {
 		ConstU32<DEFAULT_MAX_SWAP_REQUEST_DURATION_BLOCKS>,
 	>;
 
+	/// The minimum chunk size for DCA swaps. The number of chunks of a DCA swap will be reduced
+	/// so that the chunk size is greater than or equal to this value. Setting to zero will disable
+	/// the check for that asset.
+	#[pallet::storage]
+	#[pallet::getter(fn minimum_chunk_size)]
+	pub type MinimumChunkSize<T: Config> =
```
j4m1ef0rd, line 516: The `MinimumChunkSize` will default to 0 for all assets, meaning that there will be no minimum, the check is skipped. 
If we want to set an actual default, I think it will require a migration. Should I make this migration or just let governance set it once its released? @dandanlen

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1964,6 +1983,18 @@ pub mod pallet {
 				swap_amount
 			};
 
+			// Restrict the number of chunks based on the minimum chunk size.
+			let dca_params = dca_params.map(|mut dca_params| {
```
msgmaxim, line 1987: @zoheb391 just double checking that the frontend will be able to handle this (changing the number of dca chunks based on the amount)? I imagine if you are reading the number of chunks from `SwapRequested` rather than by looking at the channel, this shouldn't be a problem?

msgmaxim (approved):
LGMT, but let's see whether or not we want a migration here to set the initial limits.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -505,6 +508,14 @@ pub mod pallet {
 		ConstU32<DEFAULT_MAX_SWAP_REQUEST_DURATION_BLOCKS>,
 	>;
 
+	/// The minimum chunk size for DCA swaps. The number of chunks of a DCA swap will be reduced
+	/// so that the chunk size is greater than or equal to this value. Setting to zero will disable
+	/// the check for that asset.
+	#[pallet::storage]
+	#[pallet::getter(fn minimum_chunk_size)]
+	pub type MinimumChunkSize<T: Config> =
+		StorageMap<_, Twox64Concat, Asset, AssetAmount, ValueQuery>;
```
dandanlen, line 517: Doesn't really need to be stored as an `AssetAmount`. `u32` is plenty, and then we don't have the confusion of returning a NumberOrHex from the rpc.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -448,6 +448,7 @@ pub struct SwappingEnvironment {
 	swap_retry_delay_blocks: u32,
 	max_swap_retry_duration_blocks: u32,
 	max_swap_request_duration_blocks: u32,
+	minimum_chunk_size: any::AssetMap<NumberOrHex>,
```
dandanlen, line 451: See my other comment, `u32` is probably enough, and then we don't need NumberOrHex.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -285,6 +285,7 @@ decl_runtime_apis!(
 		fn cf_swap_retry_delay_blocks() -> u32;
 		fn cf_swap_limits() -> SwapLimits;
 		fn cf_lp_events() -> Vec<pallet_cf_pools::Event<Runtime>>;
+		fn cf_minimum_chunk_size(asset: Asset) -> AssetAmount;
```
dandanlen, line 288: This is a breaking change in the runtime interface. I'll raise an issue to handle this more gracefully (applies more generally, not just to this PR).

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Some assets like ETH have 18 decimal places. Even if we set this so something small like 0.001 ETH, it would still be a large enough number that it won't fit in u32, right?

state-chain/pallets/cf-swapping/src/lib.rs
zoheb391: @msgmaxim The front-end does not have an option to change dca interval params. It is our quoting endpoint that decides on the dca params. if the swap amount is under the configuration amount (~3000usd) we only use 1 chunk so it won't be an issue on our front-end or in our sdk if they use it like we recommend.

dandanlen: @zoheb391  The issue here is that the protocol might change the parameters: if the chosen number of chunks would result in chunks that are smaller than some minimum, we reduce the number of chunks accordingly. This can still happen with our FE. Example: User requests a quote for 10BTC and gets a dca quote for 1000 chunks. But then they only send 0.1BTC. In this case, the protocol would adjust the number of chunks to something smaller than 1000.

zoheb391: ah i see now. yeah we're using the channel for dca params. So that might be a problem for us. We will need to make that switch.

zoheb391: i'm guessing interval remains unchanged though?

dandanlen: > i'm guessing interval remains unchanged though?

Yes, no change to the interval but I would still recommend using the swap request details rather than the channel details for this.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Yes, ignore me. For some reason I was confusing chunk size with number of chunks. AssetAmount makes total sense.

dandanlen (approved):

PR #5323, opened 2024-10-09
chore: add sol min_prio_fee to ingress and egress estimation

albert-llimos:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

Adding the minimum priority fee to the ingress and egress fee calculations.

### DIFF
--- a/state-chain/chains/src/sol.rs
+++ b/state-chain/chains/src/sol.rs
@@ -192,13 +192,13 @@ impl FeeEstimationApi<Solana> for SolTrackedData {
 		);
 
 		// Match the minimum broadcast prio fee
-		let min_priority_fee = sp_std::cmp::max(self.priority_fee, MIN_COMPUTE_PRICE);
+		let priority_fee = sp_std::cmp::max(self.priority_fee, MIN_COMPUTE_PRICE);
 
 		let gas_fee = LAMPORTS_PER_SIGNATURE.saturating_add(
 			// It should never approach overflow but just in case
 			sp_std::cmp::min(
 				SolAmount::MAX as u128,
-				(min_priority_fee as u128 * compute_units_per_transfer as u128)
+				(priority_fee as u128 * compute_units_per_transfer as u128)
 					.div_ceil(MICROLAMPORTS_PER_LAMPORT.into()),
 			) as SolAmount,
 		);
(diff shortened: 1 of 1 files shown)


state-chain/chains/src/sol.rs
```diff
@@ -191,11 +191,14 @@ impl FeeEstimationApi<Solana> for SolTrackedData {
 				},
 		);
 
+		// Match the minimum broadcast prio fee
+		let priority_fee = sp_std::cmp::max(self.priority_fee, MIN_COMPUTE_PRICE);
```
kylezs, line 195: we are maxing a price and against fee? these are two different things no?

albert-llimos: Namings are not great.. I think we named Solana's "compute_units" and "compute_price" as "gas" and "priority_fee" after the EVM so it'd be easy to follow in our codebase, this way it was the same concepts across chains. But then it ends up with  some things like this that don't look right.
Priority fees is the same as compute price.

albert-llimos: To be fair, Solana's rpc calls are for "priorization fees" but then when setting them into an instruction is called "compute_price". So their naming is not consistent.

dandanlen: I think just different words for the same thing (depending on your pov, if you're charging or paying). The price is the minimum fee you need to pay.

kylezs: yeah it seems like a naming thing - we could at the very least clarify with a doc comment that it is an absolute value. when I see "price" I'm thinking ratio of two things, but that's not the case here

state-chain/chains/src/sol.rs
```diff
@@ -219,17 +212,30 @@ impl FeeEstimationApi<Solana> for SolTrackedData {
...
-				(self.priority_fee as u128 * compute_units_per_fetch as u128)
-					.div_ceil(MICROLAMPORTS_PER_LAMPORT.into()),
-			) as SolAmount,
-		)
+		calculate_gas_fee(self.priority_fee, compute_units_per_fetch)
 	}
 }
 
+// Calculate the gas fee for a transaction given its compute units and a priority fee.
+fn calculate_gas_fee(
```
kylezs, line 220: Could impl this on `SolTrackedData` and then we don't need to pass in self.priority_fee

state-chain/chains/src/sol.rs
```diff
@@ -219,17 +212,30 @@ impl FeeEstimationApi<Solana> for SolTrackedData {
...
+	priority_fee: <Solana as crate::Chain>::ChainAmount,
+	compute_units: SolComputeLimit,
+) -> <Solana as crate::Chain>::ChainAmount {
+	use compute_units_costs::*;
+
+	// Match the minimum broadcast prio fee
+	let priority_fee = sp_std::cmp::max(priority_fee, MIN_COMPUTE_PRICE);
```
dandanlen, line 227: Does this mean we been estimating too low so far? (ie. this change will raise the fees?)

state-chain/chains/src/sol.rs
albert-llimos: yes. The compute price is the literal price of a compute unit, to me it makes more sense than calling it priority fee. I'll just clarify it in a comment.

state-chain/chains/src/sol.rs
albert-llimos: yes but the priority fee is in microlamports so it's extremely low. It'd require the priority fee to go up a few orders of magnitude to make any difference. To just be on par with the base fee it'd require 2-3 orders of magnitude.

dandanlen: Sure but we're raising the prio fee to MIN_COMPUTE_PRICE here. Is the MIN_COMPUTE_PRICE really low too?

albert-llimos: yes, extremely low. For our transactions the fee will go from 5000 lamports to 5001 lamports. Can't make any difference to any user.

kylezs (approved):

PR #5325, opened 2024-10-10
fix: keyholder check should use `HistoricalActiveEpochs`

dandanlen:
# Pull Request

Closes: PRO-1721

## Checklist

## Summary

This is more reliable than using HistoricalAuthorities because the latter is not cleared when an epoch expires.

Preivously is was possible for a validator to be considered active for an expired epoch if the epochs were expired out-of-order (see updated test case).


Janislav (approved):
Much clearer 👍

PR #5326, opened 2024-10-11
[DO NOT SQUASH] Release candidate 1.6.9

dandanlen:
This is a cherry pick from main. 

Some alterations were made to the original commit to ensure this is a non-breaking change:

- Instead of adding a new variant to the PalletConfigUpdate enum, replaced MaximumSwapAmount. This ensures that encoding remains compatible. 
- No event is emitted when the min chunk size is updated.
- Some test cases were modified/removed (ie. those relating to the maaximum swap amount.

The main consequence is that we can't set a MaximumSwapAmount, but I think this is fine since we don't use it any more (can probably be removed in a future release). 

(TODO: version bumps, changelog etc.)

### DIFF
--- a/state-chain/pallets/cf-swapping/src/tests/config.rs
+++ b/state-chain/pallets/cf-swapping/src/tests/config.rs
@@ -3,61 +3,52 @@ use super::*;
 #[test]
 fn can_update_all_config_items() {
 	new_test_ext().execute_with(|| {
-		const NEW_MAX_SWAP_AMOUNT_BTC: Option<AssetAmount> = Some(100);
-		const NEW_MAX_SWAP_AMOUNT_DOT: Option<AssetAmount> = Some(69);
 		let new_swap_retry_delay = BlockNumberFor::<Test>::from(1234u32);
 		let new_flip_buy_interval = BlockNumberFor::<Test>::from(5678u32);
 		const NEW_MAX_SWAP_RETRY_DURATION: u32 = 69_u32;
 		const MAX_SWAP_REQUEST_DURATION: u32 = 420_u32;
+		const NEW_MINIMUM_CHUNK_SIZE_ETH: AssetAmount = 1;
+		const NEW_MINIMUM_CHUNK_SIZE_BTC: AssetAmount = 2;
 
 		// Check that the default values are different from the new ones
-		assert!(MaximumSwapAmount::<Test>::get(Asset::Btc).is_none());
-		assert!(MaximumSwapAmount::<Test>::get(Asset::Dot).is_none());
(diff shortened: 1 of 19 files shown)


msgmaxim:
@zoheb391 mentioned in the other PR that current frontend won't be able to handle this correctly. Just double checking that we have made the necessary changes already (or that they will be done before 1.6.8 is released)?

state-chain/pallets/cf-swapping/src/tests/config.rs
```diff
@@ -3,61 +3,46 @@ use super::*;
...
 		// Check that the default values are different from the new ones
-		assert!(MaximumSwapAmount::<Test>::get(Asset::Btc).is_none());
-		assert!(MaximumSwapAmount::<Test>::get(Asset::Dot).is_none());
 		assert_ne!(SwapRetryDelay::<Test>::get(), new_swap_retry_delay);
 		assert_ne!(FlipBuyInterval::<Test>::get(), new_flip_buy_interval);
 		assert_ne!(MaxSwapRetryDurationBlocks::<Test>::get(), NEW_MAX_SWAP_RETRY_DURATION);
 		assert_ne!(MaxSwapRequestDurationBlocks::<Test>::get(), MAX_SWAP_REQUEST_DURATION);
+		assert_ne!(MinimumChunkSize::<Test>::get(Asset::Eth), NEW_MINIMUM_CHUNK_SIZE);
 
 		// Update all config items at the same time, and updates 2 separate max swap amounts.
```
j4m1ef0rd, line 19: Stale comment.

state-chain/pallets/cf-swapping/src/tests/config.rs
```diff
@@ -3,61 +3,46 @@ use super::*;
...
 			OriginTrait::root(),
 			vec![
-				PalletConfigUpdate::MaximumSwapAmount {
-					asset: Asset::Btc,
-					amount: NEW_MAX_SWAP_AMOUNT_BTC
-				},
-				PalletConfigUpdate::MaximumSwapAmount {
-					asset: Asset::Dot,
-					amount: NEW_MAX_SWAP_AMOUNT_DOT
-				},
```
j4m1ef0rd, line 32: was good to test that it can handle 2 of the same update at once.

j4m1ef0rd (approved):

state-chain/pallets/cf-swapping/src/tests/config.rs
dandanlen: We know it can - it's tested on main and I didn't change any of that logic. None of this will be merged back into main.

dandanlen:
Tidied up a bit and added changelogs and version bumps. (Also it's 1.6.9 not 1.6.8)

PR #5328, opened 2024-10-14
fix: don't exit if we can't remove a file that already doesn't exist

msgmaxim:
# Pull Request

## Summary

Ran into this when trying to start a new network. The script shouldn't exit if the file we are trying to remove already doesn't exit, and `-f` achieves this.

### DIFF
--- a/localnet/common.sh
+++ b/localnet/common.sh
@@ -102,7 +102,7 @@ build-localnet() {
   DOT_GENESIS_HASH=$(echo $REPLY | grep -o '\"result\":\"0x[^"]*' | grep -o '0x.*')
 
   echo "🐛 Fix solana symlink issue ..."
-  rm $SOLANA_BASE_PATH/test-ledger/snapshot/100/accounts_hardlinks/account_path_0
+  rm -f $SOLANA_BASE_PATH/test-ledger/snapshot/100/accounts_hardlinks/account_path_0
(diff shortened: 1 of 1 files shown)


localnet/common.sh
```diff
@@ -102,7 +102,7 @@ build-localnet() {
   DOT_GENESIS_HASH=$(echo $REPLY | grep -o '\"result\":\"0x[^"]*' | grep -o '0x.*')
 
   echo "🐛 Fix solana symlink issue ..."
-  rm $SOLANA_BASE_PATH/test-ledger/snapshot/100/accounts_hardlinks/account_path_0
+  rm -f $SOLANA_BASE_PATH/test-ledger/snapshot/100/accounts_hardlinks/account_path_0
   ln -s $SOLANA_BASE_PATH/test-ledger/accounts/snapshot/100 $SOLANA_BASE_PATH/test-ledger/snapshot/100/accounts_hardlinks/account_path_0
```
anton-chain, line 106:
```suggestion
  ln -sf $SOLANA_BASE_PATH/test-ledger/accounts/snapshot/100 $SOLANA_BASE_PATH/test-ledger/snapshot/100/accounts_hardlinks/account_path_0
```
maybe let's apply force to `ln` command?

anton-chain (approved):

kylezs:
hit flaky case which is already tracked

PR #5329, opened 2024-10-14
feat: broker can encode btc smart contract call

j4m1ef0rd:
# Pull Request

Closes: PRO-1648

## Checklist

## Summary

- new `broker_encodeBtcSmartContractCall` rpc that will create an encoded btc start contract call with the given params.
- Extended the `SwapLimitsProvider` trait to validate the dca and refund params.
	- Added an RPC to use this param validation so that the broker api can use it. 

### TODO
- Documentation

### DIFF
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -7,7 +7,7 @@ use chainflip_api::{
 	primitives::{AccountRole, Affiliates, Asset, BasisPoints, CcmChannelMetadata, DcaParameters},
 	settings::StateChain,
 	AccountId32, AddressString, BrokerApi, OperatorApi, RefundParameters, StateChainApi,
-	SwapDepositAddress, WithdrawFeesDetail,
+	SwapDepositAddress, SwapPayload, WithdrawFeesDetail,
 };
 use clap::Parser;
 use custom_rpc::to_rpc_error;
@@ -49,16 +49,17 @@ pub trait Rpc {
 		destination_address: AddressString,
 	) -> RpcResult<WithdrawFeesDetail>;
 
-	#[method(name = "encode_btc_smart_contract_call", aliases = ["broker_encodeBtcSmartContractCall"])]
-	async fn encode_btc_smart_contract_call(
+	#[method(name = "request_swap_parameter_encoding", aliases = ["broker_requestSwapParameterEncoding"])]
+	async fn request_swap_parameter_encoding(
 		&self,
+		input_asset: Asset,
 		output_asset: Asset,
 		output_address: AddressString,
 		retry_duration: u32,
 		min_output_amount: u128,
 		boost_fee: Option<BasisPoints>,
 		dca_parameters: Option<DcaParameters>,
-	) -> RpcResult<Vec<u8>>;
+	) -> RpcResult<SwapPayload>;
 }
 
 pub struct RpcServerImpl {
@@ -132,19 +133,21 @@ impl RpcServer for RpcServerImpl {
 			.map_err(to_rpc_error)?)
 	}
 
-	async fn encode_btc_smart_contract_call(
+	async fn request_swap_parameter_encoding(
 		&self,
+		input_asset: Asset,
 		output_asset: Asset,
 		output_address: AddressString,
 		retry_duration: u32,
 		min_output_amount: u128,
 		boost_fee: Option<BasisPoints>,
 		dca_parameters: Option<DcaParameters>,
-	) -> RpcResult<Vec<u8>> {
+	) -> RpcResult<SwapPayload> {
 		Ok(self
 			.api
 			.broker_api()
-			.encode_btc_smart_contract_call(
+			.request_swap_parameter_encoding(
+				input_asset,
 				output_asset,
 				output_address,
 				retry_duration,
@@ -153,8 +156,7 @@ impl RpcServer for RpcServerImpl {
 				dca_parameters,
 			)
 			.await
-			.map_err(to_rpc_error)?
-			.raw())
+			.map_err(to_rpc_error)?)
 	}
 }
 
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -1,13 +1,12 @@
 use cf_utilities::{
 	health::{self, HealthCheckOptions},
+	rpc::NumberOrHex,
 	task_scope::{task_scope, Scope},
+	try_parse_number_or_hex,
 };
 use chainflip_api::{
 	self,
-	primitives::{
-		AccountRole, Affiliates, Asset, BasisPoints, Beneficiaries, CcmChannelMetadata,
-		DcaParameters,
-	},
+	primitives::{AccountRole, Affiliates, Asset, BasisPoints, CcmChannelMetadata, DcaParameters},
 	settings::StateChain,
 	AccountId32, AddressString, BrokerApi, OperatorApi, RefundParameters, StateChainApi,
 	SwapDepositAddress, SwapPayload, WithdrawFeesDetail,
@@ -55,14 +54,15 @@ pub trait Rpc {
 	#[method(name = "request_swap_parameter_encoding", aliases = ["broker_requestSwapParameterEncoding"])]
 	async fn request_swap_parameter_encoding(
 		&self,
-		input_asset: Asset,
-		output_asset: Asset,
-		output_address: AddressString,
+		source_asset: Asset,
+		destination_asset: Asset,
+		destination_address: AddressString,
+		broker_commission: BasisPoints,
 		retry_duration: u32,
-		min_output_amount: u128,
-		boost_fee: Option<BasisPoints>,
+		min_output_amount: NumberOrHex,
 		dca_parameters: Option<DcaParameters>,
-		broker_fees: Option<Beneficiaries<AccountId32>>,
+		boost_fee: Option<BasisPoints>,
+		affiliate_fees: Option<Affiliates<AccountId32>>,
 	) -> RpcResult<SwapPayload>;
 }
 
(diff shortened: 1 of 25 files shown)


api/lib/src/lib.rs
```diff
@@ -515,6 +523,72 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 		self.simple_submission_with_dry_run(pallet_cf_swapping::Call::deregister_as_broker {})
 			.await
 	}
+
+	async fn encode_btc_smart_contract_call(
```
CumpsD, line 527: Does this take brokers and affiliate brokers into account too?

dandanlen: Not yet. For btc it's not trivial, will be added separately.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -48,6 +48,17 @@ pub trait Rpc {
...
+	#[method(name = "encode_btc_smart_contract_call", aliases = ["broker_encodeBtcSmartContractCall"])]
+	async fn encode_btc_smart_contract_call(
+		&self,
+		output_asset: Asset,
+		output_address: AddressString,
+		retry_duration: u32,
+		min_output_amount: u128,
+		boost_fee: Option<BasisPoints>,
+		dca_parameters: Option<DcaParameters>,
+	) -> RpcResult<Vec<u8>>;
```
dandanlen, line 61: Vec<u8> is not a very json-friendly return value. At the very least it should be hex-encoded, but I think we can also do better to communicate *what* it is. For example, we could make it an object with a named field. Combined with the above suggestion, a suitable output type might be:

```
pub enum SwapPayload {
	Bitcoin {
		nulldata_utxo: Bytes, // sp_core::Bytes serializes as hex
	},
	// Other variants to be added later: Eth, Sol
}
```

api/lib/src/lib.rs
```diff
@@ -321,8 +329,8 @@ impl AddressString {
 			.map_err(|_| anyhow!("Failed to parse address"))
 	}
 
-	pub fn from_encoded_address(address: &EncodedAddress) -> Self {
-		Self(address.to_string())
+	pub fn from_encoded_address<T: std::borrow::Borrow<EncodedAddress>>(address: T) -> Self {
+		Self(address.borrow().to_string())
```
dandanlen, line 333: What's the reason for this change?

api/lib/src/lib.rs
CumpsD: > Not yet. For btc it's not trivial, will be added separately.

Ok, so to start I will just provide this functionality via BaaS, and if I am correct, BTC -> x swaps initiated this way will start out with 0% commission to the affiliate (and main broker), right?

api/lib/src/lib.rs
```diff
@@ -515,6 +523,72 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
...
+				}
+				let total_swap_request_duration = params
+					.number_of_chunks
+					.saturating_sub(1)
+					.checked_mul(params.chunk_interval)
+					.ok_or(anyhow!("Invalid DCA parameters"))?;
+
+				if total_swap_request_duration > max_swap_request_duration_blocks {
+					bail!("Invalid DCA parameters: Total swap request duration must be <= {max_swap_request_duration_blocks} blocks");
+				}
```
dandanlen, line 567: This is a duplication of logic form the ingress pallet - would be good to put this in shared function, otherwise it's easy to introduce inconsistencies.

api/lib/src/lib.rs
```diff
@@ -515,6 +523,72 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
...
+					.try_into()?,
+				chunk_interval: dca_parameters
+					.as_ref()
+					.map(|params| params.chunk_interval)
+					.unwrap_or(2)
+					.try_into()?,
+				boost_fee: boost_fee.unwrap_or_default().try_into()?,
+			},
+		};
+		Ok(encode_swap_params_in_nulldata_utxo(params))
```
dandanlen, line 590: Might be worth clarifying if we want to return the encoded utxo, or just the data (ie. whether we include the OP_RETURN or not).

api/lib/src/lib.rs
dandanlen: Adding broker fees is a separate issue from what is being added here. By the time release this, we may also have support for broker/affiliate fees.

Having said that, there is no real reason to prefer this method over the other unless for whatever reason you don't want to wait 6secs to open a deposit channel.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -48,6 +48,17 @@ pub trait Rpc {
 		asset: Asset,
 		destination_address: AddressString,
 	) -> RpcResult<WithdrawFeesDetail>;
+
+	#[method(name = "encode_btc_smart_contract_call", aliases = ["broker_encodeBtcSmartContractCall"])]
+	async fn encode_btc_smart_contract_call(
```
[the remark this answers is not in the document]
CumpsD: `sign_swap_payload`?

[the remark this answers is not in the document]
msgmaxim: How about `request_swap_parameter_encoding`?

api/lib/src/lib.rs
msgmaxim: IMO returning just the data makes more sense. Integratorss might feel more in control if they are the ones who create the actual UTXO (e.g. ensuring that it is non-spendable etc). Also, this is going to be more consistent with other chains when we add them.

api/lib/src/lib.rs
j4m1ef0rd: Just to fix rust analyser complaining about it. If i just removed the reference then clippy would complain about it, so this was the best solution that made both happy.

api/bin/chainflip-broker-api/src/main.rs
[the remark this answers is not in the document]
j4m1ef0rd: > it should be chain-agnostic

How does the user specify which encoding they want? should I add a parameter for the encodeing type/chain?
or just return all encoding types for the swap?

api/lib/src/lib.rs
kylezs, line 527: Thinking, could we add a safe mode param for this? So if safe mode is enabled, this query doesn't work. It's not bullet proof since people could construct it on their own without using this - but more than likely people will be using this, and we can prevent the creation of these memo calls

api/bin/chainflip-broker-api/src/main.rs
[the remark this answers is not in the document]
dandanlen: The input asset determines the chain. The output can be an enum with one variant per chain.

msgmaxim (commented):
Looks good (apart from things that Dan already commented on)

api/lib/src/lib.rs
```diff
@@ -515,6 +528,75 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
...
+			bail!("Safe mode is active. Swaps are disabled.");
+		}
+
+		// Validate params
+		let params = serde_json::to_string(&RefundValidationParams { retry_duration })?;
+		let raw_value = serde_json::value::RawValue::from_string(params)?;
+		self.base_rpc_api()
+			.request_raw("cf_validate_refund_params", Some(raw_value))
+			.await
+			.map_err(anyhow::Error::msg)?;
```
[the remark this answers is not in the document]
j4m1ef0rd: I couldnt find a way to use them directly from the `BrokerApi` so i added them to the `BaseRpcApi`.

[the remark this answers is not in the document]
dandanlen: Cool - would be good to find a nice way to expose the api directly instead of having to add this boilerplate... can wait for another time though.

dandanlen:
Adding the broker fees is a nice idea but we should not just silently ignore them - better to return an error message for now.

dandanlen:
> Adding the broker fees is a nice idea but we should not just silently ignore them - better to return an error message for now.

☝️ Will add this and then merge, thanks @j4m1ef0rd

dandanlen:
Actually, I'll also rename the params so that they match `request_swap_deposit_address` as closely as possible - will make documentation much easier.

dandanlen (approved):

CumpsD:
Any reason you didnt keep this (https://github.com/chainflip-io/chainflip-backend/blob/main/api/bin/chainflip-broker-api/src/main.rs#L61-L65) consistent with the order of https://github.com/chainflip-io/chainflip-backend/blob/main/api/bin/chainflip-broker-api/src/main.rs#L41-L44 ?

CumpsD:
Also, my Rust is rusty, but here you got `affiliate_fees.unwrap_or_default(),` https://github.com/chainflip-io/chainflip-backend/blob/main/api/bin/chainflip-broker-api/src/main.rs#L119C5-L119C40

While here you dont: https://github.com/chainflip-io/chainflip-backend/blob/main/api/bin/chainflip-broker-api/src/main.rs#L164

CumpsD:
Out of curiosity, how are refunds handled in this way? Since there is no refund address, where do you get it from?

j4m1ef0rd:
> Out of curiosity, how are refunds handled in this way? Since there is no refund address, where do you get it from?

The refund address is stored in the 3rd output of the bitcoin transaction. Will be explained in the docs.

> but here you got affiliate_fees.unwrap_or_default()

We would just unwrao_or_default in the next function. Does not really matter at what point it becomes non optional. You're right, we should be consistent.

> Any reason you didnt keep this consistent with the order of

At first it was because I wanted to put the optional params at the end also some of the params where different, so it could not be exactly the same, but Dan changed it. So now there is no reason they are not consistent 😅.

dandanlen:
Re. consistency:

I tried to keep it as consistent as possible. 

There are some necessary differences between requesting a deposit channel vs. requesting a transaction payload. 

The specific reason for the change in order in this case is that retry_duration and min_output_amount are non-optional and replace `Option<RefundParameters>` from the first method. As @j4m1ef0rd mentioned, we keep optional parameters to the end because when passing the parameters in an array, having optional params in the middle of the array creates some weird ambiguities (should it be `null`? should it be left out? etc.). 

Unifying this would have required changing the original api so I opted not to (for now). My main concern was for naming. 

In general, however, I would highly recommend using on object rather than an array for passing the parameters, then the order is a non-issue.

CumpsD:
Object it is! thanks

PR #5331, opened 2024-10-15
feat: witnessing btc smart contract swaps

msgmaxim:
# Pull Request

Closes: PRO-1695, PRO-1644, PRO-1643

## Summary

- `OnDeposit::on_deposit_made` no longer takes `DepositChannel` as a parameter (in case of smart contract swaps, there is no channel). It was only used by bitcoin to pass deposit address. Instead, now engines submit deposit address as part of `DepositDetails` (next to the existing `UtxoId`).
- Added `deposit_details` as a parameter to `contract_swap_request`.
- Added FoK/DCA/Boost parameters to `contract_swap_request`. FoK and DCA should already by handled correctly on the SC side, however Boost is currently ignored as it will require some potentially non-trivial refactoring. For now the engine uses default parameters for these in case of EVM smart contracts (i.e. they are still effectively disabled).
- On every BTC block the engine checks contract calls into our vault’s change addresses (current and previous), and calls `contract_swaps_request` with the decoded parameters.

## Limitations

- DCA and FoK parameters aren't yet validated (might be worth waiting for https://github.com/chainflip-io/chainflip-backend/pull/5329)
- Boost parameter is currently ignored in contract swaps (needs a potentially non-trivial refactor, so I'd rather we do it in a separate PR)
- No support for "special smart contract" channels yet. Only deposits directly into the vault's change address are witnessed.

FYI @albert-llimos

### DIFF
--- a/engine/src/witness/btc/deposits.rs
+++ b/engine/src/witness/btc/deposits.rs
@@ -144,9 +144,9 @@ fn deposit_witnesses(
 }
 
 fn map_script_addresses(
-	addresses: Vec<DepositChannelDetails<state_chain_runtime::Runtime, BitcoinInstance>>,
+	deposit_channels: Vec<DepositChannelDetails<state_chain_runtime::Runtime, BitcoinInstance>>,
 ) -> HashMap<Vec<u8>, DepositAddress> {
-	addresses
+	deposit_channels
 		.into_iter()
 		.map(|channel| {
 			assert_eq!(channel.deposit_channel.asset, btc::Asset::Btc);
--- a/engine/src/witness/btc/deposits.rs
+++ b/engine/src/witness/btc/deposits.rs
@@ -168,10 +168,7 @@ pub mod tests {
 		absolute::{Height, LockTime},
 		Amount, ScriptBuf, Txid,
 	};
-	use cf_chains::{
-		btc::{deposit_address::DepositAddress, ScriptPubkey},
-		DepositChannel,
-	};
+	use cf_chains::{btc::deposit_address::DepositAddress, DepositChannel};
 	use pallet_cf_ingress_egress::{BoostStatus, ChannelAction};
 	use rand::{seq::SliceRandom, Rng, SeedableRng};
 	use sp_runtime::AccountId32;
@@ -195,16 +192,16 @@ pub mod tests {
 	}
 
 	fn fake_details(
-		address: ScriptPubkey,
+		deposit_address: DepositAddress,
 	) -> DepositChannelDetails<state_chain_runtime::Runtime, BitcoinInstance> {
 		DepositChannelDetails::<_, BitcoinInstance> {
 			opened_at: 1,
 			expires_at: 10,
 			deposit_channel: DepositChannel {
 				channel_id: 1,
-				address,
+				address: deposit_address.script_pubkey(),
 				asset: btc::Asset::Btc,
-				state: DepositAddress::new([0; 32], 1),
+				state: deposit_address,
 			},
 			action: ChannelAction::<AccountId32>::LiquidityProvision {
 				lp_account: AccountId32::new([0xab; 32]),
@@ -214,14 +211,16 @@ pub mod tests {
 		}
 	}
 
-	pub fn fake_verbose_vouts(vals_and_scripts: Vec<(u64, Vec<u8>)>) -> Vec<VerboseTxOut> {
-		vals_and_scripts
+	pub fn fake_verbose_vouts(
+		amounts_and_addresses: Vec<(u64, &DepositAddress)>,
+	) -> Vec<VerboseTxOut> {
+		amounts_and_addresses
 			.into_iter()
 			.enumerate()
-			.map(|(n, (value, script_bytes))| VerboseTxOut {
-				value: Amount::from_sat(value),
+			.map(|(n, (amount, address))| VerboseTxOut {
+				value: Amount::from_sat(amount),
 				n: n as u64,
-				script_pubkey: ScriptBuf::from(script_bytes),
+				script_pubkey: ScriptBuf::from(address.script_pubkey().bytes()),
 			})
 			.collect()
 	}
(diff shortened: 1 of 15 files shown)


engine/src/witness/btc/deposits.rs
```diff
@@ -115,15 +145,17 @@ fn deposit_witnesses(
 		.collect()
 }
 
-fn script_addresses(
+fn map_script_addresses(
 	addresses: Vec<DepositChannelDetails<state_chain_runtime::Runtime, BitcoinInstance>>,
```
j4m1ef0rd, line 147:
```suggestion
	deposit_channels: Vec<DepositChannelDetails<state_chain_runtime::Runtime, BitcoinInstance>>,
```

engine/src/witness/btc/smart_contract.rs
```diff
@@ -220,13 +226,12 @@ mod tests {
 	fn test_extract_contract_call_from_tx() {
```
j4m1ef0rd, line 226: Would be nice to have a test to make sure the encode does not change. Like we have done in other places.

engine/src/witness/btc/smart_contract.rs
```diff
@@ -86,19 +74,21 @@ fn script_buf_to_script_pubkey(script: &ScriptBuf) -> Option<ScriptPubkey> {
 	Some(pubkey)
 }
 
+type BtcIngressEgressCall =
+	pallet_cf_ingress_egress::Call<state_chain_runtime::Runtime, BitcoinInstance>;
+
 // Currently unused, but will be used by the deposit wintesser:
```
j4m1ef0rd, line 80: stale comment

j4m1ef0rd (approved):

engine/src/witness/btc/smart_contract.rs
```diff
@@ -86,19 +74,21 @@ fn script_buf_to_script_pubkey(script: &ScriptBuf) -> Option<ScriptPubkey> {
 	Some(pubkey)
 }
 
+type BtcIngressEgressCall =
+	pallet_cf_ingress_egress::Call<state_chain_runtime::Runtime, BitcoinInstance>;
```
kylezs, line 78: do we have any protection from the last PR regarding eth and arb contract swaps over the migration, safe mode doesn't apply to contract swaps since they are CFE triggered, so we should make sure we are not messing with the witnessed calls during the migration

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1118,16 +1123,20 @@ pub mod pallet {
 					},
 				};
 
+			T::DepositHandler::on_deposit_made(deposit_details, deposit_amount);
+
+			// TODO: ensure minimum deposit?
+
+			// TODO: validate dca_params and refund_params
```
kylezs, line 1130: do we have tickets for this?

engine/src/witness/btc/deposits.rs
```diff
@@ -56,12 +56,37 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
+
+					[current_vault_address].into_iter().chain(maybe_previous_vault_address)
+				};
+
 				// TODO: Make addresses a Map of some kind?
-				let (((), txs), addresses) = header.data;
+				let (((), txs), deposit_channels) = header.data;
+
+				for vault_address in vault_addresses {
+					for tx in &txs {
```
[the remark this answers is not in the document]
msgmaxim: I prefer to keep these separate.

engine/src/witness/btc/smart_contract.rs
msgmaxim: A protection where we block the runtime upgrade if we have existing witness calls?

engine/src/witness/btc/smart_contract.rs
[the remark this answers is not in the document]
msgmaxim: I will create a ticket for this. Might be worth getting [Jamies PR](https://github.com/chainflip-io/chainflip-backend/pull/5329) merged first.

state-chain/pallets/cf-ingress-egress/src/lib.rs
j4m1ef0rd: Im working on it in my PR, if i can get it in soon, then this can be completed in this PR.

engine/src/witness/btc/deposits.rs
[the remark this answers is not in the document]
kylezs: You can still keep them separate but do it in a single loop.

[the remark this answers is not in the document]
kylezs: The story we are telling with the code here is:
We have a block with transactions and we process these transactions into two transactions types.

It is not:
We have a block with transactions which we process into one transaction type. And then we process the transactions again into another transaction type.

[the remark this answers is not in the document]
kylezs: Having a single loop directly reflects what we're doing on a conceptual level

engine/src/witness/btc/smart_contract.rs
kylezs: Yes, the issue is that if 50% of nodes have witnessed before the upgrade and 50% haven't, then after the upgrade we won't be able to reach consensus.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: Created tickets now

engine/src/witness/btc/smart_contract.rs
dandanlen: I feel like the chance of this happening is remote enough that we can ignore it. 

It can't be exploited (you just lose funds). And if it does happen we have a path to recovery (governance intervention, a bit painful but ok). 

The most straightforward way do deal with this would be to create a new extrinsic and consider the old one to be deprecated, but given that we're making refund address mandatory (and there's way to default it), there's also no simple way to convert 'old' to 'new'. We'd end up keeping and maintaining the old code until the following release.

kylezs (approved):
As discussed, can look at refining the code once the code settles a bit

msgmaxim:
As suspected the reason for hitting the limit on `RuntimeCall`'s size was the `contract_swap_request` call. In particular, the most problematic field seems to be `DepositDetails` which in case of BTC now includes `DepositAddress`, which by itself is already 128 bytes. I went with the easiest approach of wrapping `DepositDetails` in a `Box`, and now `RuntimeCall` sits at 304 (with a bit of room to grow). As the link (mentioned in the test) suggests, there shouldn't be any implication to any external consumers.

I did also consider optimising the size of `DepositAddress`. Since it can be deterministically constructed from `(pubkey: [u8;32], salt: u32)`, we shouldn't have to pass around/store the expanded version that takes almost 4x the space. However, using the more compact representation would likely lead to some breaking changes, so it's probably not what we want here. Maybe something to consider in the future though? @dandanlen

dandanlen:
> I did also consider optimising the size of `DepositAddress`. Since it can be deterministically constructed from `(pubkey: [u8;32], salt: u32)`, we shouldn't have to pass around/store the expanded version that takes almost 4x the space. However, using the more compact representation would likely lead to some breaking changes, so it's probably not what we want here. Maybe something to consider in the future though? @dandanlen

Yeah we can consider this - a simple change would be to add just `(pubkey, salt)` (or even previous/current ket as an enum) in the extrinsic payload, and then generate and store the DepositAddress on-chain, rather than submitting the constructed value.

PR #5332, opened 2024-10-15
feature/PRO-1712/reject-tx

Janislav:
# Pull Request

Closes: PRO-1712

## Checklist

## Summary

#### Non-Breaking changes


state-chain/chains/src/btc/api.rs
```diff
@@ -146,6 +147,38 @@ impl<E: ReplayProtectionProvider<Bitcoin>> ExecutexSwapAndCall<Bitcoin> for Bitc
...
 
+impl<E: ReplayProtectionProvider<Bitcoin>> RejectCall<Bitcoin> for BitcoinApi<E>
+where
+	E: ChainEnvironment<UtxoSelectionType, SelectedUtxosAndChangeAmount>
+		+ ChainEnvironment<(), AggKey>,
+{
+	type DepositDetails = <Bitcoin as Chain>::DepositDetails;
+	fn new_unsigned(deposit_details: Self::DepositDetails) -> Result<Self, RejectError> {
+		if let Some(utxo) = E::lookup(UtxoSelectionType::TakeUtxo(deposit_details.utxo_id)) {
+			if utxo.0.len() != 1 {
```
dandanlen, line 158: We should not need to do any lookups in the environment.

The deposit_details (or whatever we pass in) needs to contain the utxo that we want to reject and the refund address (probably the fees too).

We want to construct a bitcoin transaction that sends the whole output back to source. So we will need the utxo and the refund address. We take the input utxo, and construct an output utxo with destination 'refund_address' and amount 'input - fees'.

Importantly, we don't want to receive any change from this operation because it's tainted, so we don't need to add a change utxo.

state-chain/chains/src/btc/api.rs
```diff
@@ -146,6 +147,34 @@ impl<E: ReplayProtectionProvider<Bitcoin>> ExecutexSwapAndCall<Bitcoin> for Bitc
...
+	) -> Result<Self, RejectError> {
+		let agg_key = <E as ChainEnvironment<(), AggKey>>::lookup(()).ok_or(RejectError::Other)?;
+		let script_pubkey = match refund_address {
+			ForeignChainAddress::Btc(script_pubkey) => script_pubkey,
+			_ => return Err(RejectError::Other),
+		};
+		Ok(Self::NoChangeTransfer(BitcoinTransaction::create_new_unsigned(
+			&agg_key,
+			vec![Utxo {
+				amount,
```
msgmaxim, line 169: The input amount has to be different from the one specified in the output, otherwise the gas fee is 0, correct? Looks like `amount` passed here already excludes egress fee, but I think the input amount should be the full amount of the UTXO that we want to refund.

dandanlen: Bitcoin deposit details will need to be changed to Utxo.

state-chain/chains/src/lib.rs
```diff
@@ -522,10 +525,25 @@ pub enum ConsolidationError {
...
+	NotSupportedForAsset,
+	Other,
+}
+
 pub trait ConsolidateCall<C: Chain>: ApiCall<C::ChainCrypto> {
 	fn consolidate_utxos() -> Result<Self, ConsolidationError>;
 }
 
+pub trait RejectCall<C: Chain>: ApiCall<C::ChainCrypto> {
+	type DepositDetails: Member + Parameter + Unpin + BenchmarkValue;
```
dandanlen, line 539: This type seems unnecessary: should always be C::DpositDetails.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -912,6 +928,30 @@ pub mod pallet {
...
+							tx.amount - egress_fee,
+						) {
+						let result = T::Broadcaster::threshold_sign_and_broadcast(api_call);
+						Self::deposit_event(Event::<T, I>::TaintedTransactionRejected {
+							broadcast_id: result.0,
+							tx_id: tx.deposit_details.clone(),
+						});
+					} else {
+						FailedRejections::<T, I>::append(tx);
+						log_or_panic!("Failed to construct reject call.");
```
dandanlen, line 948: Let's emit an event here instead of using log_or_panic. It's not a logic error.

dandanlen (approved):

PR #5333, opened 2024-10-16
chore: unify dependencies to root cargo.toml

syan095:
# Pull Request

Closes: PRO-1720

## Checklist

- [x] I have written and tested the required migrations.

# Summary
Dependencies are defined in the root cargo.toml. In other cargol.tomls, use 
``` crate = {workspace = true } ```

Refactored single quotes into double quotes in all cargo.toml.

# Updated crates that involve code changes

* secp256k1 0.27 -> 0.29.1 (coupled with Bitcoin 0.32.3)
* Bitcoin to 0.32.3 
* Sha2 to 0.10.7
* Tokio to 1.40.0
* Subxt to 0.37

### CODE REVIEW

**Review by msgmaxim — APPROVED** [review]

**Comment by dandanlen** [comment]
Nice work - why did you leave out the substrate dependencies?

**Comment by syan095** [comment]
> Nice work - why did you leave out the substrate dependencies?

Initially, I thought PolkadotSdk only list third party dependencies in the root, but upon closer inspections, it looks like they have all dependencies listed in the root, including sub-crates. 
I'll make the additional changes.

File: api/lib/Cargo.toml [hunk 1]
```diff
@@ -7,51 +7,51 @@ edition = "2021"
...
-pallet-cf-threshold-signature = { path = "../../state-chain/pallets/cf-threshold-signature" }
-pallet-cf-validator = { path = "../../state-chain/pallets/cf-validator" }
-state-chain-runtime = { path = "../../state-chain/runtime" }
+cf-amm = { workspace = true }
+cf-chains = { workspace = true }
+cf-primitives = { workspace = true }
+chainflip-node = { workspace = true }
+custom-rpc = { workspace = true }
+pallet-cf-account-roles = { workspace = true }
+pallet-cf-environment = { workspace = true }
```
dandanlen, line 39: Strictly speaking, I think these should activate default features when it's in a std context (and to remain consistent with the previous dependency declaration). 

The way it's done in the polkadot repo is:

```
pallet-xx.workspace = true
pallet-xx.default-fetaures = true
```

I think we should follow the same convention.

This is true in a few places, not just here. For example in the node, the engine, etc. Anywhere were we previously *didn't* explicitly set `default-features = false` we now need to override our workspace default and set `default-features = true`.

syan095: Right. 
I've added `default-features = true` for al the non- `no-std` crates, as well as dev-dependencies.

dandanlen (approved):

PR #5334, opened 2024-10-16
refactor: split out composite electoral system trait

kylezs:
# Pull Request

Closes: PRO-1675

## Checklist

## Summary

Rather than have a composite "electoral system" impl Electoral system, we now have a new `ElectoralSystemRunner` trait for the composite, which is then used by the elections pallet directly. This removes some prior hacks that were required to get the traits to work for both a single and composite electoral system.

It also makes it easier to write migration utilities in the future, since storage will *always* be enum wrapped, as a composite, and we don't need to consider the individual electoral system case.

As a consequence there were other simplifications that fell out:
- Simpler Composite electoral system macro
- Some trait methods now have simpler signatures (e.g. no Result necessary in some cases, fewer params).
- Removes unnecessary complex borrowing in macros by simply using the traits and associated functions directly - which also maps better to how substrate works, data is written into a static storage layer.
- Removes translators 🙌  - much clearer what's happening here now, as we can directly use a CompositeElectoralSystem and provide it with the ElectoralSystem we want to access.
- For the tests I just took the approach of most flexibility for now, the model in the tests also maps nicely to how it actually works with respect to the storage, so can save building a separate mental model.

### DIFF
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -69,7 +69,7 @@ jobs:
     uses: ./.github/workflows/_40_post_check.yml
     secrets: inherit
     with:
-      full-bouncer: true
+      full-bouncer: false
       ngrok: true
   test-benchmarks:
     needs: [build-benchmarks]
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: true
+      run-job: false
   publish:
     needs: [package]
     uses: ./.github/workflows/_30_publish.yml
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -69,7 +69,7 @@ jobs:
     uses: ./.github/workflows/_40_post_check.yml
     secrets: inherit
     with:
-      full-bouncer: false
+      full_bouncer: true
       ngrok: true
   test-benchmarks:
     needs: [build-benchmarks]
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
   publish:
     needs: [package]
     uses: ./.github/workflows/_30_publish.yml
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -69,7 +69,7 @@ jobs:
     uses: ./.github/workflows/_40_post_check.yml
     secrets: inherit
     with:
-      full_bouncer: true
+      full-bouncer: true
       ngrok: true
   test-benchmarks:
     needs: [build-benchmarks]
--- a/state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
@@ -393,7 +393,7 @@ macro_rules! generate_electoral_system_tuple_impls {
             fn unsynchronised_state_map(
                 key: &$current::ElectoralUnsynchronisedStateMapKey,
             ) -> Result<Option<$current::ElectoralUnsynchronisedStateMapValue>, CorruptStorageError> {
-                match StorageAccess::unsynchronised_state_map(&CompositeElectoralUnsynchronisedStateMapKey::$current(key.clone()))? {
+                match StorageAccess::unsynchronised_state_map(&CompositeElectoralUnsynchronisedStateMapKey::$current(key.clone())) {
                     Some(CompositeElectoralUnsynchronisedStateMapValue::$current(value)) => Ok(Some(value)),
                     None => Ok(None),
                     _ => Err(CorruptStorageError::new()),
@@ -423,17 +423,18 @@ macro_rules! generate_electoral_system_tuple_impls {
                 unsynchronised_state: $current::ElectoralUnsynchronisedState,
             ) -> Result<(), CorruptStorageError> {
                 let ($($previous,)* _, $($remaining,)*) = StorageAccess::unsynchronised_state()?;
-                StorageAccess::set_unsynchronised_state(($($previous,)* unsynchronised_state, $($remaining,)*))
+                StorageAccess::set_unsynchronised_state(($($previous,)* unsynchronised_state, $($remaining,)*));
+                Ok(())
             }
 
             fn set_unsynchronised_state_map(
                 key: $current::ElectoralUnsynchronisedStateMapKey,
                 value: Option<$current::ElectoralUnsynchronisedStateMapValue>,
-            ) -> Result<(), CorruptStorageError> {
+            ) {
                 StorageAccess::set_unsynchronised_state_map(
                     CompositeElectoralUnsynchronisedStateMapKey::$current(key),
                     value.map(CompositeElectoralUnsynchronisedStateMapValue::$current),
-                )
+                );
             }
 
             fn mutate_unsynchronised_state<
@@ -446,7 +447,7 @@ macro_rules! generate_electoral_system_tuple_impls {
             ) -> Result<T, CorruptStorageError> {
                 let ($($previous,)* mut unsynchronised_state, $($remaining,)*) = StorageAccess::unsynchronised_state()?;
                 let t = f( &mut unsynchronised_state)?;
-                StorageAccess::set_unsynchronised_state(($($previous,)* unsynchronised_state, $($remaining,)*))?;
+                StorageAccess::set_unsynchronised_state(($($previous,)* unsynchronised_state, $($remaining,)*));
                 Ok(t)
             }
         }
--- a/state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
@@ -43,8 +43,8 @@ macro_rules! generate_electoral_system_tuple_impls {
                     ElectionIdentifierOf,
                     ConsensusStatus,
                 },
-                electoral_system_runner::{ElectoralSystemRunner, AuthorityVoteOf, RunnerStorageAccessTrait,
-                    VotePropertiesOf, CompositeConsensusVotes, CompositeElectionIdentifierOf, CompositeConsensusVote},
+                electoral_system_runner::{ElectoralSystemRunner, CompositeAuthorityVoteOf, RunnerStorageAccessTrait,
+                    CompositeVotePropertiesOf, CompositeConsensusVotes, CompositeElectionIdentifierOf, CompositeConsensusVote},
                 vote_storage::{
                     AuthorityVote,
                     VoteStorage
@@ -133,14 +133,14 @@ macro_rules! generate_electoral_system_tuple_impls {
                 fn is_vote_desired(
                     election_identifier: ElectionIdentifier<Self::ElectionIdentifierExtra>,
                     current_vote: Option<(
-                        VotePropertiesOf<Self>,
-                        AuthorityVoteOf<Self>,
+                        CompositeVotePropertiesOf<Self>,
+                        CompositeAuthorityVoteOf<Self>,
                     )>,
                 ) -> Result<bool, CorruptStorageError> {
                     match *election_identifier.extra() {
                         $(CompositeElectionIdentifierExtra::$electoral_system(extra) => {
                             <$electoral_system as ElectoralSystem>::is_vote_desired(
-                                &CompositeElectionAccess::<tags::$electoral_system, $electoral_system, StorageAccess>::new(election_identifier.with_extra(extra)),
+                                &DerivedElectionAccess::<tags::$electoral_system, $electoral_system, StorageAccess>::new(election_identifier.with_extra(extra)),
                                 current_vote.map(|(properties, vote)| {
                                     Ok((
                                         match properties {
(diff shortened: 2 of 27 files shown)


.github/workflows/ci-development.yml
```diff
@@ -69,7 +69,7 @@ jobs:
     uses: ./.github/workflows/_40_post_check.yml
     secrets: inherit
     with:
-      full_bouncer: false
+      full_bouncer: true
```
kylezs, line 72: TODO: reset these before merge

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -1278,14 +1254,11 @@ pub mod pallet {
 				};
 
 				ensure!(
-					Self::with_electoral_access(
```
kylezs, line 1281: The status check that this was previously doing is already done above in `ensure_can_vote` - so we only need to handle the corrupt storage here.

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -78,44 +57,13 @@ macro_rules! generate_electoral_system_tuple_impls {
...
-                fn on_finalize<GenericElectoralAccess, $($electoral_system_alt_name_0: Translator<GenericElectoralAccess, ElectoralSystem = $electoral_system>),*>(
-                    generic_electoral_access: &mut GenericElectoralAccess,
-                    electoral_access_translators: ($($electoral_system_alt_name_0,)*),
+                fn on_finalize(
                     election_identifiers: ($(Vec<ElectionIdentifierOf<$electoral_system>>,)*),
-                    context: &Self::OnFinalizeContext,
-                ) -> Result<Self::OnFinalizeReturn, CorruptStorageError>;
-            }
-
-            impl<OnFinalizeContext, $($electoral_system: ElectoralSystem<OnFinalizeContext = OnFinalizeContext>,)*> Hooks<$($electoral_system,)*> for DefaultHooks<OnFinalizeContext> {
```
kylezs, line 100: DefaultHooks unused so deleted for simplicity

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -417,181 +318,154 @@ macro_rules! generate_electoral_system_tuple_impls {
...
+                match StorageAccess::election_state(*self.id.unique_monotonic())? {
                     CompositeElectionState::$current(state) => {
                         Ok(state)
                     },
                     _ => Err(CorruptStorageError::new())
                 }
             }
-            #[cfg(test)]
-            fn election_identifier(&self) -> Result<ElectionIdentifierOf<Self::ElectoralSystem>, CorruptStorageError> {
-                let composite_identifier = self.ea.borrow().election_identifier()?;
```
kylezs, line 445: no longer difficult to get the id :)

.github/workflows/ci-development.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
```
dandanlen, line 82: This needs to be removed before merging I guess?

kylezs: yep just left for now in case some PR comments come in that warrant changes

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
```diff
@@ -79,36 +83,48 @@ where
...
 			self.initial_election_state.unwrap_or_default();
 
-		let election = electoral_access
-			.new_election(election_identifier_extra, election_properties, election_state)
-			.unwrap();
+		// These are synchronised with respect to an election, so for simplicity in the tests
+		// we just assign them to an election.
+		MockStorageAccess::set_electoral_settings::<ES>(setup.electoral_settings.clone());
+		MockStorageAccess::set_unsynchronised_state::<ES>(setup.unsynchronised_state.clone());
+		MockStorageAccess::set_unsynchronised_settings::<ES>(setup.unsynchronised_settings.clone());
```
dandanlen, line 98: I don't really understand the comment here. How are these 'assigned to an election'?

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
```diff
@@ -79,36 +83,48 @@ where
...
 		let setup = self.clone();
-		let mut electoral_access = MockAccess::<ES>::new(
-			self.unsynchronised_state,
-			self.unsynchronised_settings,
-			self.electoral_settings,
-		);
+
+		// We need to clear the storage at every build so if there are multiple test contexts used
+		// within a single test they do not conflict.
+		MockStorageAccess::clear_storage();
```
dandanlen, line 89: Do we really need this? Storage is not persisted between tests.

dandanlen: Maybe we encapsulate both this and the intitialisation stuff below in a method MockStorageAccess::new(..)?

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
```diff
@@ -79,36 +83,48 @@ where
...
+			election_identifier_extra,
+			election_properties,
+			election_state,
+		)
+		.unwrap();
+
+		println!(
+			"Election identifier in with initial election: {:?}",
+			election.election_identifier()
+		);
```
dandanlen, line 110: We can remove this?

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
```diff
@@ -79,36 +83,48 @@ where
 	// Useful for testing check_consensus since we already have an election.
 	pub fn build_with_initial_election(self) -> TestContext<ES> {
 		let setup = self.clone();
-		let mut electoral_access = MockAccess::<ES>::new(
-			self.unsynchronised_state,
-			self.unsynchronised_settings,
-			self.electoral_settings,
-		);
+
```
dandanlen, line 86: Is the clone of self (line above this) still necessary?
I believe it may have been there because the old `self` methods were mutating.

state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
```diff
@@ -114,183 +69,387 @@ macro_rules! impl_read_access {
...
-			.ok_or_else(CorruptStorageError::new)
-	}
-	pub fn set_consensus_status(&mut self, consensus_status: ConsensusStatus<ES::Consensus>) {
-		self.with_election_mut(|e| e.consensus_status = consensus_status)
-			.expect("Cannot set consensus status for non-existent election");
-	}
+thread_local! {
+	pub static ELECTION_STATE: RefCell<BTreeMap<Vec<u8>, Vec<u8>>> = const { RefCell::new(BTreeMap::new()) };
+	pub static ELECTION_PROPERTIES: RefCell<BTreeMap<Vec<u8>, Vec<u8>>> = const { RefCell::new(BTreeMap::new()) };
+	pub static ELECTORAL_SETTINGS: RefCell<Vec<u8>> = const { RefCell::new(Vec::new()) };
```
dandanlen, line 85: Not sure we want to use thread_locals here - what if we rely on storage access properties such as transactional access? (Don't believe we do rn. but seems worth supporting). Also things like assert_noop won't work (I think it will always succeed as long as there is an error because storage will never be touched). 

It should be simple enough to changes this using parameter_types:

```rust
parameter_types! {
	storage A: T = default_value
}
```

(And in fact we could have just used the old `MockAccess` struct and stored it instead of fragmenting everything again? Would have saved a lot of boilerplate?)

state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
```diff
@@ -114,183 +69,387 @@ macro_rules! impl_read_access {
 				previous_consensus: Option<&ES::Consensus>,
 				votes: ConsensusVotes<ES>,
 			) -> Result<Option<ES::Consensus>, CorruptStorageError> {
-				ES::check_consensus(self.identifier(), self, previous_consensus, votes)
+				println!("Calling check consensus on the electoral system struct");
```
dandanlen, line 72: Can be deleted?

state-chain/pallets/cf-elections/src/electoral_systems/tests/liveness.rs
```diff
@@ -35,15 +35,12 @@ type SimpleLiveness =
...
+			assert_eq!(post.election_identifiers.len(), 1, "Only one election should exist.");
 		},
 		hook_called_once(_pre, _post) {
-			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), 1, "Hook should have been called once so far!");
-		},
+			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), 1, "Hook should have been called once so far!"); 		},
 		hook_called_twice(_pre, _post) {
-			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), 2, "Hook should have been called twice so far!");
-		},
+			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), 2, "Hook should have been called twice so far!"); 		},
```
dandanlen, line 43: nit: formatting here (and above) is a bit odd (missing a newline I believe)

dandanlen (commented):
I'm not really convinced of the benefit of mocking the storage access trait... still feel like we could just use the actual pallet.

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -401,182 +304,156 @@ macro_rules! generate_electoral_system_tuple_impls {
...
-                let extra = match composite_identifier.extra() {
-                    CompositeElectionIdentifierExtra::$current(extra) => Ok(extra),
-                    _ => Err(CorruptStorageError::new()),
-                }?;
-                Ok(composite_identifier.with_extra(*extra))
+            fn election_identifier(&self) -> ElectionIdentifierOf<Self::ElectoralSystem> {
+                self.id
```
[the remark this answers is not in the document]
kylezs: Yeah, because we removed the passing through of the election identifier to the electoral systems, this keeps it at feature parity.

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -1,31 +1,12 @@
...
+pub struct CompositeRunner<T, ValidatorId, StorageAccess, H> {
+	_phantom: core::marker::PhantomData<(T, ValidatorId, StorageAccess, H)>,
 }
 
 /// The access wrappers need to impl the access traits once for each variant,
 /// these tags ensure these trait impls don't overlap.
-mod tags {
+pub mod tags {
 	pub struct A;
 	pub struct B;
```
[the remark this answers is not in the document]
kylezs: Sensible names is a little difficult, because every chain could be different, so if we named them "block height", "ingress" etc. depending on the chain, it could be different. Though, it's certainly on the radar as something that would be nice to improve if possible

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
kylezs: Made comment more explicit.

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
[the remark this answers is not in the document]
dandanlen: I guess what I mean is that we would use this macro at the runtime-level (ie. solana_elections.rs). We don't really need to define a generic that is only used once. But yeah - can wait for later ofc.

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
kylezs: > Do we really need this? Storage is not persisted between tests.

thread local storage is persisted between tests

> Maybe we encapsulate both this and the intitialisation stuff below in a method MockStorageAccess::new(..)?

This function is kind of doing that already, what are you suggesting exactly?

state-chain/pallets/cf-elections/src/electoral_systems/mocks.rs
kylezs: removed the clone

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -401,182 +304,156 @@ macro_rules! generate_electoral_system_tuple_impls {
...
+                match StorageAccess::election_properties(self.id.with_extra(CompositeElectionIdentifierExtra::$current(*self.id.extra())))? {
                     CompositeElectionProperties::$current(properties) => {
                         Ok(properties)
                     },
                     _ => Err(CorruptStorageError::new())
                 }
             }
             fn state(&self) -> Result<$current::ElectionState, CorruptStorageError> {
-                match self.ea.borrow().state()? {
+                match StorageAccess::election_state(*self.id.unique_monotonic())? {
```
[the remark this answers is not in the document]
kylezs: The naive way of doing this would be to read everything in and then just provide back the individual fields, which is obviously a lot less efficient when we don't need everything. Doing it lazily requires a little more complexity in creating the "Election" that's returned, in order to make this simpler, so I don't think it's an obvious win.

dandanlen (approved):

PR #5340, opened 2024-10-18
bug: change_utxo not always present

marcellorigotti:
Add logic to btc_utxos_balance rpc

# Pull Request

Closes: PRO-1716, PRO-1739

## Checklist

## Summary

### DIFF
--- a/state-chain/runtime/src/chainflip.rs
+++ b/state-chain/runtime/src/chainflip.rs
@@ -834,8 +834,7 @@ impl OnBroadcastReady<Bitcoin> for BroadcastReadyProvider {
 					);
 				} else {
 					log_or_panic!(
-						"We always expect our key to be Taproot, we found {:?} instead",
-						change_output.script_pubkey
+						"Keys do not match, no change_utxo available"
 					);
 				}
 			},
--- a/state-chain/runtime/src/chainflip.rs
+++ b/state-chain/runtime/src/chainflip.rs
@@ -70,7 +70,7 @@ use cf_traits::{
 	ScheduledEgressDetails,
 };
 
-use cf_chains::btc::ScriptPubkey;
+use cf_chains::{btc::ScriptPubkey, instances::BitcoinInstance};
 use codec::{Decode, Encode};
 use eth::Address as EvmAddress;
 use frame_support::{
@@ -819,20 +819,23 @@ impl OnBroadcastReady<Bitcoin> for BroadcastReadyProvider {
 		match api_call {
 			BitcoinApi::BatchTransfer(batch_transfer) => {
 				let tx_id = batch_transfer.bitcoin_transaction.txid();
-				let outputs = batch_transfer.bitcoin_transaction.outputs.clone();
-				let output_len = outputs.len();
-				let vout = output_len - 1;
-				let change_output = outputs.get(vout).unwrap();
-				if ScriptPubkey::Taproot(batch_transfer.change_utxo_key) ==
-					change_output.script_pubkey
-				{
-					Environment::add_bitcoin_change_utxo(
-						change_output.amount,
-						UtxoId { tx_id, vout: vout as u32 },
-						batch_transfer.change_utxo_key,
-					);
-				} else {
-					log::info!("BTC egress without change UTXO");
+				let outputs = &batch_transfer.bitcoin_transaction.outputs;
+				let btc_key = pallet_cf_threshold_signature::Pallet::<Runtime, BitcoinInstance>::keys(
+					pallet_cf_threshold_signature::Pallet::<Runtime, BitcoinInstance>::current_key_epoch()
+						.expect("We should always have an epoch set")).expect("We should always have a key set for the current epoch");
+				for (i, output) in outputs.iter().enumerate() {
+					if [
+						ScriptPubkey::Taproot(btc_key.previous.unwrap_or_default()),
+						ScriptPubkey::Taproot(btc_key.current),
+					]
+					.contains(&output.script_pubkey)
+					{
+						Environment::add_bitcoin_change_utxo(
+							output.amount,
+							UtxoId { tx_id, vout: i as u32 },
+							batch_transfer.change_utxo_key,
+						);
+					}
 				}
 			},
 			_ => unreachable!(),
--- a/state-chain/runtime/src/chainflip.rs
+++ b/state-chain/runtime/src/chainflip.rs
@@ -82,7 +82,6 @@ use frame_support::{
 	},
 	traits::{Defensive, Get},
 };
-use log::{log, Level};
 pub use missed_authorship_slots::MissedAuraSlots;
 pub use offences::*;
 use scale_info::TypeInfo;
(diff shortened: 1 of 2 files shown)


state-chain/runtime/src/chainflip.rs
```diff
@@ -819,11 +821,17 @@ impl OnBroadcastReady<Bitcoin> for BroadcastReadyProvider {
...
+				if ScriptPubkey::Taproot(batch_transfer.change_utxo_key) ==
+					change_output.script_pubkey
+				{
+					Environment::add_bitcoin_change_utxo(
+						change_output.amount,
+						UtxoId { tx_id, vout: vout as u32 },
+						batch_transfer.change_utxo_key,
+					);
+				} else {
+					log_or_panic!("Keys do not match, no change_utxo available");
```
martin-chainflip, line 833: That is a weird message and also shouldn't panic in debug? I think we should just log "Witnessed BTC egress without change UTXO"

state-chain/runtime/src/lib.rs
```diff
@@ -2086,8 +2086,26 @@ impl_runtime_apis! {
...
+			let mut btc_balance = utxos.iter().fold(0, |acc, elem| acc + elem.amount);
+			//Sum the btc balance contained in the change utxos to the btc "free_balance"
+			let btc_ceremonies = pallet_cf_threshold_signature::PendingCeremonies::<Runtime,BitcoinInstance>::iter_values().map(|ceremony|{
+				ceremony.request_context.request_id
+			}).collect::<Vec<_>>();
+			for ceremony in btc_ceremonies {
+				if let RuntimeCall::BitcoinBroadcaster(pallet_cf_broadcast::pallet::Call::on_signature_ready{ api_call, ..}) = pallet_cf_threshold_signature::RequestCallback::<Runtime, BitcoinInstance>::get(ceremony).unwrap() {
+					if let BitcoinApi::BatchTransfer(batch_transfer) = *api_call {
+						let outputs = batch_transfer.bitcoin_transaction.outputs;
+						let change_output = outputs.last().unwrap();
```
ramizhasan111, line 2098: could just directly be 
```rust
 let change_output = batch_transfer.bitcoin_transaction.outputs.last().unwrap();
 ```

**Comment by ramizhasan111** [comment]
small comment, otherwise looks good (apart from Martin's comment).

**Review by ramizhasan111 — APPROVED** [review]

File: state-chain/runtime/src/chainflip.rs [hunk 3]
```diff
@@ -819,11 +821,17 @@ impl OnBroadcastReady<Bitcoin> for BroadcastReadyProvider {
...
 				let change_output = outputs.get(vout).unwrap();
-				Environment::add_bitcoin_change_utxo(
-					change_output.amount,
-					UtxoId { tx_id, vout: vout as u32 },
-					batch_transfer.change_utxo_key,
-				);
+				if ScriptPubkey::Taproot(batch_transfer.change_utxo_key) ==
+					change_output.script_pubkey
+				{
+					Environment::add_bitcoin_change_utxo(
```
dandanlen, line 827: We should also consider checking against the previous_pubkey: imagine we send a tx before but witness it after a rotation/key handover. 

Also: Maybe we should do this check for each of the outputs? 

For example imagine someone sends funds to the vault (I don't know why...)

ie we could do:

```rust
for output in outputs {
	if [vault.current_key, vault.previous_key].contains(output.address) {
		// add to utxos
	}
}
```


@martin-chainflip ?

dandanlen: Alternatively: we encode the change amount directly in the BatchTransfer type, then we don't need to parse anything from the utxos. 

(ie. we change the type to:

```rust
pub struct BatchTransfer {
	pub bitcoin_transaction: BitcoinTransaction,
	// Use a struct:
	pub change_utxo_details: (
		[u8; 32], VOut, TxId, Amount
	),
}
```

martin-chainflip: i think Dan's first suggestions is best.

dandanlen: Anyone can open a channel sending funds to the vault. (could be by accident, could be on purpose: maybe we want to top up the vault?)

ramizhasan111: But why would that change the order of the outputs of the tx? When we create a btc tx, we select some utxos from the vault, and create output utxos (transfers, and possibly change utxo appended at the end), and save it this output utxo list in the ```BitcoinTransaction``` type. So when the tx is successfully signed, we then take the change utxo and add it to the available utxo list which is what we are doing here. I dont see how someone sending funds to the vault is relevant here.

martin-chainflip: Imagine I request a swap with the BTC vault as the destination address. Then we construct a transaction with two outputs, both of which will go into the vault. Pretty theoretical example, but this code clearly does the right thing, whereas just checking the last output adds a lot of unnecessary assumptions that might break in the future.

ramizhasan111: In that theoretical example, i would say that the right thing to do is not egress at all because you are basically taking a utxo from the vault, splitting it in two and sending it back to the vault (which is pointless if not actually harmful because we are splitting utxos) and so instead of allowing the egress to happen and deal with it here in an indirect way, I would just not egress.

The code here is specifically to add the change utxo to the available utxo list and is tightly coupled with the code of creating btc tx. I dont think we are making assumptions here because we construct the btc tx a certain way (putting the change utxo at the end) and until we change how we construct btc tx, this should always be the case.

state-chain/runtime/src/lib.rs
```diff
@@ -2106,8 +2106,32 @@ impl_runtime_apis! {
...
+			//Sum the btc balance contained in the change utxos to the btc "free_balance"
+			let btc_ceremonies = pallet_cf_threshold_signature::PendingCeremonies::<Runtime,BitcoinInstance>::iter_values().map(|ceremony|{
+				ceremony.request_context.request_id
+			}).collect::<Vec<_>>();
+			let btc_key = pallet_cf_threshold_signature::Pallet::<Runtime, BitcoinInstance>::keys(
+				pallet_cf_threshold_signature::Pallet::<Runtime, BitcoinInstance>::current_key_epoch()
+				.expect("We should always have an epoch set")).expect("We should always have a key set for the current epoch");
+			for ceremony in btc_ceremonies {
+				if let RuntimeCall::BitcoinBroadcaster(pallet_cf_broadcast::pallet::Call::on_signature_ready{ api_call, ..}) = pallet_cf_threshold_signature::RequestCallback::<Runtime, BitcoinInstance>::get(ceremony).unwrap() {
+					if let BitcoinApi::BatchTransfer(batch_transfer) = *api_call {
```
dandanlen, line 2119: Instead of doing all this it might be nicer to just store the change utxo somewhere else on-chain temporarily and then move it back to the utxo set when it's ready...

This also would avoid needing to reconstruct it, removing another source of assumptions.

(can wait for another time though)

dandanlen (approved):

PR #5341, opened 2024-10-21
feat: update custom_rpc, runtime_api and broker api for broker level screening

MxmUrw:
# Pull Request

Closes: PRO-1706

## Checklist

## Summary

In this PR all changes to the broker API which are required for broker level screening are going to be merged. Currently based on @Janislav 's PR ~~(https://github.com/chainflip-io/chainflip-backend/pull/5310)~~ (https://github.com/chainflip-io/chainflip-backend/pull/5332) which implements the extrinsic for rejecting transactions.

Once that PR is merged, rebase this on main.

This PR includes the following changes:
 - Add new methods to `custom_rpc`:
     - `cf_get_tainted_transaction_events` which gets all events (for a given block) related with tainted transactions occuring in the ingress egress pallet.
     - `cf_get_open_deposit_channels` which returns all open channel addresses for a given account id.
 - These methods forward to corresponding implementations in the runtime api.
 - Changes to the broker API:
     - Add `mark_transaction_as_tainted` endpoint which submits the corresponding extrinsic.
     - Add `get_open_deposit_channels` method. 
     - Add `subscribe_tainted_btc_transaction_events` method which creates a subscription for calling the `cf_get_tainted_transaction_events` custom rpc method.
#### Non-Breaking changes

### DIFF
--- a/api/bin/chainflip-broker-api/Cargo.toml
+++ b/api/bin/chainflip-broker-api/Cargo.toml
@@ -23,6 +23,7 @@ workspace = true
 
 [dependencies]
 chainflip-api = { workspace = true }
+cf-chains = { workspace = true, default-features = true }
 cf-utilities = { workspace = true, default-features = true }
 custom-rpc = { workspace = true }
 
@@ -34,6 +35,7 @@ jsonrpsee = { workspace = true, features = ["full"] }
 serde = { workspace = true, default-features = true, features = ["derive"] }
 sp-core = { workspace = true }
 sp-rpc = { workspace = true }
+sc-rpc = { workspace = true, default-features = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
--- a/api/bin/chainflip-broker-api/Cargo.toml
+++ b/api/bin/chainflip-broker-api/Cargo.toml
@@ -33,8 +33,8 @@ futures = { workspace = true }
 hex = { workspace = true, default-features = true }
 jsonrpsee = { workspace = true, features = ["full"] }
 serde = { workspace = true, default-features = true, features = ["derive"] }
-sp-core = { workspace = true }
-sp-rpc = { workspace = true }
+sp-core = { workspace = true, default-features = true }
+sp-rpc = { workspace = true, default-features = true }
 sc-rpc = { workspace = true, default-features = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -12,8 +12,8 @@ use chainflip_api::{
 	},
 	settings::StateChain,
 	AccountId32, AddressString, BlockUpdate, BrokerApi, ChainApi, DepositMonitorApi, OperatorApi,
-	RefundParameters, SignedExtrinsicApi, StateChainApi, SwapDepositAddress,
-	SwapPayload, TransactionInId, WithdrawFeesDetail,
+	RefundParameters, SignedExtrinsicApi, StateChainApi, SwapDepositAddress, SwapPayload,
+	TransactionInId, WithdrawFeesDetail,
 };
 use clap::Parser;
 use custom_rpc::CustomApiClient;
@@ -269,7 +269,7 @@ impl RpcServer for RpcServerImpl {
 						// We only want to send a notification if there have been events.
 						// If other chains are added, they have to be considered here as
 						// well.
-						if events.btc_events.len() == 0 {
+						if events.btc_events.is_empty() {
 							continue;
 						}
 
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -11,9 +11,9 @@ use chainflip_api::{
 		AccountRole, Affiliates, Asset, BasisPoints, CcmChannelMetadata, DcaParameters,
 	},
 	settings::StateChain,
-	AccountId32, AddressString, BlockUpdate, BrokerApi, DepositMonitorApi, OperatorApi,
-	RefundParameters, SignedExtrinsicApi, StateChainApi, SwapDepositAddress, SwapPayload,
-	TransactionInId, WithdrawFeesDetail,
+	AccountId32, AddressString, BlockUpdate, BrokerApi, ChainApi, DepositMonitorApi, OperatorApi,
+	RefundParameters, SignedExtrinsicApi, StateChainApi, SwapDepositAddress,
+	SwapPayload, TransactionInId, WithdrawFeesDetail,
 };
 use clap::Parser;
 use custom_rpc::CustomApiClient;
@@ -251,28 +251,46 @@ impl RpcServer for RpcServerImpl {
 		&self,
 		pending_sink: PendingSubscriptionSink,
 	) -> SubscriptionResult {
-		let subscription_stream = self
-			.api
-			.state_chain_client
-			.base_rpc_client
-			.raw_rpc_client
-			.cf_subscribe_tainted_transaction_events()
-			.await?
-			.take_while(|item| {
-				futures::future::ready(
-					item.as_ref()
-						.inspect_err(|err| {
-							tracing::error!("Subscription error, could not deserialize message: {err}. Closing subscription.")
-						})
-						.is_ok(),
-				)
-			})
-			.map(Result::unwrap);
+		let state_chain_client = self.api.state_chain_client.clone();
+		let sink = pending_sink.accept().await.map_err(|e| e.to_string())?;
+		tokio::spawn(async move {
+			// Note we construct the subscription here rather than relying on the custom-rpc
+			// subscription. This is because we want to use finalized blocks.
+			// TODO: allow custom rpc subscriptions to use finalized blocks.
+			let mut finalized_block_stream = state_chain_client.finalized_block_stream().await;
+			while let Some(block) = finalized_block_stream.next().await {
+				match state_chain_client
+					.base_rpc_client
+					.raw_rpc_client
+					.cf_get_tainted_transaction_events(Some(block.hash))
+					.await
+				{
+					Ok(events) => {
+						// We only want to send a notification if there have been events.
+						// If other chains are added, they have to be considered here as
+						// well.
+						if events.btc_events.len() == 0 {
+							continue;
+						}
 
-		tokio::spawn(async {
-			sc_rpc::utils::pipe_from_stream(pending_sink, subscription_stream).await
+						let block_update = BlockUpdate {
+							block_hash: block.hash,
+							block_number: block.number,
+							data: events,
+						};
+
+						if sink
+							.send(sc_rpc::utils::to_sub_message(&sink, &block_update))
+							.await
+							.is_err()
+						{
+							break;
+						}
+					},
+					Err(_) => break,
+				}
+			}
 		});
-
 		Ok(())
 	}
 }
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -7,29 +7,30 @@ use cf_utilities::{
 use chainflip_api::{
 	self,
 	primitives::{
-		state_chain_runtime::runtime_apis::TaintedBtcTransactionEvent, AccountRole, Affiliates,
-		Asset, BasisPoints, CcmChannelMetadata, DcaParameters,
+		state_chain_runtime::runtime_apis::{ChainAccounts, TaintedTransactionEvents},
+		AccountRole, Affiliates, Asset, BasisPoints, CcmChannelMetadata, DcaParameters,
 	},
 	settings::StateChain,
 	AccountId32, AddressString, BlockUpdate, BrokerApi, DepositMonitorApi, OperatorApi,
 	RefundParameters, SignedExtrinsicApi, StateChainApi, SwapDepositAddress, SwapPayload,
-	WithdrawFeesDetail,
+	TransactionInId, WithdrawFeesDetail,
 };
 use clap::Parser;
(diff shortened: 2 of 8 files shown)


MxmUrw:
To be discussed: error handling in the subscription forwarding in the broker api (`subscribe_tainted_btc_transaction_events`). Currently a bit ad-hoc, forwarding a `Result<_,String>` whose error-variant is used for the case where the subscription message from the engine cannot be decoded.

api/bin/chainflip-broker-api/Cargo.toml
```diff
@@ -34,6 +35,7 @@ jsonrpsee = { workspace = true, features = ["full"] }
 serde = { workspace = true, default-features = true, features = ["derive"] }
 sp-core = { workspace = true }
 sp-rpc = { workspace = true }
+sc-rpc = { workspace = true, default-features = true }
```
dandanlen, line 38: Duplicate 👀 

FWIW I think sp-core should also have `default-features=true`

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -100,6 +106,17 @@ pub trait Rpc {
 		affiliate_fees: Option<Affiliates<AccountId32>>,
 		dca_parameters: Option<DcaParameters>,
 	) -> RpcResult<SwapPayload>;
+
+	#[method(name = "mark_btc_transaction_as_tainted", aliases = ["broker_markBtcTransactionAsTainted"])]
```
dandanlen, line 110: Can we make this more general, just to future-proof it? 

We could use an enum to make it a little less painful:

```rust
enum TxId {
	Bitcoin {
		tx_id: cf_chains::btc::Hash,
	}
	// other variants reserved for other chains.
}

fn mark_transaction_as_tainted(&self, tx_id: TxId) {
	// ...
}
```

dandanlen: (Same for the other methods)

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -100,6 +106,17 @@ pub trait Rpc {
...
+	async fn mark_btc_transaction_as_tainted(&self, tx_id: cf_chains::btc::Hash) -> RpcResult<()>;
+
+	#[method(name = "open_btc_deposit_channels", aliases = ["broker_openBtcDepositChannels"])]
+	async fn open_btc_deposit_channels(
```
dandanlen, line 114: nit: this sounds a bit like a method for opening channels. Maybe `get_open_deposit_channels` ?

(and return a chain-variant enum as above)

api/lib/src/lib.rs
```diff
@@ -628,6 +634,20 @@ pub fn clean_foreign_chain_address(chain: ForeignChain, address: &str) -> Result
...
+#[async_trait]
+pub trait DepositMonitorApi:
+	SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'static
+{
+	async fn mark_btc_transaction_as_tainted(&self, tx_id: cf_chains::btc::Hash) -> Result<()> {
+		let _ = self
+			.submit_signed_extrinsic(state_chain_runtime::RuntimeCall::BitcoinIngressEgress(
+				pallet_cf_ingress_egress::Call::mark_transaction_as_tainted { tx_id },
+			))
+			.await;
```
dandanlen, line 646: Can we return the result here instead of ignoring it?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1730,6 +1742,18 @@ where
 	) -> RpcResult<Vec<u8>> {
 		self.with_runtime_api(at, |api, hash| api.cf_filter_votes(hash, validator, proposed_votes))
 	}
+
+	fn cf_subscribe_tainted_btc_transaction_events(
+		&self,
+		sink: jsonrpsee::PendingSubscriptionSink,
+	) {
+		self.new_subscription(
+			false, /* only_on_changes */
```
dandanlen, line 1751: What does 'only on changes = false' mean? We would get an update per block? I think only pushing an update when there is an event should be enough?

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -183,6 +183,24 @@ pub struct FailingWitnessValidators {
 	pub validators: Vec<(cf_primitives::AccountId, String, bool)>,
 }
 
+#[derive(Serialize, Deserialize, Encode, Decode, Eq, PartialEq, TypeInfo, Debug, Clone)]
+pub enum TaintedBtcTransactionEvent {
+	TaintedTransactionReportReceived {
+		account_id: <Runtime as frame_system::Config>::AccountId,
+		tx_id: <<cf_chains::Bitcoin as cf_chains::Chain>::ChainCrypto as cf_chains::ChainCrypto>::TransactionInId,
```
dandanlen, line 190: nit: we can use cf_chains::Chain/Chaincrypto to make this a bit more succinct.

state-chain/runtime/src/lib.rs
```diff
@@ -2072,8 +2072,41 @@ impl_runtime_apis! {
 		fn cf_validate_refund_params(retry_duration: u32) -> Result<(), DispatchErrorWithMessage> {
 			pallet_cf_swapping::Pallet::<Runtime>::validate_refund_params(retry_duration).map_err(Into::into)
 		}
+
+		fn cf_open_btc_deposit_channels(account_id: AccountId) -> Result<Vec<<cf_chains::Bitcoin as cf_chains::Chain>::ChainAccount>, DispatchErrorWithMessage> {
```
dandanlen, line 2076: Can we make the account_id optional - we might want to get all channels irrespective of who opened them.

state-chain/runtime/src/lib.rs
```diff
@@ -2072,8 +2072,41 @@ impl_runtime_apis! {
...
+			Ok(pallet_cf_ingress_egress::DepositChannelLookup::<Runtime,BitcoinInstance>::iter()
+				.map(|(_, value)| value)
+				.filter(|channel_details| channel_details.owner == account_id)
+				.filter(|channel_details| match channel_details.action {
+					ChannelAction::Swap {..} => true,
+					ChannelAction::CcmTransfer {..} => true,
+					ChannelAction::LiquidityProvision {..} => false,
+				})
+				.map(|channel_details| channel_details.deposit_channel.address)
```
dandanlen, line 2085: I don't really mind, but clippy will probably complain and tell you merge this filter.map.map sequence into a single filter_map.

dandanlen (commented):
Looking good - I have one larger (and very annoying) suggestion and some minor ones...

api/bin/chainflip-broker-api/Cargo.toml
MxmUrw: Unfortunately one is `sP-` and the other `sC-`... The latter was required for the `sc_rpc::utils::pipe_from_stream()` utility.

I am not sure about the default-features though, I copied it from the `custom_rpc` lib. In fact it compiles if I remove the `default-features=true` for `sc-rpc`.

Should I enable default-features for all three of them anyways?

api/lib/src/lib.rs
MxmUrw: The return type of that `.await` is `(H256, (<Self as SignedExtrinsicApi>::UntilInBlockFuture, <Self as SignedExtrinsicApi>::UntilFinalizedFuture))`, I am now returning the first `tx_hash: H256` element.

api/bin/chainflip-broker-api/src/main.rs
MxmUrw: Yes, done.

state-chain/custom-rpc/src/lib.rs
MxmUrw: I agree, now '=true'.

state-chain/runtime/src/lib.rs
MxmUrw: Clippy is happy as is, and so am I :)

api/bin/chainflip-broker-api/Cargo.toml
dandanlen: Yes we should enable it for all three - it might not be a problem now, but this kind of thing tends to cause problems unexpectedly.

dandanlen (commented):
Some final comments then we're good.

api/lib/src/lib.rs
MxmUrw: Yes I see, done.

state-chain/runtime/src/lib.rs
```diff
@@ -2072,8 +2074,48 @@ impl_runtime_apis! {
...
 		}
+
+		fn cf_get_open_deposit_channels(account_id: Option<AccountId>) -> ChainAccounts {
+			let btc_chain_accounts = pallet_cf_ingress_egress::DepositChannelLookup::<Runtime,BitcoinInstance>::iter()
+				.map(|(_, value)| value)
+				.filter(|channel_details| account_id.is_none() || Some(&channel_details.owner) == account_id.as_ref())
+				.filter(|channel_details| match channel_details.action {
+					ChannelAction::Swap {..} => true,
+					ChannelAction::CcmTransfer {..} => true,
+					ChannelAction::LiquidityProvision {..} => false,
```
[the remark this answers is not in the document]
dandanlen: Yes but this rpc is called `cf_get_open_deposit_channels` so it should probably return all of the open deposit channels for the given id. 

If a broker submits this request via the broker api it will implicitly filter out an LP channels anyway. And later on when we want to be able to monitor *all* channels, we won't get any surprises in the behaviour of this rpc.

dandanlen:
https://github.com/chainflip-io/chainflip-backend/pull/5341#discussion_r1824680102

☝️ disagree? Apart from this I think we're good to go.

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -486,27 +486,34 @@ impl RpcServer for RpcServerImpl {
...
-						Err(_) => break,
-					}
-				}
-			}
-		});
+	async fn subscribe_order_fills(
+		&self,
+		pending_sink: PendingSubscriptionSink,
+	) -> SubscriptionResult {
+		let sub = self
```
kylezs, line 493: 👀  this change was reverted in a separate PR, and now it's back again?

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -194,6 +219,62 @@ impl RpcServer for RpcServerImpl {
...
+			.cf_get_open_deposit_channels(account_id, None)
+			.await
+			.map_err(BrokerApiError::ClientError)
+	}
+
+	async fn subscribe_tainted_transaction_events(
+		&self,
+		pending_sink: PendingSubscriptionSink,
+	) -> SubscriptionResult {
+		let subscription_stream = self
```
kylezs, line 254: same as subscription in the lp-api, this should probably used the finalised stream from the sc_client since we then have a guarantee that we can't skip blocks

api/bin/chainflip-lp-api/src/main.rs
kylezs: ah, you need to rebase - and then we want the same for the new broker-api subscription

state-chain/runtime/src/lib.rs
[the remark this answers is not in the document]
MxmUrw: Agree, sounds good.

api/bin/chainflip-lp-api/src/main.rs
MxmUrw: Rebased. And re-implemented the `tainted_events_subscription` to work similar to the lp api.

api/bin/chainflip-broker-api/src/main.rs
MxmUrw: Done.

state-chain/runtime/src/lib.rs
```diff
@@ -2100,8 +2102,43 @@ impl_runtime_apis! {
 		fn cf_validate_refund_params(retry_duration: u32) -> Result<(), DispatchErrorWithMessage> {
 			pallet_cf_swapping::Pallet::<Runtime>::validate_refund_params(retry_duration).map_err(Into::into)
 		}
+
+		fn cf_get_open_deposit_channels(account_id: Option<AccountId>) -> ChainAccounts {
+			let btc_chain_accounts = pallet_cf_ingress_egress::DepositChannelLookup::<Runtime,BitcoinInstance>::iter()
+				.map(|(_, value)| value)
```
kylezs, line 2108:
```suggestion
			let btc_chain_accounts = pallet_cf_ingress_egress::DepositChannelLookup::<Runtime,BitcoinInstance>::iter_values()
```

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -183,6 +184,39 @@ pub struct FailingWitnessValidators {
...
+		tx_id: TxId,
+	},
+
+	TaintedTransactionReportExpired {
+		account_id: <Runtime as frame_system::Config>::AccountId,
+		tx_id: TxId,
+	},
+
+	TaintedTransactionRejected {
+		broadcast_id: BroadcastId,
```
kylezs, line 207: nit: a comment that this is for the refund broadcast, maybe even renaming to `refund_broadcast_id`

state-chain/runtime/src/lib.rs
kylezs: I'm glad, in these cases I think I prefer filter().map(), reads more nicely than shoving an if statement in a filter_map...

kylezs (approved):
Looks good to me 👍  would await Dan's review

state-chain/runtime/src/runtime_apis.rs
MxmUrw: True, renamed.

dandanlen (approved):

PR #5351, opened 2024-10-24
chore: consistent naming prewitnessed

kylezs:
Everywhere else we treat "prewitnessed as one word" so making it consistent.


Janislav (approved):
👍

PR #5354, opened 2024-10-24
fix: separate lp account for lp api

kylezs:
# Pull Request

Closes: PRO-1530

## Checklist

- [x] I am confident that the code works - as confident as one can be, it passed first try - we'll see if this helps prevent the issues when in main. 

## Summary

We would get LP_API failures sometimes due to nonce conflicts with the usual bouncer runs - this doesn't always happen of course, depending on timing etc. Hence the flakiness.


localnet/init/keys/LP_API
```diff
@@ -0,0 +1 @@
+fa4e0729180ead2fb9194f96857350af06054a27178ef5cd1265812d74e7fdc0
```
kylezs, line 1: this is //LP_API

dandanlen:
From the bouncer error: 

```

=== LP-API::Withdraw-Asset test failed (15:22:12) ===
=== LP-API test failed (15:22:12) ===
Error: JSON Rpc request {"jsonrpc":"2.0","method":"lp_withdraw_asset","params":[100000000000000000,{"chain":"Ethereum","asset":"ETH"},"0x1594300cbd587694affd70c933b9ee9155b186d9","InBlock"],"id":9} failed: Module error ‘InsufficientBalance‘ from pallet ‘AssetBalances‘: ‘The user does not have enough funds.‘
```

Looks like we might have some mutually dependent tests that do/don't use the lp-api...

PR #5321, opened 2024-10-08
feat: Open/Close Solana contract swap accounts

ramizhasan111:
# Pull Request

Closes: PRO-1522

## Checklist

Please conduct a thorough self-review before opening the PR.

- [ ] I am confident that the code works.
- [ ] I have written sufficient tests.
- [ ] I have written and tested required migrations.
- [ ] I have updated documentation where appropriate.

## Summary

*Please include a succinct description of the purpose and content of the PR. What problem does it solve, and how? Link issues, discussions, other PRs, and anything else that will help the reviewer.*

#### Non-Breaking changes

If this PR includes non-breaking changes, select the `non-breaking` label. On merge, CI will automatically cherry-pick the commit to a PR against the release branch.

PR #5355, opened 2024-10-24
feat: Solana Vault Swaps Elections

ramizhasan111:
# Pull Request

Closes: PRO-1521 , PRO-1522

## Checklist

## Summary

- introduces elections for tracking vault swap accounts on Solana and handles initiating swaps and closing swap accounts.
- This impl make the other PR (https://github.com/chainflip-io/chainflip-backend/pull/5321) redundant and so that PR is not needed anymore.

~~note: please ignore some of the naming around Contract Swaps. That will be fixed before merging.~~ Renaming done.


engine/src/witness/sol.rs
```diff
@@ -140,6 +141,26 @@ impl VoterApi<SolanaEgressWitnessing> for SolanaEgressWitnessingVoter {
...
+impl VoterApi<SolanaVaultSwapTracking> for SolanaVaultSwapsVoter {
+	async fn vote(
+		&self,
+		_settings: <SolanaVaultSwapTracking as ElectoralSystem>::ElectoralSettings,
+		_properties: <SolanaVaultSwapTracking as ElectoralSystem>::ElectionProperties,
+	) -> Result<
+		<<SolanaVaultSwapTracking as ElectoralSystem>::Vote as VoteStorage>::Vote,
+		anyhow::Error,
+	> {
+		todo!()
```
kylezs, line 160: This should be implemented alongside the ElectoralSystem, is there a reason you haven't done so here? It's kind of impossible to tell if the electoral system is appropriate if this side of it isn't implemented

foreign-chains/solana/sol-prim/src/consts.rs
```diff
@@ -42,3 +42,8 @@ pub const LAMPORTS_PER_SIGNATURE: u64 = 5000u64;
 pub const NONCE_ACCOUNT_LENGTH: u64 = 80u64;
 
 pub const SOL_USDC_DECIMAL: u8 = 6u8;
+
+// todo: confirm this
+pub const MAX_BATCH_SIZE_OF_CONTRACT_SWAP_ACCOUNT_CLOSURES: usize = 10;
+pub const MAX_WAIT_BLOCKS_FOR_SWAP_ACCOUNT_CLOSURE_APICALLS: u32 = 14400; // 1 day
+pub const NONCE_AVAILABILITY_THRESHOLD_FOR_INITIATING_SWAP_ACCOUNT_CLOSURES: usize = 4; // revisit this
```
kylezs, line 49: what needs confirming / revisiting? can we do this before this PR is merged?

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+				if *count >= success_threshold {
+					Some(vote.clone())
+				} else {
+					None
+				}
+			})
+		} else {
+			None
+		})
+	}
```
kylezs, line 213: this is the same in 3 electoral systems, should look at factoring this out

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1104,7 +1104,9 @@ pub mod pallet {
 			destination_address: EncodedAddress,
 			tx_hash: TransactionHash,
 		) -> DispatchResult {
-			T::EnsureWitnessed::ensure_origin(origin)?;
+			// This extrinsic can either be witnessed called or called from within the runtime (for
```
kylezs, line 1107: Is a bit nicer to create an inner function than do this. The fact we need a comment suggests it's confusing. Then in the solana_elections we can just do:
```rust
SolanaIngressEgress::contract_swap_request(
			swap_details.from,
			swap_details.to,
			swap_details.deposit_amount,
			swap_details.destination_address,
			swap_details.tx_hash,
		)
```
which makes more sense - we shouldn't need an origin at that point.

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -224,8 +268,13 @@ impl
...
 				>,
 			>,
+			Vec<
+				ElectionIdentifier<
+					<SolanaVaultSwapTracking as ElectoralSystem>::ElectionIdentifierExtra,
+				>,
+			>,
 		),
-		_context: &Self::OnFinalizeContext,
+		context: &Self::OnFinalizeContext,
```
kylezs, line 277: 👀 I'm about to delete the `context` for the hooks in my refactor PR - we can instead just get the block number directly since we're in the runtime.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+
+pub trait SolanaVaultSwapAccountsHook<Account, SwapDetails, E> {
+	fn close_accounts(accounts: Vec<Account>) -> Result<(), E>;
+	fn initiate_vault_swap(swap_details: SwapDetails);
+	fn get_number_of_available_sol_nonce_accounts() -> usize;
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize, TypeInfo, Encode, Decode)]
+pub struct SolanaVaultSwapsElectoralState<Account: Ord, BlockNumber> {
+	pub block_number_last_closed_accounts: BlockNumber,
```
kylezs, line 39: nit:
```suggestion
	pub accounts_last_closed_at: BlockNumber,
```

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+
+#[derive(
+	Clone, PartialEq, Eq, Debug, Serialize, Deserialize, TypeInfo, Encode, Decode, Ord, PartialOrd,
+)]
+pub struct SolanaVaultSwapsVote<Account: Ord, SwapDetails: Ord> {
+	pub new_accounts: BTreeSet<(Account, SwapDetails)>,
+	pub confirm_closed_accounts: BTreeSet<Account>,
+}
+
+pub struct SolanaVaultSwapAccounts<
```
kylezs, line 63: there are no tests?

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+	) -> Result<Option<Self::Consensus>, CorruptStorageError> {
+		let num_authorities = consensus_votes.num_authorities();
+		let success_threshold = success_threshold_from_share_count(num_authorities);
+		let active_votes = consensus_votes.active_votes();
+		let num_active_votes = active_votes.len() as u32;
+		Ok(if num_active_votes >= success_threshold {
+			let mut counts = BTreeMap::new();
+			for vote in active_votes {
+				counts.entry(vote).and_modify(|count| *count += 1).or_insert(1);
+			}
```
kylezs, line 202: will this actually work? If nodes are querying at different times then their votes may not be identical, and therefore wouldn't come to consensus

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+					});
+
+					Ok(())
+				})?;
+			}
+		} else {
+			electoral_access.new_election((), (), ())?;
+		}
+
+		let mut unsynchronised_state = electoral_access.unsynchronised_state()?;
```
kylezs, line 149: When we're in specific implementations we can use more specific and descriptive variable names than "unsynchronised state" etc.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+{
+	type ValidatorId = ValidatorId;
+	type ElectoralUnsynchronisedState = SolanaVaultSwapsElectoralState<Account, BlockNumber>;
+	type ElectoralUnsynchronisedStateMapKey = ();
+	type ElectoralUnsynchronisedStateMapValue = ();
+
+	type ElectoralUnsynchronisedSettings = ();
+	type ElectoralSettings = Settings;
+	type ElectionIdentifierExtra = ();
+	type ElectionProperties = ();
```
dandanlen, line 101: I'm surprised there are no election properties. 

I'm guessing the intention is to have the engine read the `ElectoralUnsynchronisedState` but this is not how the electoral systems are intended to work. 

All the required context for the vote (i'm guessing this is: 'currently opened accounts') should be stored in the properties: see the fn signature of the voting api.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
dandanlen: I agree - we need to count each account addition/removal individually and then keep only the additions/removals that have consensus votes.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,214 @@
...
+	Parameter,
+};
+use itertools::Itertools;
+use sp_std::vec::Vec;
+
+pub trait SolanaVaultSwapAccountsHook<Account, SwapDetails, E> {
+	fn close_accounts(accounts: Vec<Account>) -> Result<(), E>;
+	fn initiate_vault_swap(swap_details: SwapDetails);
+	fn get_number_of_available_sol_nonce_accounts() -> usize;
+}
```
dandanlen, line 35: I don't think we should be coupling this so tightly to Solana. 

Can we not follow the same patterns/conventions as we previously have, with the hooks  defined externally rather than in the pallet?

engine/src/witness/sol.rs
ramizhasan111: Albert is doing the engine side so I left it as a todo. Gonna merge albert's PR into this and hook up the elections to it as discussed in discord.

foreign-chains/solana/sol-prim/src/consts.rs
ramizhasan111: this carried over from the base PR. This is mostly confirmed if @albert-llimos doesnt have any objection to it.

state-chain/pallets/cf-ingress-egress/src/lib.rs
ramizhasan111: will do it.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: wanted to first nail down the approach based on the reviews before writing any tests

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: I originally wanted to have engines vote on individual accounts but the way Bitmap Components work doesn't allow that (we cant vote for more than one bitmap component in a single vote). So I instead took the route of coming to consensus over the whole list. But I can see that it is possible we cant come to a consensus over the whole list. Will modify the consensus to parse the votes and count votes for individual accounts.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: I considered using Properties since we have access to it in the Voting api but didnt want to store the list twice (once in the unsynchronised state and once in properties). But I can see now that if we use the properties, we dont need to use the unsynchronised state.

foreign-chains/solana/sol-prim/src/consts.rs
albert-llimos: sounds good

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
kylezs: You can think of Properties as a way to pass arguments to the engine voting rather than storage.

state-chain/runtime/src/chainflip/solana_elections.rs
ramizhasan111: My first approach was to directly get the block number but couldn't do so. The problem is that we dont have access to the Runtime here since it is a standalone module and is not part of the Pallet impl which means it doesnt have access to the runtime generic ```T```

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
kylezs: The bitmap should generally only be used for when we expect identical votes

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
kylezs: That's true, but to expand on *why* the properties are there and separate: The issue is in the name "unsynchronised" - the engines need to receive the same election inputs to provide consistent witness data. By using the properties it is then synchronised by the election identifier.

ramizhasan111: good to know 👍 However, in this case the unsynchronised state is also synchronised since there is only one election at a given moment and the only way it changes is when a consensus is reached and we then immediately delete that election and start a new one. 

But in any case, we can use properties here and then we dont need to use the unsynchronised state. will do that.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: updated the consensus.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: the impl has changed now and so it is not the same as other electoral systems.

state-chain/runtime/src/chainflip/solana_elections.rs
kylezs: you don't have access because it's not in a generic context, you don't need T, you can just use `Runtime`

ramizhasan111: We dont have access to the concrete ```Runtime``` type either since we are at pallet level and the Runtime is constructed at the Runtime level (one level higher). Thats the problem. The way to get access to the Runtime at the pallet level is the Runtime generic ```T``` which we cant use here either because of the reason i mentioned.

ramizhasan111: clarified in dms. I thought kyle was talking about the ```context``` in the ```ElectoralSystem``` trait whereas he was talking about the ```context``` in ```Hooks``` trait.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
dandanlen: What I mean is that this electoral system includes a lot of non-election/consensus related logic inside of itself. ie. the interfaces for other electoral systems is defined in term of hooks that trigger externally-defined logic. However here the 'hooks' are not really hooks: they are accessors to specific external functionality, and the logic for how to process those things is internal to the election. We're mixing contexts.

Rather than `get_number_of_available_sol_nonce_accounts` and `initiate_vault_swap`, I think it should just have a single hook that says something like `on_account_change(added, removed)`; and the logic of *what* to do with that information should be defined externally.

ramizhasan111: Given that its a custom election which is not intended to be re-used, I dont see why we cant have some custom logic in it especially if that logic is closely tied to what the election is about. We can rename the trait to not indicate that it is a hook but a general trait that calls some external code. The idea of injecting a trait that provides external code access is general enough here and the ElectoralSystem trait doesnt force us that the trait injected necessarily has to fit the definition of a hook (whatever that might be).

Even if we move out some of the functionality we would then need that logic to return some information back to the electoral system which it then uses to create a new election which would still mean it would still not function like a hook. 

In other electoral systems as well the code in on_finalize of the ElectoralSystem includes logic of first checking for consensus (calling the check_consensus) and then executing some logic to then decide whether to start a new election and with what properties. It is also the same here with just some external logic accessed through the Hook.

albert-llimos:
I just merged #5375 on top of this so it has the entire vault swap feature so it can be reviewed together (CCM, vaultSwapParameters).
The bouncer changes can be reviewed at the end after we merge this to the base Solana Vault branch if needed, no need to do here necessarily.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,263 @@
...
+	) -> Result<VotePropertiesOf<Self>, CorruptStorageError> {
+		Ok(())
+	}
+
+	fn is_vote_desired<ElectionAccess: ElectionReadAccess<ElectoralSystem = Self>>(
+		_election_identifier_with_extra: crate::electoral_system::ElectionIdentifierOf<Self>,
+		_election_access: &ElectionAccess,
+		_current_vote: Option<(VotePropertiesOf<Self>, AuthorityVoteOf<Self>)>,
+	) -> Result<bool, CorruptStorageError> {
+		Ok(true)
```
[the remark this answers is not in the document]
ramizhasan111: Defined is_vote_needed().
The right thing to do is to check if their vote has changed rather than checking if it is empty.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: working on unit tests right now. @albert-llimos added some end to end tests in the bouncer.

state-chain/node/src/chain_spec/sisyphos.rs
```diff
@@ -158,4 +158,4 @@ pub fn extra_accounts() -> Vec<(AccountId, AccountRole, FlipBalance, Option<Vec<
 pub const BITCOIN_SAFETY_MARGIN: u64 = 5;
 pub const ETHEREUM_SAFETY_MARGIN: u64 = 6;
 pub const ARBITRUM_SAFETY_MARGIN: u64 = 1;
-pub const SOLANA_SAFETY_MARGIN: u64 = 1; //TODO: put correct value
+pub const SOLANA_SAFETY_MARGIN: u64 = 1; // Unused - we use "finalized" instead
```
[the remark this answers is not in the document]
albert-llimos: It's currently used in the chain spec as we have a safety margin for all chains. I'd also remove it but that should be part of another PR.

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,544 @@
...
+									}
+								};
+
+								Some(Ok((VaultSwapAccountAndSender {
+									vault_swap_account: account,
+									swap_sender: data.sender.into()
+								}, SolanaVaultSwapDetails {
+									from: if data.src_token.is_none() {SolAsset::Sol} else {SolAsset::SolUsdc},
+									deposit_amount: data.amount,
+									destination_address: EncodedAddress::from_chain_bytes(data.dst_chain.try_into().map_err(|e| warn!("error while parsing destination chain for solana vault swap:{}. Omitting swap", e)).ok()?, data.dst_address.to_vec()).map_err(|e| warn!("failed to decode the destination chain address for solana vault swap:{}. Omitting swap", e)).ok()?,
```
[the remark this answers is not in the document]
albert-llimos: @ramizhasan111 will be doing some small change to this code to handle decoding errors. We can do it when that's up.

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,544 @@
...
+					},
+					_ => {
+						warn!("Unsupported input token for the witnessed solana vault swap, omitting the swap and the swap account.");
+						None
+					},
+				},
+			))
+		})
+		.try_flatten()
+		.try_collect()
```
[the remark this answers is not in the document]
albert-llimos: Good point. Another one that should be handled when the changes are done like above. We should end up with no error @ramizhasan111

state-chain/node/src/chain_spec/sisyphos.rs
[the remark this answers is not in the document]
dandanlen: We don't have one for polkadot (set to None in the chainspec). It's not important but if we're going to change it we might as well change it something correct, otherwise it's just confusing.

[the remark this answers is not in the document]
albert-llimos: Sure, I just updated the comment because the old one was just wrong. It's true that we should clean it up but at least this comment is not wrong.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,277 @@
...
+						MAX_BATCH_SIZE_OF_VAULT_SWAP_ACCOUNT_CLOSURES
+					{
+						known_accounts
+							.witnessed_open_accounts
+							.drain(..MAX_BATCH_SIZE_OF_VAULT_SWAP_ACCOUNT_CLOSURES)
+							.collect()
+					} else {
+						sp_std::mem::take(&mut known_accounts.witnessed_open_accounts)
+					};
+					match Hook::close_accounts(accounts_to_close.clone()) {
```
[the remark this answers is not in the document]
ramizhasan111: depending on whether we were successfully able to close accounts, we have to set the unsynchronized state so it makes sense to have this logic here.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,277 @@
...
+						Ok(()) => {
+							known_accounts.closure_initiated_accounts.extend(accounts_to_close);
+							electoral_access.set_unsynchronised_state(*current_block_number)?;
+						},
+						Err(e) => {
+							log::error!(
+								"failed to build Solana CloseSolanaVaultSwapAccounts apicall: {:?}",
```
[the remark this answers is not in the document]
ramizhasan111: will improve the error message.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1243,7 +1243,9 @@ pub mod pallet {
 			_destination_address: EncodedAddress,
 			_tx_hash: TransactionHash,
 		) -> DispatchResult {
-			T::EnsureWitnessed::ensure_origin(origin)?;
+			// This extrinsic can either be witnessed called or called from within the runtime (for
+			// solana contract swaps).
+			ensure_root(origin.clone()).or(T::EnsureWitnessed::ensure_origin(origin).map(|_| ()))?;
```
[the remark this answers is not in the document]
ramizhasan111: no we dont. will revert.

engine/src/witness/sol/program_swaps_witnessing.rs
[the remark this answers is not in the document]
ramizhasan111: We can see in the code that it will never return an ```Err```. The items that error are filtered out in the inner ```filter_map``` and the iterator returned should always have non-error items and so it should be ok.

That said, I think this can be optimized as I feel there is needless layering of iterators. for example, I am not sure what purpose ```map_ok``` is serving here.

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,508 @@
...
+										});
+										(deposit_metadata, vault_swap_parameters)
+									}
+								};
+
+								Some(Ok((VaultSwapAccountAndSender {
+									vault_swap_account: account,
+									swap_sender: data.sender.into()
+								}, SolanaVaultSwapDetails {
+									from: if data.src_token.is_none() {SolAsset::Sol} else {SolAsset::SolUsdc},
```
kylezs, line 109: this is pretty high risk - I would say we explicitly check that it's USDC here - I know we check it above, but seems something easy to miss - or ideally structure the code so it's clear. like have a `get_src_asset -> Option<asset>` function on the data... don't think this works super well, since the `usdc_token_mint_pubkey` is only available in this function's context, but yeah, worth seeing if we can make it a bit clearer/more explicit

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,508 @@
...
+							data.src_token.is_some_and(|addr| addr == usdc_token_mint_pubkey.into())) => {
+
+								let (deposit_metadata, vault_swap_parameters) = match data.ccm_parameters {
+									None => {
+										let CfParameters { ccm_additional_data: (), vault_swap_parameters } =
+											CfParameters::decode(&mut &data.cf_parameters[..]).map_err(|e| warn!("error while decoding CfParameters for solana vault swap: {}. Omitting swap", e)).ok()?;
+										(None, vault_swap_parameters)
+									},
+									Some(ccm_parameters) => {
+										let CfParameters { ccm_additional_data, vault_swap_parameters } =
```
kylezs, line 86: Just for clarity's sake
```suggestion
										let CcmCfParameters { ccm_additional_data, vault_swap_parameters } =
```

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -838,3 +893,100 @@ impl ElectionEgressWitnesser for SolanaEgressWitnessingTrigger {
...
+#[derive(
+	Clone, PartialEq, Eq, Debug, Serialize, Deserialize, TypeInfo, Encode, Decode, PartialOrd, Ord,
+)]
+pub struct SolanaVaultSwapDetails {
+	pub from: SolAsset,
+	pub to: Asset,
+	pub deposit_amount: SolAmount,
+	pub destination_address: EncodedAddress,
+	pub deposit_metadata: Option<CcmDepositMetadata>,
+	// TODO: These two will potentially be a TransactionId type
```
kylezs, line 906: which two? (better to explicitly name in these cases in case code gets moved too, more obvious if it's a stale comment)
```suggestion
	// TODO: These two will potentially be a TransactionId type
```

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -652,6 +702,11 @@ impl
 			ingress_identifiers,
 			&block_height,
 		)?;
+		SolanaVaultSwapTracking::on_finalize(
+			&mut vault_swap_translator.translate_electoral_access(generic_electoral_access),
+			vault_swap_identifiers,
+			&crate::System::current_block_number(),
```
kylezs, line 708: when rebased, the liveness PR already calls this, so we can pass the same block number in rather than call twice

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+						.and_modify(|c| *c += *count)
+						.or_insert(*count);
+				});
+			});
+
+			counts_new_accounts.retain(|_, count| *count >= success_threshold);
+			let new_accounts = counts_new_accounts.into_keys().cloned().collect::<BTreeSet<_>>();
+			counts_confirm_closed_accounts.retain(|_, count| *count >= success_threshold);
+			let confirm_closed_accounts =
+				counts_confirm_closed_accounts.into_keys().cloned().collect::<BTreeSet<_>>();
```
kylezs, line 263: for each of new_account and confirm_closed accounts there's dupe logic - could factor out

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+						Hook::initiate_vault_swap((*swap_details).clone());
+						(*account).clone()
+					},
+				));
+				consensus.confirm_closed_accounts.into_iter().for_each(|acc| {
+					known_accounts.closure_initiated_accounts.remove(&acc);
+				});
+
+				if Hook::get_number_of_available_sol_nonce_accounts() >
+					NONCE_AVAILABILITY_THRESHOLD_FOR_INITIATING_SWAP_ACCOUNT_CLOSURES &&
```
kylezs, line 180: comment here would be useful, just so it's clear that it's just to provide better service to users, there's not some fundamental reason

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+					(known_accounts.witnessed_open_accounts.len() >=
+						MAX_BATCH_SIZE_OF_VAULT_SWAP_ACCOUNT_CLOSURES ||
+						(*current_block_number)
+							.checked_sub(&electoral_access.unsynchronised_state()?)
+							.expect("current block number is always greater than when apicall was last created")
```
kylezs, line 185: while, true, could we make this saturating - just makes it more difficult to trigger a panic accidentally with code changes later (and when searching for places in the code that can panic, it's nicer when there are fewer)

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+			.map_err(|_| CorruptStorageError::new())?
+		{
+			let mut election_access = electoral_access.election_mut(election_identifier)?;
+			if let Some(consensus) = election_access.check_consensus()?.has_consensus() {
+				let mut known_accounts = election_access.properties()?;
+				election_access.delete();
+				known_accounts.witnessed_open_accounts.extend(consensus.new_accounts.iter().map(
+					|(account, swap_details)| {
+						Hook::initiate_vault_swap((*swap_details).clone());
+						(*account).clone()
```
kylezs, line 172:
```suggestion
						account.clone()
```

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+			.at_most_one()
+			.map_err(|_| CorruptStorageError::new())?
+		{
+			let mut election_access = electoral_access.election_mut(election_identifier)?;
+			if let Some(consensus) = election_access.check_consensus()?.has_consensus() {
+				let mut known_accounts = election_access.properties()?;
+				election_access.delete();
+				known_accounts.witnessed_open_accounts.extend(consensus.new_accounts.iter().map(
+					|(account, swap_details)| {
+						Hook::initiate_vault_swap((*swap_details).clone());
```
kylezs, line 171:
```suggestion
						Hook::initiate_vault_swap(swap_details.clone());
```

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+						Ok(()) => {
+							known_accounts.closure_initiated_accounts.extend(accounts_to_close);
+							electoral_access.set_unsynchronised_state(*current_block_number)?;
+						},
+						Err(e) => {
+							log::error!(
+								"failed to close accounts: {:?}",
+								e
+							);
+							known_accounts.witnessed_open_accounts.extend(accounts_to_close);
```
kylezs, line 205: not really sure I understand why we do this?

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+					.collect();
+					match Hook::close_accounts(accounts_to_close.clone()) {
+						Ok(()) => {
+							known_accounts.closure_initiated_accounts.extend(accounts_to_close);
+							electoral_access.set_unsynchronised_state(*current_block_number)?;
+						},
+						Err(e) => {
+							log::error!(
+								"failed to close accounts: {:?}",
+								e
```
kylezs, line 203:
```suggestion
							log::error!(
								"Failed to initiate account closure: {:?}",
								e
```

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
```diff
@@ -0,0 +1,274 @@
...
+	) -> Result<Self::OnFinalizeReturn, CorruptStorageError> {
+		if let Some(election_identifier) = election_identifiers
+			.into_iter()
+			.at_most_one()
+			.map_err(|_| CorruptStorageError::new())?
+		{
+			let mut election_access = electoral_access.election_mut(election_identifier)?;
+			if let Some(consensus) = election_access.check_consensus()?.has_consensus() {
+				let mut known_accounts = election_access.properties()?;
+				election_access.delete();
```
kylezs, line 168: I would move this to right before new_election, just so it's easier to see the fact we only expect (and always expect) one election to open at any one time.

kylezs (commented):
Still going through it, but thought I'd drop some comments now is looking good so far 👍

state-chain/runtime/src/chainflip/solana_elections.rs
albert-llimos: done

state-chain/primitives/src/lib.rs
```diff
@@ -418,10 +429,15 @@ pub struct Beneficiary<Id> {
...
+	Ord,
 )]
 pub struct DcaParameters {
 	/// The number of individual swaps to be executed
 	pub number_of_chunks: u32,
 	/// The interval in blocks between each swap.
 	pub chunk_interval: u32,
 }
+
+// TODO: Define this / implement it on the SC - PRO-1743.
```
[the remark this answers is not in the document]
albert-llimos: Actually just removed the comment. It will be used more after implementing PRO-1743 but it's already in the right place.

engine/src/witness/sol/program_swaps_witnessing.rs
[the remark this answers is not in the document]
ramizhasan111: ignore the above comment. while the inner closure always returns Ok, we would still need to deal with errors in rpc calls (basically ignore queries that errored).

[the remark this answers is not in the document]
ramizhasan111: see commit [`ac586b6` (#5355)](https://github.com/chainflip-io/chainflip-backend/pull/5355/commits/ac586b60eaae369d9d1eeb2c8203fb79d9d7ad77)

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
albert-llimos: Ramiz can confirm but my understanding is that we are readding the `accounts_to_close` to the `known_accounts` in the scenario where the closing of accounts has failed. Since we had drained that account from the list we are readding it back.

kylezs: sorry yeah, didn't delete the comment... was more a comment to self when going through 👍

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
kylezs, line 1: could we name this `solana_vault_swap_accounts` so it matches the Electoral system name

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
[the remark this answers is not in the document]
kylezs: If all the logic were in the close_accounts hook, then, we wouldn't need to have two separate hooks for `get_number_of_available_sol_nonce_accounts` and `close_accounts` - close accounts would already have access to it -> which makes it easier to mock and test, and for the writing of state, all you need is to return a result as is already the case, and just set the block number.

Then for testing this ElectoralSystem's on_finalize, it's very simple, you can just check "has hook been called" - we already have tests that do this for other ElectoralSystems so it would be very similar. Then you could also write tests for the close_accounts hook implementation itself, separately.

bouncer/tests/all_swaps.ts
```diff
@@ -68,5 +68,14 @@ async function main() {
       });
   });
 
+  // Not doing BTC due to encoding complexity in vault_swap. Will be fixed once SDK supports it.
```
kylezs, line 71: we have a ticket for this?

engine/src/witness/sol.rs
```diff
@@ -175,6 +183,41 @@ impl VoterApi<SolanaLiveness> for SolanaLivenessVoter {
 	}
 }
 
+#[allow(dead_code)]
```
kylezs, line 186: is not dead code anymore

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,505 @@
...
+		{
+			new_program_swap_accounts.push(*account);
+		}
+	}
+
+	for account in sc_closure_initiated_accounts {
+		if !open_event_accounts.contains(&account.vault_swap_account) {
+			closed_accounts.push(account);
+		}
+	}
```
kylezs, line 186:
```suggestion
	let new_program_swap_accounts: Vec<_> = open_event_accounts
		.iter()
		.filter(|account| {
			!sc_opened_accounts_hashset.contains(account) &&
				!sc_closure_initiated_accounts.iter().any(|x| &x.vault_swap_account == *account)
		})
		.cloned()
		.collect();

	let closed_accounts: Vec<_> = sc_closure_initiated_accounts
		.into_iter()
		.filter(|account| !open_event_accounts.contains(&account.vault_swap_account))
		.collect();
```

kylezs: death to mutable variables 💀

engine/src/witness/sol.rs
```diff
@@ -175,6 +183,41 @@ impl VoterApi<SolanaLiveness> for SolanaLivenessVoter {
...
+		anyhow::Error,
+	> {
+		program_swaps_witnessing::get_program_swaps(
+			&self.client,
+			settings.swap_endpoint_data_account_address,
+			properties
+				.witnessed_open_accounts
+				.into_iter()
+				.map(|VaultSwapAccountAndSender { vault_swap_account, .. }| vault_swap_account)
+				.collect(),
```
kylezs, line 209: we can collect into a hashset here instead of a vec, and pass through a HashSet to these functions, we end up just doing `let sc_opened_accounts_hashset: HashSet<_> = sc_opened_accounts.iter().collect();` inside `get_changed_program_swap_accounts`

engine/src/witness/sol.rs
```diff
@@ -175,6 +183,41 @@ impl VoterApi<SolanaLiveness> for SolanaLivenessVoter {
...
+			properties.closure_initiated_accounts,
```
kylezs, line 210: doesn't look like we use the `swap_sender` from here either? By removing the sender here, it could allow us to simplify the code in `get_changed_program_swap_accounts` even further than my suggestion by using set operators instead of filter collect()

bouncer/tests/all_concurrent_tests.ts
```diff
@@ -56,6 +57,8 @@ async function runAllConcurrentTests() {
 
   await Promise.all([broadcastAborted.stop(), feeDeficitRefused.stop()]);
 
+  await checkSolEventAccountsClosure();
+
   await checkAvailabilityAllSolanaNonces();
```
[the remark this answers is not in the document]
albert-llimos: Technically yes but we don't really want/need to start checking Solana nonce until the accounts have been closed, it doesn't make sense. Same for checking nonces/accounts while swaps are going one. We'd maximum win a few seconds and I can see scenarios where we introduce flakyness.

bouncer/shared/vault_swap.ts
```diff
@@ -184,20 +228,17 @@ export async function executeSolContractSwap(
...
           .xSwapToken({
-            amount: new BN(
-              amountToFineAmount(defaultAssetAmounts(srcAsset), assetDecimals(srcAsset)),
-            ),
-            dstChain: Number(destChain),
-            dstAddress: Buffer.from(destAddress),
-            dstToken: Number(stateChainAssetFromAsset(destAsset)),
+            amount,
+            dstChain: chainContractId(destChain),
+            dstAddress: Buffer.from(destinationAddress.slice(2), 'hex'),
```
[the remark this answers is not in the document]
albert-llimos: It's not the prettiest lines of code but this will be removed once this logic is added to the SDK so it's fine, it's still easy to understand.

engine/src/witness/sol.rs
albert-llimos: good point, will do

bouncer/tests/all_swaps.ts
albert-llimos: This should be part of adding the logic in the SDK and broker api to create Bitcoin Vault swaps and other vault swaps for that matter. It's encapsulated into PRO-1642 but not sure if there is some ticket on the product side to expand vault swaps in the SDK. There should be or at least they must have planned for it.

engine/src/witness/sol/program_swaps_witnessing.rs
ramizhasan111: changed so that we do the check at one place and use the Asset from there.

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
ramizhasan111: this is not possible i guess since the rust compiler does not allow you to borrow immutably while we have already borrowed the same thing mutably. We have that kind of a case here: we have to call the ```				election_access.delete();``` before calling ```electoral_access.unsynchronised_state()```

kylezs: interesting... might be possible after my refactor - but ok to leave it for now, is a tiny thing

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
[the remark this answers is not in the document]
ramizhasan111: I considered moving all the logic concerning all three hooks inside one hook but realized it would not actually make testing easier. This is because in that case, writing the test for one big hook impl would then be difficult since the logic in hook impl calls external code (constructing apicall, accessing sol nonces etc). So we would have to mock those. So basically we are back where we started. We wanted to avoid create 3 mocks of 3 election hooks, instead we create one. However, that means we create those extra mocks while testing the hook impl so I dont think we gain anything. In fact, creating the mocks for the hook impl is even harder since we dont have a framework for it at the Runtime level. creating 3 mocks instead of one while writing tests for the on_finalize shouldnt be a big deal.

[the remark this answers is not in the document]
kylezs: Having the apicall construction be a separate trait function is probably reasonable, it's also kind of a different part of the flow, so conceptually could make things easier to follow. Though I think putting the get nonce and close accounts in one should definitely be simpler

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -72,53 +72,59 @@ pub async fn get_program_swaps(
...
-										(None, vault_swap_parameters)
+					Some(data) => {
+						let from_asset = if data.src_token.is_none() {
+							SolAsset::Sol
+						} else if data.src_token.unwrap() == usdc_token_mint_pubkey.into() {
+							SolAsset::SolUsdc
+						} else {
+							warn!("Unsupported input token for the witnessed solana vault swap, omitting the swap and the swap account.");
+							None?
+						};
```
kylezs, line 83: We can avoid the unwrap:
```rust
						let asset = if let Some(token) = data.dst_token {
							if token == usdc_token_mint_pubkey.into() {
								warn!("Unsupported output token for the witnessed solana vault swap, omitting the swap and the swap account.");
								SolAsset::SolUsdc
							} else {
								return None;
							}
						} else {
							SolAsset::Sol
						};
```

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,502 @@
...
+					// 2. If due to rpc load management the get event accounts rpc is queried at a
+					//    slot < get swap endpoint data rpc slot, the min_context_slot will prevent
+					//    it from being executed before that.
+					// This could only happen if an engine is behind and were to see the account
+					// opened and closed between queries. That's not reallistic as it takes minutes
+					// for an account to be closed and even if it were to happen it's not
+					// problematic as we'd have reached consensus and the engine would just filter
+					// it out.
+					None => {
+						warn!("Event account not found for solana event account");
```
kylezs, line 142: probably want to include which account in this log message?

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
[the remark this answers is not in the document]
ramizhasan111: I am not sure I understand what you mean. The apicall construction is inside the ```close_accounts``` function. trying to combine apicall construction and get nonces into one hook would mean we have to move the whole logic inside that hook which creates the problem of testability that I mentioned above.

In general, I think it is a pretty arbitrary and subjective choice here between having the logic in one place or splitting it. In the end logic is logic, one has to follow it to make sure it works. Its not like we don't have long pieces of logic inside single functions in other places in the codebase.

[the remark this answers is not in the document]
kylezs: Sorry, should've been clearer. I mean leaving initiate_vault_swap separate makes sense. But since the only reason we need the number of sol accounts is to decide whether we construct the close accounts call, seems to make sense to include it.

I'm not against having long pieces of logic, it's just about keeping conceptually similar things together and the boundaries between things clean. Happy to see how the tests look keeping it like this

engine/src/witness/sol/program_swaps_witnessing.rs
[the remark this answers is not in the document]
ramizhasan111: did some formatting as part of this: https://github.com/chainflip-io/chainflip-backend/pull/5405
cargo fmt does not work well with the long string literals so I had to manually do the formatting. hopefully its easy to read.

engine/src/witness/sol/program_swaps_witnessing.rs
ramizhasan111: done

engine/src/witness/sol/program_swaps_witnessing.rs
kylezs: Yeah ofc, better to use the language features so you don't even need to think about if it's safe to unwrap or not 👍

state-chain/pallets/cf-elections/src/electoral_systems/solana_swap_accounts_tracking.rs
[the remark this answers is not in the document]
ramizhasan111: yes of course but in this case we cant test the hook impl easily as i said before so it makes sense to keep it here  since we want to test and keep the hook impl trivial.

ramizhasan111:
Apart from tests, which I am doing as another PR, this PR should be ready and I would like to merge it if there are no more open points.

engine/src/witness/sol.rs
kylezs: This conversation was resolved but nothing was changed here?

albert-llimos: I did simplify that function but that was for another comment and I missed this one.
I think not passing the sender downstream might be not be very clean to implement as we do require the sender for each of those accounts in the vote. Maybe there is a nice way to do it in Rust that I'm missing.
What I think could be a better option is to not require the sender address when voting on a closed account. We are anyway not doing anything with it since we report just the sender address that the SC has told the engine to witness.
I see how that might require some update in the pallet as we currently store those accounts in a `BTreeSet<(Address, Sender)>` in `closure_initiated_accounts`. We could technically also not store the sender in there and just have a `BTreeSet<Address>` as we don't need the anymore once the closure transaction has been initiated.
What do you think? @kylezs @ramizhasan111

kylezs: Fair, probably simpler to just leave it as is I think

albert-llimos:
Just fyi, when this gets rebased I expect the Solana Vault swaps to fail because the encoding of `cf_parameters` have changed in main. I will fix that after the rebasing is done.

state-chain/pallets/cf-elections/src/electoral_systems/solana_vault_swap_accounts.rs
```diff
@@ -165,8 +165,10 @@ impl<
 				let mut known_accounts = election_access.properties()?;
 				election_access.delete();
 				known_accounts.witnessed_open_accounts.extend(consensus.new_accounts.iter().map(
-					|(account, swap_details)| {
-						Hook::initiate_vault_swap(swap_details.clone());
+					|(account, maybe_swap_details)| {
+						if let Some(swap_details) = maybe_swap_details.as_ref() {
+							Hook::initiate_vault_swap(swap_details.clone());
+						}
```
dandanlen, line 171: If someone submits invalid data they will never find out - we log it in the engines but maybe we should also notify somehow (an event?).

albert-llimos: That sounds reasonable but at this stage we could only emit the event account address, which will be closed afterwards. We will only be able to fetch that on-chain information for a limited period of time.
I guess the most useful part would be for external parties to have a way to identify if their swap has been detected and has failed or not. Having said that, for other chains (at least EVM) I believe the data doesn't even make it to the SC if it fails decoding.
Therefore I'm not sure this adds more information that just monitoring the engine for decoding errors.

ramizhasan111:
There is an edge case where if we initate some vault swaps that are below the account closure threshold, but then if there are no new vault swaps, the account closure will not initiate automatically (which it should after the waiting limit of one day has passed). It will initiate whenever the next new vault swap is created after the waiting limit has passed. This is because the logic of initiating account closures is inside the conditional code block of when we have consensus. So if the time limit has passed and there is no new consensus on new swaps, we wont initiate closures.

To solve this, we need to revert back to the older impl I had where the known_accounts were stored in the UnsynchronizedState and not in the ElectionProperties which would be some work to do. I dont think this is a critical edge case and so maybe we can leave it as is. Basically, it will take a bit longer than a day for account closures to initiate (whenever the next new swap happens). Thoughts? @kylezs @dandanlen @kylezs

albert-llimos:
> To solve this, we need to revert back to the older impl I had where the known_accounts were stored in the UnsynchronizedState and not in the ElectionProperties which would be some work to do. I dont think this is a critical edge case and so maybe we can leave it as is. Basically, it will take a bit longer than a day for account closures to initiate (whenever the next new swap happens). Thoughts? @kylezs @dandanlen @kylezs

I don't think this is a big issue because in normal circumstances the rent will be a small amount of money compared to the amount swapped. Moreover, we are anyway promising some relatively large amount of time (1 day), not instantaneous or any close to that.

However, I do expect to hit this scenario fairly regularly just by looking at the usage of EVM vault swaps (very low until integrators ramp up). I'm not familiar with the advantages or reason of making that change in the first place but I'd say that if it takes some work to revert then it's fine as it is. On the other hand, if it's relatively easy to revert and there is not much of an advantage of having this current implementation then we could revert it.

dandanlen:
I think the correct way to fix it is to pull the account closure trigger outside of the consensus (eg. poll the state from some on_idle hook) instead of using consensus as a trigger. 

But I agree this isn't critical.

state-chain/pallets/cf-elections/src/electoral_systems/solana_vault_swap_accounts.rs
```diff
@@ -0,0 +1,287 @@
...
+			} else {
+				Some(SolanaVaultSwapsVote { new_accounts, confirm_closed_accounts })
+			}
+		} else {
+			None
+		})
+	}
+}
+
+pub fn count_votes<T: Ord + Clone>(
```
kylezs, line 276:
```suggestion
fn count_votes<T: Ord + Clone>(
```

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -472,3 +510,103 @@ impl ElectionEgressWitnesser for SolanaEgressWitnessingTrigger {
...
+	> for SolanaVaultSwapsHandler
+{
+	fn initiate_vault_swap(swap_details: SolanaVaultSwapDetails) {
+		SolanaIngressEgress::process_vault_swap_request(
+			swap_details.from,
+			swap_details.deposit_amount,
+			swap_details.to,
+			swap_details.destination_address,
+			swap_details.deposit_metadata,
+			Default::default(), // TODO txHash PRO-1760
```
kylezs, line 569: pro-1760 is done now, so I guess we can do this now?

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -472,3 +510,103 @@ impl ElectionEgressWitnesser for SolanaEgressWitnessingTrigger {
...
+#[derive(
+	Clone, PartialEq, Eq, Debug, Serialize, Deserialize, TypeInfo, Encode, Decode, PartialOrd, Ord,
+)]
+pub struct SolanaVaultSwapDetails {
+	pub from: SolAsset,
+	pub to: Asset,
+	pub deposit_amount: SolAmount,
+	pub destination_address: EncodedAddress,
+	pub deposit_metadata: Option<CcmDepositMetadata>,
+	// TODO: swap_account and creation_slot will be pulled into TransactionInId type (PRO-1760)
```
kylezs, line 523: same ^, pro-1760 is done now

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,590 @@
...
+async fn get_changed_program_swap_accounts(
+	sol_rpc: &SolRetryRpcClient,
+	sc_opened_accounts: HashSet<SolAddress>,
+	sc_closure_initiated_accounts: BTreeSet<VaultSwapAccountAndSender>,
+	swap_endpoint_data_account_address: SolAddress,
+) -> Result<(Vec<SolAddress>, Vec<VaultSwapAccountAndSender>, u64), anyhow::Error> {
+	let (_historical_number_event_accounts, open_event_accounts, slot) =
+		get_swap_endpoint_data(sol_rpc, swap_endpoint_data_account_address)
+			.await
+			.expect("Failed to get the event accounts");
```
kylezs, line 235: The expect message should state why we expect this not to fail. If it might fail, then we should throw the error up - it does look like it might fail, given we handle some different encodings, and we've seen some RPCs act weirdly on this, so I would say we throw an error to avoid panicking the whole engine

state-chain/runtime/src/chainflip/solana_elections.rs
albert-llimos: I'll do that afterwards in a separate PR on top of this or after it's merged to main. Already opened a ticket for that: https://linear.app/chainflip/issue/PRO-1790/solana-vault-swaps-update-transactioninid

state-chain/runtime/src/chainflip/solana_elections.rs
albert-llimos: same ^

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -0,0 +1,590 @@
...
+
+						SwapEvent::check_and_deserialize(&bytes[..])
+							.map_err(|e| anyhow!("Failed to deserialize data: {}", e))
+					},
+					Some(other) => Err(anyhow!(
+						"Expected UiAccountData::Binary(String, UiAccountEncoding::Base64), got {}",
+						match other.data {
+							UiAccountData::Binary(_, other) => format!("{:?}", other),
+							UiAccountData::Json(_) => "Json".to_string(),
+							UiAccountData::LegacyBinary(_) => "LegacyBinary".to_string(),
```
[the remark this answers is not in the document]
albert-llimos: for context, @dandanlen changed this in this PR: https://github.com/chainflip-io/chainflip-backend/pull/5405
Maybe he can comment on that.

engine/src/witness/sol/program_swaps_witnessing.rs
ramizhasan111: also found another expect where we should also return an error. corrected both.

kylezs (approved):

engine/src/witness/sol/program_swaps_witnessing.rs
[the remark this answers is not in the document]
dandanlen: I think my reasoning was roughly this: Using the debug impl would print the encoding and all the data. If the encoding is wrong, it's enough to know that the rpc returned data with the wrong encoding, and what that encoding was. The actual data is irrelevant.

But I can see where the confusion comes from, i'll make it a bit clearer.

dandanlen (approved):

kylezs:
For visibility @ramizhasan111 is going to do the missing migrations

kylezs:
The upgrade test might pass without those version changes until you run it in the queue because it'll be put on top of main, which will then require them

.github/workflows/ci-development.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
```
kylezs, line 82: reminder: needs to be reverted before merge

albert-llimos:
I'm looking into the Vault swap failures after upgrade.

albert-llimos (approved):

PR #5357, opened 2024-10-25
chore: rename contract swaps -> vault swaps

msgmaxim:
# Pull Request

Closes: PRO-1730

## Summary

- Renamed most mentions of "contract swaps" -> "vault swaps" (at least what I could find, and where it made sense)
- While renaming, noticed that we `contract_swap_request` benchmark is now unused since we recently unified `contract_swap_request` with `contract_swap_ccm_request` (and we can use the "ccm" version to benchmark the new extrinsic), so I removed the other one.


dandanlen (approved):

martin-chainflip (approved):

msgmaxim:
I briefly looked into the bouncer failures, and looks like the tests fail even on main 🤷‍♂️ (although on main the failing tests were LP-API). Will try again tomorrow.

PR #5358, opened 2024-10-25
test: fix submission watcher test

j4m1ef0rd:
# Pull Request

Closes: PRO-1718

## Checklist

## Summary

Changed the `jsonrpsee::Subscription` into a `Stream` so we can return an empty stream in the test.


engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
```diff
@@ -31,6 +31,7 @@ use crate::state_chain_observer::client::{
 	storage_api::{CheckBlockCompatibility, StorageApi},
 	SUBSTRATE_BEHAVIOUR,
 };
+use futures::StreamExt;
```
msgmaxim, line 34: Nit: is this only used in the test file? Might make sense to move the import there if so.

msgmaxim (approved):

kylezs (approved):

engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
j4m1ef0rd: It is required by` transaction_status_stream.next()`.

PR #5359, opened 2024-10-25
fix: take into account only active suspensions

marcellorigotti:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

Changed the way we count for suspensions, we now keep into account if the suspension is still valid, based on the current block.


state-chain/runtime/src/lib.rs
```diff
@@ -2147,7 +2147,8 @@ impl_runtime_apis! {
 		}
 
 		fn cf_suspended_validators() -> Vec<(Offence, u32)> {
-			pallet_cf_reputation::Suspensions::<Runtime>::iter().map(|(key, elem)| (key, elem.len() as u32)).collect()
+			let current_block = <frame_system::Pallet<Runtime>>::block_number();
+			pallet_cf_reputation::Suspensions::<Runtime>::iter().map(|(key, elem)| (key, elem.iter().filter(|(suspended_until, _)| *suspended_until < current_block).collect::<Vec<_>>().len() as u32)).collect()
```
dandanlen, line 2151: I think it should be `>` (or `>=`)

dandanlen:
```suggestion
			pallet_cf_reputation::Suspensions::<Runtime>::iter().map(|(key, elem)| (key, elem.iter().filter(|(suspended_until, _)| *suspended_until > current_block).count() as u32)).collect()
```

marcellorigotti:
Is there a better way to filter out duplicates without collecting to a set?
I tried to use https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.unique, but with no std features those method are disabled..

msgmaxim:
> Is there a better way to filter out duplicates without collecting to a set? I tried to use https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.unique, but with no std features those method are disabled..

Not really, and I believe even `unique` from itertools uses a hashset internally.

state-chain/runtime/src/lib.rs
```diff
@@ -2147,7 +2147,25 @@ impl_runtime_apis! {
...
+				if key == pallet_cf_threshold_signature::PalletOffence::FailedKeygen.into() {
+					return (key, suspended_for_keygen);
+				}
+				(key, elem.iter().filter_map(|(suspended_until, validator)| {
+					if *suspended_until > current_block {
+						return Some(validator);
+					}
+					None
+				}).collect::<BTreeSet<_>>().len() as u32)
+			}).collect()
```
dandanlen, line 2168: FYI there is a pallet method `Reputation::<T>::validators_suspended_for` that might make this a bit more readable.

(`pub fn validators_suspended_for(offences: &[T::Offence]) -> BTreeSet<T::ValidatorId>`)

PR #5360, opened 2024-10-29
fix: `request_swap_deposit_address_with_affiliates` refund address is consistent with destination address type

syan095:
# Pull Request

Closes: PRO-1709

## Checklist

## Summary

Make Refund parameter's address to be consistent with the destination_address


dandanlen:
I think the reason this is failing in the bouncer is because the SDK needs to be updated...

dandanlen (approved):

syan095:
Merge queue is failing due to failed Solana CCM Swaps after runtime migration. Not sure if this is related to this PR as no migration was added.

PR #5361, opened 2024-10-29
feat: support for opening/closing private broker channels

msgmaxim:
# Pull Request

Closes: PRO-1748

## Checklist

## Summary

- Added a storage map `AccountId` -> `ChannelId`
- Added Extrinsics to open/close private channels + tests + benchmarks

Out of scope (these seem like may need some further discussion):
- Charging fee/rent for opening a private channel/keeping it open
- Closing private channels upon account deletion (maybe this isn't necessary for an MVP?)
(will create separate linear tickets for these if we agree to leave them out of this PR)

May need to rebase this after #5357 is merged.


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -582,6 +582,10 @@ pub mod pallet {
 	pub(crate) type ScheduledTxForReject<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, Vec<TaintedTransactionDetails<T, I>>, ValueQuery>;
 
+	#[pallet::storage]
+	pub(crate) type BrokerPrivateChannels<T: Config<I>, I: 'static = ()> =
+		StorageMap<_, Twox64Concat, T::AccountId, ChannelId, OptionQuery>;
```
dandanlen, line 587: This can use the Identity hasher.

```suggestion
		StorageMap<_, Identity, T::AccountId, ChannelId, OptionQuery>;
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1310,6 +1326,50 @@ pub mod pallet {
...
+				!BrokerPrivateChannels::<T, I>::contains_key(&broker_id),
+				Error::<T, I>::PrivateChannelExistsForBroker
+			);
+
+			// TODO: burn fee for opening a channel?
+			let next_channel_id =
+				ChannelIdCounter::<T, I>::try_mutate::<_, Error<T, I>, _>(|id| {
+					*id = id.checked_add(1).ok_or(Error::<T, I>::ChannelIdsExhausted)?;
+					Ok(*id)
+				})?;
```
kylezs, line 1345: we should make this a function, is duplicate

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1310,6 +1326,50 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		#[pallet::call_index(14)]
+		#[pallet::weight(T::WeightInfo::open_private_channel())]
+		pub fn open_private_channel(origin: OriginFor<T>) -> DispatchResult {
+			let broker_id = T::AccountRoleRegistry::ensure_broker(origin)?;
+
+			ensure!(
+				!BrokerPrivateChannels::<T, I>::contains_key(&broker_id),
```
kylezs, line 1336: Should we also restrict this to be just for Bitcoin - it's not useful for any other chains right? probably more importantly is that not restricting it to Bitcoin would be confusing when reading the code, since it's an instantiable pallet, we'd assume it's for all instances.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1310,6 +1326,50 @@ pub mod pallet {
...
+				Error::<T, I>::PrivateChannelExistsForBroker
+			);
+
+			// TODO: burn fee for opening a channel?
```
kylezs, line 1340: I see this burning/rent is mentioned in the ticket, but also that this PR closes that ticket? so maybe we create another to address this separately.

dandanlen:
> Closing private channels upon account deletion (maybe this isn't necessary for an MVP?)
(will create separate linear tickets for these if we agree to leave them out of this PR)

I feel like we can just add this here. We need it and it's not a lot of work to add. Not sure what we gain by waiting. 

(Sorry was sure I left this comment yesterday...)

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: I will create tickets once we decide what should be in the scope of this PR and what isn't (I mentioned this in the description).

kylezs: > Closing private channels upon account deletion (maybe this isn't necessary for an MVP?)
(will create separate linear tickets for these if we agree to leave them out of this PR)

Ah ok, that seemed to suggest was just about the account deletion

dandanlen:
Forgot to mention - I think charging/not charging a fee needs a little more discussion, happy to leave this out for now.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: Deduplicated

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: Added the check.

msgmaxim:
Addressed all comments now. Ready for re-review @dandanlen @kylezs

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1054,6 +1075,34 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::open_private_channel())]
+		pub fn open_private_channel(origin: OriginFor<T>, chain: ForeignChain) -> DispatchResult {
+			let broker_id = T::AccountRoleRegistry::ensure_broker(origin)?;
+
+			ensure!(chain == ForeignChain::Bitcoin, Error::<T>::NoPrivateChannelsForChain);
```
kylezs, line 1084: Now we only inject the BitcoinIngressEgress as the PrivateChannelManager, so it's only possible to create bitcoin channel, or inversely, if we injected the wrong instance of the ingress egress then we would put bitcoin in here, and then open an Ethereum channel. 
 So perhaps would be better to have this be:
```fn open_private_btc_channel(origin)``` and we can remove this ensure, same below.

File: state-chain/pallets/cf-swapping/src/tests.rs [hunk 6]
```diff
@@ -1618,3 +1652,64 @@ mod swap_batching {
...
+			System::assert_has_event(RuntimeEvent::Swapping(
+				Event::<Test>::PrivateBrokerChannelOpened { broker_id: BROKER, channel_id: 1 },
+			));
+		});
+	}
+
+	#[test]
+	fn close_private_channel() {
+		new_test_ext().execute_with(|| {
+			// Only brokers can  channels
```
kylezs, line 1690:
```suggestion
			// Only brokers can open channels
```

kylezs (approved):
Nice tests 👌  - would let Dan approve before merge

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Was going to suggest the same thing - the only benefit of the `chain` parameter is in case we want to add another utxo chain... but that seems unlikely right now and in the meantime it's not used for anything.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1054,6 +1075,34 @@ pub mod pallet {
...
+
+			Ok(())
+		}
+
+		#[pallet::call_index(13)]
+		#[pallet::weight(T::WeightInfo::close_private_channel())]
+		pub fn close_private_channel(origin: OriginFor<T>, chain: ForeignChain) -> DispatchResult {
+			let broker_id = T::AccountRoleRegistry::ensure_broker(origin)?;
+
+			ensure!(chain == ForeignChain::Bitcoin, Error::<T>::NoPrivateChannelsForChain);
```
dandanlen, line 1098: Same here, including the `chain` param seems unnecessary.

dandanlen (commented):
I think we might as well make it btc-specific and remove the chain params to make it clearer.

msgmaxim:
Renamed the extrinsics in cf-swapping and removed the chain parameter. Note that I chose not to rename the methods inside cf-ingress-egress as I thought it would look odd (it would be misleading for non-btc instances to have a btc-specific method), but I added a comment to say that this functionality is expected to be used by bitcoin only. Importantly, we don't have a chain parameter and we made it more explicit to external users that this is bitcoin only feature. @dandanlen

kylezs:
> Note that I chose not to rename the methods inside cf-ingress-egress

I think that makes sense 👍

kylezs (approved):

dandanlen (commented):
LGTM - only thing is that looking at the code now, i'm not sure why we put the storage in the ingess-egress pallet. All it needs to do is return an incremented channel id. We could store it in the swapping pallet and then we have a smaller internal API surface (no need for close/lookup methods in the trait). 

Will leave it up to you @msgmaxim whether to make this change or not. I think it would reduce complexity but having the 'channel' (even though it's just the id) in the same place as other channels also kind of makes sense.

msgmaxim:
> LGTM - only thing is that looking at the code now, i'm not sure why we put the storage in the ingess-egress pallet. All it needs to do is return an incremented channel id. We could store it in the swapping pallet and then we have a smaller internal API surface (no need for close/lookup methods in the trait).
> 
> Will leave it up to you @msgmaxim whether to make this change or not. I think it would reduce complexity but having the 'channel' (even though it's just the id) in the same place as other channels also kind of makes sense.

Made these changes now @dandanlen

dandanlen (approved):

PR #5363, opened 2024-10-29
fix: remove is_vote_valid

kylezs:
# Pull Request

Closes: PRO-1757

## Checklist

## Summary

`is_vote_valid` was only used once, and was used erroneously. Plus, if a single vote within a group is invalid then the whole extrinsic vails, effectively invalidating all the votes. We have is_vote_needed to prevent engines from submitting too many votes. So have removed it, which simplifies things too. No need to hash the data at all, we can compare the values directly in check_consensus.

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
@@ -101,7 +101,8 @@ impl<
 		),
 	) -> bool {
 		match current_vote {
-			AuthorityVoteOf::<Self>::Vote(current_vote) => current_vote != proposed_vote,
+			AuthorityVoteOf::<Self>::Vote(current_vote) =>
+				current_vote.value != proposed_vote.value,
 			// Could argue for either true or false. If the `PartialVote` is never reconstructed and
 			// becomes invalid, then this function will be bypassed and the vote will be considered
 			// needed. So false is safe, and true will likely result in unneeded voting.
--- a/state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
@@ -12,7 +12,7 @@ use frame_support::{
 	pallet_prelude::{MaybeSerializeDeserialize, Member},
 	Parameter,
 };
-use sp_std::{collections::btree_map::BTreeMap, vec, vec::Vec};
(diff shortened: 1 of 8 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs
```diff
@@ -80,17 +76,19 @@ fn generate_votes(
...
 fn generate_votes_with_different_slots(
 	correct_voters: AuthorityCount,
-	correct_value: MonotonicChangeVote<Value, Slot>,
+	correct_value: Value,
 	incorrect_voters: AuthorityCount,
 	incorrect_value: MonotonicChangeVote<Value, Slot>,
 ) -> ConsensusVotes<SimpleMonotonicChange> {
 	ConsensusVotes {
+		// we start from 1 so don't get a Default::default() conflict on vote validity.
 		votes: (0..correct_voters)
```
dandanlen, line 85: This comment is kind of confusing - says 'start from 1' but the range starts from 0?

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
dandanlen, line 1: is_vote_needed is currently this: 

```rust
match current_vote {
			AuthorityVoteOf::<Self>::Vote(current_vote) => current_vote != proposed_vote,
			_ => false,
		}
```

Should it not be:

```rust
match current_vote {
			AuthorityVoteOf::<Self>::Vote(current_vote) => current_vote.value != proposed_vote.value,
			_ => false,
		}
```

ie. we only want a cote if the value has changed, not if the block has changed? 

Otherwise the engines will just vote almost continuously.

state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs
kylezs: stale, removing

state-chain/pallets/cf-elections/src/electoral_systems/monotonic_change.rs
kylezs: Yeah good point, better that way

state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs
kylezs: stale, removed

PR #5371, opened 2024-10-31
fix: expire previous epochs and rename epoch duration

dandanlen:
~~TEMP: opening this PR to investigate recent issues.~~

Appears to be solved. There were a couple of issues:

- It looks like the migration was incorrect.
- Doing the expiry in the on_initialsed hook was causing the block to be filled up with expiry weight. Expiry has now been moved to the on_idle hook and only runs if there is enough remaining weight in the block.

### DIFF
--- a/state-chain/custom-rpc/src/monitoring.rs
+++ b/state-chain/custom-rpc/src/monitoring.rs
@@ -1,8 +1,10 @@
 use super::pass_through;
 use crate::{BlockT, CustomRpc, RpcAccountInfoV2, RpcFeeImbalance, RpcMonitoringData, RpcResult};
 use cf_chains::{dot::PolkadotAccountId, sol::SolAddress};
+use cf_utilities::rpc::NumberOrHex;
 use jsonrpsee::proc_macros::rpc;
 use sc_client_api::{BlockchainEvents, HeaderBackend};
+use serde::{Deserialize, Serialize};
 use sp_api::ApiExt;
 use sp_core::{bounded_vec::BoundedVec, ConstU32};
 use state_chain_runtime::{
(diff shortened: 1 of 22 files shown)


kylezs:
what's the status here @dandanlen ? seems stale now? or potentially the epoch duration change could be pulled out.

dandanlen:
I'd still like to understand what the issue was...

.github/workflows/ci-development.yml
```diff
@@ -69,7 +69,7 @@ jobs:
     uses: ./.github/workflows/_40_post_check.yml
     secrets: inherit
     with:
-      full-bouncer: false
+      full-bouncer: true
```
kylezs, line 72: these flags need to be reset

state-chain/custom-rpc/src/monitoring.rs
```diff
@@ -15,6 +17,44 @@ use state_chain_runtime::{
...
+	pub rotation_phase: String,
+}
+
+// Temporary struct to hold the deprecated blocks_per_epoch field.
+// Can be deleted after v1.7 is released (meaning: after the version is bumped to 1.8).
+#[derive(Serialize, Deserialize, Clone)]
+pub struct RpcEpochStateV2 {
+	#[deprecated(
+		since = "1.8.0",
+		note = "This field is deprecated and will be removed in v1.8. Use blocks_per_epoch instead."
```
kylezs, line 46:
```suggestion
		note = "This field is deprecated and will be removed in v1.8. Use epoch_duration instead."
```

kylezs (approved):

j4m1ef0rd (approved):

kylezs:
two in a row, something still up here I guess?

PR #5375, opened 2024-11-01
Feat: Expand Solana Vault Swap functionality and fix bouncer

albert-llimos:
# Pull Request

Closes: PRO-1697

## Checklist

## Summary

- Update image with the new SwapEndpoint that has the `creation_slot`
- Add support for Vault Swap Attributes
- Add support for Solana Vault CCM swaps
- Fix bouncer issues all around.
- Properly import new types from the Solana program repo after PRO-1710.

Solana Vault swaps are now working end to end. The SDK doesn't have the vault swap logic but it does have the encoding of CfParameters. For the time being we're just manually doing the swap calls in the bouncer and using some dummy parameters for Vault Swap parameters. When it's moved to the SDK I'll expand the testing to also check for `VaultSwapParameters` and some other small things worth testing.

This code might require some tweaks when PRO-1760 and PRO-1743 are implemented but we'll do that separately.


ramizhasan111 (approved):

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -290,7 +334,8 @@ async fn get_program_swap_event_accounts_data(
 				Err(anyhow!("Expected UiAccountData::Binary(String, UiAccountEncoding::Base64)")),
 			None => Ok((account, None)),
```
ramizhasan111, line 335: why this change?

ramizhasan111 (approved):

engine/src/witness/sol/program_swaps_witnessing.rs
```diff
@@ -290,7 +334,8 @@ async fn get_program_swap_event_accounts_data(
...
 		})
-		.collect()
+		.collect();
+	program_swap_event_accounts_iter
```
ramizhasan111, line 338: why this change?

albert-llimos: oh, it's just because I had added some printing before. Will revert.

PR #5377, opened 2024-11-04
tests: Bouncer test for broker level screening

Janislav:
# Pull Request

Closes: PRO-1746

## Checklist

## Summary

Adds a bouncer tests for testing broker level screening on-chain functionality. We are covering the follwing scenarios in this test:

1. No boost and early tx report -> Tainted tx is reported early and the swap is refunded.
2. Boost and early tx report -> Tainted tx is reported early and the swap is refunded.
3. Boost and late tx report -> Tainted tx is reported late and the swap is not refunded.

#### Non-Breaking changes


bouncer/tests/broker_level_screening.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+  console.error(`BTC balance for ${address} did not change after ${MAX_RETRIES} seconds.`);
+  return Promise.resolve(false);
+}
+
+/**
+ * Submits a transaction as tainted to the extrinsic on the state chain.
+ *
+ * @param txId - The txId to submit as tainted as byte array in the correct order.
```
kylezs, line 47:
```suggestion
 * @param txId - The txId to submit as tainted as byte array in the order it is on the Bitcoin chain - which is reverse of how it's normally displayed in block explorers.
```
correct is a bit ambiguous

bouncer/tests/broker_level_screening.ts
```diff
@@ -0,0 +1,194 @@
...
+  // Note: We expect the swap to be executed and not refunded because the tainted tx was to late reported.
+  testBrokerLevelScreening.log('Testing broker level screening with boost and a late tx report...');
+  btcRefundAddress = await newAddress('Btc', randomBytes(32).toString('hex'));
+
+  const channelId = await brokerLevelScreeningTestScenario(
+    '0.2',
+    true,
+    btcRefundAddress,
+    0,
+    MILLI_SECS_PER_BLOCK * 2,
```
kylezs, line 186: ?
```suggestion
    MILLI_SECS_PER_BLOCK * BLOCKS_TO_WAIT,
```

also these are all the same. if we need to wait the same duration, then would just factor that out

Janislav: For now. But maybe not forever. It's likely to change if we maybe want to extend the test suite, modify it or play with different values to see what the results are. We usually define this kind of constants in test suites. Don't see why it should be bad...

kylezs: just unnecessary complexity atm, we can add the parameterisation when we need it (which might be never) - i won't block the the PR because of it, but yeah, is unnecessary

bouncer/tests/all_concurrent_tests.ts
```diff
@@ -39,6 +40,7 @@ async function runAllConcurrentTests() {
     testLpApi.run(),
     testBrokerFeeCollection.run(),
     testBoostingSwap.run(),
+    testBrokerLevelScreening.run(),
```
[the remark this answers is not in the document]
Janislav: I extended the `all_concurrent_tests.ts` to take the `LOCALNET` flag like we do with the number of nodes and just add the test to the execution if we can assume we are in a localnet environment.

bouncer/tests/all_concurrent_tests.ts
```diff
@@ -14,13 +14,16 @@ import { testCancelOrdersBatch } from './create_and_delete_multiple_orders';
...
+import { testBrokerLevelScreening } from './broker_level_screening';
 
 async function runAllConcurrentTests() {
   // Specify the number of nodes via providing an argument to this script.
   // Using regex because the localnet script passes in "3-node".
   const match = process.argv[2] ? process.argv[2].match(/\d+/) : null;
   const givenNumberOfNodes = match ? parseInt(match[0]) : null;
   const numberOfNodes = givenNumberOfNodes ?? 1;
+  // If the third argument is not explicitly false, we assume it's true and we are in a localnet environment.
+  const addConcurrentLocalNetTests = process.argv[3] !== 'false';
```
kylezs, line 26:
```suggestion
  const addConcurrentLocalnetTests = process.argv[3] !== 'false';
```

kylezs (approved):

PR #5382, opened 2024-11-04
test: solana vault swap electoral system tests

ramizhasan111:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

#### Non-Breaking changes


state-chain/pallets/cf-elections/src/electoral_system/mock.rs
kylezs, line 1: `state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs`

We have one defined already

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -990,3 +990,181 @@ impl BenchmarkValue for SolanaVaultSwapsSettings {
...
+	pub const NEW_ACCOUNT_2: SolAddress = SolAddress([2u8; 32]);
+	pub const NEW_ACCOUNT_3: SolAddress = SolAddress([3u8; 32]);
+
+	pub const CLOSED_ACCOUNT_1: SolAddress = SolAddress([4u8; 32]);
+	pub const CLOSED_ACCOUNT_2: SolAddress = SolAddress([5u8; 32]);
+
+	pub const DUMMY_VALIDATOR: [u8; 32] = [0u8; 32];
+
+	#[test]
+	fn test_solana_vault_swaps_consensus() {
```
kylezs, line 1011: would have a look at some tests for the other electoral systems for some ideas on how to tidy this up a bit.
e.g. state-chain/pallets/cf-elections/src/electoral_systems/tests/liveness.rs
state-chain/pallets/cf-elections/src/electoral_systems/tests/monotonic_change.rs

kylezs:
Also we definitely want some tests for the on_finalize too

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
```diff
@@ -0,0 +1,309 @@
...
+				Check::<MinimalVaultSwapAccounts>::close_accounts_hook_not_called(),
+				Check::<MinimalVaultSwapAccounts>::get_sol_nonces_hook_not_called(),
+			],
+		)
+		.expect_consensus(
+			generate_votes_n_to_m_accounts(0, 15),
+			Some(SolanaVaultSwapsVote {
+				new_accounts: (0..15u64).map(|i| (i, ())).collect::<BTreeSet<_>>(),
+				confirm_closed_accounts: BTreeSet::from([CLOSED_ACCOUNT_1, CLOSED_ACCOUNT_2]),
+			}),
```
kylezs, line 126: The construction here is a little strange, we are duplicating code here and then inside generate_votes_n_to_m_accounts. Makes it hard to see why this expectation is actually true just from looking at the test. It's only true because it happens to exactly match what's in `generate_votes_n_to_m_accounts` but this could be explicit at the level of the test

kylezs (commented):
Tests looking good overall 👌

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
```diff
@@ -0,0 +1,309 @@
...
+				Check::<MinimalVaultSwapAccounts>::only_one_election(),
+				Check::<MinimalVaultSwapAccounts>::initiate_vault_swap_hook_called_15_times(),
+				Check::<MinimalVaultSwapAccounts>::close_accounts_hook_called_once(),
+				Check::<MinimalVaultSwapAccounts>::get_sol_nonces_hook_called_once(),
+			],
+		);
+}
+
+#[test]
+fn on_finalize_time_limit_reached() {
```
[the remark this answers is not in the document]
ramizhasan111: Its passing for me  :thinking:

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
ramizhasan111: This test is actually about testing the on_finalize(). It is not testing consensus and so It is not so important if consensus is correct. We just need to provide some consensus value which is then used to test on_finalize so I think it is fine.

kylezs: `.expect_consensus(` means it is testing consensus. If the check_consensus result changes then it impacts the on_finalize so it's no less important

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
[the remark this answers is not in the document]
kylezs: interesting, perhaps something cached in my local env, I clean compiled to be sure, and it works now

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
ramizhasan111: I meant that the test is not suppose to test the consensus so it is not important. I was using the ```expect_consensus``` function to update the consensus. I changed to using the dedicated function to update the consensus (```force_consensus_update```) so that this fact is clear.

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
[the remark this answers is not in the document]
ramizhasan111: added tests for these cases.

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
```diff
@@ -0,0 +1,437 @@
...
+			)
+			.chain(
+				(0..no_of_each_vote[3])
+					.map(|_| ConsensusVote { vote: Some(((), vote_4.clone())), validator_id: () }),
+			)
+			.collect::<Vec<_>>(),
+	}
+}
+
+fn generate_votes_for_account_range(n: u64, m: u64) -> SolanaVaultSwapsVote<Account, SwapDetails> {
```
kylezs, line 432: the main part on the comment here is that we can use the `Range` type rather than provide two values like this (there are a couple available, I think you want the core::ops::Range)

state-chain/pallets/cf-elections/src/electoral_systems/tests/solana_vault_swap_accounts.rs
```diff
@@ -0,0 +1,309 @@
...
+				(0..no_of_each_vote[2])
+					.map(|_| ConsensusVote { vote: Some(((), vote_3.clone())), validator_id: () }),
+			)
+			.chain(
+				(0..no_of_each_vote[3])
+					.map(|_| ConsensusVote { vote: Some(((), vote_4.clone())), validator_id: () }),
+			)
+			.collect::<Vec<_>>(),
+	}
+}
```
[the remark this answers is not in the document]
kylezs: Think you missed this comment?

kylezs (approved):

PR #5383, opened 2024-11-05
Feat: Private Broker Channel Witnessing

msgmaxim:
# Pull Request

Closes: PRO-1749

## Summary

- Added a new adapter (similar to the existing `deposit_addresses`) which monitors the new storage for private channels `BrokerPrivateBtcChannels` on every block.
- `btc_deposits` now has access to private channel ids which it uses to derive deposit addresses (based on the current and previous epoch keys) and checks transactions for vault deposits into them (similarly to how it is already done for channel_id = 0 aka "change address").
- Based on #5361, will need to rebase after that one is merged.

### Discussion

- Do we still want to allow vault deposits directly into the "change address" skipping the broker?

### DIFF
--- a/engine/src/state_chain_observer/client/storage_api.rs
+++ b/engine/src/state_chain_observer/client/storage_api.rs
@@ -180,6 +180,13 @@ pub trait StorageApi {
 			.map(|(_k, v)| v)
 			.collect())
 	}
+
+	async fn storage_map_entries<StorageMap: StorageMapAssociatedTypes + 'static>(
+		&self,
+		block_hash: state_chain_runtime::Hash,
+	) -> RpcResult<Vec<(StorageMap::Key, StorageMap::Value)>> {
+		Ok(self.storage_map::<StorageMap, Vec<_>>(block_hash).await?.into_iter().collect())
+	}
 }
 
 #[async_trait]
--- a/engine/src/state_chain_observer/client/storage_api.rs
+++ b/engine/src/state_chain_observer/client/storage_api.rs
@@ -180,13 +180,6 @@ pub trait StorageApi {
 			.map(|(_k, v)| v)
 			.collect())
 	}
-
-	async fn storage_map_entries<StorageMap: StorageMapAssociatedTypes + 'static>(
-		&self,
-		block_hash: state_chain_runtime::Hash,
-	) -> RpcResult<Vec<(StorageMap::Key, StorageMap::Value)>> {
-		Ok(self.storage_map::<StorageMap, Vec<_>>(block_hash).await?.into_iter().collect())
-	}
 }
 
 #[async_trait]
--- a/engine/src/witness/btc.rs
+++ b/engine/src/witness/btc.rs
@@ -128,6 +128,8 @@ where
 		.chunk_by_vault(vaults.clone(), scope)
 		.deposit_addresses(scope, unfinalised_state_chain_stream, state_chain_client.clone())
 		.await
+		.private_deposit_channels(scope, state_chain_stream.clone(), state_chain_client.clone())
+		.await
 		.btc_deposits(prewitness_call)
 		.logging("pre-witnessing")
 		.spawn(scope);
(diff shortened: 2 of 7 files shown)


engine/src/state_chain_observer/client/storage_api.rs
```diff
@@ -180,6 +180,13 @@ pub trait StorageApi {
 			.map(|(_k, v)| v)
 			.collect())
 	}
+
+	async fn storage_map_entries<StorageMap: StorageMapAssociatedTypes + 'static>(
+		&self,
+		block_hash: state_chain_runtime::Hash,
+	) -> RpcResult<Vec<(StorageMap::Key, StorageMap::Value)>> {
+		Ok(self.storage_map::<StorageMap, Vec<_>>(block_hash).await?.into_iter().collect())
```
kylezs, line 188: could just inline this, we can just do
```rust
state_chain_client
							.storage_map::<pallet_cf_swapping::BrokerPrivateBtcChannels<
								state_chain_runtime::Runtime,
							>, Vec<_>>(block_hash)
							.await
							.expect(STATE_CHAIN_CONNECTION)
```

engine/src/witness/btc.rs
```diff
@@ -128,6 +128,8 @@ where
 		.chunk_by_vault(vaults.clone(), scope)
 		.deposit_addresses(scope, unfinalised_state_chain_stream, state_chain_client.clone())
 		.await
+		.private_deposit_channels(scope, state_chain_stream.clone(), state_chain_client.clone())
```
kylezs, line 131: 👀 
```suggestion
		.private_deposit_channels(scope, unfinalised_state_chain_stream, state_chain_client.clone())
```

engine/src/witness/btc/deposits.rs
```diff
@@ -55,25 +55,30 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
 
 					let key: &AggKey = &epoch.info.0;
 
-					let maybe_previous_vault_address =
-						key.previous.map(|key| DepositAddress::new(key, CHANGE_ADDRESS_SALT));
-					let current_vault_address =
-						DepositAddress::new(key.current, CHANGE_ADDRESS_SALT);
-
-					[current_vault_address].into_iter().chain(maybe_previous_vault_address)
+					[key.current].into_iter().chain(key.previous).flat_map(|key| {
```
kylezs, line 68: this block is reasonably clear if you know the context, but without it, might be a little hard to parse, so maybe a little comment reasoning it through here

engine/src/witness/common/chunked_chain_source/chunked_by_vault/private_deposit_channels.rs
```diff
@@ -0,0 +1,73 @@
...
+							.storage_map_entries::<pallet_cf_swapping::BrokerPrivateBtcChannels<
+								state_chain_runtime::Runtime,
+							>>(block_hash)
+							.await
+							.expect(STATE_CHAIN_CONNECTION)
+					}
+				},
+				// Private channels are not reusable (at least at the moment), so we
+				// don't need to check for their expiration:
+				|_index, addresses: &BrokerPrivateChannels| addresses.clone(),
```
kylezs, line 67: we can/should probably still add the first assert that's in `deposit_addresses` as a sanity check:

`assert!(<Inner::Chain as Chain>::is_block_witness_root(index));`

kylezs (commented):
This seems like the simplest approach 👍  just some smaller comments

engine/src/witness/btc.rs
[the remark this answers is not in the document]
msgmaxim: I was going to follow your suggestion (though I didn't entirely agree), but as I was making changes I realised that this would be problematic: `btc_deposits` expects private channels to be present. We could probably think of a workaround, but seems not worth it considering that we are planning to add support for this very soon (probably even in the same release).

msgmaxim: good catch!

msgmaxim:
@kylezs I rebased to clean up after the base PR got merged. All new changes are the new/separate commit.

msgmaxim:
> > Do we still want to allow vault deposits directly into the "change address" skipping the broker?
> 
> Hmmm, my first thought was why not, but I think removing a bit of logic that will more than likely never be used since they'll be going through the broker is probably better. This could be done in a separate PR though (or here, will leave that to you).

Do you agree with this @dandanlen? If so, I will go ahead and remove it here.

dandanlen:
As discussed - I don't think this is a technical decision. We can merge this as is and make the decision to remove at a later point.

engine/src/witness/btc.rs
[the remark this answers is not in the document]
kylezs: yes true

PR #5385, opened 2024-11-06
feat: add version and affiliates to UTXO encoding

msgmaxim:
# Pull Request

Closes: PRO-1759, PRO-1770

## Summary

- Added version byte (currently set to 0); decoding will fail if it is not 0
- Added broker fee field (as a u8)
- Added affiliate fees. Supports up to 2 affiliates. Each entry takes 2 bytes (1 byte for affiliate idx, and another byte for their fee)

I briefly considered inlining `SharedCfParameters` (since we are not going to share them after all), but it seemed helpful to have a separate struct for fields that we don't want to decode independently (unlike the version, output asset and address), so I left it for now.

### DIFF
--- a/state-chain/chains/src/btc/vault_swap_encoding.rs
+++ b/state-chain/chains/src/btc/vault_swap_encoding.rs
@@ -2,12 +2,15 @@ use crate::address::EncodedAddress;
 use cf_primitives::{Asset, AssetAmount, ForeignChain};
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
+use sp_core::ConstU32;
+use sp_runtime::BoundedVec;
 use sp_std::vec::Vec;
 
 use super::{BitcoinOp, BitcoinScript};
 
 // The maximum length of data that can be encoded in a nulldata utxo
 const MAX_NULLDATA_LENGTH: usize = 80;
+const CURRENT_VERSION: u8 = 0;
 
 #[derive(Clone, PartialEq, Debug, TypeInfo)]
 pub struct UtxoEncodedData {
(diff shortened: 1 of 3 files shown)


state-chain/chains/src/btc/vault_swap_encoding.rs
```diff
@@ -20,6 +23,8 @@ impl Encode for UtxoEncodedData {
 	fn encode(&self) -> Vec<u8> {
 		let mut r = Vec::with_capacity(MAX_NULLDATA_LENGTH);
 
+		r.push(CURRENT_VERSION);
```
dandanlen, line 26: We should encode the version here.

```suggestion
		CURRENT_VERSION.encode_to(&mut r);
```

dandanlen:
> I briefly considered inlining SharedCfParameters (since we are not going to share them after all)

What did I miss? Why are they no longer shared?

dandanlen (approved):
I applied my suggestion, I don't think it's controversial.

state-chain/chains/src/btc/vault_swap_encoding.rs
msgmaxim: What do you mean? Encoding of a single byte is just the byte itself... If this is just about style, I'm okay either way, although i did use push intentionally to make it slightly more explicit, but I also understand the consistency argument.

msgmaxim:
> > I briefly considered inlining SharedCfParameters (since we are not going to share them after all)
> 
> What did I miss? Why are they no longer shared?

See https://discord.com/channels/775961728608895008/1288753906494472202/1296348630919155734

msgmaxim:
I'm going to hit merge so we can start using this code. Feel free to still leave comments @kylezs, althought I don't think there is anything controversial (and this PR is small anyway).

PR #5387, opened 2024-11-06
feat: Custom TypeInfo derivation macro for generic pallets

martin-chainflip:
# Pull Request

Closes: PRO-1587

## Checklist

## Summary

This change introduces a new derive macro "GenericTypeInfo" that can replace the existing "TypeInfo" macro for structs or enums in generic pallets that have different types depending on the chain the pallet is implemented for.


state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+		type Identity = Self;
+		fn type_info() -> scale_info::Type {
+			let expanded_ident = scale_info::prelude::format!("{}For{}", ::core::stringify!(#ident), T::NAME).into_boxed_str();
+			scale_info::Type::builder()
+			.path(scale_info::Path::new(Box::leak(expanded_ident), ::core::module_path!()))
+			.variant(
+				scale_info::build::Variants::new() #( #variants )* )
+		}
+	})
+}
```
kylezs, line 90: can dedupe more:
```suggestion
fn derive_type_info(
	ident: Ident,
	impl_generics: ImplGenerics,
	ty_generics: TypeGenerics,
	body: TokenStream2,
) -> TokenStream2 {
	quote!(impl #impl_generics TypeInfo for #ident #ty_generics {
		type Identity = Self;
		fn type_info() -> scale_info::Type {
			let expanded_ident = scale_info::prelude::format!("{}For{}", ::core::stringify!(#ident), T::NAME).into_boxed_str();
			scale_info::Type::builder()
				.path(scale_info::Path::new(Box::leak(expanded_ident), ::core::module_path!()))
				#body
		}
	})
}

fn derive_struct(
	d: DataStruct,
	ident: Ident,
	impl_generics: ImplGenerics,
	ty_generics: TypeGenerics,
) -> TokenStream2 {
	let fields = derive_fields(d.fields);
	derive_type_info(ident, impl_generics, ty_generics, quote!(.composite(#fields)))
}

fn derive_enum(
	d: DataEnum,
	ident: Ident,
	impl_generics: ImplGenerics,
	ty_generics: TypeGenerics,
) -> TokenStream2 {
	let variants: Vec<_> = d
		.variants
		.into_iter()
		.enumerate()
		.map(|(index, variant)| {
			let name = variant.ident;
			let fields = derive_fields(variant.fields);
			quote!(.variant(::core::stringify!(#name), |v| {
				v.index(#index as u8).fields(#fields)
			}))
		})
		.collect();
	derive_type_info(
		ident,
		impl_generics,
		ty_generics,
		quote!(.variant(scale_info::build::Variants::new() #( #variants )* )),
	)
}
```

state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+fn derive_fields(d: Fields) -> TokenStream2 {
+	match d {
+		syn::Fields::Named(fields_named) => {
+			let fields: Vec<TokenStream2> = fields_named.named.iter().map(|field| {
+                let ty = field.clone().ty;
+                let typename = clean_type_string(&quote!(#ty).to_string());
+                let name = field.clone().ident.unwrap();
+                quote!(.field(|f| {f.ty::<#ty>().type_name(scale_info::prelude::format!("{}For{}", #typename, T::NAME).leak()).name(::core::stringify!(#name))}))
```
dandanlen, line 28: Am I right that this assumes that the type of the field *also* has `#[derive(GenericTypeInfo)]`? 

For example this code:

```
#[derive(GenericTypeInfo)]
struct A<T, I> {
	inner: B<T, I>,
}
```

The inner type will be set to `BForT` - but what if we don't derive GenericTypeInfo for B? Would this be caught at compile time? Or does it somehow propagate magically through the typeinfo registry?


(Side not - tbh. I would also prefer some more explicit distinguishing constant than just `NAME`. Like `TYPE_INFO_SUFFIX` or something.)

state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,110 @@
+extern crate proc_macro;
```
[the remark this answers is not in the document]
dandanlen: Yes - That or a test :)

state-chain/pallets/cf-broadcast/Cargo.toml
```diff
@@ -22,6 +22,7 @@ cf-primitives = { workspace = true }
 cf-traits = { workspace = true }
 cfe-events = { workspace = true }
 cf-runtime-utilities = { workspace = true }
+generic-typeinfo-derive = { path = '../../generic-typeinfo-derive'}
```
dandanlen, line 25: Please put it in the workspace toml :)

```suggestion
generic-typeinfo-derive = { workspace = true }
```

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -122,6 +122,9 @@ pub mod pallet {
 	#[pallet::config]
 	#[pallet::disable_frame_system_supertrait_check]
 	pub trait Config<I: 'static = ()>: Chainflip {
+		/// Name of the chain that this Config is implemented for
+		const NAME: &'static str;
```
dandanlen, line 126: Can we constrain this to match the TargetChain::NAME?

(Although I think simply creating a trait or passing the name in as a derive arg would be nicer, for example 
```rust
#[derive(GenericTypeInfo)]
#[generic_type_info("<T as TargetChain>::NAME")]
```
Then we don't need to make any assumptions on what the traits contain.
)

state-chain/generic-typeinfo-derive/src/lib.rs
dandanlen: Also btw. we're assuming that the type param will always be `T` - probably ok for now but we can probably also pull the actual type param from the parsed input.

dandanlen, line 28: Does this work? (to avoid the `leak`)

```suggestion

				let typename = std::format!("{}For{}", typename, quote!(T::NAME));
				quote!(.field(|f| f.ty::<#ty>().type_name(#typename)))
```

Also, i'm surprised we need to stringify the name - can syn not handle this? Ie. we convert the Ident to a string: 
instead of `let name = field.clone().ident.unwrap();`
we just use `let name = field.ident.map(|i| i.to_string()).unwrap();`

And then we use `.name(#name)` without the stringify.

kylezs:
@martin-chainflip this crate might help: https://crates.io/crates/paste

dandanlen:
I also came across this (should already be in scope): https://docs.rs/quote/latest/quote/macro.format_ident.html

martin-chainflip:
Replaced with #5775

PR #5389, opened 2024-11-07
feat: Affiliate Broker Registration and Lookup for Vault Swaps

msgmaxim:
# Pull Request

Closes: PRO-1750, PRO-1751

## Summary

- Brokers can register affiliates through the swapping pallet to assign them short ids. Brokers are expected to provide short id that they want to use (rather than SC deciding on their behalf), which also means that they can overwrite existing entries if they want (this can be useful since we only support 256 entries). The new `AffiliateRegistrationUpdated` event includes (among other things) `Option<AccountId>` for the entry that is being replaced (most likely to be used as a sanity check to see that they haven't overwritten anything).

- `vault_swap_request` extrinsic now takes `Option<(Beneficiary<T::AccountId>, Affiliates<AffiliateShortId>)>` where the primary broker is AccountId and affiliate brokers are "short ids" (`u8`). Short affialite ids are converted to full `AccountId`s using affiliate registry (added a new trait for this).

- BTC witnesser now passes decoded broker id and affiliate short ids into the `vault_swap_request` call. This hasn't been done for other chains, but should be trivial to add ones primary broker id can be decoded there (I believe @albert-llimos is working on this).

- A small refactor in the testing/mock code: moved `mock_eth` and `mock_btc` under the new `mocks` module so that they can share code in a more sensible way (they now share the new `MockAffiliateRegistry`, which I didn't put under `cf_traits` sincie it is unlikely that it will be used by any other pallet).


albert-llimos:
> BTC witnesser now passes decoded broker id and affiliate short ids into the vault_swap_request call. This hasn't been done for other chains, but should be trivial to add ones primary broker id can be decoded there (I believe @albert-llimos is working on this).
Yes, I'll get that done.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1111,6 +1131,31 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Associates `short_id` with `affiliate_id` for a given broker. Overwrites the record
+		/// under `short_id` if already taken by another affiliate.
+		#[pallet::call_index(14)]
+		#[pallet::weight(T::WeightInfo::register_affiliate())]
+		pub fn register_affiliate(
+			origin: OriginFor<T>,
+			short_id: u8,
```
kylezs, line 1141:
```suggestion
			short_id: AffiliateShortId,
```

engine/src/witness/btc/deposits.rs
```diff
@@ -83,11 +83,13 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 				};
 
 				// TODO: provide broker id (along with broker fees) in the call
-				for (_maybe_broker_id, vault_address) in vault_addresses {
+				for (maybe_broker_id, vault_address) in vault_addresses {
```
dandanlen, line 86: The TODO comment can be removed now?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1285,7 +1290,7 @@ pub mod pallet {
 			deposit_metadata: Option<CcmDepositMetadata>,
 			tx_hash: TransactionHash,
 			deposit_details: Box<<T::TargetChain as Chain>::DepositDetails>,
-			broker_fees: Beneficiaries<T::AccountId>,
+			broker_fees: Option<(Beneficiary<T::AccountId>, Affiliates<AffiliateShortId>)>,
```
dandanlen, line 1293: Already discussed with @albert-llimos I believe, but I think the consensus was that we should make the broker mandatory.

kylezs (approved):

state-chain/pallets/cf-ingress-egress/src/mocks.rs
```diff
@@ -0,0 +1,40 @@
...
+use sp_runtime::BoundedBTreeMap;
+
+pub mod mock_btc;
+pub mod mock_eth;
+
+parameter_types! {
+	pub storage AffiliateMapping:  BoundedBTreeMap<(u64, AffiliateShortId), u64, ConstU32<100>> = Default::default();
+}
+
+pub struct MockAffiliateRegistry {}
```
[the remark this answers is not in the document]
msgmaxim: I intentionally didn't put in in cf_traits (see my note in the description to this PR). I prefer limiting code's visibility to only where it is needed. We can move it later if actually needed.

> We're likely to need it in other places (for example when it comes to processing payouts).

We will have full ids at that point.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2193,6 +2198,36 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+								account,
+								&primary_broker
+							);
+							None
+						}
+					}))
+					.collect();
+
+				fees.try_into().expect(
+					"must fit since affiliates are limited to 1 fewer element than beneficiaries",
```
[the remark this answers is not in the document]
msgmaxim: Updated `MAX_BENEFICIARIES` to be derived as `MAX_AFFILIATES + 1` and now `MAX_BENEFICIARIES` is used when defining `Beneficiaries` type.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: Sure, but does this also mean that we won't allow swaps into our change address? Also, I believe what @albert-llimos was discussing was mostly extenral interfaces? Can we not leave Option in this extrinsic call which is only used by our own engines? I also saw mentions of using "fallback" brokers in some edge cases and None seems like a good candidate for this.

albert-llimos: That's right, my idea was to require a broker Account to be witnessed as part of a Vault swap. Then we would need to check if that broker AccountId exists. If so, it's all good, and for now if it doesn't exist we would fallback to some broker or for now just let it through in some way.
However, I imagine the checking of the existence of the broker AccountId is better done in the SC, not in the engine. Therefore, it makes sense to not make this optional as we are making it mandatory in the engine witnessing. The SC will then check it's existence and fallback if necessary. So imo there is no real use of the engine passing an Option here. If in the future we make it mandatory to pass a valid AccountId, we would anyway make that change in the SC.

msgmaxim: Well, the use case for None right now is supporing deposits into the change address. I never got a definitive answer on whether or not we want to allow them, but I guess we are not supporting after all since the broker in mandatory?

albert-llimos: I don't know what the change address is :sweat_smile:

state-chain/pallets/cf-ingress-egress/src/mocks.rs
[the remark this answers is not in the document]
dandanlen: The convention of implementing the mock alongside the trait exists for good reason. We can implement it once, and it can serve as a 'reference implementation' of how the trait should behave. We know where to find it and we avoid duplicate implementations.

The 'only implement it where it's needed' approach works for you right now (and probably also for you in the future) because you know where it's used. If someone *does* want to use it later (and we can't anticipate every possible use case of a trait), they may not know that there is already a mock defined in the ingress-egress pallet, and decide to 'only implement it where it's needed'. Now we have two different mock implementations.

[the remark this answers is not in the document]
msgmaxim: >  they may not know that there is already a mock defined in the ingress-egress pallet, and decide to 'only implement it where it's needed'.

To make this assessment I would (reasonably) expect them to see how/where it is used 🤷‍♂️. Can't say I'm convinced, but don't have a strong opinion. I will move the mock like you suggest.

msgmaxim:
Addressed all comments now @dandanlen. I also added a check that primary broker is registered as a broker. I chose to just broker fields if the broker is unknown (after emitting a new event), which seems like a reasonable fallback behaviour. If we decide to refund instead, I think this can probably done separately.

albert-llimos:
> If we decide to refund instead, I think this can probably done separately.

For now I'd use the fallback as there doesn't seem to be any strong argument (or anyone) pushing for refunding instead. The current behavior sounds good to me.

state-chain/pallets/cf-ingress-egress/src/mocks.rs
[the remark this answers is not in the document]
dandanlen: Thank you.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -748,6 +753,13 @@ pub mod pallet {
 		FailedToRejectTaintedTransaction {
 			tx_id: <T::TargetChain as Chain>::DepositDetails,
 		},
+		UnknownPrimaryBroker {
+			broker_id: T::AccountId,
+		},
+		UnknownAffiliateBroker {
```
dandanlen, line 759:
```suggestion
		UnknownBroker {
			broker_id: T::AccountId,
		},
		UnknownAffiliate {
```

Do we need to introduce this new distinction between 'primary broker' vs 'affiliate broker'? Is broker/affiliate not enough?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2206,6 +2221,40 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				.into_iter()
+				.chain(affiliate_fees.into_iter().filter_map(
+					|Beneficiary { account: short_affiliate_id, bps }| {
+						if let Some(affiliate_id) =
+							T::AffiliateRegistry::lookup(&primary_broker, short_affiliate_id)
+						{
+							Some(Beneficiary { account: affiliate_id, bps })
+						} else {
+							// In case the entry not found, we ignore the entry, but process the
+							// swap (to avoid having to refund it).
```
dandanlen, line 2238: 👌

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2206,6 +2221,40 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+					},
+				))
+				.collect::<Vec<_>>()
+				.try_into()
+				.expect("must fit since affiliates are limited to 1 fewer element than beneficiaries")
+			} else {
+				Self::deposit_event(Event::<T, I>::UnknownPrimaryBroker {
+					broker_id: primary_broker,
+				});
+				Default::default()
```
dandanlen, line 2255: Do we have a Linear issue to deal with this ie. what to do as fallback? We should probably store the data somewhere so we can reject the deposit / trigger a swap.

dandanlen (approved):
👌

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: I'm not sure. To me "affiliate" is short for "affiliate broker", which means simply "broker" is ambiguous 🤷‍♂️. Not a problem if we stick to the convention, but there is also value in being explicit. No strong opinion though, I will make the change.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: But we already do trigger a swap. We can consider refunding of course, but I figure this needs some discussion first. Storing which data are you talking about tough?

dandanlen: As discussed I've opened a triage issue PRO-1804 

I think we might be able to take the same path as we do for transaction screening.

PR #5392, opened 2024-11-07
feat: add cf-parameters versioning

albert-llimos:
# Pull Request

Closes: PRO-1767

## Checklist

## Summary

Updated SDK with changes in: https://github.com/chainflip-io/chainflip-sdk-monorepo/pull/1037/commits/26c1406a1fa21e1bffd468ff816498ab5867dad3

### DIFF
--- a/engine/src/witness/common/cf_parameters.rs
+++ b/engine/src/witness/common/cf_parameters.rs
@@ -4,14 +4,18 @@ use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 
 #[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
+pub enum VersionedCfParameters<CcmData = ()> {
+	V0(CfParameters<CcmData>),
+}
+
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
 pub struct CfParameters<CcmData = ()> {
-	/// CCMs may require additional data (for example CCMs to Solana require adding a list of
-	/// addresses).
+	/// CCMs may require additional data (e.g. CCMs to Solana requires a list of addresses).
 	pub ccm_additional_data: CcmData,
 	pub vault_swap_parameters: VaultSwapParameters,
 }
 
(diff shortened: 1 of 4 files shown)


engine/src/witness/common/cf_parameters.rs
```diff
@@ -3,15 +3,19 @@ use cf_primitives::{BasisPoints, Beneficiaries, DcaParameters};
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
+pub enum VersionedCfParameters<CcmData = ()> {
```
msgmaxim, line 7: Do we have a test to check that the way this is encoded is what we expect (and that we don't accidentally change it in the future)?

albert-llimos: Not a specific unit test but the bouncer does test this as it encodes the data following this type. If this changes without the SDK (in future the broker api) changing it will break.

dandanlen (approved):

engine/src/witness/common/cf_parameters.rs
msgmaxim: I don't know if we should rely on bouncer for things that are easy to put in a unit test. "We can decode" doesn't provide the same level of assurance as "the encoding hasn't changed". Like, it is possilbe that some parameters, like min price or DCA, are decoded wrong without resulting in an error, but bouncer will still succeed because we might see a successful swap still. This will be more important as we allow multiple versions of the encoding (and the bouncer will likely only test one).

I also like to use unit tests in cases like this as a form of documentation: you can see at a glance what the encoded data will look like. E.g. the reason I asked about the test is because I wanted to double check that enum will be encoded as a single byte following the variant without looking up the scale's documentation.

albert-llimos: I did add DCA and FoK for vault swaps to be able to test that. Having said that, it totally makes sense what you said, just added a ticket to myself: https://linear.app/chainflip/issue/PRO-1780/add-encodedecode-tests-for-versionedcfparameters

PR #5398, opened 2024-11-08
chore: build and run script for localnet

j4m1ef0rd:
# Pull Request

## Summary

Simple script to:  cargo build -> recreate localnet -> setup for test. I have been doing this exact chain of commands for a while now when developing. I think others might too, so this is just a shortcut for us.
Usage: `./localnet/build_and_run.sh`

### DIFF
--- a/localnet/build_and_run.sh
+++ b/localnet/build_and_run.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# Does a cargo build and then recreates the localnet using the new binaries.
+# Usage: ./localnet/build_and_run.sh
+
+cargo build && ./localnet/recreate.sh -d && cd bouncer && ./setup_for_test.sh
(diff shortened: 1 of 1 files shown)


localnet/build_and_run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# Does a cargo build and then recreates the localnet using the new binaries.
+# Usage: ./localnet/build_and_run.sh
+
+cargo build && ./localnet/recreate.sh -d && cd bouncer && ./setup_for_test.sh
```
kylezs, line 5: So it doesn't actually change your directory
```suggestion
cargo build && ./localnet/recreate.sh -d && (cd bouncer && ./setup_for_test.sh)
```

kylezs (approved):

PR #5407, opened 2024-11-11
chore: bump redis dependency

msgmaxim:
# Pull Request

## Summary

Redis 0.27.4 has been yanked, which makes cargo audit sad. Bumping to 0.27.5 should make it happy again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9475,9 +9475,9 @@ dependencies = [
 
 [[package]]
 name = "redis"
-version = "0.27.4"
+version = "0.27.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


j4m1ef0rd (approved):

msgmaxim:
No idea why cargo audit fails now. I have fixed the redis issue (and verified with cargo cf-audit locally), but in CI I get: 

```
error: couldn't check if the package is yanked: registry: Failed to retrieve scale-encode from crates.io index: registry: status code '503 Service Unavailable': the status code is invalid for this protocol
```

Any ideas? Maybe @ahasna?

kylezs:
There seemed to be another couple of unmaintained reports, fixed those now. Still getting that error in the logs for a bit before it passes though.

PR #5408, opened 2024-11-11
Feat/cf parameters broker affiliates

albert-llimos:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

Add affiliates and broker account support for EVM Vault Swaps as a followup of #5389

### DIFF
--- a/engine/src/witness/arb.rs
+++ b/engine/src/witness/arb.rs
@@ -177,7 +177,8 @@ struct ArbCallBuilder {}
 
 use cf_chains::{address::EncodedAddress, CcmDepositMetadata, ChannelRefundParameters};
 use cf_primitives::{
-	Asset, AssetAmount, BasisPoints, Beneficiaries, DcaParameters, TransactionHash,
+	AccountId, Affiliates, Asset, AssetAmount, BasisPoints, Beneficiary, DcaParameters,
+	TransactionHash,
 };
 
 impl super::evm::vault::IngressCallBuilder for ArbCallBuilder {
@@ -190,7 +191,8 @@ impl super::evm::vault::IngressCallBuilder for ArbCallBuilder {
 		destination_address: EncodedAddress,
 		deposit_metadata: Option<CcmDepositMetadata>,
 		tx_hash: TransactionHash,
-		_broker_fees: Beneficiaries<ShortId>,
+		_broker_fees: Beneficiary<AccountId>,
+		_affiliate_fees: Affiliates<ShortId>,
 		refund_params: Option<ChannelRefundParameters>,
 		dca_params: Option<DcaParameters>,
 		// This is only to be checked in the pre-witnessed version
@@ -205,8 +207,9 @@ impl super::evm::vault::IngressCallBuilder for ArbCallBuilder {
 				deposit_metadata,
 				tx_hash,
 				deposit_details: Box::new(DepositDetails { tx_hashes: Some(vec![tx_hash.into()]) }),
-				// Defaulting to no broker fees until PRO-1743 is completed.
+				// TODO: To update after PRO-1751 for both broker_fees and affiliate_fees
 				broker_fees: Default::default(),
+				// affiliate_fees: Default::default(),
 				boost_fee: boost_fee.unwrap_or_default(),
 				dca_params,
 				refund_params: refund_params.map(Box::new),
--- a/engine/src/witness/common/cf_parameters.rs
+++ b/engine/src/witness/common/cf_parameters.rs
@@ -4,14 +4,18 @@ use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 
 #[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
+pub enum VersionedCfParameters<CcmData = ()> {
+	V0(CfParameters<CcmData>),
+}
+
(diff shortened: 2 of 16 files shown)


engine/src/witness/common/cf_parameters.rs
```diff
@@ -22,18 +26,26 @@ pub struct VaultSwapParameters {
 	pub refund_params: ChannelRefundParameters,
 	pub dca_params: Option<DcaParameters>,
 	pub boost_fee: Option<BasisPoints>,
```
dandanlen, line 28: Unrelated to this PR but just noticed this is `Option<BasisPoints>` - should it not just be `BasisPoints`? We can save a byte 😅

engine/src/witness/arb.rs
```diff
@@ -205,15 +199,17 @@ impl super::evm::vault::IngressCallBuilder for ArbCallBuilder {
...
+				affiliate_fees: vault_swap_parameters
+					.affiliate_fees
+					.into_iter()
+					.map(|entry| Beneficiary { account: entry.affiliate.into(), bps: entry.fee.into() })
+					.collect_vec()
+					.try_into()
+					.expect("runtime supports at least as many affiliates as we allow in cf_parameters encoding"),
+				boost_fee: vault_swap_parameters.boost_fee.unwrap_or_default(),
+				dca_params: vault_swap_parameters.dca_params,
+				refund_params: Some(Box::new(vault_swap_parameters.refund_params)),
```
dandanlen, line 212: How come refund params are optional on the extrinsic but we never use None?

dandanlen (commented):
LGTM but would like @msgmaxim to look over it too.

engine/src/witness/common/cf_parameters.rs
albert-llimos: Originally when I wrote this my thought was that 1 byte makes no difference in any chain (not BTC as we don't use that there) and that it actually makes more sense to use None instead of boost of zero. However seeing that the extrinsic doesn't take an option it makes sense to remove that for consistency.
I'll change that, I can push it on top of this now that I'm also modifying the SDK.

engine/src/witness/arb.rs
albert-llimos: I guess it's because we wanted to have some flexibility as we've been going back and forth between making this mandatory or not. I think we could consider removing the option since we are now enforcing it in Vault swaps. Wdyt? @dandanlen @msgmaxim

engine/src/witness/common/cf_parameters.rs
albert-llimos: Also technically the option saves you one byte compared to using `BasisPoints` when not being used, which is actually for all swaps using this as we don't support boost for any of these chains (EVM/SOL).

engine/src/witness/arb.rs
msgmaxim: Yeah, let's remove `Option`

engine/src/witness/common/cf_parameters.rs
```diff
@@ -21,19 +25,27 @@ pub type VersionedCcmCfParameters = VersionedCfParameters<CcmAdditionalData>;
...
 mod tests {
 	use super::*;
-	use cf_chains::MAX_CCM_ADDITIONAL_DATA_LENGTH;
+	use cf_chains::{ChannelRefundParameters, ForeignChainAddress, MAX_CCM_ADDITIONAL_DATA_LENGTH};
 
 	const MAX_VAULT_SWAP_PARAMETERS_LENGTH: u32 = 1_000;
 	const MAX_CF_PARAM_LENGTH: u32 =
 		MAX_CCM_ADDITIONAL_DATA_LENGTH + MAX_VAULT_SWAP_PARAMETERS_LENGTH;
 
+	const REFERENCE_EXPECTED_ENCODED: &[u8] = &[
```
msgmaxim, line 42: Maybe subjective, but I kind of like the way I broke down individual fields in `check_utxo_encoding` (https://github.com/chainflip-io/chainflip-backend/blob/main/state-chain/chains/src/btc/vault_swap_encoding.rs#L144) so we could see which bytes correspond to which field. Probably more important for BTC though where we were trying to compress the encoding.

engine/src/witness/common/cf_parameters.rs
```diff
@@ -44,4 +56,41 @@ mod tests {
...
+		let ccm_cf_parameters = CfParameters {
+			ccm_additional_data: CcmAdditionalData::default(),
+			vault_swap_parameters,
+		};
+
+		encoded = VersionedCcmCfParameters::V0(ccm_cf_parameters).encode();
+
+		// Extra byte for the empty ccm metadata
+		let mut expected_encoded = vec![0];
+		expected_encoded.extend_from_slice(REFERENCE_EXPECTED_ENCODED);
```
msgmaxim, line 92: Nit: I believe you can just convert to Vec with `Vec::from` (or alternatively compare slices)

msgmaxim (approved):
Left some suggestions, but looks good, and happy to merge this either way.

engine/src/witness/common/cf_parameters.rs
msgmaxim: Or you could just use a single byte (`u8`) like we did for BTC 🙂

engine/src/witness/arb.rs
albert-llimos: Done

PR #5409, opened 2024-11-11
Refactor: vault_swap should take TransactionInId

Janislav:
# Pull Request

Closes: PRO-1760

## Checklist

## Summary

- Replaces the TransactionHash with TransactionInId
- Did all the necessary refactor to handle the convert between the usage of generic and none-generic pallets
- Fixed all tests and benchmarks
- Change the engine where necessary

#### Non-Breaking changes

This is a breaking change, but I don't think we touched anything that is live yet. That means we should check with the frontends. The following things have chanded:

- The vault_swap_request extrinsic in the ingress egress pallet -> We are changed the name of tx_hash to tx_id
- The SwapRequested event in the swapping pallet -> tx_hash is now tx_id and the SwapOrigin is SwapOrigin<AnyChainType> now

In terms of migration, I think we are good. I didn't see any usage SwapOrigin in or former tx_hash in storage, which means there is no need for a migration.


albert-llimos:
@Janislav I know this is a draft but I've seen that you've used `SolHash` for Solana, which seems to be the obvious option. However, due to witnessing shortcomings, we will need to use `(SolAddress, u64)` for Solana as AccountAddress and slot number. I wrote it in the Linear comments but I can see how it was not too clear.
https://linear.app/chainflip/issue/PRO-1760/vault-swaps-should-take-transactioninid-instead-of-transactionhash#comment-26f67a54
Let me know if anything is unclear.

Janislav:
Mhhh... We will and can use whatever we decide the TransactionInId for Solana should be, but I think this should be not part of this PR, right? I assume changing this has also some other side effects and deserves it's own PR/Ticket I guess.

albert-llimos:
> Mhhh... We will and can use whatever we decide the TransactionInId for Solana should be, but I think this should be not part of this PR, right? I assume changing this has also some other side effects and deserves it's own PR/Ticket I guess.

I was thinking that you can already declare it as the correct type for Solana, as the SolHash is not what we want. The work on the engine to report the correct value for that will of course be in another PR.

bouncer/shared/utils.ts
```diff
@@ -499,7 +499,9 @@ export async function observeSwapRequested(
           (typeof id === 'number' &&
             'DepositChannel' in data.origin &&
             Number(data.origin.DepositChannel.channelId.replaceAll(',', '')) === id) ||
-          (typeof id === 'string' && 'Vault' in data.origin && data.origin.Vault.txHash === id);
+          (typeof id === 'string' &&
+            'Vault' in data.origin &&
+            data.origin.Vault.txHash.ByteHash === id);
```
Janislav, line 504: FYI: For now, it's safe to assume that we always have to deal with EVM or BTC, so it will be always a H256 hash string. If this should change, we have to handle this somehow.

albert-llimos: We have the Solana Vault swaps PR in the pipeline that will definitely break that. I'll make sure to update that after this is complete. Thanks for pointing that out.

dandanlen (commented):
I don't think we need a generic SwapOrigin: The only reason we ever construct a `SwapOrigin<TxInId>` is to convert it to a `SwapOrigin<TxInIdAnyChain>`. Instead of building what we don't want and then converting it to what we need, we could just build what we need. We just need a way to wrap the chain-specific id to the the 'any chain' enum.

PR #5410, opened 2024-11-12
feat: Btc vault swap bouncer test can open a private channel

j4m1ef0rd:
# Pull Request

Closes: PRO-1752

## Checklist

## Summary

- Added RPC's to the broker to open and close private btc channels.
- The Btc vault swap now opens a private channel
- Added the btc vault swap bouncer test to the concurrent tests

### DIFF
--- a/api/lib/src/lib.rs
+++ b/api/lib/src/lib.rs
@@ -477,10 +477,10 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 
 	async fn open_private_btc_channel(&self) -> Result<ChannelId> {
 		let (_, events, ..) = self
-			.submit_signed_extrinsic(RuntimeCall::from(
+			.submit_signed_extrinsic_with_dry_run(RuntimeCall::from(
 				pallet_cf_swapping::Call::open_private_btc_channel {},
 			))
-			.await
+			.await?
 			.until_in_block()
 			.await?;
 
@@ -502,10 +502,10 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 
 	async fn close_private_btc_channel(&self) -> Result<ChannelId> {
 		let (_, events, ..) = self
-			.submit_signed_extrinsic(RuntimeCall::from(
+			.submit_signed_extrinsic_with_dry_run(RuntimeCall::from(
 				pallet_cf_swapping::Call::close_private_btc_channel {},
 			))
-			.await
+			.await?
 			.until_in_block()
 			.await?;
 
--- a/api/lib/src/lib.rs
+++ b/api/lib/src/lib.rs
@@ -474,6 +474,56 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 		self.simple_submission_with_dry_run(pallet_cf_swapping::Call::deregister_as_broker {})
 			.await
 	}
+
+	async fn open_private_btc_channel(&self) -> Result<ChannelId> {
+		let (_, events, ..) = self
+			.submit_signed_extrinsic(RuntimeCall::from(
+				pallet_cf_swapping::Call::open_private_btc_channel {},
+			))
+			.await
+			.until_in_block()
+			.await?;
+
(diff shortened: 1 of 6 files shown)


api/lib/src/lib.rs
```diff
@@ -474,6 +474,56 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
...
+
+		if let Some(state_chain_runtime::RuntimeEvent::Swapping(
+			pallet_cf_swapping::Event::PrivateBrokerChannelOpened { channel_id, .. },
+		)) = events.iter().find(|event| {
+			matches!(
+				event,
+				state_chain_runtime::RuntimeEvent::Swapping(
+					pallet_cf_swapping::Event::PrivateBrokerChannelOpened { .. }
+				)
+			)
```
kylezs, line 495: we use this pattern a lot here, seems like it should be a util

bouncer/tests/btc_vault_swap.ts
```diff
@@ -112,8 +114,25 @@ async function testVaultSwap(depositAmountBtc: number, brokerUri: string, destin
...
+  // TODO: Use chainflip SDK instead so we can support any broker uri
+  assert.strictEqual(brokerUri, '//BROKER_1', 'Support for other brokers is not implemented');
+
+  // TODO: use chainflip SDK to check if the channel is already open
+  try {
+    await jsonRpc('broker_open_private_btc_channel', [], 'http://127.0.0.1:10997');
+    testBtcVaultSwap.log('Private Btc channel opened');
+  } catch (error) {
+    // We expect this to fail if the channel already exists from a previous run
+    testBtcVaultSwap.debugLog('Failed to open private Btc channel', error);
```
kylezs, line 127: If this is the expectation when we fail then I think we should check that the channel already exists, or inspect the error so we know etc.
else if we fail because the request params are wrong, this function will still pass

api/lib/src/lib.rs
```diff
@@ -474,6 +474,56 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 		self.simple_submission_with_dry_run(pallet_cf_swapping::Call::deregister_as_broker {})
 			.await
 	}
+
+	async fn open_private_btc_channel(&self) -> Result<ChannelId> {
+		let (_, events, ..) = self
+			.submit_signed_extrinsic(RuntimeCall::from(
```
dandanlen, line 480: nit: we could use a dry run to avoid submitting the extrinsic if it won't succeed.

bouncer/tests/btc_vault_swap.ts
```diff
@@ -112,8 +114,25 @@ async function testVaultSwap(depositAmountBtc: number, brokerUri: string, destin
   testBtcVaultSwap.log(`Balance increased, Vault Swap Complete`);
 }
 
+async function openPrivateBtcChannel(brokerUri: string) {
+  // TODO: Use chainflip SDK instead so we can support any broker uri
+  assert.strictEqual(brokerUri, '//BROKER_1', 'Support for other brokers is not implemented');
+
+  // TODO: use chainflip SDK to check if the channel is already open
+  try {
+    await jsonRpc('broker_open_private_btc_channel', [], 'http://127.0.0.1:10997');
```
dandanlen, line 123: Does the jsonRpc util not have a constant (or similar) to avoid having to explicitly pass the api url/port??

dandanlen (approved):

bouncer/tests/btc_vault_swap.ts
j4m1ef0rd: yes, thats why i put a TODO there to check if its open once we use the chainflip SDK.

bouncer/tests/btc_vault_swap.ts
j4m1ef0rd: Didnt seem worth adding when the code will be replaced later by the chainflip SDK.

api/lib/src/lib.rs
j4m1ef0rd: I can't see a good way to make it a util. Same problem we hit in the LP api.

msgmaxim (approved):

PR #5412, opened 2024-11-13
refactor: deduplicate event finding logic

kylezs:
# Pull Request

## Checklist

## Summary

Creating the util to reduce the extra duplication that was introduced in #5410

### DIFF
--- a/api/lib/src/lib.rs
+++ b/api/lib/src/lib.rs
@@ -346,13 +346,13 @@ impl fmt::Display for WithdrawFeesDetail {
 }
 
 macro_rules! extract_event {
-    ($events:expr, $event_variant:path, $pattern:tt, $result:expr) => {
-        if let Some(state_chain_runtime::RuntimeEvent::Swapping($event_variant $pattern)) = $events.iter().find(|event| {
-            matches!(event, state_chain_runtime::RuntimeEvent::Swapping($event_variant { .. }))
+    ($events:expr, $pallet_event:path, $event_variant:path, $pattern:tt, $result:expr) => {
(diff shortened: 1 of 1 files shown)


api/lib/src/lib.rs
```diff
@@ -345,6 +345,18 @@ impl fmt::Display for WithdrawFeesDetail {
 	}
 }
 
+macro_rules! extract_event {
+    ($events:expr, $event_variant:path, $pattern:tt, $result:expr) => {
+        if let Some(state_chain_runtime::RuntimeEvent::Swapping($event_variant $pattern)) = $events.iter().find(|event| {
+            matches!(event, state_chain_runtime::RuntimeEvent::Swapping($event_variant { .. }))
```
dandanlen, line 351: We can probably make this work for any pallet with a bit of into/try_into gymnastics:

```suggestion
        if let Some($event_variant $pattern) = $events.iter()
        	.filter_map(|runtime_event| runtime_event.try_into().ok())
        	.find(|pallet_event| { // <-- might need explicit type annotation
            	matches!(pallet_event, $event_variant { .. })
```

Might have to split the pallet event type and variant.

dandanlen (approved):

api/lib/src/lib.rs
```diff
@@ -345,6 +345,18 @@ impl fmt::Display for WithdrawFeesDetail {
 	}
 }
 
+macro_rules! extract_event {
+    ($events:expr, $pallet_event:path, $event_variant:path, $pattern:tt, $result:expr) => {
```
dandanlen, line 349:
```suggestion
    ($events:expr, $runtime_event_variant:path, $pallet_event_variant:path, $pattern:tt, $result:expr) => {
```

PR #5413, opened 2024-11-14
feat: btc vault swap with affiliates

j4m1ef0rd:
# Pull Request

Closes: PRO-1815

## Checklist

## Summary

- Completed the encoding of the affiliates in the `cf_get_vault_swap_details` rpc.
- Added a RPC to the broker api to register an affiliate.
- Added an affiliate to the bouncer btc vault swap test.


dandanlen (commented):
This is a good first step but I would like us to think about how brokers will interact with this. 

In particular, as it stands, when they call `register_affiliate`, they need to know which ids are available to be assigned, otherwise they can accidentally overwrite existing mappings.

The idea with requiring the id to be specified explicitly in the extrinsic was that (a) it allows ids to be replaced and (b) ids can be assigned 'optimistically' ie. you can author a payload with some affiliate id that you haven't registered yet (or that is pending finalisation).

However I expect the most common usage to simply be "assign a new id, I don't care what it is", so we should default to that.

I feel like we could use something like this for the api:

```rust
async fn register_affiliate(
	&self,
	affiliate_id: AccountId32,
	// if None, then we choose the lowest unassigned id.
	// if Some, then we overwrite any already existing mappings.
	short_id: Option<AffiliateShortId>,
// Return the mapping even if it wasn't changed
) -> RpcResult<AffiliateMapping>; 
// Alternatively, always return the full list of affiliate mappings 
) -> RpcResult<Vec<AffiliateMapping>>; 

struct AffiliateMapping {
	id: AccountId,
	short_id: AffiliateShortId,
}
```

An endpoint for querying a list of mappings would be useful too.

bouncer/tests/btc_vault_swap.ts
```diff
@@ -12,30 +12,51 @@ import {
...
   testBtcVaultSwap.debugLog('Broker:', broker.address);
   testBtcVaultSwap.debugLog(`Btc endpoint is set to`, BTC_ENDPOINT);
 
+  const affiliates: Beneficiary[] = [];
+  if (affiliateUri1) {
+    const affiliate1 = createStateChainKeypair(affiliateUri1);
+    affiliates.push({ account: affiliate1.address, bps: commissionBps });
+  }
+  if (affiliateUri2) {
+    const affiliate1 = createStateChainKeypair(affiliateUri2);
```
msgmaxim, line 56:
```suggestion
    const affiliate2 = createStateChainKeypair(affiliateUri2);
```

Alghough maybe instead of passing each affiliate as an individual parameter, you could pass an array instead?

bouncer/tests/btc_vault_swap.ts
```diff
@@ -12,30 +12,51 @@ import {
...
+  bps: number;
+}
+
 async function buildAndSendBtcVaultSwap(
   depositAmountBtc: number,
   brokerUri: string,
   destinationAsset: Asset,
   destinationAddress: string,
   refundAddress: string,
+  affiliateUri1: string | null = null,
```
msgmaxim, line 41: Maybe just pass the addresses directly here and avoid having to call `createStateChainKeypair` both here and at the caller's side? It is not like these accounts need to sign anything.

msgmaxim (approved):

state-chain/runtime/src/lib.rs
```diff
@@ -2161,8 +2160,14 @@ impl_runtime_apis! {
...
-							// TODO: lookup affiliate mapping to convert affiliate ids and use them here
-							affiliates: Default::default(),
+							affiliates: affiliate_fees.into_iter().map(|beneficiary|
+							{
+								Result::<AffiliateAndFee, DispatchErrorWithMessage>::Ok(AffiliateAndFee{
+									affiliate: Swapping::get_short_id(&broker_id, &beneficiary.account).ok_or(pallet_cf_swapping::Error::<Runtime>::AffiliateNotRegistered)?,
+									fee: beneficiary.bps.try_into().map_err(|_| pallet_cf_swapping::Error::<Runtime>::AffiliateFeeTooHigh)?
+								})
+							},
+							).collect::<Result<Vec<AffiliateAndFee>,_>>()?.try_into().map_err(|_| pallet_cf_swapping::Error::<Runtime>::TooManyAffiliates)?,
```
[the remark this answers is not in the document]
j4m1ef0rd: I am using auto formatting. I have to copy the code to another location to format it, but its correct.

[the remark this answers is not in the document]
dandanlen: I wasn't suggesting it was incorrect, just that it's hard to read. rustfmt can't always format code inside macros / closures, so manually adding a few newlines here and there can help readability.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -323,10 +326,28 @@ impl RpcServer for RpcServerImpl {
...
+		// Check that the affiliate was registered and get its short id
+		let affiliates = self.api.query_api().get_affiliates(None, None).await?;
+		if let Some((assigned_short_id, _)) = affiliates.iter().find(|(_, id)| *id == affiliate_id)
+		{
+			if let Some(given_short_id) = short_id {
+				if given_short_id != *assigned_short_id {
+					return Err(anyhow!("Failed to update affiliate registration").into());
+				}
+			}
+			return Ok(*assigned_short_id).map_err(BrokerApiError::Other);
```
dandanlen, line 343: Why do it like this instead of using the register_affiliate extrinsic return value / events?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1141,17 +1143,41 @@ pub mod pallet {
...
-		/// under `short_id` if already taken by another affiliate.
+		/// under `short_id` if already taken by another affiliate. Leave `short_id` empty to auto
+		/// assign one.
 		#[pallet::call_index(14)]
 		#[pallet::weight(T::WeightInfo::register_affiliate())]
 		pub fn register_affiliate(
 			origin: OriginFor<T>,
-			short_id: AffiliateShortId,
 			affiliate_id: T::AccountId,
+			short_id: Option<AffiliateShortId>,
```
dandanlen, line 1153: I don't think we should change the extrinsic. 

My suggestion was to do the checks before calling the extrinsic.

dandanlen (commented):
Sorry, maybe the original comment wasn't very clear. 

I think the extrinsic definition was fine as it was. 

We should make the ShortId optional on the outer interfaces (ie. the broker api) and if not provided on the broker api, we choose an id and then register that id. 
If it *is* provided, we just pass it through to the extrinsic.

PR #5416, opened 2024-11-15
Feature: bond brokers for private channels

Janislav:
# Pull Request

Closes: PRO-1761

## Checklist

## Summary

- Bond a broker by a 100 FLIP if he wants to open a private broker channel
- Added an extrinsic to update the required bond
- Extended the Flip pallet to bond an account under the condition of having enough liquid balance in his account
- Added a benchmark and tests
- Increased the amount of FLIP for BROKER accounts in the test net to have enough funds in the bouncer tests

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -143,8 +143,6 @@ pub mod pallet {
 		InsufficientReserves,
 		/// No pending redemption for this ID.
 		NoPendingRedemptionForThisID,
-		/// Account does not exist.
-		AccountDoesNotExist,
 	}
 
 	#[pallet::call]
@@ -447,29 +445,6 @@ impl<T: Config> Bonding for Bonder<T> {
 		Account::<T>::mutate_exists(authority, |maybe_account| {
 			if let Some(account) = maybe_account.as_mut() {
 				account.bond = bond
-			}
-		})
-	}
-
-	fn try_bond(authority: &Self::ValidatorId, bond: Self::Amount) -> DispatchResult {
-		Account::<T>::mutate_exists(authority, |maybe_account| {
-			if let Some(account) = maybe_account.as_mut() {
-				if account.balance >= bond {
-					account.bond = bond;
-					Ok(())
-				} else {
-					Err(Error::<T>::InsufficientLiquidity.into())
-				}
-			} else {
-				Err(Error::<T>::AccountDoesNotExist.into())
-			}
-		})
-	}
-
-	fn unbond(authority: &Self::ValidatorId) {
-		Account::<T>::mutate_exists(authority, |maybe_account| {
-			if let Some(account) = maybe_account.as_mut() {
-				account.bond = 0u128.into()
 			}
 		})
 	}
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -448,6 +448,29 @@ impl<T: Config> Bonding for Bonder<T> {
 			}
 		})
 	}
+
+	fn try_bond(authority: &Self::ValidatorId, bond: Self::Amount) -> bool {
+		Account::<T>::mutate_exists(authority, |maybe_account| {
+			if let Some(account) = maybe_account.as_mut() {
+				if account.balance >= bond {
+					account.bond = bond;
+					true
+				} else {
+					false
+				}
+			} else {
+				false
+			}
+		})
+	}
+
+	fn kill_bond(authority: &Self::ValidatorId) {
+		Account::<T>::mutate_exists(authority, |maybe_account| {
+			if let Some(account) = maybe_account.as_mut() {
+				account.bond = 0u128.into()
+			}
+		})
+	}
 }
 
 pub struct FlipIssuance<T>(PhantomData<T>);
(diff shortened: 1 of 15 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1162,6 +1184,15 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Updates the broker's bond to the new amount.
+		#[pallet::call_index(15)]
+		#[pallet::weight(T::WeightInfo::update_broker_bond())]
+		pub fn update_broker_bond(origin: OriginFor<T>, new_bond: T::Amount) -> DispatchResult {
```
dandanlen, line 1189: We can put this in `update_pallet_config` instead of creating a new extrinsic.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -731,6 +748,8 @@ pub mod pallet {
 		PrivateChannelExistsForBroker,
 		/// The Broker does not have an open private channel.
 		NoPrivateChannelExistsForBroker,
+		/// The Bonder does not have enough Funds to cover the bond.
+		InsufficientFunds,
```
dandanlen: This is unused. (Although i think we can use it instead of defining a new error in the flip pallet)

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -448,6 +450,29 @@ impl<T: Config> Bonding for Bonder<T> {
...
+		})
+	}
+
+	fn unbond(authority: &Self::ValidatorId) {
+		Account::<T>::mutate_exists(authority, |maybe_account| {
+			if let Some(account) = maybe_account.as_mut() {
+				account.bond = 0u128.into()
+			}
+		})
+	}
```
dandanlen, line 475: I don't think we need to add two new methods here, we can just use existing methods. (see my other comments)

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1114,11 +1133,12 @@ pub mod pallet {
 				Error::<T>::PrivateChannelExistsForBroker
 			);
 
-			// TODO: burn fee for opening a channel?
 			let channel_id = T::ChannelIdAllocator::allocate_private_channel_id()?;
 
 			BrokerPrivateBtcChannels::<T>::insert(broker_id.clone(), channel_id);
 
+			T::Bonder::try_bond(&broker_id, BrokerBond::<T>::get())?;
+
```
dandanlen, line 1139: Adding a new trait method seems unnecessary:

```rust
// use cf_traits::AccountInfo, can check this at the top:
ensure!(<T as Chainflip>::FundingInfo::balance() > x);
// Once other checks pass:
T::Bonder::update_bond(x);
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1133,6 +1153,8 @@ pub mod pallet {
 				return Err(Error::<T>::NoPrivateChannelExistsForBroker.into())
 			};
 
+			T::Bonder::unbond(&broker_id);
```
dandanlen, line 1154: Similary, this is just: `T::Bonder::update_bond(&broker_id, 0);`

state-chain/pallets/cf-validator/src/mock.rs
```diff
@@ -112,25 +114,6 @@ thread_local! {
 	pub static AUTHORITY_BONDS: RefCell<HashMap<ValidatorId, Amount>> = RefCell::new(HashMap::default());
 }
 
-pub struct MockBonder;
```
dandanlen, line 115: 👌 thanks for moving this.

state-chain/traits/src/mocks/bonding.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::{cell::RefCell, collections::HashMap};
+
+use crate::DispatchResult;
+
+pub type Amount = u128;
+pub type ValidatorId = u64;
+
+thread_local! {
+	pub static AUTHORITY_BONDS: RefCell<HashMap<ValidatorId, Amount>> = RefCell::new(HashMap::default());
+}
```
dandanlen, line 11: Prefer to use storage otherwise you might see some strange side effects (for example, thread_locals are not transactional so they don't obey `assert_noop`).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1133,6 +1151,8 @@ pub mod pallet {
...
```
Janislav, line 1154: I think it's worth taking a deeper think what it means to do this here. So far I don't see there should be an edge case since this extrinsic is only callable by a broker. Still, we should care about any edge cases here in which an account could benefit from it by changing account roles between broker and validator.

dandanlen: It shouldn't be possible to change account roles if the account is still bonded.

dandanlen: Currently this is ensured because the only way to be bonded is to open a private channel, and the only way to unbond is by closing it. And we check for closure in `deregister_as_broker`.

dandanlen (approved):

PR #5430, opened 2024-11-19
chore: imporove `swap_rate_v3` rpc

syan095:
# Pull Request

Closes: PRO-1830
Closes: PRO-1370

## Checklist

## Summary

Improved the swap_rate_v3 RPC call and runtime API `cf_pool_simulate_swap`

* Added boost_fee (taken first before all other fees)
* Added swap_input (input - ingress_fee - boost_fee)
* Intermediary returns None/Null if output == stable asset


dandanlen (commented):
I would have rather we kept these two issues separate. If not two PRs then at least two commits. 

I don't think this addresses PRO-1370 as the issue is intended.

I think the idea is that we should check *how much* boost liquidity is available and whether there is enough to boost the swap, and at what price. However it's not clear to me how useful this is or whether / how the producy would make use of it.

I'll extract the code related to (PRO-1830) and will move it into the release branch.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1421,6 +1426,7 @@ where
 		to_asset: Asset,
 		amount: U256,
 		broker_commission: BasisPoints,
+		boost_fee: BasisPoints,
```
niklasnatter, line 1429: this will allow the caller to pass the expected boost fee in basis points, right?

niklasnatter: for us it would be awesome to just pass a boolean if the quote should be boosted, and the rpc would estimate the required fee based on the current boost pool liquidity

dandanlen: My feeling is that we don't gain much by adding this (we can take the discussion to Linear, i've already written plenty there). We already have a boost pool depth rpc that can be used to estimate this. What do we gain by adding yet another way to get this information? 

Btw. What if boost == true and there isn't enough liquidity? Return an error?

niklasnatter: what we would gain is that we dont need to do the `cf_boost_pools_depth` before we can request a quote. but its not a strong opinion, we already have that implemented anyway, so its more about what might be useful for somebody else

> Btw. What if boost == true and there isn't enough liquidity? Return an error?

yes, i think would have expected a similar error to what happens if there is not enough liquidity to swap the given amount

dandanlen: Do we really need to know the boost pools depth before getting a quote? At most this will change the total input amount by 0.1%, which is unlikely to impact the swap price in a any noticeable way.

To me it seems much more flexible to provide an rpc for price, and and rpc for available boost liquidity, and let people do with it what they want. We can't anticipate all the ways in which people might want to use this. For eaxmple what if we want to display both the price in boosted an unboosted scenarios?

syan095:
> I don't think this addresses [PRO-1370](https://linear.app/chainflip/issue/PRO-1370) as the issue is intended.

Ah, I missed all the conversation in that linear ticket - it was added as a last-minute thing and I should not have been so presumptuous. 
I will hold off on this until we can get spec locked down.

dandanlen:
I'll close this for now, we can come back to it when we have some agreement.

PR #5436, opened 2024-11-21
fix: bouncer btc client mutex for vault swaps

j4m1ef0rd:
# Pull Request

Closes: PRO-1832

## Checklist

## Summary

Using the btc client mutex around selecting inputs and sending transaction to avoid double spending error.

### DIFF
--- a/bouncer/shared/send_btc.ts
+++ b/bouncer/shared/send_btc.ts
@@ -21,6 +21,8 @@ export async function fundAndSendTransaction(
     const fundedTx = (await btcClient.fundRawTransaction(rawTx, {
       changeAddress,
       feeRate: feeRate ?? 0.00001,
+      lockUnspents: true,
+      changePosition: 2,
     })) as { hex: string };
     const signedTx = await btcClient.signRawTransactionWithWallet(fundedTx.hex);
     const txId = (await btcClient.sendRawTransaction(signedTx.hex)) as string | undefined;
--- a/bouncer/shared/send_btc.ts
+++ b/bouncer/shared/send_btc.ts
@@ -12,25 +12,33 @@ export const btcClient = new Client({
 });
 
 export async function selectInputs(amount: number) {
-  // List unspent UTXOs
-  const utxos = await btcClient.listUnspent();
+  return btcClientMutex.runExclusive(async () => {
+    // List unspent UTXOs
+    const utxos = await btcClient.listUnspent();
 
-  // Find a UTXO with enough funds
-  const utxo = utxos.find((u) => u.amount >= amount);
-  if (!utxo) throw new Error('Insufficient funds');
-  // TODO: be able to select more than one UTXO
+    // Find a UTXO with enough funds
+    const utxo = utxos.find((u) => u.amount >= amount);
+    if (!utxo) throw new Error('Insufficient funds');
+    // TODO: be able to select more than one UTXO
 
-  const change = utxo.amount - amount;
+    // Lock the selected UTXO to prevent it from being used in another transaction
+    await btcClient.lockUnspent(false, [{ txid: utxo.txid, vout: utxo.vout }]);
 
-  // Prepare the transaction inputs and outputs
-  const inputs = [
-    {
-      txid: utxo.txid,
-      vout: utxo.vout,
-    },
-  ];
+    const change = utxo.amount - amount;
 
-  return { inputs, change };
+    // Prepare the transaction inputs
+    const inputs = [
+      {
+        txid: utxo.txid,
+        vout: utxo.vout,
+      },
+    ];
+
+    return {
+      inputs,
+      change,
+    };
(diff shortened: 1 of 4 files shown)


bouncer/shared/send_btc.ts
```diff
@@ -12,25 +12,33 @@ export const btcClient = new Client({
...
+    // List unspent UTXOs
+    const utxos = await btcClient.listUnspent();
 
-  // Find a UTXO with enough funds
-  const utxo = utxos.find((u) => u.amount >= amount);
-  if (!utxo) throw new Error('Insufficient funds');
-  // TODO: be able to select more than one UTXO
+    // Find a UTXO with enough funds
+    const utxo = utxos.find((u) => u.amount >= amount);
+    if (!utxo) throw new Error('Insufficient funds');
```
kylezs, line 21: this introduces another issue right? If we have a single UTXO, one of the processes locks it, and then another process tries to send, and it gets an error because there are none there.

bouncer/shared/send_btc.ts
```diff
@@ -12,25 +12,33 @@ export const btcClient = new Client({
 });
 
 export async function selectInputs(amount: number) {
-  // List unspent UTXOs
-  const utxos = await btcClient.listUnspent();
+  return btcClientMutex.runExclusive(async () => {
+    // List unspent UTXOs
+    const utxos = await btcClient.listUnspent();
```
dandanlen, line 17: We don't need to break up the vault swap construction into so many steps (in btc_vault_swap). Putting a mutex around just the input selection doesn't solve our problem. 

We should create method on the btc client that takes the nulldata utxo, deposit address and amount as args, and then does all the steps with the mutex acquired.

```
fn build_vault_tx(nulldata, amount, address):
	acquire_mutex()
		get_inputs()
		create_raw_tx()
		sign_tx()
		broadcast_tx()
	release_mutex()
```

In fact maybe only the first two steps need to be in the mutex, once we build the tx I guess the btc node is smart enough to know we can use the output (??) but to be safe we migth as well put everything inside.

bouncer/shared/send_btc.ts
j4m1ef0rd: Seems to be plenty available on a localnet. I tested with a bunch sending funds at the exact same time. We can solve that later if we ever hit the error.

bouncer/shared/send_btc.ts
j4m1ef0rd: >  Putting a mutex around just the input selection doesn't solve our problem.

The mutex and the locking of the utxo should solve it. I don't understand.

> We should create method on the btc client

Ok, that's fine, but I don't see the need. Just so we don't have to lock the utxo?

dandanlen: It's what Kyle said: if we lock the utxos, then request the payload, then build and send the tx, there is a chance that some other process fails because there are none available. We can eliminate this race condition quite easily though: We don't need to lock the inputs at all, do we? We can request the payload, then build and send the tx with whatever utxos are available at that point. 

I don't see why we wouldn't just do this once and do it properly instead of kicking the can down the road until the next time all the PRs are blocked.

dandanlen:
I was playing around with this on Friday, and it looks like the `selectInputs` method is not necessary. 

Bitcoin nodes have a `fundRawTransaction` method that selects the inputs for us.

This would also solve the (still remaining) issue that we can only select a single utxo. I'll see if it works on localnet and if yes, will push it.

dandanlen (approved):

bouncer/tests/lp_api_test.ts
```diff
@@ -21,7 +21,7 @@ import { getChainflipApi, observeEvent } from '../shared/utils/substrate';
 import { ExecutableTest } from '../shared/executable_test';
 
 /* eslint-disable @typescript-eslint/no-use-before-define */
-export const testLpApi = new ExecutableTest('LP-API', main, 200);
+export const testLpApi = new ExecutableTest('LP-API', main, 1000);
```
dandanlen, line 24: I also increased the timeout of this test - it contains multiple other tests with a 500 second timeout. We don't want the timeout of the parent to be stricter.

j4m1ef0rd: Actually the 200 sec timeout was correct. I just set the child tests timeouts to 500 so they would never get hit and could be ignored. Was easier than setting correct timeouts for each.

dandanlen: Surely it's better to allow the child timeouts to be hit? This way we get more granular feedback in terms of what went wrong. If we want 200, then let's set all the child timeouts to 200, and the parent to 500 or something.

dandanlen: Looks like you updated it - much better, thanks.

kylezs (approved):

PR #5437, opened 2024-11-21
Feat: Boost Support for Vault Swaps

msgmaxim:
# Pull Request

Closes: PRO-1753

## Summary

- Added additional fields to `vault_swap_request` (height, deposit_address, channel_id). This is needed for DepositBoosted/DepositFinalised events. I also made it take a vec of deposits (represented as `VaultDepositWitness`) instead to be consistent with "regular" deposits and as a better (compared to box-ing individual fields) solution to the "max call size" problem that occurs when we have too many paramters not on a heap. Right now engines still vote on individual deposits (i.e. vec of size 1), but we can change this in the future and submit batches of deposits like we do for regular deposits (@dandanlen mentions that this might require some discussion, so not done here).

- Added `BoostedVaultTxs` field to track the status of vault deposits (regular deposits use `DepositChannelLookup` for this)

- Added `DepositOrigin` which mimicks `SwapOrigin` but excludes `Internal` variant and makes more sense in contexts where we are not necessarily dealing with swaps, but deposits in general. I considered merging them into one to simplify, but seemed a bit cleaner to have explicit types (but happy to reconsider).

- Factored out `process_prewitness_deposit_inner` and `process_full_witness_deposit_inner` so that the core logic is used regardless of the "origin". Deposit extrinsics now just assemble the parameters for the channel action, perform validation where needed, and read/update the boost status.

- `DepositBoosted/DepositFinalised` events now include `DepositOriginType` so the product knows whether `channel_id` should be used to lookup deposit channel details for example. @niklasnatter 

Edit: I might need to check unit-tests/bouncer tests to see why they fail, but otherwise this PR should be read for review.

### DIFF
--- a/engine/src/witness/btc/deposits.rs
+++ b/engine/src/witness/btc/deposits.rs
@@ -6,8 +6,12 @@ use itertools::Itertools;
 use pallet_cf_ingress_egress::{DepositChannelDetails, DepositWitness};
 use state_chain_runtime::BitcoinInstance;
 
-use super::super::common::chunked_chain_source::chunked_by_vault::{
-	builder::ChunkedByVaultBuilder, private_deposit_channels::BrokerPrivateChannels, ChunkedByVault,
+use super::{
+	super::common::chunked_chain_source::chunked_by_vault::{
+		builder::ChunkedByVaultBuilder, private_deposit_channels::BrokerPrivateChannels,
+		ChunkedByVault,
+	},
+	vault_swaps::BtcIngressEgressCall,
 };
 use crate::{
 	btc::rpc::VerboseTransaction,
@@ -71,6 +75,7 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 						private_channels.clone().into_iter().map(move |(broker_id, channel_id)| {
 							(
 								broker_id,
+								channel_id,
 								DepositAddress::new(
 									key,
 									channel_id.try_into().expect("BTC channel id must fit in u32"),
@@ -80,14 +85,23 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 					})
 				};
 
-				for (broker_id, vault_address) in vault_addresses {
+				for (broker_id, channel_id, vault_address) in vault_addresses {
 					for tx in &txs {
-						if let Some(call) = super::vault_swaps::try_extract_vault_swap_call(
+						if let Some(deposit) = super::vault_swaps::try_extract_vault_swap_witness(
 							tx,
 							&vault_address,
+							channel_id,
 							&broker_id,
 						) {
-							process_call(call.into(), epoch.index).await;
+							process_call(
+								BtcIngressEgressCall::vault_swap_request {
+									block_height: header.index,
+									deposits: vec![deposit],
+								}
+								.into(),
+								epoch.index,
+							)
+							.await;
 						}
 					}
 				}
--- a/engine/src/witness/btc/deposits.rs
+++ b/engine/src/witness/btc/deposits.rs
@@ -96,7 +96,7 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 							process_call(
 								BtcIngressEgressCall::vault_swap_request {
 									block_height: header.index,
-									deposits: vec![deposit],
+									deposit: Box::new(deposit),
 								}
 								.into(),
 								epoch.index,
--- a/engine/src/witness/eth.rs
+++ b/engine/src/witness/eth.rs
@@ -8,10 +8,13 @@ use cf_chains::{
 	evm::{DepositDetails, H256},
 	CcmDepositMetadata, Ethereum,
 };
-use cf_primitives::{chains::assets::eth::Asset as EthAsset, Asset, AssetAmount, EpochIndex};
+use cf_primitives::{
+	chains::assets::eth::Asset as EthAsset, Asset, AssetAmount, ChannelId, EpochIndex,
+};
 use cf_utilities::task_scope::Scope;
 use futures_core::Future;
 use itertools::Itertools;
+use pallet_cf_ingress_egress::VaultDepositWitness;
 use sp_core::H160;
 
 use crate::{
@@ -232,7 +235,10 @@ impl super::evm::vault::IngressCallBuilder for EthCallBuilder {
 	type Chain = Ethereum;
 
 	fn vault_swap_request(
+		block_height: u64,
 		source_asset: Asset,
+		deposit_address: cf_chains::eth::Address,
+		channel_id: ChannelId,
 		deposit_amount: AssetAmount,
 		destination_asset: Asset,
 		destination_address: EncodedAddress,
@@ -242,24 +248,30 @@ impl super::evm::vault::IngressCallBuilder for EthCallBuilder {
 	) -> state_chain_runtime::RuntimeCall {
 		state_chain_runtime::RuntimeCall::EthereumIngressEgress(
 			pallet_cf_ingress_egress::Call::vault_swap_request {
-				input_asset: source_asset.try_into().expect("invalid asset for chain"),
-				output_asset: destination_asset,
-				deposit_amount,
-				destination_address,
-				deposit_metadata,
-				tx_id,
-				deposit_details: Box::new(DepositDetails { tx_hashes: Some(vec![tx_id]) }),
-				broker_fee: vault_swap_parameters.broker_fee,
-				affiliate_fees: vault_swap_parameters
-					.affiliate_fees
-					.into_iter()
-					.map(Into::into)
-					.collect_vec()
-					.try_into()
-					.expect("runtime supports at least as many affiliates as we allow in cf_parameters encoding"),
-				boost_fee: vault_swap_parameters.boost_fee.into(),
-				dca_params: vault_swap_parameters.dca_params,
-				refund_params: Box::new(vault_swap_parameters.refund_params),
+				block_height,
+				deposits: vec![ VaultDepositWitness {
+					input_asset: source_asset.try_into().expect("invalid asset for chain"),
+					output_asset: destination_asset,
+					deposit_amount,
+					destination_address,
+					deposit_metadata,
+					tx_id,
+					deposit_details: DepositDetails { tx_hashes: Some(vec![tx_id]) },
+					broker_fee: vault_swap_parameters.broker_fee,
+					affiliate_fees: vault_swap_parameters
+						.affiliate_fees
+						.into_iter()
+						.map(Into::into)
+						.collect_vec()
+						.try_into()
+						.expect("runtime supports at least as many affiliates as we allow in cf_parameters encoding"),
+					boost_fee: vault_swap_parameters.boost_fee.into(),
+					dca_params: vault_swap_parameters.dca_params,
+					refund_params: vault_swap_parameters.refund_params,
+					channel_id,
+					deposit_address,
+				}
+				],
 			},
 		)
 	}
--- a/engine/src/witness/eth.rs
+++ b/engine/src/witness/eth.rs
@@ -257,7 +257,7 @@ impl super::evm::vault::IngressCallBuilder for EthCallBuilder {
 		state_chain_runtime::RuntimeCall::EthereumIngressEgress(
 			pallet_cf_ingress_egress::Call::vault_swap_request {
 				block_height,
-				deposits: vec![deposit],
+				deposit: Box::new(deposit),
 			},
 		)
 	}
--- a/engine/src/witness/eth.rs
+++ b/engine/src/witness/eth.rs
@@ -31,7 +31,10 @@ use crate::{
 	},
 };
 
-use super::{common::epoch_source::EpochSourceBuilder, evm::source::EvmSource};
+use super::{
+	common::epoch_source::EpochSourceBuilder,
+	evm::{source::EvmSource, vault::vault_deposit_witness},
+};
 use crate::witness::common::chain_source::extension::ChainSourceExt;
 
 use anyhow::{Context, Result};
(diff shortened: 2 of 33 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2162,116 +2440,96 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-			source_asset.into(),
-			deposit_amount.into(),
-			destination_asset,
-			request_type,
-			broker_fees,
-			Some(refund_params),
-			dca_params,
-			swap_origin,
-		);
+		let (action, source_address) = if let Some(deposit_metadata) = deposit_metadata {
```
msgmaxim, line 2491: I think a cleaner solution for extracting/passing `source_address` might be to create a different struct e.g. `DepositAction` which will include `CcmDepositMetadata` instead of `CcmChannelMetadata` in `ChannelAction` (and/or making ChannelAction generic over metadata, though this would make the code less readable), but I chose to go with the solution that requires fewer code for now.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1301,37 +1403,19 @@ pub mod pallet {
...
-			deposit_metadata: Option<CcmDepositMetadata>,
-			tx_id: TransactionInIdFor<T, I>,
-			deposit_details: Box<<T::TargetChain as Chain>::DepositDetails>,
-			broker_fee: Beneficiary<T::AccountId>,
-			affiliate_fees: Affiliates<AffiliateShortId>,
-			refund_params: Box<ChannelRefundParameters>,
-			dca_params: Option<DcaParameters>,
-			boost_fee: BasisPoints,
+			block_height: TargetChainBlockNumber<T, I>,
+			deposits: Vec<VaultDepositWitness<T, I>>,
```
albert-llimos, line 1407: I imagine the benefit is to be able to submit multiple Vault swaps in the same extrinsic, so potentially decreasing the numbers of extrinsics. That improvement is probably irrelevant for any chain besides BTC as we most likely won't have many if this instances.
On the other hand, I do see some room for this being problematic. My understanding is that to have consensus well need engines to submit the exact same extrinsic. In Solana where we can only query the tip of the chain, submitting all Vault swaps as a vector will be problematic:
- Start at block 0
- At block 10 a Vault swap is executed
- At block 15 a Vault Swap is executed
- Some engines query between 10 and 11 submitting the extrinsic with one Vault Swap and then later on another one with the second Vault swap.
- Other engines query at block 16 submitting the extrinsic with two Vault swaps.
Then we wouldn't reach consensus.
We can choose to have the engine handle that (e.g in Solana never submit a vector) but it could be prone to error in the future. I'm fine with keeping it like this, since there is some upside to this, but I can see the argument of aiming for a more safe approach. I just wanted to highlight it and also to check that my understanding is correct, especially since we'll also have to update this for Solana Vault swaps.

albert-llimos: Fwiw I think this is only problematic for Solana atm but I wouldn't be surprised if other fast chains we might integrate in the future would also fall into that category.

kylezs: For solana it's fine, the way the elections are implemented is that it destructures them to process each deposit independently, regardless of how they're submitted e.g.
validator 1 submits [1, 2, 3], validator 2 submits [1, 2] and validator 3 submits [2], then [1 ,2] go through since they both reached 2/3. 

For bitcoin we witness for a particular block so every engine should agree.

ramizhasan111: yes, thats correct.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1965,6 +1969,284 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+							prewitnessed_deposit_id,
+							channel_id,
+							deposit_details: deposit_details.clone(),
+							ingress_fee,
+							boost_fee: boost_fee_amount,
+							action,
+							origin_type: origin.into(),
+						});
+					} else {
+						// TODO: emit error?
```
msgmaxim, line 2129: I looked into how this can fail, and it looks like it can only happen when `try_credit_account` fails, which itself only fails when balance overflows... Maybe we could just make `try_credit_account` infallible instead? We don't really have any place where this is handled by the caller in a meaningful way: we either unwrap, print error or propogate it to the point where we no longer know what to do with it and simply ignore (as has been the case here even prior to this PR). Maybe we could just make it infallible and either saturate and/or use `log_or_panic` internally? Would simplify a lot of code. @dandanlen

dandanlen: Yeah I agree - I think assuming we can't overflow is almost always a fair assumption. I think it was left as fallible because of uncertainty in the AMM implementation (eg. at the extreme ends of possible pools state / swap rates). So as a compromise I would suggest creating a separate infallible (saturating) method, and then use the fallible one when crediting AMM funds, and the saturating one when crediting funds from deposits, for example .

msgmaxim: OK, though I still don't quite understand the scenario where it would overflow. For AMM to return an amount that might overflow, it has to exist in the pool in the first place... What am I missing?

albert-llimos:
`channel_id` is not needed and actually makes no sense outside of Bitcoin, as the premise of Vault swaps is to not need a channel. Therefore passing a `channel_id` as a mandatory parameter to `vault_swap_request` and arbitrarily hardcoding that to zero for other chains seems misleading to me. It could even be just wrong for chains were channels are reused as we could currently be using that channel.
I'd suggest making that an Option.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2162,116 +2439,96 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				deposit_details,
+				source_address,
+				&broker,
+				boost_status,
+				channel_id,
+				action,
+				block_height,
+				deposit_origin,
+			) {
+			// This allows the channel to be boosted again:
```
kylezs, line 2529: incorrectly copied comment?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2096,59 +2374,58 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-					deposit_details,
-					ingress_fee: fees_withheld,
-					action: deposit_action,
-					channel_id,
-				});
+					source_address,
+					amount_after_fees,
+					origin.clone(),
+				) {
+					// TODO: this needs to include deposit type (vault/channel)
```
kylezs, line 2388: origin_type ? looks like a stale comment

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1965,6 +1969,284 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+					});
+				},
+				Err(err) => {
+					Self::deposit_event(Event::InsufficientBoostLiquidity {
+						prewitnessed_deposit_id,
+						asset,
+						amount_attempted: amount,
+						channel_id,
+					});
+					log::debug!(
```
kylezs, line 2145: I don't think we need a log if we have an event

kylezs (commented):
pretty minor comments overall 👌

engine/src/witness/btc/deposits.rs
```diff
@@ -80,14 +85,23 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
 							tx,
 							&vault_address,
+							channel_id,
 							&broker_id,
 						) {
-							process_call(call.into(), epoch.index).await;
+							process_call(
+								BtcIngressEgressCall::vault_swap_request {
+									block_height: header.index,
+									deposits: vec![deposit],
```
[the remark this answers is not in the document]
msgmaxim: Yeah, originally @dandanlen and I talked about this as something we could do, but would be outside of scope of this PR, but now that I have changed the extrinsic to take a Vec, I suppose we might as well batch them.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2013,17 +2293,22 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 
-		ScheduledEgressFetchOrTransfer::<T, I>::append(FetchOrTransfer::<T::TargetChain>::Fetch {
-			asset,
-			deposit_address: deposit_address.clone(),
-			deposit_fetch_id: None,
-			amount: deposit_amount,
-		});
-		Self::deposit_event(Event::<T, I>::DepositFetchesScheduled { channel_id, asset });
+		// Vault deposits don't need to be fetched:
+		if !matches!(origin, DepositOrigin::Vault { .. }) {
```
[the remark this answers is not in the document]
msgmaxim: I now changed to an exhaustive match so that we are forced to handle any new variant.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -140,6 +139,76 @@ pub enum DepositIgnoredReason {
...
+	// TODO: replace SwapOrigin with this?
+	#[derive(Clone)]
+	pub(super) enum DepositOrigin {
+		DepositChannel {
+			deposit_address: EncodedAddress,
+			channel_id: ChannelId,
+			deposit_block_height: u64,
+		},
+		Vault {
+			tx_id: TransactionInIdForAnyChain,
```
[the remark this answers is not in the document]
msgmaxim: Changed it to `TransactionInIdFor<T, I>` by moving the conversion to a later point (when `SwapOrigin` is constructed).

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1965,6 +1969,284 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+					destination_address: destination_address_internal,
+					broker_fees,
+					refund_params: Some(refund_params),
+					dca_params,
+				},
+				None,
+			)
+		};
+
+		let boost_status = BoostedVaultTxs::<T, I>::get(&tx_id).unwrap_or(BoostStatus::NotBoosted);
```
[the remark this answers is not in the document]
msgmaxim: > Also we use the full word Transaction/s

OK, renamed.

> we can pull this out a level higher, is duped between the prewitness and full witness

Not sure it is worth it tbh. I prefer keeping the signature of `process_vault_swap_request_full_witness` and `process_vault_swap_request_prewitness` simple where the processing of a deposit is done entirely within one of them, especially considering that we are talking about a single/trivial line of code.

engine/src/witness/eth.rs
```diff
@@ -232,7 +235,10 @@ impl super::evm::vault::IngressCallBuilder for EthCallBuilder {
 	type Chain = Ethereum;
```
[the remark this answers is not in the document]
msgmaxim: I created a macro for creating VaultDepositWitness from params (couldn't quite do it with generic parameters alone). There is a bit of boilerplate code that I'm sure could also be deduplicated with a macro, but I don't think it is worth it considering there is no logic there.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: updated the comment

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: removed

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: removed

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1965,6 +1969,284 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				Ok(address) => address,
+				Err(err) => {
+					log::warn!("Failed to process vault swap due to invalid destination address. Tx hash: {tx_id:?}. Error: {err:?}");
+					return;
+				},
+			};
+
+		let broker = broker_fee.account.clone();
+
+		let broker_fees = Self::assemble_broker_fees(broker_fee, affiliate_fees);
```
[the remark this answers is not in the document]
msgmaxim: I prefer to keep it the way it is (for the same reason as the other as in the other comment). I also wouldn't inline even if we only used it once since it makese sense to keep it separate conceptually and it makes the rest of the code cleaner.

> though wouldn't mind keeping the function and writing a test

Added a test that uses max affiliates

engine/src/witness/btc/deposits.rs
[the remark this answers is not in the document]
msgmaxim: Update: as mentioned on discord (https://discord.com/channels/775961728608895008/1311562365828857947/1311939330188644413), I changed the signature to take a single deposit rather than a Vec for now as doing otherwise was causing some weird bouncer behaviour.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1969,6 +1943,275 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				match opt.as_mut() {
+					// Transaction has been reported, mark it as pre-witnessed.
+					Some(status @ TransactionPrewitnessedStatus::Unseen) => {
+						*status = TransactionPrewitnessedStatus::Prewitnessed;
+						true
+					},
+					// Pre-witnessing twice is unlikely but possible. Either way we don't want
+					// to change the status and we don't want to allow boosting.
+					Some(TransactionPrewitnessedStatus::Prewitnessed) => true,
+					// Transaction has not been reported, mark it as boosted to prevent further
```
kylezs, line 2041: This comment is a little confusing to me, we're not really marking anything here? we're just continuing in the function (which might then return a Boosted status)

kylezs (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1153,7 +1153,10 @@ pub mod pallet {
 			block_height: TargetChainBlockNumber<T, I>,
 		) -> DispatchResult {
 			if T::EnsurePrewitnessed::ensure_origin(origin.clone()).is_ok() {
-				Self::add_prewitnessed_deposits(deposit_witnesses, block_height)?;
+				for deposit_witness in deposit_witnesses {
+					// TODO: emit event on error?
+					let _ = Self::process_channel_deposit_prewitness(deposit_witness, block_height);
```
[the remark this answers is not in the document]
msgmaxim: (added the test)

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: (wans't introduced here, but I removed the confusing/inaccurate part as discussed)

state-chain/chains/src/lib.rs
```diff
@@ -785,6 +791,8 @@ impl<Address> CcmDepositMetadataGeneric<Address> {
 
 		let principal_swap_amount = deposit_amount.saturating_sub(gas_budget);
 
+		// TODO: we already check ccm support when opening a channel (and we have to).
+		// If we can also check this in vault swaps, we should be able to remove this here.
```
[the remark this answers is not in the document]
msgmaxim: (has been discussed in the office)

PR #5440, opened 2024-11-21
feat: try-state hook for elections pallet

nmammeri:
# Pull Request

Closes: PRO-1726

## Checklist

## Summary

This PR adds try-state hook checks to the Elections pallet. For these hooks to be fired up, the try-runtime-cli must be called with --checks all (or at- least try-state). The ci and troubleshooting guide were updated accordingly.

This PR implemented try-state hook only for the Elections pallet, but it was agreed that another issue will be raised to handle try-state hooks in a more generic way and enable it for all other pallets.

### DIFF
--- a/state-chain/pallets/cf-elections/src/lib.rs
+++ b/state-chain/pallets/cf-elections/src/lib.rs
@@ -2159,7 +2159,7 @@ pub mod pallet {
 			let shared_ref_count_keys =
 				SharedDataReferenceCount::<T, I>::iter_keys().collect::<Vec<_>>();
 			for shared_data_key in SharedData::<T, I>::iter_keys() {
-				if shared_ref_count_keys.iter().find(|(h, _)| *h == shared_data_key).is_none() {
+				if !shared_ref_count_keys.iter().any(|(h, _)| *h == shared_data_key) {
 					Err(DispatchError::Other(
 						"All keys in SharedData should have an entry in SharedDataReferenceCount",
 					))?;
--- a/state-chain/pallets/cf-elections/src/lib.rs
+++ b/state-chain/pallets/cf-elections/src/lib.rs
@@ -1622,6 +1622,11 @@ pub mod pallet {
 			}
 		}
 
+		#[cfg(feature = "try-runtime")]
+		fn try_state(_n: BlockNumberFor<T>) -> Result<(), DispatchError> {
+			Self::do_try_state()
+		}
+
 		fn integrity_test() {
 			let properties_keys = ElectionProperties::<T, I>::iter_keys()
 				.map(|id| *id.unique_monotonic())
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -2127,4 +2132,63 @@ pub mod pallet {
...
+		pub fn do_try_state() -> Result<(), DispatchError> {
+			let properties_keys = ElectionProperties::<T, I>::iter_keys()
+				.map(|id| *id.unique_monotonic())
+				.collect::<BTreeSet<_>>();
+
+			if properties_keys != ElectionState::<T, I>::iter_keys().collect::<BTreeSet<_>>() {
+				Err(DispatchError::Other(
+					"All keys in ElectionProperties and ElectionState should match",
+				))?;
+			}
```
kylezs, line 2157:
```suggestion
			if properties_keys != ElectionState::<T, I>::iter_keys().collect::<BTreeSet<_>>() {
				Err(DispatchError::Other(
					"All keys in ElectionProperties and ElectionState should match",
				))?;
			}
```

The substrate-y way to do this is to use ensure!() - is also nicer because we don't need to wrap it in Err()

kylezs: same for all of these conditions

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -2127,4 +2132,63 @@ pub mod pallet {
...
+				if !properties_keys.contains(&election_id) {
+					Err(DispatchError::Other(
+						"BitmapComponents should have a corresponding entry in ElectionProperties",
+					))?;
+				}
+			}
+			for (election_id, _) in IndividualComponents::<T, I>::iter_keys() {
+				if !properties_keys.contains(&election_id) {
+					Err(DispatchError::Other(
+						"BitmapComponents should have a corresponding entry in ElectionProperties",
```
kylezs, line 2186:
```suggestion
						"IndividualComponents should have a corresponding entry in ElectionProperties",
```

state-chain/pallets/cf-elections/src/mock.rs
```diff
@@ -102,6 +102,9 @@ pub fn election_test_ext(test_setup: TestSetup) -> TestRunner<TestContext> {
 			}
 			MockEpochInfo::next_epoch(test_setup.all_authorities());
 
+			Pallet::<Test, _>::do_try_state()
```
kylezs, line 105: this will only run when we start up the test, would be good to have this run at least on the end of each test too

nmammeri: as discussed with Dan we will create a separate issue to properly handle this and make it available for all pallets

kylezs (approved):

PR #5441, opened 2024-11-21
feat: allow for providing params to ElectoralSystem test checks

kylezs:
# Pull Request

## Checklist

## Summary

Allows us to avoid duplicating the "hook_called" etc. I'm also going to need to this to, for example, check how many elections are open at any one time, and would rather not write a hook specific to the number of times I expect, as it will change throughout the test.

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/tests/liveness.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/tests/liveness.rs
@@ -1,9 +1,10 @@
 use sp_std::collections::btree_set::BTreeSet;
 
-use super::{mocks::*, register_checks};
+use super::mocks::*;
 use crate::{
 	electoral_system::{ConsensusVote, ConsensusVotes},
 	electoral_systems::liveness::*,
+	register_checks,
 };
 
 pub type ChainBlockHash = u64;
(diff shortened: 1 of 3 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/tests/liveness.rs
```diff
@@ -37,18 +38,8 @@ register_checks! {
...
-			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), 2, "Hook should have been called twice so far!");
-		},
-		hook_not_called(_pre, _post) {
-			assert_eq!(
-				HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()),
-				0,
-				"Hook should not have been called!"
-			);
+		hook_called_n_times(_pre, _post, n: u8) {
+			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), n, "Hook should have been called n so far!");
```
dandanlen, line 42:
```suggestion
			assert_eq!(HOOK_CALLED_COUNT.with(|hook_called| hook_called.get()), n, "Hook should have been called {} times so far!", n);
```

Does this work?

dandanlen (approved):

state-chain/pallets/cf-elections/src/electoral_systems/tests/liveness.rs
kylezs: yep

PR #5444, opened 2024-11-22
feat: check all restricted balances are above minimum when redeeming

nmammeri:
# Pull Request

Closes: PRO-1821

## Checklist

## Summary

* Added a check that makes sure we are not left with the situation where a restricted address is below MINIMUM_FUNDING.
* This broke some other tests which were updated to take the new checks into consideration

### DIFF
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -354,13 +354,6 @@ pub mod pallet {
 
 			let mut restricted_balances = RestrictedBalances::<T>::get(&account_id);
 
-			// ensure that all restricted balances are above MinimumFunding, to avoid
-			// situations where it is not possible to redeem max
-			ensure!(
-				restricted_balances.values().all(|v| *v >= MinimumFunding::<T>::get()),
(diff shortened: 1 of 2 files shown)


state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -351,6 +354,13 @@ pub mod pallet {
 
 			let mut restricted_balances = RestrictedBalances::<T>::get(&account_id);
 
+			// ensure that all restricted balances are above MinimumFunding, to avoid
+			// situations where it is not possible to redeem max
+			ensure!(
+				restricted_balances.values().all(|v| *v >= MinimumFunding::<T>::get()),
+				Error::<T>::RestrictedBalanceBelowMinimumFunding
+			);
```
dandanlen, line 362: I think we need to do the check *after* we have deducted the redemption amount. 

The problem was that there was dust remaining in the account after the withdrawal.

nmammeri: agree, plz see updated code

dandanlen (approved):
Good work, thanks.

dandanlen (approved):

PR #5445, opened 2024-11-25
test: Solana's delta-based-ingress electoral System

syan095:
# Pull Request

Closes: PRO-1581

## Checklist

## Summary

Added unit tests for the delta based ingress electoral system.
Improved the mock system for the tests.
Fixed a bug where on revert, on_finalize will cause underflow panic 
Log a warning on ingress reverts

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
@@ -239,7 +239,7 @@ where
 							);
 						},
 						Ordering::Greater => {
-							log::warn!("Deposit channels on Solana chain has reverted! Account: {:?}, Asset: {:?}, amount: {:?}", account, details.asset, previous_amount - ingress_total.amount);
+							log::warn!("Deposit channels on Solana chain has reverted! Account: {:?}, Asset: {:?}, Previous ingressed total: {:?}, new ingressed total: {:?}", account, details.asset, previous_amount, ingress_total.amount);
 						},
 						Ordering::Equal => (),
 					}
--- a/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
@@ -239,11 +239,7 @@ where
 							);
 						},
 						Ordering::Greater => {
-							Sink::on_ingress_reverted(
-								account.clone(),
-								details.asset,
-								previous_amount - ingress_total.amount,
-							);
+							log::warn!("Deposit channels on Solana chain has reverted! Account: {:?}, Asset: {:?}, amount: {:?}", account, details.asset, previous_amount - ingress_total.amount);
 						},
 						Ordering::Equal => (),
 					}
--- a/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
@@ -242,7 +242,7 @@ where
 							Sink::on_ingress_reverted(
 								account.clone(),
 								details.asset,
-								ingress_total.amount - previous_amount,
+								previous_amount - ingress_total.amount,
 							);
 						},
 						Ordering::Equal => (),
--- a/state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
@@ -378,8 +378,7 @@ impl MockStorageAccess {
 		ELECTORAL_UNSYNCHRONISED_STATE_MAP.with(|old_state_map| {
 			let state_map_ref = old_state_map.borrow();
 			state_map_ref
-				.get(&key.encode())
-				.expect("Key should exist")
+				.get(&key.encode())?
 				.clone()
 				.map(|v| ES::ElectoralUnsynchronisedStateMapValue::decode(&mut &v[..]).unwrap())
 		})
--- a/state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
@@ -154,7 +154,7 @@ impl<ES: ElectoralSystem> ElectoralReadAccess for MockAccess<ES> {
 		Option<<Self::ElectoralSystem as ElectoralSystem>::ElectoralUnsynchronisedStateMapValue>,
 		CorruptStorageError,
 	> {
-		Ok(MockStorageAccess::unsynchronised_state_map_state::<ES>(key))
+		Ok(MockStorageAccess::unsynchronised_state_map::<ES>(key))
 	}
 }
 
@@ -185,7 +185,7 @@ impl<ES: ElectoralSystem> ElectoralWriteAccess for MockAccess<ES> {
 			<Self::ElectoralSystem as ElectoralSystem>::ElectoralUnsynchronisedStateMapValue,
 		>,
 	) {
-		MockStorageAccess::set_unsynchronised_state_map_state::<ES>(key, value);
+		MockStorageAccess::set_unsynchronised_state_map::<ES>(key, value);
 	}
 }
 
@@ -372,7 +372,7 @@ impl MockStorageAccess {
 			ES::ElectoralUnsynchronisedState::decode(&mut &state_ref[..]).unwrap()
 		})
 	}
-	pub fn unsynchronised_state_map_state<ES: ElectoralSystem>(
+	pub fn unsynchronised_state_map<ES: ElectoralSystem>(
 		key: &ES::ElectoralUnsynchronisedStateMapKey,
 	) -> Option<ES::ElectoralUnsynchronisedStateMapValue> {
 		ELECTORAL_UNSYNCHRONISED_STATE_MAP.with(|old_state_map| {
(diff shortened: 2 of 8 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
```diff
@@ -304,19 +308,22 @@ impl MockStorageAccess {
...
 		identifier: ElectionIdentifierOf<ES>,
-		properties: ES::ElectionProperties,
+		properties: Option<ES::ElectionProperties>,
 	) {
 		ELECTION_PROPERTIES.with(|old_properties| {
 			let mut properties_ref = old_properties.borrow_mut();
-			properties_ref.insert(identifier.encode(), properties.encode());
+			match properties {
+				Some(properties) => properties_ref.insert(identifier.encode(), properties.encode()),
+				None => properties_ref.remove(&identifier.encode()),
```
kylezs, line 325: maybe this is related to the question you asked this morning, but this shouldn't be necessary. should only need to remove properties as part of the election deletion

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
```diff
@@ -0,0 +1,383 @@
...
+		.force_consensus_update(ConsensusStatus::Gained {
+			most_recent: None,
+			new: channel_state_ingressed().to_state(),
+		})
+		.test_on_finalize(
+			&900,
+			|_| (),
+			vec![Check::started_at_default_state(), Check::ended_at_ingressed_state()],
+		)
+		.then(|| {
```
kylezs, line 299: I recently had a PR merge to main which allows these to be turned into `Check`s by allowing passing a parameter to a check. Seems we dupe this ingress amount check, so we could definitely make a Check with a parameter for that. 
the PR that adds it, and converts some existing Checks: https://github.com/chainflip-io/chainflip-backend/pull/5441

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
```diff
@@ -0,0 +1,383 @@
...
+			assert_eq!(
+				pre_finalize.unsynchronised_state_map,
+				to_raw_state_map(initial_channel_state().to_state_map()),
+				"Expected initial state incorrect pre-finalization."
+			);
+		},
+		ended_at_ingressed_state(_pre, post_finalize) {
+			assert_eq!(
+				post_finalize.unsynchronised_state_map,
+				to_raw_state_map(channel_state_ingressed().to_state_map()),
```
kylezs, line 262: we could also probably use the parameterised checks here too, since we're creating local state in the test and comparing to a global const, we can pass through the local expected state as a param inside the test so it's all nice and contained

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -242,7 +242,7 @@ where
 							Sink::on_ingress_reverted(
 								account.clone(),
```
kylezs, line 243: think decision was to just log here and remove this hook

state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
```diff
@@ -110,8 +110,12 @@ impl<ES: ElectoralSystem> ElectionWriteAccess for MockWriteAccess<ES> {
 		new_extra: <Self::ElectoralSystem as ElectoralSystem>::ElectionIdentifierExtra,
 		properties: <Self::ElectoralSystem as ElectoralSystem>::ElectionProperties,
 	) -> Result<(), CorruptStorageError> {
+		// Remove the old properties and replace it with the new one.
+		MockStorageAccess::set_election_properties::<ES>(self.identifier(), None);
```
kylezs, line 114: we don't need to explicitly remove, it'll get overwritten by setting to the new properties

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
```diff
@@ -0,0 +1,383 @@
...
+							block_number: channel.block_number,
+						},
+					),
+				)
+			})
+			.collect::<BTreeMap<_, _>>()
+	}
+}
+
+fn to_raw_state_map<K: Encode, V: Encode, M: IntoIterator<Item = (K, V)>>(
```
kylezs, line 137: I think it's better to decode the state in the utility and then we compare the typed data rather than the encoded, so think we should change the utility to pass through the decoded pre and post state.

state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
syan095: I had to do this because on "refresh()", the a new identifier is created (with extra incremented by 1), which creates a new entry in the "election properties" map. In "force_consensus" we have to ensure there's only 1 property identifier, which will fail if we don't remove the previous identifier from the properties map. This logic is also consistent with the actual implementation of the "refresh_elections()" function

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
syan095: it turns out that returning the decoded version requires the Key to implement `Ord` trait, which is not required by the ElectoralSystem trait (and not implemented by the type used for DeltaBasedIngress for Solana: `(Account, Asset)`. 
I think this is why the code originally stores the encoded version of everything

syan095: I managed to change this! it turned out I just needed to add the trait bounds in. 

I also made the StateMap<Vec, Vec> instead of <Vec, Option<Vec>> just to make everything easier

state-chain/pallets/cf-elections/src/electoral_systems/mocks/access.rs
syan095: Same as before - the "extra" will just add a new entry instead of overwriting the old one, since Extra has changed.

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
syan095: Now logging a warning instead.

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
syan095: I ended up keeping this `then` function, because I wanted to test `open_channel`. 
I thought opening new channels as part of "finalize checks" feels weird, but more critically, open_channel modifies the states of the Electoral system, (e.g. starts a new election) and I need the "identifier" field to refresh after an election has started. It would be easier to have this function where code that mutates Electoral state can execute.

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
```diff
@@ -0,0 +1,487 @@
...
+			&950,
+			|_| (),
+			vec![
+				Check::started_at_state(channel_state_ingressed()),
+				Check::ended_at_state(channel_state_final()),
+				Check::ingressed(vec![
+					(1u32, Asset::Sol, 1_000u64),
+					(2u32, Asset::SolUsdc, 2_000u64),
+					(1u32, Asset::Sol, 1_000u64),
+					(2u32, Asset::SolUsdc, 2_000u64),
```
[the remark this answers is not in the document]
syan095: Done

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
```diff
@@ -0,0 +1,487 @@
...
+
+#[test]
+fn only_trigger_ingress_on_witnessed_blocks() {
+	with_setup(None)
+		.build_with_initial_election()
+		.force_consensus_update(ConsensusStatus::Gained {
+			most_recent: None,
+			new: channel_state_ingressed().to_state(),
+		})
+		.test_on_finalize(&699, |_| (), vec![Check::assert_unchanged()])
```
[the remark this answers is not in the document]
syan095: Added a new test that tests these use cases.

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -239,11 +239,7 @@ where
 							);
 						},
 						Ordering::Greater => {
```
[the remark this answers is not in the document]
syan095: Done.

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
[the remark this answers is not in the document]
syan095: Added a new test that tests these use cases.

state-chain/pallets/cf-elections/src/electoral_systems/tests/delta_based_ingress.rs
```diff
@@ -221,6 +221,9 @@ register_checks! {
 		ended_at_state(_pre, post, election_state: BTreeMap<AccountId, ChannelTotalIngressedFor<MockIngressSink>>) {
 			assert_eq!(*post.election_state.get(post.election_identifiers[0].unique_monotonic()).unwrap(), election_state, "Expected election state incorrect.");
 		},
+		ended_at_empty_state(_pre, post) {
```
kylezs, line 224: 👍  nice additional safety check

kylezs (approved):
Nice tests 🚀  thanks for knocking this one off 🫡

PR #5446, opened 2024-11-25
chore: update deps and remove stale audit exceptions

dandanlen:
Make Audit Great Again


kylezs (approved):
Rust compiler got a bit better at lifetime inference it seems 🚀

PR #5450, opened 2024-11-26
feat: cleanup VaultStartBlockNumbers upon epoch expiry

nmammeri:
# Pull Request

Closes: PRO-1699

## Checklist

## Summary

* Implemented cleaning VaultStartBlockNumbers upon epoch expiry
* It's implemented in a simple way where it removes all epochs <= expired_epoch. This means that it would be expensive the first time we run this.


dandanlen (approved):
👌

PR #5455, opened 2024-11-27
feat: add assethub to engine

MxmUrw:
# Pull Request

Closes: PRO-1858
Closes: PRO-1879

## Checklist

## Summary

#### Non-Breaking changes


engine/src/witness/hub.rs
```diff
@@ -0,0 +1,367 @@
...
+	ExtrinsicSuccess,
+}
+
+use assethub::{
+	balances::events::Transfer, proxy::events::ProxyAdded, system::events::ExtrinsicSuccess,
+	transaction_payment::events::TransactionFeePaid,
+};
+
+pub fn filter_map_events(
+	res_event_details: Result<EventDetails<PolkadotConfig>, subxt::ext::subxt_core::error::Error>,
```
kylezs, line 64: ?
```suggestion
	res_event_details: Result<EventDetails<AssethubConfig>, subxt::ext::subxt_core::error::Error>,
```

engine/src/witness/hub.rs
```diff
@@ -0,0 +1,367 @@
...
+		.map(|event| (event_details.phase(), event)),
+		Err(err) => {
+			error!("Error while parsing event: {:?}", err);
+			None
+		},
+	}
+}
+
+pub async fn proxy_added_witnessing(
+	epoch: Vault<cf_chains::Polkadot, PolkadotAccountId, ()>,
```
kylezs, line 99: and below
```suggestion
	epoch: Vault<cf_chains::Assethub, PolkadotAccountId, ()>,
```

engine/src/witness/hub.rs
```diff
@@ -0,0 +1,367 @@
...
+		.egress_items(scope, state_chain_stream.clone(), state_chain_client.clone())
+		.await
+		.then({
+			let process_call = process_call.clone();
+			let dot_client = dot_client.clone();
+			move |epoch, header| {
+				process_egress(epoch, header, process_call.clone(), dot_client.clone())
+			}
+		})
+		.continuous("Polkadot".to_string(), db)
```
kylezs, line 252: this is a particularly important change, otherwise polkadot and assethub would be writing to the same database table
```suggestion
		.continuous("Assethub".to_string(), db)
```

engine/src/lib.rs
```diff
@@ -315,13 +315,27 @@ async fn run_main(
 				.await?
 			};
 
+			let hub_client = {
+				let expected_hub_genesis_hash = PolkadotHash::from(
+					state_chain_client
+						.storage_value::<pallet_cf_environment::PolkadotGenesisHash<state_chain_runtime::Runtime>>(
+							// Change to Assethub
```
kylezs, line 322: guess we can make this change now

engine/src/witness/hub.rs
kylezs: we might want to define an explicit Config for Assethub so it's not tied to the Polkadot one defined in subxt

MxmUrw: In `witness/hub.rs` we only use the Polkadot Config in order to pass it to the `EventDetails<>` and `Events<>` structs, which themselves access the associated `PolkadotConfig::Hash` type or just put the Config in a phantom data, respectively.

Its more interesting in the (dot) `rpc` file(s) where we also access the associated `PolkadotConfig::Header` type.

The `PolkadotConfig` struct itself inherits all fields from `SubstrateConfig` (https://github.com/paritytech/subxt/blob/f5e9ce0d2ca59ecd716027f3faac7fcc0875ac1b/core/src/config/polkadot.rs#L19-L28). 

I think there are two options how to proceed:
1. Make the rpc functions parametric over the config type, i.e. `function<T: Config>(...)`, and instantiate it with `PolkadotConfig` for dot and with a newly created `AssethubConfig` for assethub.
2. Rename all occurences of `PolkadotConfig` in the rpc files to `SubstrateConfig` and create a type alias `type AssethubConfig = SubstrateConfig`.

engine/src/witness/hub.rs
MxmUrw: I see, thanks for catching that!

engine/src/witness/hub.rs
kylezs: As discussed, think we can leave it as is, or use option 2. If we need to update we'll need to update it whether we parameterise or not, so doesn't make much of a difference in the end.

MxmUrw:
This PR has been superseded by https://github.com/chainflip-io/chainflip-backend/pull/5473, closing.

PR #5462, opened 2024-12-03
Feat: improve ccm gas calculations and refactors

albert-llimos:
# Pull Request

Closes: PRO-1314, PRO-1869, PRO-1870, PRO-1871, PRO-1872, PRO-1873, PRO-1741, PRO-1637, PRO-1771

## Checklist

## Summary

- Improve CCM gas budget
- Unify CCM swaps with regular swaps
- Unify `CcmFailed` and `DepositIgnored` into` DepositFailed`.
- Update Arbitrum Chain Tracking.
- Extend cf_swap_rate to quote CCM swaps
- Update bouncer accordingly
- Added missing parameters to events as requested by Product.


msgmaxim (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -320,21 +322,14 @@ pub mod pallet {
 			destination_asset: Asset,
 			destination_address: ForeignChainAddress,
 			broker_fees: Beneficiaries<AccountId>,
+			channel_metadata: Option<CcmChannelMetadata>,
```
msgmaxim, line 325: I think this will require a migration since `ChannelAction` is stored as part of (at least) `DepositChannelDetails` from `DepositChannelLookup`.

Also, not strictly necessary for this PR, but maybe we could make `refund_params` non optional if you are writing a migration anyway?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1515,86 +1481,23 @@ pub mod pallet {
...
+							swap.swap_request_id,
+							*accumulated_output_amount,
+							request.output_asset,
+							output_address.clone(),
+							None,  /* ccm */
+							false, /* refund */
+						);
 					}
+
+					true
```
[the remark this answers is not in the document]
albert-llimos: True, I missed that when removing code. Will do.

state-chain/pallets/cf-swapping/src/tests/dca.rs
```diff
@@ -111,20 +112,17 @@ fn get_dca_state(request_id: SwapRequestId) -> DcaState {
...
-		.expect("request state does not exist")
-		.state
-	{
-		ccm.gas_swap_state
-	} else {
-		panic!("Not a CCM swap");
-	}
+#[test]
+fn dca_happy_path_ccm() {
+	dca_happy_path(true);
```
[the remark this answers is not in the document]
albert-llimos: I wanted to do it especially to test that all existing functionality would work the same for regular swaps and DCA swaps after the big refactoring. Just as a healthy code check.
Having said that, I think you're right that it 's not strictly needed but given that CCM and non-CCM are different types of swaps "offered to users" I'd keep the tests to ensure that they have the same behavior and that no modification of the code in the future alters that, or that at least it has to be a conscious change.
Not a very strong opinion though, I wouldn't mind removing. Any opinion @dandanlen ?

state-chain/pallets/cf-ingress-egress/src/lib.rs
albert-llimos: Right, that must be at least one of the migrations needed, not sure if that's what's causing the upgrade test to fail. Will do.

albert-llimos: > Also, not strictly necessary for this PR, but maybe we could make refund_params non optional if you are writing a migration anyway?
Afaik that we've decided on Vault swaps and deposit channels to have FoK so I'd agree on remove it. However, if we remove it on both places we need to update the SDK and some more updates on the bouncer so it'll clutter this PR a bit. I've opened this ticket, please check it out and add a comment: https://linear.app/chainflip/issue/PRO-1883/make-the-refund-parameters-non-optional

albert-llimos:
There is some overlap in the types introduced in this PR and the newly merged #5437 . I'll take a look and see if we can merge them in a nice way.

dandanlen (commented):
Nice work

state-chain/chains/src/eth.rs
```diff
@@ -83,13 +80,43 @@ impl EthereumTrackedData {
...
+		// Adding one extra gas unit per message's length (byte) for the extra gas overhead of
+		// passing the message through the Vault. The extra l2 gas per message's calldata byte
+		// should be included in the user's gas budget.
+		(gas_budget
+			.saturating_add(if is_native_asset {
+				CCM_VAULT_NATIVE_GAS_OVERHEAD
+			} else {
+				CCM_VAULT_TOKEN_GAS_OVERHEAD
+			})
+			.saturating_add(message_length as u128))
```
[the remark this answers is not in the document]
albert-llimos: Here we are only calculating the gas limit, which is the user's specified gas budget + gas overhead. In Arbitrum the calculation is way more complex because the gas limit depends on a bunch of factors like the `L1_GAS_PER_BYTES`.
The Ethereum gas per bytes is part of the user's specified gas budget. It's a bit cumbersome.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -343,11 +338,27 @@ pub mod pallet {
...
+		pub affiliate_fees: Affiliates<AffiliateShortId>,
+		pub deposit_metadata: Option<CcmDepositMetadataEncoded>,
+		pub refund_params: Option<ChannelRefundParameters>,
+		pub dca_params: Option<DcaParameters>,
+	}
+
+	#[derive(Clone, RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+	pub enum DepositInfo<AccountId> {
+		DepositChannel { action: ChannelAction<AccountId> },
+		VaultSwap { details: VaultSwapDetails<AccountId> },
```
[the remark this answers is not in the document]
albert-llimos: Makes sense. Also I saw in the recent PR merged to main the addition of `VaultDepositWitness` that adds more similar (if not duplicated) code. I looked at rebasing and due to that PR we probably need some refactoring around this too so probably worth rethinking it together with that code.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -642,12 +653,13 @@ pub mod pallet {
...
-		DepositIgnored {
-			deposit_address: Option<TargetChainAccount<T, I>>,
+		DepositFailed {
+			reason: DepositFailedReason,
+			origin: SwapOrigin,
 			asset: TargetChainAsset<T, I>,
 			amount: TargetChainAmount<T, I>,
 			deposit_details: <T::TargetChain as Chain>::DepositDetails,
-			reason: DepositIgnoredReason,
+			details: Box<DepositInfo<T::AccountId>>,
```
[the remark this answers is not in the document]
albert-llimos: yes...not ideal.. I'll reassess depending on how we address this (https://github.com/chainflip-io/chainflip-backend/pull/5462#discussion_r1869557071).

state-chain/custom-rpc/src/lib.rs
```diff
@@ -781,6 +781,18 @@ pub trait CustomApi {
...
 	) -> RpcResult<RpcSwapOutputV2>;
+	#[method(name = "swap_rate_v4")]
+	fn cf_pool_swap_rate_v4(
+		&self,
+		from_asset: Asset,
+		to_asset: Asset,
+		amount: U256,
+		broker_commission: BasisPoints,
+		dca_parameters: Option<DcaParameters>,
+		ccm_data: Option<CcmData>,
```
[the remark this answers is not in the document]
albert-llimos: > we can consider this a non-breaking change and just update the v3 rpc
But it will break anyone using v3 if we update it like you suggest, no?

v2 and v3 were added and just had one or two new fields, that's why I went for v4. I don't have a strong opinion there, happy to just add it to v3 if you think that's better.

state-chain/runtime/src/chainflip.rs
```diff
@@ -249,28 +244,20 @@ macro_rules! impl_transaction_builder_for_evm_chain {
...
-						.tracked_data
-						.max_fee_per_gas(One::one());
-					Some(gas_budget
-						.checked_div(current_fee_per_gas)
-						.unwrap_or_else(||{
-							log::warn!("Current gas price for {} is 0. This should never happen. Please check Chain Tracking data.", $chain::NAME);
-							Default::default()
-						}).min($max_gas_limit)
-						.into())
+				if let (Some(gas_budget), Some(message_length), Some(transfer_asset), Some(native_asset)) = (call.gas_budget(), call.message_length(), call.transfer_asset(), <$env as EvmEnvironmentProvider<$chain>>::token_address($chain::GAS_ASSET)) {
```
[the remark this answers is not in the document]
albert-llimos: > is there any reason these all have to be accessed separately?

Just fixed that.

> calculate_ccm_gas_limit could also take an asset instead of bool

The issue there is that from the `call.transfer_asset()` we just get H160 (address of the token) but not the `Asset`, given that the call has the encoded data already for EVM.  Comparing them here, where we can get the token addresses from the `EvmEnvironmentProvider` is quite easy and then pass a bool. Passing the token address to that `calculate_ccm_gas_limit` is quite ugly too. Is there an existing way to get the asset from a token address so we can compare them here and pass the asset to `calculate_gas_limit`?
Otherwise I guess I can just add a function to the `EvmEnvironmentProvider` like `token_address_to_asset`.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
albert-llimos: Alright Maxim's rebase and refactor made it fit perfectly so we don't have to overcomplicate this :ok_hand:

state-chain/custom-rpc/src/lib.rs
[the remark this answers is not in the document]
dandanlen: v2 changed the output type in a breaking way.
v3 added a new *required* input param (breaking change)
v4 adds a new optional param -> this is not a breaking change, so we can just add it to the v3 rpc.

state-chain/runtime/src/chainflip.rs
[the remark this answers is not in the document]
dandanlen: Ok yeah i didn't quite realise we were dealing with addresses here.

state-chain/custom-rpc/src/lib.rs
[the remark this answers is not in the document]
albert-llimos: Oh right - I've just been too long in the blockchain encoding trenches where everything in the encoding is a breaking change :facepalm: but of course this isn't. Will fix.

dandanlen:
@albert-llimos is PRO-1820 solved by this too?

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -324,15 +324,6 @@ decl_runtime_apis!(
 			amount: AssetAmount,
 			additional_limit_orders: Option<Vec<SimulateSwapAdditionalOrder>>,
 		) -> Result<SimulatedSwapInformation!["1.0.0"], DispatchErrorWithMessage>;
-		#[changed_in(3)]
-		fn cf_pool_simulate_swap(
```
dandanlen: FYI I think you need to keep this - it's not a breaking change in json, but it *is* breaking in SCALE :/

albert-llimos:
> @albert-llimos is [PRO-1820](https://linear.app/chainflip/issue/PRO-1820) solved by this too?

No, there's still bunch of warning loggings when decoding and checking values which should be emitted as events instead.

albert-llimos:
Resolving all the comments about changes that impact product. I collected them in a linear ticket: https://linear.app/chainflip/issue/PRO-1889/update-product-side-after-changes-in-the-protocol-related-to-ccm

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -376,6 +380,15 @@ pub mod pallet {
 		pub boost_fee: BasisPoints,
 	}
 
+	#[derive(
+		CloneNoBound, RuntimeDebugNoBound, PartialEqNoBound, EqNoBound, Encode, Decode, TypeInfo,
+	)]
+	#[scale_info(skip_type_params(T, I))]
+	pub enum DepositFailedDetails<T: Config<I>, I: 'static> {
```
msgmaxim, line 387: I'm a little concerned about exposing *everything* we receive in an event. Feels like we should be more intentional about what we put in the public interface. I see the rationale for providing as much info as possible, but this also means we are more likely break compatibility (either intentionally or not) whenever these mostly internal structs are modified (I know CFEs also uses them, but could still be considered internal).

As a thought, maybe there is a way to separate fields that can be dependant upon externally from "extra info" that's just for informational purposes (i.e. it could be dumped somewhere in a text field in the explorer). Not sure if the consumers of these events will be disciplined enough to not start depending on that info as well.

Interested to hear your thoughts @dandanlen.

CumpsD: > Not sure if we (or other consumers of these events) will be disciplined enough to not start depending on that info as well.

Assume whatever is public will be used by someone eventually :)

msgmaxim: > > Not sure if we (or other consumers of these events) will be disciplined enough to not start depending on that info as well.
> 
> Assume whatever is public will be used by someone eventually :)

Yes, this essentially: https://www.hyrumslaw.com/

state-chain/runtime/src/lib.rs
```diff
@@ -1293,13 +1293,22 @@ type PalletMigrations = (
...
+	VersionedMigration<
+		2,
+		3,
+		migrations::solana_vault_swaps_migration::SolanaVaultSwapsMigration,
+		pallet_cf_elections::Pallet<Runtime, SolanaInstance>,
+		DbWeight,
+	>,
+	VersionedMigration<
+		3,
+		4,
```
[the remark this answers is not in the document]
ramizhasan111: added a No op migration for other pallet instances. For SolanaVaultSwaps migration it should not be a problem since we have only one instance of elections pallet.

state-chain/pallets/cf-swapping/src/migrations.rs
```diff
@@ -1,4 +1,16 @@
...
-pub type PalletMigration<T> = PlaceholderMigration<6, Pallet<T>>;
+pub type PalletMigration<T> = (
+	VersionedMigration<
+		6,
+		7,
+		swap_request_migration::SwapRequestMigration<T>,
+		Pallet<T>,
+		<T as frame_system::Config>::DbWeight,
+	>,
+	PlaceholderMigration<17, Pallet<T>>,
```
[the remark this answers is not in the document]
albert-llimos: Well spotted, thanks!

ramizhasan111:
As mentioned to @albert-llimos, since we changed the Apicall types for ethereum and arbitrum (because of the GasAmount change), I think we also need the migration for ```PendingApiCalls``` in the broadcast pallet.

state-chain/pallets/cf-ingress-egress/src/lib.rs
CumpsD: Didn't know that one, thanks! It's very correct indeed :)

state-chain/pallets/cf-swapping/src/migrations/swap_request_migration.rs
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		crate::SwapRequests::<T>::translate_values::<old::SwapRequest<T>, _>(|old_swap_requests| {
+			Some(SwapRequest {
+				id: old_swap_requests.id,
+				input_asset: old_swap_requests.input_asset,
+				output_asset: old_swap_requests.output_asset,
+				refund_params: old_swap_requests.refund_params,
+				state: match old_swap_requests.state {
```
msgmaxim, line 74: It ocurred to me just now (missed this initially, sorry @albert-llimos) that the old state here in case of CCM contains `GasSwapState::ToBeScheduled { gas_budget, .. }` where `gas_budget` is what we set aside for swapping into the gas asset. As things stand, looks like this amount will be lost, and we probably want to put it back to `dca_state.remaining_input_amount` during the migration?

(I think `GasSwapState::ToBeScheduled` is going to be the only possible state/variant since we will have safe mode on, and if we get a vault swap during the upgrade it will be waiting in its initial state, which means the migration is much more straightforward.)

albert-llimos: I pushed this after your review so you hadn't even reviewed it ;) 
I thought about this. With the new logic the gas swap is done on the egress and CCM fees will be taking there (aka the new gas swap). I don't think it's much of a problem if we happen to swallow a CCM gas amount for a swap and then the egress fee that is taken is different. Also there is no CCM swaps going on and we will put safe mode, so as long it can't cause some failure I think it's fine as it is.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: While I tend to agree, I also feel like this can reasonably be considered an exception to this rule. With errors we do want to provide as much context as possible, and errors should not be (and generally are not) used for 'control flow' - only for display. (like you mentioned - just 'dump it somewhere' for debugging)

In this case, I feel like it's reasonable to consider the action/intent of the failed operation to be useful context. Maybe we can be less vague about the purpose of this field and just call it 'error context' or 'metadata' or something.

I'm open to being more deliberate and only exposing the necessary context as long as it's obvious what that necessary context is. For example, as an alternative, we could use a richer type for the error reason, that includes the value that caused the issue? 

To recap - where this all came from was that we wanted access to the destination asset and the transaction_in_id in DepositIgnored and CcmFailed events. Then we realised that there is lot of duplicate information, so we merged the two events and added the extra context for vault swaps. I feel like the new approach is more consistent and structured than what we had previously.

state-chain/pallets/cf-swapping/src/migrations/swap_request_migration.rs
msgmaxim: If I understand correctly, it is not a matter of charging a different fee, but essentially charging the fee twice. Sure, we would have to be unlucky to hit this, but it also seems very simple to just add the gas amount back to input. Not a strong opinion, I will leave it to you to decide whether it is worth it or not.

albert-llimos: We also then need to change the state of another storage slot (`SwapState`), I think it's just unnecessary. This would only happen in a CCM+DCA before the first chunk of DCA, since the gas swap is scheduled after that. Besides the main point that we have never even had a swap like that. The downside is just charging egress fees twice in that extremely rare scenario, which is fine.

dandanlen: I agree, I think we can live with this.

albert-llimos:
For visibility, I wrote two migrations here but there is more migrations needed. Writing them in https://github.com/chainflip-io/chainflip-backend/pull/5490

dandanlen (approved):
LFG

PR #5463, opened 2024-12-04
feat: Expose vault swaps from ingress-egress-tracker

j4m1ef0rd:
# Pull Request

Closes: PRO-1825

## Checklist

## Summary

- Refactored the tracker code to use proper traits instead of exploiting the from/into impls.
- Added a new data enum `VaultDeposit` that gets stored in the database. It is stored using the key `vault_deposit:{chain}:{tx_id}`.
- Added a unit test for the vault deposit.
- For every vault deposit the tracker will call an RPC to get the brokers registered affiliates and map them to the short id's in the vault deposit data.
- New Wrapper `TrackerStateChainClient` to assist with mocking RPC calls.


api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -84,6 +104,22 @@ enum WitnessInformation {
 		tx_out_id: TransactionId,
 		tx_ref: TransactionRef,
 	},
+	VaultDeposit {
```
niklasnatter, line 107: think it makes sense to call this `Swap` as it has an output asset
```suggestion
	VaultSwap {
```

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -84,6 +104,22 @@ enum WitnessInformation {
...
+		tx_id: TransactionInIdForAnyChain,
+		deposit_chain_block_height: <AnyChain as Chain>::ChainBlockNumber,
+		input_asset: cf_chains::assets::any::Asset,
+		output_asset: cf_chains::assets::any::Asset,
+		amount: NumberOrHex,
+		destination_address: EncodedAddress,
+		deposit_metadata: Option<CcmDepositMetadata>,
+		deposit_details: Option<DepositDetails>,
+		broker_fee: Beneficiary<AccountId32>,
+		affiliate_fees: Affiliates<AffiliateShortId>,
```
niklasnatter, line 118: would be great to have the account ids instead of the short ids here

api/bin/chainflip-ingress-egress-tracker/src/witnessing/snapshots/chainflip_ingress_egress_tracker__witnessing__state_chain__tests__handle_vault_deposit_calls.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
+expression: "store.storage.get(\"vault_deposit:Ethereum:b5c8bd9430b6cc87a0e2fe110ece6bf527fa4f170a4bc8cd032f768fc5219838\").unwrap()"
+snapshot_kind: text
+---
+{"affiliate_fees":[],"amount":"0x64","boost_fee":0,"broker_fee":{"account":"5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM","bps":0},"dca_params":null,"deposit_chain_block_height":1,"deposit_details":null,"deposit_metadata":null,"destination_address":{"Eth":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]},"input_asset":{"asset":"ETH","chain":"Ethereum"},"output_asset":{"asset":"FLIP","chain":"Ethereum"},"refund_params":{"min_price":"0x0","refund_address":{"Eth":"0x541f563237a309b3a61e33bdf07a8930bdba8d99"},"retry_duration":0}}
```
niklasnatter, line 6: it looks like the destination_address is formatted differently than the refund_adress here. think it would make sense to have them both exposed as string

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -84,6 +104,22 @@ enum WitnessInformation {
...
 	},
+	VaultDeposit {
+		#[serde(skip_serializing)]
+		tx_id: TransactionInIdForAnyChain,
+		deposit_chain_block_height: <AnyChain as Chain>::ChainBlockNumber,
+		input_asset: cf_chains::assets::any::Asset,
+		output_asset: cf_chains::assets::any::Asset,
+		amount: NumberOrHex,
+		destination_address: EncodedAddress,
+		deposit_metadata: Option<CcmDepositMetadata>,
```
niklasnatter, line 115:
```suggestion
		ccm_deposit_metadata: Option<CcmDepositMetadata>,
```

api/bin/chainflip-ingress-egress-tracker/src/witnessing/snapshots/chainflip_ingress_egress_tracker__witnessing__state_chain__tests__handle_vault_deposit_calls.snap
niklasnatter: would be cool to have some affiliates, dca params, ccm params in here to see the encoding

niklasnatter (commented):
looks good to me from a product perspective otherwise!

api/bin/chainflip-ingress-egress-tracker/src/witnessing/snapshots/chainflip_ingress_egress_tracker__witnessing__state_chain__tests__handle_vault_deposit_calls.snap
acdibble: and the account id uses the substrate ss58 prefix instead of the cf prefix

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
j4m1ef0rd: I like `VaultDeposit` better because it is more consistent with `VaultDepositWitness`.

kylezs: Don't have a lot of context but, If we have a vault deposit for an LP it would be called a VaultDeposit? which would make more sense, and then this being VaultSwap makes sense

niklasnatter: i dont have a strong opinion on this - `VaultDeposit` is fine for me too if you prefer it. feel free to resolve

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
j4m1ef0rd: This is turning out to be a non-trivial change. Can we leave this for a later update?

niklasnatter: unfortunately we will need to do a separate rpc call to handle the short ids on our side, so i would really prefer to have them in redis

niklasnatter: its fine for me to do that in a follow up pr obv, but that follow up should land before we start using it 🙂

j4m1ef0rd: I managed to get it working. Not the nicest, but it will do. At least we are one step closer to storing all swap data in redis. Shouldn't be too hard now to fetch the deposit channel data and get it in the db.

api/bin/chainflip-ingress-egress-tracker/src/witnessing/snapshots/chainflip_ingress_egress_tracker__witnessing__state_chain__tests__handle_vault_deposit_calls.snap
kylezs: Looks fixed now

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -96,20 +136,43 @@ impl WitnessInformation {
...
+					TransactionInIdForAnyChain::Bitcoin(hash) => hex::encode(hash.as_bytes()),
+					TransactionInIdForAnyChain::Evm(hash) => hex::encode(hash.as_bytes()),
+					TransactionInIdForAnyChain::Polkadot(transaction_id) => format!(
+						"{}-{}",
+						transaction_id.block_number, transaction_id.extrinsic_index
+					),
+					TransactionInIdForAnyChain::Solana((address, id)) => format!("{address}-{id}",),
+					TransactionInIdForAnyChain::MockEthereum(_) |
+					TransactionInIdForAnyChain::None => {
+						return Err(anyhow!("Invalid transaction id: {tx_id:?}"));
```
dandanlen, line 172: We should just impl Display for TransactionInIdForAnyChain ?

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -122,95 +185,256 @@ impl Storable for WitnessInformation {
...
+trait DepositIntoWitnessInformation<C: Chain> {
+	async fn into_witness_information<StateChainClient>(
+		self,
+		height: <C as Chain>::ChainBlockNumber,
+		network: NetworkEnvironment,
+		state_chain_client: Arc<StateChainClient>,
+	) -> anyhow::Result<WitnessInformation>
+	where
+		StateChainClient: StorageApi + ChainApi + TrackerApi + 'static + Send + Sync;
+}
```
dandanlen, line 222: This looks like it can be simplified - no need for async or sc client? (Or result??)

```suggestion
trait DepositIntoWitnessInformation<C: Chain> {
	async fn into_witness_information<StateChainClient>(
		self,
		height: <C as Chain>::ChainBlockNumber,
		network: NetworkEnvironment,
	) -> WitnessInformation;
}
```

dandanlen: (Also, not sure we really need a macro)

dandanlen: Actually - this is pretty confusing - we have two macros for implementing the same trait 😅

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -122,95 +185,256 @@ impl Storable for WitnessInformation {
...
 		}
+	};
+}
+
+impl_vault_deposit_into_witness_information!(Ethereum);
+impl_vault_deposit_into_witness_information!(Bitcoin);
+impl_vault_deposit_into_witness_information!(Polkadot);
+impl_vault_deposit_into_witness_information!(Arbitrum);
+
+impl BroadcastIntoWitnessInformation for BroadcastDetails<Ethereum> {
```
dandanlen, line 322: ... but no macro for this?

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -122,95 +185,256 @@ impl Storable for WitnessInformation {
...
+			broadcast_id: self.broadcast_id,
+			tx_out_id: TransactionId::Polkadot {
+				signature: DotSignature(*self.tx_out_id.aliased_ref()),
+			},
+			tx_ref: TransactionRef::Polkadot { transaction_id: self.tx_ref },
+		})
+	}
+}
+
+async fn save_deposit_witnesses<
```
dandanlen, line 364: I feel like we could simplify a lot if we query for a map of affiliate IDs in this method, and then pass that through, rather than threading the sc client (and attendant type constraints and async-ness) down through the whole call stack.

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -122,95 +185,256 @@ impl Storable for WitnessInformation {
...
+where
+	I: cf_chains::instances::ChainInstanceAlias + Chain + 'static,
+	StateChainClient: StorageApi + ChainApi + 'static + Send + Sync,
+	state_chain_runtime::Runtime: pallet_cf_broadcast::Config<ChainInstanceFor<I>>,
+	BroadcastDetails<I>: BroadcastIntoWitnessInformation,
+{
+	if let Some(broadcast_details) =
+		get_broadcast_id::<I, StateChainClient>(state_chain_client, &tx_out_id)
+			.await
+			.map(|broadcast_id| BroadcastDetails::<I> { broadcast_id, tx_out_id, tx_ref })
```
dandanlen, line 411: (...A bit like this approach)

dandanlen:
@niklasnatter @acdibble I made some changes :)

Polkadot addresses should now serialize to ss58 instead of a list of integers (see snapshot diff) - let me know if this is ok. Notably, this is different than the way it's done for channel deposits: currently they serialize polkadot addresses as hex. It would be good to be consistent - all other addresses are stored in the humand readable format (hex for eth/arb, base58 for Solana, bitcoin address). Let me know if you have a preference (use hex for polkadot or ss58). 

I also noticed that eth *hashes* are serialized as hex *without* the leading `0x`. Again, would be good to know if this is ok / consistent with other uses...

kylezs (approved):

api/bin/chainflip-ingress-egress-tracker/src/store.rs
dandanlen, line 1: Looking at this code - the save_to_array method looks like it pushes to an array... meaning that calling this with an array as argument just creates a single-valued array of arrays. (ie. `save_to_array(key, "[1,2,3]")` will result in a stored value of `[[1,2,3]]`). 

Is this intentional? If not, I would suggest just using save_singleton and removing this `save_to_array` method. 

cc @acdibble @niklasnatter

dandanlen: Actually never mind, I see it now. Deposits are keyed by the address so there can be multiple 👌

niklasnatter:
> Note there's an inconsistency in how polkadot destination addresses are stored: they use hex instead of ss58. All other addresses are stored in their respective 'humanreadable' format (hex for eth/arb, base58 for Solana, bitcoin address). @dandanlen 

No strong opinion on my side. Id say doing it consistent with what we already have makes sense

> I also noticed that eth hashes are serialized as hex without the leading 0x. Again, would be good to know if this is ok / consistent with other uses...

I think we have it with a leading `0x` on the events (i remembered) at the moment. So might make sense to have it with the `0x` here too.
- DepositFinalised: https://scan.chainflip.io/events/5791964-1224
- SwapRequested: https://scan.chainflip.io/events/5792253-747

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -84,6 +103,22 @@ enum WitnessInformation {
...
+		output_asset: cf_chains::assets::any::Asset,
+		amount: NumberOrHex,
+		destination_address: String,
+		ccm_deposit_metadata: Option<CcmDepositMetadata>,
+		deposit_details: Option<DepositDetails>,
+		broker_fee: Beneficiary<AccountId32>,
+		affiliate_fees: Affiliates<AccountId32>,
+		refund_params: ChannelRefundParameters,
+		dca_params: Option<DcaParameters>,
+		boost_fee: BasisPoints,
```
niklasnatter, line 120: think it has the `max_` prefix on some events like `DepositFinalised`, but not too important
```suggestion
		max_boost_fee: BasisPoints,
```

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
```diff
@@ -96,121 +131,316 @@ impl WitnessInformation {
...
+	<T::TargetChain as Chain>::DepositDetails: IntoDepositDetailsAnyChain,
+{
+	fn into_witness_information(
+		self,
+		height: <T::TargetChain as Chain>::ChainBlockNumber,
+		affiliate_mapping: &BTreeMap<AffiliateShortId, AccountId32>,
+	) -> WitnessInformation {
+		WitnessInformation::VaultDeposit {
+			tx_id: <<<T::TargetChain as Chain>::ChainCrypto as ChainCrypto>::TransactionInId as IntoTransactionInIdForAnyChain<
+				<T::TargetChain as Chain>::ChainCrypto,
```
niklasnatter, line 251: as somebody that does not fully understand rust, this looks fun

api/bin/chainflip-ingress-egress-tracker/src/witnessing/snapshots/chainflip_ingress_egress_tracker__witnessing__state_chain__tests__handle_vault_deposit_calls.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
+expression: "store.storage.get(\"vault_deposit:Ethereum:b5c8bd9430b6cc87a0e2fe110ece6bf527fa4f170a4bc8cd032f768fc5219838\").unwrap()"
+---
+{"affiliate_fees":[{"account":"cFHtoB6DrnqUVY4DwMHCVCtgCLsiHvv98oGw8k66tazF2ToFv","bps":10}],"amount":"0x64","boost_fee":5,"broker_fee":{"account":"cFHsUq1uK5opJudRDczhdPVj6LGoVTqYsfj71tbHfKsTAzkJJ","bps":10},"ccm_deposit_metadata":null,"dca_params":{"chunk_interval":100,"number_of_chunks":5},"deposit_chain_block_height":1,"deposit_details":null,"destination_address":"0x0000000000000000000000000000000000000000000000000000000000000000","input_asset":{"asset":"ETH","chain":"Ethereum"},"output_asset":{"asset":"FLIP","chain":"Ethereum"},"refund_params":{"min_price":"0x0","refund_address":{"Eth":"0x541f563237a309b3a61e33bdf07a8930bdba8d99"},"retry_duration":0}}
```
niklasnatter, line 5: would be awesome to have a `ccm_deposit_metadata` in here to check how it looks. but not a blocker

dandanlen: Added

dandanlen:
@niklasnatter I added another snapshot test with the destination address format for each chain. LMK if this looks ok.

api/bin/chainflip-ingress-egress-tracker/src/witnessing/snapshots/chainflip_ingress_egress_tracker__witnessing__state_chain__tests__handle_vault_deposit_calls.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
+expression: "store.storage.get(\"vault_deposit:Ethereum:b5c8bd9430b6cc87a0e2fe110ece6bf527fa4f170a4bc8cd032f768fc5219838\").unwrap()"
+---
+{"affiliate_fees":[{"account":"cFHtoB6DrnqUVY4DwMHCVCtgCLsiHvv98oGw8k66tazF2ToFv","bps":10}],"amount":"0x64","broker_fee":{"account":"cFHsUq1uK5opJudRDczhdPVj6LGoVTqYsfj71tbHfKsTAzkJJ","bps":10},"ccm_deposit_metadata":{"channel_metadata":{"ccm_additional_data":"4d4f5245","gas_budget":"0x3039","message":"48454c4c4f"},"source_address":{"Eth":"0xcfcfcfcfcfcfcfcfcfcfcfcfcfcfcfcfcfcfcfcf"},"source_chain":"Ethereum"},"dca_params":{"chunk_interval":100,"number_of_chunks":5},"deposit_chain_block_height":1,"deposit_details":null,"destination_address":"0x0000000000000000000000000000000000000000000000000000000000000000","input_asset":{"asset":"ETH","chain":"Ethereum"},"max_boost_fee":5,"output_asset":{"asset":"FLIP","chain":"Ethereum"},"refund_params":{"min_price":"0x0","refund_address":{"Eth":[84,31,86,50,55,163,9,179,166,30,51,189,240,122,137,48,189,186,141,153]},"retry_duration":0}}
```
niklasnatter, line 5: it looks like the `refund_address` is encoded differently to the other addresses here 😕

niklasnatter: created an issue for this here: https://linear.app/chainflip/issue/PRO-1933/ingress-egress-tracker-refund-address-encoding-differs-from-other

dandanlen: Sorry - forgot that this already had an approval, shouldn't have set auto-merge...

PR #5465, opened 2024-12-04
refactor: remove op codes from btc vault swap encoding

j4m1ef0rd:
# Pull Request

Closes: PRO-1845

## Checklist

## Summary

- Fixed a bug in the auto assigning of affiliate short id's in the broker api.
- Removed the op codes from the nulldata utxo.
- Renamed the `nulldata_utxo` -> `nulldata_payload` seems it's no longer a complete uxto.

### DIFF
--- a/api/lib/src/lib.rs
+++ b/api/lib/src/lib.rs
@@ -521,7 +521,7 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 		affiliate_id: AccountId32,
 		short_id: Option<AffiliateShortId>,
 	) -> Result<AffiliateShortId> {
-		let next_id = if let Some(short_id) = short_id {
+		let register_as_id = if let Some(short_id) = short_id {
 			short_id
 		} else {
 			let affiliates =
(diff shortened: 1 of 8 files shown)


api/lib/src/lib.rs
```diff
@@ -537,16 +537,24 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
...
+				used_ids.sort_unstable();
+				AffiliateShortId::from(
+					used_ids
+						.into_iter()
+						.enumerate()
+						.find(|(index, assigned_id)| {
+							AffiliateShortId::from(*index as u8) != *assigned_id
+						})
+						.map(|(index, _)| index)
+						.unwrap_or(used_id_len) as u8,
```
dandanlen, line 555: I think if we have exactly 255 used ids we might just overwrite that id?

Maybe a test would help 😅

dandanlen (approved):

bouncer/shared/send_btc.ts
```diff
@@ -36,7 +36,7 @@ export async function fundAndSendTransaction(
 }
 
 export async function sendVaultTransaction(
-  nulldataUtxo: string,
+  nulldata_payload: string,
```
kylezs, line 39: javascript convention
```suggestion
  nulldataPayload: string,
```

bouncer/tests/btc_vault_swap.ts
```diff
@@ -22,7 +22,7 @@ const commissionBps = 100;
 
 interface VaultSwapDetails {
   chain: string;
-  nulldata_utxo: string;
+  nulldata_payload: string;
```
kylezs, line 25: same

j4m1ef0rd: This one has to remain like this because its an interface for a rust struct.

api/lib/src/lib.rs
j4m1ef0rd: I wrote a test, seems fine. But yeah it needed one.

dandanlen (approved):

PR #5466, opened 2024-12-04
Feature/pro 1768/custom rpc finalized block subscriptions

nmammeri:
# Pull Request

Closes: PRO-1768

## Checklist

## Summary

#### Non-Breaking changes


state-chain/custom-rpc/src/lib.rs
```diff
@@ -1984,76 +1984,43 @@ where
...
+					log::warn!("ChainHead subscription error {:?}", e);
+					None
+				},
+				_ => None,
+			}
+		})
+		.filter_map(move |event| async move {
+			match (only_finalized, event) {
+				(
+					true,
```
dandanlen, line 2000: @nmammeri actually maybe this condition should be relaxed? ie. we return the events from the initialized message for both options. 

The reason for this event is so that clients can 'catch up' if they need to reset the connection.

nmammeri: Yes, we can relax this condition.  Before we were always starting from the current best block and tried to keep the same behaviour.

dandanlen (approved):

PR #5474, opened 2024-12-06
refactor: move broker api subscription to custom-rpc

nmammeri:
# Pull Request

Closes: PRO-1886

## Checklist

## Summary

* Added cf_subscribe_transaction_screening_events to custom_rpc that uses the new finalized notification behavior
* Modified broker-api subscribe_transaction_screening_events to use the custom-rpc equivalent function
* Also, changed the logic slightly so that it returns an empty events vector if there are no events detected, making it consistent with other custom-rpc subscriptions


MxmUrw (approved):
Broker API changes look good, tested with the deposit monitor, works as expected. For custom-rpc stuff maybe someone else should also have a look.

dandanlen (approved):

PR #5475, opened 2024-12-09
Feat/kyle height witnesser

kylezs:


MxmUrw (approved):
Looking good, thanks for the voterapi implementation!

**EDIT**: Maybe I shouldn't have marked it as ready for review.

PR #5476, opened 2024-12-09
Chore: extend DepositFailed and merge with DepositWitnessRejected

albert-llimos:
# Pull Request

Closes: PRO-1820

## Checklist

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -136,7 +136,7 @@ pub enum DepositFailedReason {
 	/// required to process the requisite transactions.
 	NotEnoughToPayFees,
 	TransactionRejectedByBroker,
-	DepositWitnessRejected,
+	DepositWitnessRejected(DispatchError),
 	InvalidDestinationAddress,
 	InvalidBrokerFees,
 	InvalidRefundParameters,
@@ -1859,9 +1859,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		block_height: TargetChainBlockNumber<T, I>,
 	) {
 		Self::process_channel_deposit_full_witness_inner(&deposit_witness, block_height)
-			.unwrap_or_else(|_| {
+			.unwrap_or_else(|e| {
 				Self::deposit_event(Event::<T, I>::DepositFailed {
-					reason: DepositFailedReason::DepositWitnessRejected,
+					reason: DepositFailedReason::DepositWitnessRejected(e),
 					details: DepositFailedDetails::DepositChannel { deposit_witness },
 				});
 			})
@@ -2323,7 +2323,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 	pub fn process_vault_swap_request_full_witness(
 		block_height: TargetChainBlockNumber<T, I>,
-		VaultDepositWitness {
+		vault_deposit_witness: VaultDepositWitness<T, I>,
+	) {
+		let VaultDepositWitness {
 			input_asset: source_asset,
 			deposit_address,
 			channel_id,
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2109,6 +2112,23 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
+			let destination_chain: ForeignChain = output_asset.into();
+			if !destination_chain.ccm_support() {
+				log::warn!(
+					"Failed to process vault swap due to destination chain not supporting CCM"
+				);
+				return;
+			}
+		}
+
```
albert-llimos, line 2131: I moved this code that checks CCM from the perform_channel action to here. As I see it we only need this check for vault swaps because for channel  swaps this is already checked when opening the deposit channel.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1832,18 +1834,6 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-					Self::deposit_event(Event::<T, I>::DepositFailed {
-						reason: DepositFailedReason::CcmUnsupportedForTargetChain,
-						origin: origin.into(),
-						asset,
-						amount: amount_after_fees,
-						deposit_details,
-						action: Box::new(action),
-					});
-					return DepositAction::NoAction;
-				}
```
albert-llimos, line 1846: I moved this code that checks CCM. As I see it we only need this check for vault swaps because for channel swaps this is already checked when opening the deposit channel. Doing this we avoid having to pass unnecessary paramters all the way down to `perform_channel_action`.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2340,106 +2338,123 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 		let boost_status =
 			BoostedVaultTransactions::<T, I>::get(&tx_id).unwrap_or(BoostStatus::NotBoosted);
 
+		let emit_deposit_failed_event = |reason: DepositFailedReason| {
+			Self::deposit_event(Event::<T, I>::DepositFailed {
+				reason,
+				details: DepositFailedDetails::Vault {
+					vault_witness: Box::new(VaultDepositWitness {
+						input_asset: source_asset,
+						deposit_address: deposit_address.clone(),
```
dandanlen: We can make this a bit nicer by destructuring a cloned value instead of doing it in the function arg.

dandanlen:
@albert-llimos LGTM, I just made some cosmetic changes, mentioned in the comments above. Feel free to merge.

dandanlen (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -375,6 +380,15 @@ pub mod pallet {
 		pub boost_fee: BasisPoints,
 	}
 
+	#[derive(
+		CloneNoBound, RuntimeDebugNoBound, PartialEqNoBound, EqNoBound, Encode, Decode, TypeInfo,
+	)]
+	#[scale_info(skip_type_params(T, I))]
+	pub enum DepositFailedDetails<T: Config<I>, I: 'static> {
+		DepositChannel { deposit_witness: DepositWitness<T::TargetChain> },
+		Vault { vault_witness: Box<VaultDepositWitness<T, I>> },
```
msgmaxim, line 389: I'm assuming we put VaultDepositWitness in a Box to reduce the size of `DepositFailedDetails` on the stack? If so, why not also box `DepositWitness`? I know `VaultDepositWitness` is larger, but the other struct isn't particularly small either. Just curious if we have a rule of thumb for things like this @dandanlen. Or alternatively, are we enforcing max size for the error enum like we do for runtime calls?

albert-llimos: Fwiw the reason I used a Box was just clippy's advise due to the usual "large size difference between variants".

PR #5480, opened 2024-12-09
chore: Solana skip preflight only for CCM

albert-llimos:
# Pull Request

## Checklist

## Summary

Following the approach of EVM, where the transaction specifies some of the broadcasting parameters (`gas_limit`, `max_fee_per_gas` ...), I've added the `skip_preflight` to the Solana transactions.
For CCM transactions can revert so we do want to force broadcast it. For non-CCM transactions we expect them to succeed so we'd rather broadcast them only if they succeed (aka not skip preflight).


engine/src/sol/retry_rpc.rs
```diff
@@ -241,7 +252,7 @@ impl SolRetryRpcApi for SolRetryRpcClient {
 			.request_with_limit(
 				RequestLog::new(
 					"sendTransaction".to_string(),
-					Some(format!("{:?}, {:?}", transaction, config)),
+					Some(format!("{:?}, {:?}", transaction.serialized_transaction, config)),
```
dandanlen: While we're changing this anyway - can we make sure this is formatted as hex?

dandanlen (commented):
Nice - I made a couple of small changes (see commits).

dandanlen (approved):

PR #5481, opened 2024-12-10
feat: vault swap broker-api / rpc call

syan095:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

Tidied up the vault swap RPC call + broker api call so it is more compatible with other chains.
Added "extra_parameter" that contains chain-specific parameters so it is configurable per chain.
Added support for Solana assets for Vault Swap call encoding

### DIFF
--- a/state-chain/chains/src/lib.rs
+++ b/state-chain/chains/src/lib.rs
@@ -20,7 +20,9 @@ use address::{
 	IntoForeignChainAddress, ToHumanreadableAddress,
 };
 use cf_amm_math::Price;
-use cf_primitives::{AssetAmount, BroadcastId, ChannelId, EgressId, EthAmount, GasAmount, TxId};
+use cf_primitives::{
+	AssetAmount, BlockNumber, BroadcastId, ChannelId, EgressId, EthAmount, GasAmount, TxId,
+};
 use codec::{Decode, Encode, FullCodec, MaxEncodedLen};
 use frame_support::{
 	pallet_prelude::{MaybeSerializeDeserialize, Member, RuntimeDebug},
@@ -64,6 +66,7 @@ use cf_primitives::chains::assets::any::GetChainAssetMap;
 pub use deposit_channel::*;
 use strum::IntoEnumIterator;
 pub mod ccm_checker;
+pub mod cf_parameters;
 pub mod instances;
 
 pub mod mocks;
@@ -847,6 +850,7 @@ impl RetryPolicy for DefaultRetryPolicy {
 	}
 }
 
+/// Refund parameter used within the swapping pallet.
 #[derive(
 	Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen, Serialize, Deserialize,
 )]
@@ -885,7 +889,8 @@ impl<A: BenchmarkValue> BenchmarkValue for ChannelRefundParametersGeneric<A> {
 		}
 	}
 }
-
+#[cfg(feature = "std")]
+pub type RefundParameters = ChannelRefundParametersGeneric<crate::address::AddressString>;
 pub type ChannelRefundParameters = ChannelRefundParametersGeneric<ForeignChainAddress>;
 pub type ChannelRefundParametersEncoded = ChannelRefundParametersGeneric<EncodedAddress>;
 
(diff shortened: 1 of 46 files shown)


state-chain/chains/src/lib.rs
```diff
@@ -977,3 +982,116 @@ pub trait DepositDetailsToTransactionInId<C: ChainCrypto> {
...
+		from: Address,
+		event_data_account: Address,
+		input_amount: Number,
+		refund_parameters: ChannelRefundParametersGeneric<Address>,
+	},
+	SolUsdc {
+		from: Address,
+		from_token_account: Address,
+		event_data_account: Address,
+		input_amount: Number,
```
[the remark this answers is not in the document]
syan095: Yeah its because for RPC we use NumberOrHex type, but internally we use AssetAmount.
Having the Number as generic allow the same type to be used everywhere

syan095:
I've rebased this branch onto the latest main. 
I think we should merge this. I don't think we gain anything by waiting for EVM vault swap work to finish before merging.

state-chain/runtime/src/chainflip/vault_swap.rs
```diff
@@ -0,0 +1,232 @@
...
+	dca_parameters: Option<DcaParameters>,
+) -> Result<VaultSwapDetails<String>, DispatchErrorWithMessage> {
+	// Load up environment variables.
+	let api_environment = SolEnvironment::api_environment()
+		.map_err(|_| DispatchErrorWithMessage::from("Failed to load Solana API environment"))?;
+
+	let agg_key = SolEnvironment::current_agg_key()
+		.map_err(|_| DispatchErrorWithMessage::from("Failed to load Solana Agg key"))?
+		.into();
+
```
albert-llimos, line 130: I was talking to @martin-chainflip and he was raising the concern that during a period of time on a rotation we might be encoding transactions that will be invalid. Those transacitons will revert on chain and there is no risk of user losing funds either. Worst case scenario the user transaction reverts (basically zero cost), and that's why the API will also return the lifetime of that transaction per say, but it will be impossible for integrators to craft a valid transaction.
The period of time would be after a new key is created and before the rotation is finished and the rotation transaction is included on chain. The biggest risk there is some engines failing to broadcast. It's safe to say that it could easily go up to minutes.
I thought about it and I don't think there is any way we can reliably craft valid transactions all the time, even if we were to change the programs. However, one thing that could be done (if it's even possible) to mitigate this is to use the `SolanaBroadcaster::CurrentOnChainKey` instead of the environment one. In that way we would decrease that time to the finality of Solana (12 sec). Wdyt, is it possible? @syan095 @dandanlen

syan095: The change would be easy to implement. 

Do we want to this this for Vault swaps only? or for all Solana transactions we craft? 
i.e. currently we have lookup function to get the current Aggkey:
`let epoch = SolanaThresholdSigner::current_key_epoch()?;
SolanaThresholdSigner::keys(epoch)`

I can replace this to use `SolanaBroadcaster::CurrentOnChainKey` instead.

Alternatively I can add one more function that returns the on chain key, and only use this for VaultSwap transactions, so nothing else is affected.

syan095:
Rebased this onto the latest main. This branch is now in a mergable state

state-chain/runtime/src/chainflip/vault_swap.rs
syan095: >We will also return the expiry of that transaction. 

Do we want to return the expiry (end of current epoch I assume?) block as part of the response?
Currently the response only contain the instruction. We can easily add the expiry block.

albert-llimos: We can do the expiry separately, there is a ticket already: https://linear.app/chainflip/issue/PRO-1851/add-expires-at-field-for-request-swap-parameter-encoding

albert-llimos: > Do we want to this this for Vault swaps only? or for all Solana transactions we craft?
If we do this change we should only do it for the Solana Vault swaps. Otherwise there'll be scenarios where the aggKey used in the payload will be different to the one used to sign and that is problematic.

> The change would be easy to implement.
In that case I think this is the cleanest solution we have. Estimating the expiry time during a rotation can't really be done in an exact way but at least there will only be ~12 sec every 3 days of crafted payloads that will not be valid and there is no risk of losing funds.

albert-llimos: Resolving as this is being addressed in https://github.com/chainflip-io/chainflip-backend/pull/5512

dandanlen:
Just FYI - I'm being a bit slow but still reviewing this.

dandanlen:
I made some small change to the structure, happy to merge this now 👌

dandanlen (approved):

albert-llimos:
For visibility, we are ready to merge this but the latest PR/commits merged here are unverified so it won't be merged automatically. Dan/Roy can fix it on Monday.

PR #5482, opened 2024-12-10
feat: vault swap rpc support for Solana

syan095:
# Pull Request

Closes: PRO-1650

## Checklist

## Summary

Added support for Solana Vault Swap rpc call.
Added SolanaInstructionBuilder for building Vault swap calls of the Solana chains.


albert-llimos:
Regarding the tests you added, unfortunately it's a bit complex to try that the encoded data matches what would work in a localnet due to blockhashes and other small stuff. All this will be thoroughly tested when the SDK uses it in the bouncer and also the crafting of the instruction is enforced by the macro.
Therefore I think it's fine to have these tests as they are right now to check that the encoding doesn't change without us noticing. We can reassess after we plug in the SDK if we face any issues.

state-chain/runtime/src/chainflip/vault_swap.rs
```diff
@@ -109,3 +119,134 @@ pub fn bitcoin_vault_swap(
...
+	broker_commission: BasisPoints,
+	extra_parameters: VaultSwapExtraParametersEncoded,
+	channel_metadata: Option<CcmChannelMetadata>,
+	boost_fee: BasisPoints,
+	affiliate_fees: Affiliates<AccountId>,
+	dca_parameters: Option<DcaParameters>,
+) -> Result<VaultSwapDetails<String>, DispatchErrorWithMessage> {
+	// Load up environment variables.
+	let api_environment = SolEnvironment::api_environment().map_err(|_| {
+		DispatchErrorWithMessage::Other("Failed to load Solana API environment".into())
```
dandanlen, line 137: Avoid using `DispatchErrorWithMessage::Other("Error string".into())`: The string gets converted to `DispatchError::Other`, which erases the string... 

(See #5485, maybe copy the changes from there)

utilities/src/with_std.rs
```diff
@@ -109,6 +109,10 @@ pub fn make_periodic_tick(duration: Duration, yield_immediately: bool) -> tokio:
 	interval
 }
 
+pub fn bs58_address<const S: usize>(b: [u8; S]) -> String {
```
dandanlen, line 112: nit: it's not necessarily an address, it can convert an array of any size.

state-chain/runtime/src/chainflip/vault_swap.rs
syan095: I've cherry-picked the change into this feature branch and made the appropriate changes.

albert-llimos (approved):
LGTM now. We can wait for Dan to do a last check.

PR #5483, opened 2024-12-10
Feat: add boost_fee and broker_fees to events

msgmaxim:
# Pull Request

Closes: PRO-1824

## Summary

- Added `max_boost_fee_bps` field to both `DepositFinalised` and `DepositBoosted` events

- Added `broker_fees` field to the `SwapRequested` event FYI @niklasnatter 

- While updating tests, I noticed some tests used unnecessary annotations like `crate::Event::<Test, Instance1>` where `Event` would suffice, so I cleaned up the tests where this applies. This change is done in a separate commit.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -2153,7 +2153,6 @@ pub mod pallet {
 				dca_params
 			});
 
-			// MAXIM: explose: broker/affiliate fees, boost_fee_bps
 			Self::deposit_event(Event::<T>::SwapRequested {
 				swap_request_id: request_id,
 				input_asset,
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -564,6 +564,7 @@ pub mod pallet {
 			output_asset: Asset,
 			origin: SwapOrigin,
 			request_type: SwapRequestTypeEncoded,
+			broker_fees: Beneficiaries<T::AccountId>,
 			refund_parameters: Option<ChannelRefundParametersEncoded>,
 			dca_parameters: Option<DcaParameters>,
 		},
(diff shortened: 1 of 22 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2152,6 +2153,7 @@ pub mod pallet {
 				dca_params
 			});
 
+			// MAXIM: explose: broker/affiliate fees, boost_fee_bps
```
dandanlen, line 2156: Stale comment?

dandanlen (approved):
LGTM - although not quite sure why we call it `max_boost_fee`?

niklasnatter:
> LGTM - although not quite sure why we call it `max_boost_fee`?

I like the name as it is the maximum bps a user was willing to pay and the `DepositBoosted` also has a `boost_fee` which is the effective boost fee that was actually paid.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -564,6 +564,7 @@ pub mod pallet {
 			output_asset: Asset,
 			origin: SwapOrigin,
 			request_type: SwapRequestTypeEncoded,
+			broker_fees: Beneficiaries<T::AccountId>,
```
niklasnatter, line 567: just to be safe - we can be sure that the first entry of this array will always be the main broker?

CumpsD:
> > LGTM - although not quite sure why we call it `max_boost_fee`?
> 
> I like the name as it is the maximum bps a user was willing to pay and the `DepositBoosted` also has a `boost_fee` which is the effective boost fee that was actually paid.

Totally agree, it was one of the remarks on my Shapeshift PR too, I had `boostFee` and they suggested to rename to `maxBoostFee` because that is what it really is

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Yes, it will always be the main broker

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: Removed 👍

PR #5484, opened 2024-12-10
chore: prepare localnet for backspin usage 💾

ahasna:


localnet/manage.sh
```diff
@@ -47,6 +47,7 @@ main() {
         npm install -g wscat
     fi
     if [ -z $CI ]; then
+      export START_TRACKER="y"
```
anton-chain, line 50: is this part related to the PR? looks like $CI is always empty..

anton-chain (approved):

PR #5485, opened 2024-12-10
fix: expose error detail when swap simulation fails

dandanlen:
This fixes an issue observed during the recent upgrade, when liquidity is low. Error messages were being erased due to the implementation of DispatchError.


syan095:
Is there still value in keepping the dispatcherror::Others variant? I believe that the original intent was for a string message to be passed through RPC back to the user. Maybe we should just replace the Other with your new type.

syan095:
Also a nitpick - the word "adhoc" isn't very clear on its intended purpose, or the difference between this and the other 2 variant. Maybe something like "message" or "string" or "utf8" would be clearer?

syan095:
> Is there still value in keepping the dispatcherror::Others variant?

I suppose the variant is just there to blanket cover all the other cases (non-module, non-str)

syan095 (approved):
Nice. I'm already using this in the Vault swap branch!

PR #5487, opened 2024-12-10
Feat: add exclude fee quotes for Vault swaps

albert-llimos:
# Pull Request

Closes: PRO-1892

## Checklist

## Summary

Implemented on top of the ccm gas improvement since it's adding a new parameter to the already changed functions. No need to to the same versioning update in a separate PR.

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -63,7 +63,7 @@ use state_chain_runtime::{
 	Block, Hash, NetworkFee, SolanaInstance,
 };
 use std::{
-	collections::{BTreeMap, HashMap},
+	collections::{BTreeMap, BTreeSet, HashMap},
 	marker::PhantomData,
 	sync::Arc,
 };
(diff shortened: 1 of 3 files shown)


dandanlen (commented):
Approach LGTM but I'd like to take a closer look at the runtime api changes - looks like a breaking change in the runtime api. If/How we need to deal with this depends on how it compares to the current release version. 

@albert-llimos if it's ok i'll take a look later and will update the PR directly.

albert-llimos:
> Approach LGTM but I'd like to take a closer look at the runtime api changes - looks like a breaking change in the runtime api. If/How we need to deal with this depends on how it compares to the current release version.
> 
> @albert-llimos if it's ok i'll take a look later and will update the PR directly.

Notice that the change adding `exclude_fees` is on the funcitons that the PR underneath has also added `ccm_data`. If you have reviewed this I'd suggest merging it to the main PR and then review those api changes all together. Are you ok with that @dandanlen ?

dandanlen:
Let me make some small changes first.

dandanlen:
@albert-llimos all done ☝️ 

Feel free to merge unless you have any objections to the changes. 

- I felt including the broker fee in the enum was not necessary since it's implicit in the amount: if amount is greater than zero that means you want to include it. 
- Using a btreeSet is helpful since it already deduplicates the values. 
- Using an option in the runtime api doesn't add anything (in contrast with the json api where using an Option<> makes the field optional, SCALE encoding still requires a `None`)
- Factored out the !exluded checks into a closure `included`

albert-llimos:
LGTM, thanks!

PR #5488, opened 2024-12-11
Divide logic between enum variants

kylezs:


state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser.rs
```diff
@@ -172,123 +169,104 @@ impl<
...
+				// Delete any elections that are ongoing for any blocks in the reorg range.
+				for (i, election_identifier) in election_identifiers.into_iter().enumerate() {
+					let election = ElectoralAccess::election_mut(election_identifier);
+					let properties = election.properties()?;
+					if properties.0.into_range_inclusive() == *reorg_range {
+						election.delete();
+						open_elections = open_elections.saturating_sub(1);
+						remaining_election_identifiers.remove(i);
+					}
+				}
```
MxmUrw, line 192: Actually, when I detect a reorg, I have two ranges: `removed` and `added`. Depending on circumstances either one could be "longer" then the other. `removed` are the headers which I have previously witnessed and forwarded to you and are now invalid due to the reorg, `added` are the replacement headers now valid after the reorg.

It seems to me that since you want to delete the previously created elections, and recreate elections for the new blocks, you actually want to get both ranges. So I propose to change the variant to be `ChainProgress::Reorg { removed_range: InclusiveRange<>, added_range: InclusiveRange<> }`

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser.rs
```diff
@@ -172,123 +169,104 @@ impl<
...
 
-					open_elections = open_elections.saturating_sub(1);
+				// NB: We do not clear any of the unprocessed data here. This is because we need to
+				// prevent double dispatches. By keeping the state, if we have a reorg we can check
+				// against the state in the process_block_data hook to ensure we don't double
+				// dispatch.
+				*reorg_range.end()
+			},
+			ChainProgress::None(last_block_root_seen) => *last_block_root_seen,
+			ChainProgress::Continuous(witness_range) => *witness_range.start(),
```
MxmUrw, line 218: Are you sure you want `.start()` here? You assign this to last seen root, so shouldn't this also be `.end()`? Of course this element should be a "witness root".

kylezs: yeah should be end 👍

PR #5489, opened 2024-12-11
feat: Add private channel address to account info rpc

Janislav:
# Pull Request

Closes: PRO-1862

## Checklist

## Summary

Adds the deposit address of a private broker channel to the response of cf_account_info.

#### Non-Breaking changes

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -252,7 +252,7 @@ pub enum RpcAccountInfo {
 	Broker {
 		flip_balance: NumberOrHex,
 		earned_fees: any::AssetMap<NumberOrHex>,
-		channel_address: Option<ForeignChainAddressHumanreadable>,
+		channel_address: Option<String>,
 	},
 	LiquidityProvider {
 		balances: any::AssetMap<NumberOrHex>,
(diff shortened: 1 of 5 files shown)


state-chain/runtime/src/lib.rs
```diff
@@ -1858,11 +1859,32 @@ impl_runtime_apis! {
...
+			let earned_fees: Vec<_> = Asset::all().map(|asset|
 				(asset, AssetBalances::get_balance(&account_id, asset))
 			).collect();
-
-			BrokerInfo { earned_fees }
+			if let Some(channel_id) = pallet_cf_swapping::BrokerPrivateBtcChannels::<Runtime>::get(&account_id) {
+				let EpochKey { key, .. } = BitcoinThresholdSigner::active_epoch_key().expect("We should always have a key for the current epoch.");
+				BrokerInfo {
+					earned_fees,
+					deposit_address: if let Ok(salt) = channel_id.try_into() {
```
Janislav, line 1869: I thought it is acceptable to return None if the conversion should fail here, since this should not happen anyway. We can also return an error in this case, but I don't know how valuable this is for the consumer of the API, and this means more overhead.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -252,6 +252,7 @@ pub enum RpcAccountInfo {
 	Broker {
 		flip_balance: NumberOrHex,
 		earned_fees: any::AssetMap<NumberOrHex>,
+		deposit_address: Option<String>,
```
dandanlen, line 255: Let's make this clear and unambiguous: 

```suggestion
		btc_vault_deposit_address: Option<String>,
```

state-chain/runtime/src/lib.rs
dandanlen: Hmm.. Not sure if I agree. If this happens we definitely want to know about it instead of failing silently. I agree this should never happen though, so maybe this is a one of those cases where it's ok to `expect`? 

Btw. Maybe we can deduplicate this address-derivation code: this is very similar to what is in `cf_get_vault_swap_details`.

Janislav: Agree, I think expect makes sense.

Janislav:
Seems to be there is some stubborn bouncer flakiness... Let's try again and pray 🙏

PR #5495, opened 2024-12-12
Feat: Network Fee from Boost

msgmaxim:
# Pull Request

Closes: PRO-1778

## Summary

- New storage item: `NetworkFeeDeductionFromBoostPercents` representing the % of boost fee to be taken as network fee. Its value is initially 0, and can be updated via `update_pallet_config` governance extrinsic.
- When a deposit is boosted, the full boost fee is "deducted" from the deposit, but we now apply network fee reduction when recording the amounts to be credited to boosters upon finalisation.
- When a boosted deposit is finalised, we add up all funds credited to boosters, and any non-zero excees amount is swaped into FLIP as network fee. (The swap request's type being NetworkFee takes care of the resulting flip being scheduled for burning later.)
- `DepositFinalised` (when action is `BoostersCredited`) now additionally exposes `network_fee_swap_request_id` (of the network fee swap), and `network_fee_from_boost` (the amount of input asset to be swapped).
- Added `network_fee_deduction` to `cf_boost_pool_details` rpc.
- Added tests that should cover all new behaviour.

I haven't added network_fee_decution to `DepositBoosted` (there still seems to be an ongoing discussion whether this should be necessary), but it shouldn't be too difficult to add if needed (though maybe with a name indicating that it is not confirmed?).

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -49,7 +49,7 @@ use sp_api::{ApiError, ApiExt, CallApiAt};
 use sp_core::U256;
 use sp_runtime::{
 	traits::{Block as BlockT, Header as HeaderT, UniqueSaturatedInto},
-	Permill,
+	Percent, Permill,
 };
 use sp_state_machine::InspectState;
 use state_chain_runtime::{
@@ -565,6 +565,7 @@ mod boost_pool_rpc {
 		available_amounts: Vec<AccountAndAmount>,
 		deposits_pending_finalization: Vec<PendingBoost>,
 		pending_withdrawals: Vec<PendingWithdrawal>,
+		network_fee_deduction_percents: Percent,
 	}
 
 	impl BoostPoolDetailsRpc {
@@ -602,6 +603,7 @@ mod boost_pool_rpc {
 						pending_deposits,
 					})
 					.collect(),
+				network_fee_deduction_percents: details.network_fee_deduction_percents,
 			}
 		}
 	}
@@ -2422,6 +2424,7 @@ mod test {
 				(1, BTreeMap::from([(ID_1.clone(), OwedAmount { total: 1_000, fee: 50 })])),
 			]),
 			pending_withdrawals: Default::default(),
+			network_fee_deduction_percents: Percent::from_percent(40),
 		}
 	}
 
@@ -2439,6 +2442,7 @@ mod test {
 				(ID_1.clone(), BTreeSet::from([0])),
 				(ID_2.clone(), BTreeSet::from([0])),
 			]),
+			network_fee_deduction_percents: Percent::from_percent(0),
 		}
 	}
 
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -565,7 +565,7 @@ mod boost_pool_rpc {
 		available_amounts: Vec<AccountAndAmount>,
 		deposits_pending_finalization: Vec<PendingBoost>,
 		pending_withdrawals: Vec<PendingWithdrawal>,
-		network_fee_deduction_percents: Percent,
+		network_fee_deduction_percent: Percent,
 	}
 
 	impl BoostPoolDetailsRpc {
@@ -603,7 +603,7 @@ mod boost_pool_rpc {
 						pending_deposits,
 					})
 					.collect(),
-				network_fee_deduction_percents: details.network_fee_deduction_percents,
+				network_fee_deduction_percent: details.network_fee_deduction_percent,
 			}
 		}
 	}
@@ -2424,7 +2424,7 @@ mod test {
 				(1, BTreeMap::from([(ID_1.clone(), OwedAmount { total: 1_000, fee: 50 })])),
 			]),
 			pending_withdrawals: Default::default(),
-			network_fee_deduction_percents: Percent::from_percent(40),
+			network_fee_deduction_percent: Percent::from_percent(40),
 		}
 	}
 
@@ -2442,7 +2442,7 @@ mod test {
 				(ID_1.clone(), BTreeSet::from([0])),
 				(ID_2.clone(), BTreeSet::from([0])),
 			]),
-			network_fee_deduction_percents: Percent::from_percent(0),
+			network_fee_deduction_percent: Percent::from_percent(0),
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


state-chain/custom-rpc/src/lib.rs
```diff
@@ -2422,6 +2424,7 @@ mod test {
 				(1, BTreeMap::from([(ID_1.clone(), OwedAmount { total: 1_000, fee: 50 })])),
 			]),
 			pending_withdrawals: Default::default(),
+			network_fee_deduction_percents: Percent::from_percent(40),
```
kylezs, line 2427: ?
```suggestion
			network_fee_deduction_percent: Percent::from_percent(40),
```

state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
```diff
@@ -240,7 +250,12 @@ where
 			(provided_amount, fee)
 		};
 
-		self.use_funds_for_boosting(prewitnessed_deposit_id, provided_amount, fee_amount)?;
+		// NOTE: only a portion of the boost fee goes to the boost pool,
```
kylezs, line 253: nit: the comment reads the other way around to the code. in the code below we're taking the network fee first

state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
```diff
@@ -146,6 +146,15 @@ pub struct BoostPool<AccountId, C: Chain> {
 	pending_withdrawals: BTreeMap<AccountId, BTreeSet<PrewitnessedDepositId>>,
 }
 
+#[derive(DefaultNoBound, DebugNoBound, PartialEqNoBound)]
+pub struct DepositFinalisationOutcomeForPool<AccountId, C: Chain>
+where
+	AccountId: PartialEq + core::fmt::Debug,
+{
+	pub unlocked_funds: Vec<(AccountId, C::ChainAmount)>,
+	pub amount_credited: C::ChainAmount,
```
kylezs, line 155: Initially I assumed this included crediting the network fee to the protocol, before reading the rest of the code, since it didn't seem specific to boosters
```suggestion
	pub amount_credited_to_boosters: C::ChainAmount,
```

state-chain/pallets/cf-ingress-egress/src/boost_pool/tests.rs
```diff
@@ -160,19 +162,82 @@ fn boosting_with_fees() {
...
+	const BOOSTER_1_FEE: u128 = 2;
+	const BOOSTER_2_FEE: u128 = 4;
+
+	const TOTAL_BOOSTERS_FEE: u128 = BOOSTER_1_FEE + BOOSTER_2_FEE;
+
+	const NETWORK_FEE_FROM_BOOST: u128 = FULL_BOOST_FEE * NETWORK_FEE_PORTION_PERCENT as u128 / 100;
+
+	// NOTE: we subtract 1 to account for a rounding "error" (in real code any remaining fee will be
+	// used as network fee, so all atomic units will be accounted for)
+	assert_eq!(TOTAL_BOOSTERS_FEE, FULL_BOOST_FEE - NETWORK_FEE_FROM_BOOST - 1);
```
kylezs, line 216: Imo, this is a little strange that we're testing something that's explicitly stated can't happen in the real code. Seems like we could write this test in such a way that demonstrates that the statement in the comment is true, maybe capturing init_swap_request's input amount?

state-chain/pallets/cf-ingress-egress/src/boost_pool/tests.rs
```diff
@@ -160,19 +162,82 @@ fn boosting_with_fees() {
...
+	);
 
 	check_pool(&pool, [(BOOSTER_1, 1003), (BOOSTER_2, 2006)]);
 }
 
+#[test]
+fn boosting_with_fees_including_network_fee_portion() {
+	const NETWORK_FEE_PORTION_PERCENT: u8 = 30;
+
+	let mut pool = TestPool::new(100); // 1%
```
kylezs, line 185: Can use a const for this 100 and then use it to derive the FULL_BOOST_FEE

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -870,6 +897,9 @@ pub mod pallet {
 			broker_id: T::AccountId,
 			short_affiliate_id: AffiliateShortId,
 		},
+		NetworkFeeDeductionFromBoostSet {
+			deduction_percents: Percent,
```
kylezs, line 901: same here
```suggestion
			deduction_percent: Percent,
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2302,14 +2345,18 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 			for boost_tier in used_pools {
 				BoostPools::<T, I>::mutate(asset, boost_tier, |maybe_pool| {
 					if let Some(pool) = maybe_pool {
-						for (booster_id, finalised_withdrawn_amount) in
-							pool.process_deposit_as_finalised(prewitnessed_deposit_id)
-						{
+						let DepositFinalisationOutcomeForPool { unlocked_funds, amount_credited } =
+							pool.process_deposit_as_finalised(prewitnessed_deposit_id);
+
+						total_amount_credited += amount_credited;
```
kylezs, line 2357: nit: saturating? just to be defensive (though, it would be quite bad if we even got close to saturating here...)

state-chain/pallets/cf-ingress-egress/src/boost_pool/tests.rs
```diff
@@ -160,19 +162,82 @@ fn boosting_with_fees() {
...
+		pool.process_deposit_as_finalised(BOOST_1),
+		DepositFinalisationOutcomeForPool { amount_credited: 1010, unlocked_funds: vec![] }
+	);
 
 	check_pool(&pool, [(BOOSTER_1, 1003), (BOOSTER_2, 2006)]);
 }
 
+#[test]
+fn boosting_with_fees_including_network_fee_portion() {
+	const NETWORK_FEE_PORTION_PERCENT: u8 = 30;
```
kylezs, line 183: Worth testing the extreme case of 100% too, just to see it behaves rationally.
(0 is tested in some of the other tests already).

kylezs (commented):
Nice one 👌 only pretty minor things

dandanlen (commented):
👌

kylezs (approved):

PR #5500, opened 2024-12-12
chore: bump old CFE version to 1.7.5

kylezs:


dandanlen (approved):

dandanlen:
FYI I think this is failing because of another issue with pallet versions that @ramizhasan111 is fixing.

PR #5503, opened 2024-12-13
chore: increase ccm message max to 15k

albert-llimos:
# Pull Request

## Checklist

## Summary

Increase allowed ccm length to 15k.
While testing with higher payloads I've encountered some issues:
- In localnet payloads bigger than several hundred bytes end up with gas limits higher than Arbitrum block gas limit so we need to limit the size. Pretty sure this is due to localnet settings, this won't happen in a real network.
- Vault swaps, particularly in Solana, run into issues due to transaction size limitations. I've added a the corresponding limitations under the `newVaultSwapCcmMetadata` function.
- Arbitrum gas proved extremely unreliable. While this is just a localnet issue, it made me rethink how estimations should be done. Part of the gas estimation that takes into account Arbitrum's gas limit variability should be done for the user so only the CF vault overhead should be taken into consideration when estimating the Arbitrum's gas. Then in the bouncer I have added a gas estimation that is what the integrators should be doing.

### DIFF
--- a/bouncer/shared/evm_vault_swap.ts
+++ b/bouncer/shared/evm_vault_swap.ts
@@ -75,7 +75,7 @@ export async function executeEvmVaultSwap(
   } as const;
   const txOptions = {
     // This is run with fresh addresses to prevent nonce issues. Will be 1 for ERC20s.
-    gasLimit: srcChain === Chains.Arbitrum ? 100000000n : 5000000n,
+    gasLimit: srcChain === Chains.Arbitrum ? 32000000n : 5000000n,
   } as const;
 
   const receipt = await executeSwap(
--- a/bouncer/shared/evm_vault_swap.ts
+++ b/bouncer/shared/evm_vault_swap.ts
@@ -75,7 +75,7 @@ export async function executeEvmVaultSwap(
   } as const;
   const txOptions = {
     // This is run with fresh addresses to prevent nonce issues. Will be 1 for ERC20s.
-    gasLimit: srcChain === Chains.Arbitrum ? 10000000n : 200000n,
+    gasLimit: srcChain === Chains.Arbitrum ? 50000000n : 1000000n,
   } as const;
 
   const receipt = await executeSwap(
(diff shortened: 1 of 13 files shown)


albert-llimos:
It's been pointed out that since this value is included in extrinsics and events... I think it's possible (likely) that if there's an event with a too-large payload then it would fail to decode.

Putting it back to draft because I might want to test locally that very large messages don't cause any issue

bouncer/shared/evm_vault_swap.ts
```diff
@@ -75,7 +75,7 @@ export async function executeEvmVaultSwap(
   } as const;
   const txOptions = {
     // This is run with fresh addresses to prevent nonce issues. Will be 1 for ERC20s.
-    gasLimit: srcChain === Chains.Arbitrum ? 10000000n : 200000n,
+    gasLimit: srcChain === Chains.Arbitrum ? 32000000n : 5000000n,
```
albert-llimos, line 78: Arbitrum's "soft" gas limit is 32M. Increased Eth's gas limit because larger messages need more gas.

bouncer/shared/swapping.ts
```diff
@@ -15,56 +14,8 @@ import { BtcAddressType } from '../shared/new_btc_address';
...
-        variables.push(randomAsHex(20));
-        break;
-      // Add more cases for other Solidity types as needed
-      default:
-        throw new Error(`Unsupported Solidity type: ${typesArray[i]}`);
-    }
-  }
-  return web3.eth.abi.encodeParameters(typesArray, variables);
-}
-
```
albert-llimos, line 67: I originally had this to check that abi encoded messages work but it's actually better to have generalize to any arbitrary bytes for this end to end testing.

engine/src/evm/retry_rpc.rs
```diff
@@ -380,7 +380,9 @@ impl<Rpc: EvmSigningRpcApi> EvmRetrySigningRpcApi for EvmRetryRpcClient<Rpc> {
 							Some(gas_limit) =>
 								if estimated_gas > gas_limit {
 									return Err(anyhow::anyhow!(
-										"Estimated gas is greater than the gas limit"
+										"Estimated gas ({}) is greater than the gas limit ({})",
+										estimated_gas,
+										gas_limit
```
albert-llimos, line 385: I added this for debugging and I think it's useful to have when CCM starts being used. I assume even if we release this change by itself we need some engine update so making this change in the engine doesn't introduce any new overhead.

dandanlen (approved):

PR #5504, opened 2024-12-13
fix: lp_cancel_orders_batch to accepts OrderId as string/hex

marcellorigotti:
# Pull Request

Closes: PRO-1897

## Checklist

## Summary

Updated the LP-API cancel_orders_batch to accepts OrderId as string/hex as well 

#### Non-Breaking changes

### DIFF
--- a/api/bin/chainflip-lp-api/src/main.rs
+++ b/api/bin/chainflip-lp-api/src/main.rs
@@ -101,6 +101,7 @@ pub mod rpc_types {
 	}
 
 	#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+	#[serde(untagged)]
 	pub enum CloseOrderJson {
 		Limit { base_asset: any::Asset, quote_asset: any::Asset, side: Side, id: OrderIdJson },
 		Range { base_asset: any::Asset, quote_asset: any::Asset, id: OrderIdJson },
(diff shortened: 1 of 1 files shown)


kylezs (approved):

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -601,13 +622,20 @@ impl RpcServer for RpcServerImpl {
...
+		orders: BoundedVec<CloseOrderJson, ConstU32<MAX_ORDERS_DELETE>>,
 		wait_for: Option<WaitFor>,
 	) -> RpcResult<ApiWaitForResult<Vec<LimitOrRangeOrder>>> {
+		let mut final_orders: BoundedVec<CloseOrder, ConstU32<MAX_ORDERS_DELETE>> =
+			BoundedVec::new();
+		for order in orders {
+			final_orders
+				.try_push(order.try_into()?)
+				.expect("Impossible to fail, given the same MAX_ORDERS_DELETE");
+		}
```
kylezs, line 634: can remove the mut by collecting
```suggestion
Ok(self
			.api
			.lp_api()
			.cancel_orders_batch(
				orders
					.into_iter()
					.map(TryInto::try_into)
					.collect::<Result<Vec<_>, _>>()?
					.try_into()
					.expect("Impossible to fail, given the same MAX_ORDERS_DELETE"),
				wait_for.unwrap_or_default(),
			)
			.await?)
```

marcellorigotti: Doesn't this require an extra memory allocation for the `Vec<>` when collecting?
(or maybe this gets optimized away by the compiler?)

kylezs: Yes, I believe so. But given the performance gain is negligible, I think having without a mut var is a bit simpler and less likely to be mistakenly broken with future changes (definitely isn't a strong opinion which way is preferred here)

kylezs (approved):

kylezs (approved):

PR #5506, opened 2024-12-16
feat: add expires_at field to request swap parameter encoding

nmammeri:
# Pull Request

Closes: PRO-1851

## Checklist

## Summary

* Return `expires_at` field determining when the payload is considered expired.
* Only implemented for Bitcoin for now, Solana implementation should be straightforward when Solana vault swapping is in main
* Updated bouncer btc_vault_swap test to validate `expires_at`


kylezs:
See my comment on the linear issue, I think having an RPC endpoint will be much more robust and error prone, and probably best not to have integrators depend on something we know will change.

dandanlen (approved):

PR #5507, opened 2024-12-17
refactor: vault swap refund instead of ignoring the deposit

Janislav:
# Pull Request

Closes: PRO-1804, PRO-2139

## Checklist

## Summary

Adding refund functionality if there is no broker provided in the vault swap. If we identify on the prewitness stage that there is no broker provided or the account provided is no broker, we mark the tx as aborted. When the deposit gets eventually fully witnessed, we check if it was identified as aborted and directly try to refund it. If the refund should not be possible (failing to create the API call for e.x) we push the details of the tx to another storage item to remember the tx.

#### Non-Breaking changes

## Summary by CodeRabbit

- **New Features**
  - Introduced a refund mechanism for vault swaps to ensure users receive prompt recovery for aborted transactions.
  - Added stage tracking to provide clearer insight into the progress of vault swap processes.
- **Enhancements**
  - Improved fee processing by assigning a dedicated fee recipient for more accurate fee management.
- **Tests**
  - Updated automation to validate the new refund and fee handling capabilities, ensuring reliable swap transactions.

### DIFF
--- a/state-chain/cf-integration-tests/src/solana.rs
+++ b/state-chain/cf-integration-tests/src/solana.rs
@@ -33,7 +33,7 @@ use cf_chains::{
 	ExecutexSwapAndCallError, ForeignChainAddress, RequiresSignatureRefresh, SetAggKeyWithAggKey,
 	SetAggKeyWithAggKeyError, Solana, SwapOrigin, TransactionBuilder,
 };
-use cf_primitives::{AccountRole, AuthorityCount, ForeignChain, SwapRequestId};
+use cf_primitives::{AccountRole, AuthorityCount, Beneficiary, ForeignChain, SwapRequestId};
 use cf_test_utilities::{assert_events_match, assert_has_matching_event};
 use cf_utilities::{assert_matches, bs58_array};
 use codec::Encode;
@@ -46,7 +46,7 @@ use pallet_cf_elections::{
 	AuthorityVoteOf, ElectionIdentifierOf, MAXIMUM_VOTES_PER_EXTRINSIC,
 };
 use pallet_cf_ingress_egress::{
-	DepositFailedReason, DepositWitness, FetchOrTransfer, VaultDepositWitness,
+	DepositAction, DepositWitness, FetchOrTransfer, RefundReason, VaultDepositWitness,
 };
 use pallet_cf_validator::RotationPhase;
 use sp_core::ConstU32;
@@ -385,7 +385,7 @@ fn vault_swap_deposit_witness(
 		deposit_metadata,
 		tx_id: Default::default(),
 		deposit_details: (),
-		broker_fee: None,
+		broker_fee: Some(Beneficiary { account: BROKER.into(), bps: 0 }),
 		affiliate_fees: Default::default(),
 		refund_params: ChannelRefundParameters {
 			retry_duration: REFUND_PARAMS.retry_duration,
@@ -483,8 +483,8 @@ fn solana_ccm_fails_with_invalid_input() {
 				RuntimeEvent::SolanaIngressEgress(pallet_cf_ingress_egress::Event::<
 					Runtime,
 					SolanaInstance,
-				>::DepositFailed {
-					reason: DepositFailedReason::CcmInvalidMetadata,
+				>::DepositFinalised {
+					action: DepositAction::Refund { reason: RefundReason::CcmInvalidMetadata, .. },
 					..
 				}),
 			);
@@ -536,7 +536,7 @@ fn solana_ccm_fails_with_invalid_input() {
 					Runtime,
 					SolanaInstance,
 				>::CcmEgressInvalid {
-					egress_id: (ForeignChain::Solana, 1u64),
+					egress_id: (ForeignChain::Solana, 2u64),
 					error: ExecutexSwapAndCallError::FailedToBuildCcmForSolana(
 						SolanaTransactionBuildingError::InvalidCcm(
 							CcmValidityError::CcmAdditionalDataContainsInvalidAccounts
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -169,7 +169,7 @@ enum FullWitnessDepositOutcome {
 }
 
 #[derive(RuntimeDebug, Clone)]
-pub struct ValidatedVaultSwap<AccountId> {
+pub struct ValidatedVaultSwapParams<AccountId> {
 	pub broker_fees: BoundedVec<Beneficiary<AccountId>, ConstU32<6>>,
 	pub channel_metadata: Option<CcmChannelMetadata>,
 	pub source_address: Option<ForeignChainAddress>,
@@ -2352,7 +2352,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			broker_fee.as_ref().map(|Beneficiary { account, .. }| account.clone()),
 		);
 
-		if let Ok(ValidatedVaultSwap {
+		if let Ok(ValidatedVaultSwapParams {
 			broker_fees,
 			channel_metadata,
 			source_address,
@@ -2589,7 +2589,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 	fn try_validate_vault_swap(
 		vault_deposit_witness: VaultDepositWitness<T, I>,
-	) -> Result<ValidatedVaultSwap<T::AccountId>, RefundReason> {
+	) -> Result<ValidatedVaultSwapParams<T::AccountId>, RefundReason> {
 		let VaultDepositWitness {
 			output_asset: destination_asset,
 			destination_address,
@@ -2604,7 +2604,6 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		let Some(broker_fees) =
 			Self::assemble_broker_fees(broker_fee.clone(), affiliate_fees.clone())
 		else {
-			log::warn!("Failed to process vault swap due to invalid broker fees");
 			return Err(RefundReason::InvalidBrokerFees);
 		};
 
@@ -2620,7 +2619,6 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			)
 			.is_err()
 			{
-				log::warn!("Failed to process vault swap due to invalid CCM metadata");
 				return Err(RefundReason::CcmInvalidMetadata);
 			}
 
@@ -2637,13 +2635,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		if let Err(_err) =
 			T::SwapLimitsProvider::validate_refund_params(refund_params.retry_duration)
 		{
-			log::warn!("Failed to process vault swap due to invalid refund parameters");
 			return Err(RefundReason::InvalidRefundParameters);
 		}
 
 		if let Some(params) = &dca_params {
 			if T::SwapLimitsProvider::validate_dca_params(params).is_err() {
-				log::warn!("Failed to process vault swap due to invalid DCA parameters");
 				return Err(RefundReason::InvalidDcaParameters);
 			}
 		}
@@ -2655,12 +2651,11 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			) {
 				Ok(address) => address,
 				Err(_) => {
-					log::warn!("Failed to process vault swap due to invalid destination address");
 					return Err(RefundReason::InvalidDestinationAddress);
 				},
 			};
 
-		Ok(ValidatedVaultSwap {
+		Ok(ValidatedVaultSwapParams {
 			broker_fees,
 			channel_metadata,
 			source_address,
@@ -2689,7 +2684,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		let (action, source_address) =
 			match Self::try_validate_vault_swap(vault_deposit_witness.clone()) {
-				Ok(ValidatedVaultSwap {
+				Ok(ValidatedVaultSwapParams {
 					broker_fees,
 					channel_metadata,
 					source_address,
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -151,7 +151,6 @@ pub enum DepositFailedReason {
 	NotEnoughToPayFees,
 	TransactionRejectedByBroker,
 	DepositWitnessRejected(DispatchError),
-	CanNotRejectRefunds,
 }
 
 #[derive(RuntimeDebug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo)]
@@ -2466,10 +2465,6 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 						// Collect to ensure that the iterator is fully consumed.
 						.collect::<Vec<_>>()
 						.is_empty();
-                    
-					if matches!(action, ChannelAction::Refund { .. }) {
-						return Err(DepositFailedReason::CanNotRejectRefunds);
-					}
 
 					if is_marked_by_broker_or_screening_id {
 						let refund_address = match &action {
@@ -2477,7 +2472,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 								refund_params.refund_address.clone(),
 							ChannelAction::LiquidityProvision { refund_address, .. } =>
 								refund_address.clone(),
-                            ChannelAction::Refund { .. } => unreachable!("Condition checked above!"),
+							ChannelAction::Refund { refund_address, .. } =>
+								refund_address.clone().into_foreign_chain_address(),
 						};
 
 						ScheduledTransactionsForRejection::<T, I>::append(
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -151,17 +151,30 @@ pub enum DepositFailedReason {
 	NotEnoughToPayFees,
 	TransactionRejectedByBroker,
 	DepositWitnessRejected(DispatchError),
-	InvalidDestinationAddress,
+	CanNotRejectRefunds,
+}
+
+#[derive(RuntimeDebug, Eq, PartialEq, Clone, Encode, Decode, TypeInfo)]
+pub enum RefundReason {
 	InvalidBrokerFees,
 	InvalidRefundParameters,
 	InvalidDcaParameters,
 	CcmUnsupportedForTargetChain,
 	CcmInvalidMetadata,
+	InvalidDestinationAddress,
 }
 
 enum FullWitnessDepositOutcome {
 	BoostFinalised,
 	DepositActionPerformed,
+}
+
+#[derive(RuntimeDebug, Clone)]
+pub struct ValidatedVaultSwap<AccountId> {
+	pub broker_fees: BoundedVec<Beneficiary<AccountId>, ConstU32<6>>,
+	pub channel_metadata: Option<CcmChannelMetadata>,
+	pub source_address: Option<ForeignChainAddress>,
+	pub destination_address: ForeignChainAddress,
 }
 
 mod deposit_origin {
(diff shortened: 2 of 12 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2161,6 +2193,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			boost_fee,
 		}: VaultDepositWitness<T, I>,
 	) {
+		if Self::should_reject_vault_swap(&broker_fee) {
```
dandanlen, line 2196: We only check this when we pre-witness. Why not on full deposits? What about chains where we don't have pre-witnessing - am I missing something?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -766,6 +766,18 @@ pub mod pallet {
...
+	/// Stores the tx_id of an aborted vault transaction. At the moment we consider a vault
+	/// transaction as aborted if
+	#[pallet::storage]
+	pub(crate) type AbortedVaultTransaction<T: Config<I>, I: 'static = ()> =
+		StorageMap<_, Identity, TransactionInIdFor<T, I>, ()>;
+
+	/// Stores the details of an aborted vault transaction. In case we fail to refund we store the
+	/// details in this storage item.
+	#[pallet::storage]
+	pub(crate) type AbortedVaultTransactionDetails<T: Config<I>, I: 'static = ()> =
```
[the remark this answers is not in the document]
Janislav: We can also unify it.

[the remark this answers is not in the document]
Janislav: EDIT: We can not unify it and it's better to keep it separate. We could even argue to just remove this and not save this, since this is a very theoretical case. After we enforced refund parameters, the only case a refund could fail is when fail to build an API call, but that should actually never happen.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Pretty sure we only prewitness for Bitcoin. Also, sometimes we miss the pre-witness (for example if there's a reorg).

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: Yes exactly - we don't need the map with a `()` value.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: Okay, but checking it only on full deposit is not an option since it could be a boosted BTC transaction. We have to check it on both stages but only refund it on full deposit

Janislav:
Debugging a decode/encode issue of the vault_swap parameter in test in bouncer right now.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2466,6 +2500,34 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+						},
+					) {
+					T::Broadcaster::threshold_sign_and_broadcast(api_call);
+					Self::deposit_event(Event::VaultSwapRefunded { tx_id: tx_id.clone() });
+				} else {
+					log_or_panic!("Failed to create refund api call for vault swap.");
+				}
+			};
+
+			Self::deposit_event(Event::VaultSwapIgnored {
```
dandanlen, line 2522: It's a bit strange that we say it was ignored even though we just refunded it. We could just use the already-existing `DepositFailed` event? 

Then we can also attach some information about *why* it failed (ie `DepositFailedReason::InvalidBroker`).

This will require fewer changes to the product.

dandanlen: In fact, we could maybe do the refunding in the error handling of `emit_deposit_failed_event`? This way we would attempt to refund any failed vault swaps (of course it might depend on the failure reason).

state-chain/chains/src/cf_parameters.rs
```diff
@@ -108,8 +108,65 @@ mod tests {
 		);
 	}
 
+	#[test]
+	fn proper_address_conversion() {
```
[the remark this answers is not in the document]
albert-llimos: He just pushed some of this code so I could help out debugging some encoding issues. I've now pushed the fix. This can be cleaned up now :+1:

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2466,6 +2500,34 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			boost_fee,
 		} = vault_deposit_witness.clone();
 
+		if Self::should_reject_vault_swap(&broker_fee) {
+			// TODO: We are going to make refund_params mandatory so we can remove this if let check
+			// as soon the other PR is merged.
```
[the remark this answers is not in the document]
Janislav: I would not do that. The other PR is big enough already. If the other one should get in first, we do this anyway, since we have to rebase, but it's probably the other way around.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: It's for debugging. Going to remove this event again

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: Sure, whichever you prefer.

dandanlen (commented):
I pushed a commit fixing some of these issues, and also with a todo!() and some comments.

@Janislav can you please look at where I added the todo!() and fix the listed issues. We need to use the normal processing pipeline for these refunds instead of exiting early. This means: 
- adding a new variant ChannelAction::Refund
- handling this in perform_channel_action
- it should use a normal egress with BatchAll, not TransferFallback (schedule_egress should take care of this)
- We might need a new event to link the refund with the egress id so the product can track it. 

I think it makes sense to still emit the DepositFailed event.

Janislav:
@dandanlen one thought from my side regarding the DepoistFaild event. I think if we want to keep the current pipeline as it is with just adding a new DepositAction, it's easier to not handle it as an error case and instead as a successful deposit, which it also is. We can return an error if the provided Action was a refund, but we first have to modify the current behaviour and second still have to do everything we would if the deposit was successful. I assume it would be easier for product to handle it if it would be an error?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -104,6 +104,12 @@ pub struct BoostOutput<C: Chain> {
 	total_fee: C::ChainAmount,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub enum VaultSwapStage {
+	Prewitnessed,
+	Deposit,
+}
+
```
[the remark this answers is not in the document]
dandanlen: Still unused as far as I can tell?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1935,6 +1959,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+						}
+					},
+					_ => {
+						log_or_panic!("Refund action with invalid asset or address.");
+						DepositAction::Refund {
+							egress_id: None,
+							amount: amount_after_fees,
+							channel_id,
+							refund_success: false,
+						}
```
dandanlen, line 1990: Is this equivalent to `DepositFailed`? 

If yes, then instead of `DepositAction::Refund { .., success: false }` we could have a `DepositAction::Failed` variant.

dandanlen (commented):
Thanks for this. Will take a closer look on Monday. Just some initial comments for now.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: No actually not. It is a successful deposit. What it says is: "The Deposit was successful, and we took the money, but as action of this deposit we scheduled a refund over the same amount". In the following code path, we also don't swap but return early. I had a chat with product, and they said they could handle it like that.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1935,6 +1959,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 				DepositAction::Swap { swap_request_id }
 			},
+			ChannelAction::Refund { asset, refund_address, channel_id } => {
+				match (asset.try_into(), refund_address.try_into()) {
+					(Ok(asset), Ok(address)) => {
+						let egress_id =
+							match Self::schedule_egress(asset, amount_after_fees, address, None) {
+								Ok(egress_details) => Some(egress_details.egress_id),
+								Err(e) => {
+									log_or_panic!(
```
dandanlen, line 1969: We should not log_or_panic here - this can happen if the egress is below the dust limit (which is legitimate: not a logic problem). It would make more sense to emit an event (maybe emit it inside schedule_egress, then that fn can be made infallible)

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1935,6 +1953,38 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+						}
+					},
+					_ => {
+						log_or_panic!("Refund action with invalid asset or address.");
+						DepositAction::Refund {
+							egress_id: None,
+							amount: amount_after_fees,
+							channel_id,
+							refund_success: false,
+						}
```
dandanlen, line 1984: Actually i'm pretty sure this can't happen - if we change the types on DepositAction to the chain-specific types, then we can enforce this in the compiler.

Janislav: I don't think so, we have only try_from implemented for primitive::Asset. I think this is because of some internal ID matching things that could be wrong. I agree, it can (actually) not happen in reality (as long we don't change any surrounding assumptions). Still, we have to handle it.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: Not sure if it is possible to make the function infallible. We expect it to schedule an egress under the assumptions that the dust requirements are held. Based on that, it's always possible to fail. 

We emit Refund {success: false} together with the DepoistFinalized event, so technically we emit this information as an event, but it's possible to emit another event to make it more clear.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -2391,3 +2371,66 @@ fn ignore_change_of_minimum_deposit_if_deposit_is_not_boosted() {
...
+			dca_params: None,
+			boost_fee: 0,
+		};
+
+		IngressEgress::process_vault_swap_request_prewitness(0, vault_swap.clone());
+
+		assert_eq!(
+			ScheduledEgressFetchOrTransfer::<Test, ()>::get().len(),
+			0,
+			"Refund broadcast should have been scheduled!"
```
dandanlen, line 2407:
```suggestion
			"Refund broadcast should not have been scheduled!"
```

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -2391,3 +2371,66 @@ fn ignore_change_of_minimum_deposit_if_deposit_is_not_boosted() {
...
+		assert!(
+			MockSwapRequestHandler::<Test>::get_swap_requests().is_empty(),
+			"No swaps should have been triggered!"
+		);
+
+		assert_eq!(
+			ScheduledEgressFetchOrTransfer::<Test, ()>::get().len(),
+			1,
+			"Refund broadcast should have been scheduled!"
+		);
```
dandanlen, line 2421: This doesn't check that it's a refund... it just checks that there is *some* egress scheduled. It would be good to check the destination of the egress.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -939,6 +956,13 @@ pub mod pallet {
 		NetworkFeeDeductionFromBoostSet {
 			deduction_percent: Percent,
 		},
+		VaultSwapRefundRequested {
+			channel_id: Option<ChannelId>,
+			asset: TargetChainAsset<T, I>,
+			amount: TargetChainAmount<T, I>,
+			refund_address: TargetChainAccount<T, I>,
+			egress_id: EgressId,
+		},
```
dandanlen, line 959: This event is unused.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Fair enough. In that case I would suggest removing refund_success anyway since it's already indicated by the egress_id (if it's None, then the egress failed).

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: In the deposit verification code, we call `into` to convert the asset and address into the `any::Asset` and `ForeignChainAddress` types. If we avoid doing this, then we don't have to convert them back via `try_into()`, and we don't have to handle this failure case.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2483,7 +2544,55 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		else {
+			match Self::process_full_witness_deposit_inner(
+				None,
+				source_asset,
+				deposit_amount,
+				deposit_details.clone(),
+				None,
+				BoostStatus::NotBoosted, /* We immeditly abort the prewitness processing if the
+				                          * broker fees are invalid. In this case the depoist
+				                          * can never be boosted. */
```
dandanlen, line 2557: It's not clear why we're hard-coding some of these to `None`. 

There's also lot of duplication. The other `process_full_witness_deposit_inner` call is (or should be) effectively identical except for the ChannelAction. 

I feel like it would be nice here to handle *all* the possible errors (invalid destination address etc) in the same way, by attempting a refund. There's nothing special about a missing broker.

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: The problem is that we are passing the reund_address and asset other the ChannelAction which is not generic over T and I. Because of that we have to convert the address and asset when we pass it via ChannelAction. The alternative would be making ChannelAction generic using the associated type directly.

Janislav: *but that requires a lot of refactoring

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: Yeah, you're right. I think in all of this validation cases (validate_refund_params, validate_dca_params, validate_broker_fees, validate_address) we should trigger a refund instead of aborting the deposit which would potentially lead to lose of funds.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: We can make the ChannelAction generic over Account and chain to avoid the full dependency on T and I. (`ChannelAction<AccountId, C: Chain>`).

state-chain/pallets/cf-ingress-egress/src/lib.rs
Janislav: Added another ticket here: https://linear.app/chainflip/issue/PRO-2139/refactor-processing-of-vault-swaps where I describe how we do it for all the possible other cases.

state-chain/cf-integration-tests/src/solana.rs
```diff
@@ -545,7 +546,7 @@ fn solana_ccm_fails_with_invalid_input() {
 					Runtime,
 					SolanaInstance,
 				>::CcmEgressInvalid {
-					egress_id: (ForeignChain::Solana, 1u64),
+					egress_id: (ForeignChain::Solana, 2u64),
```
Janislav, line 549: I assume this changed to 2 because we also trigger a refund, that is querying an egress id 🤔

state-chain/pallets/cf-ingress-egress/src/tests.rs
Janislav: There is actually no difference between a refund or a normal egress, and we check in the event under this that it was a refund.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2302,11 +2351,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				// Only consider rejecting a transaction if we haven't already boosted it,
 				// since by boosting the protocol is committing to accept the deposit.
 				if TransactionsMarkedForRejection::<T, I>::take(broker_id, &tx_id).is_some() {
+					if matches!(action, ChannelAction::Refund { .. }) {
+						return Err(DepositFailedReason::CanNotRejectRefunds);
+					}
```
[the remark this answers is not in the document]
Janislav: In this case, we do nothing and ignore the deposit. In reality, this can (actually) not happen since a ChannelAction::Refund is something that can only be triggered from within the system. Somehow we still have to handle it somehow since we need to derive the refund address from the enum. The alternative would be to just put `unreacable!`, but I really don't like anything that could panic inside the runtime.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -492,6 +516,11 @@ pub mod pallet {
 			lp_account: AccountId,
 			refund_address: ForeignChainAddress,
 		},
+		Refund {
+			asset: C::ChainAsset,
+			reason: RefundReason,
```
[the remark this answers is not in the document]
Janislav: Agree with asset, don't agree with reason since we are building the DepositAction from it which needs the reason. If we don't need there, we don't need at all.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: Imagine a scenario where we receive a vault deposit with invalid broker data, and the deposit is also flagged for rejection by the deposit monitor: we want to process the rejection, we don't want to do nothing.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
dandanlen: The reason I suggested it's not really necessary is that we already emit it in `DepositFailed` event. If we add it to the DepositAction, it just means that it will be emitted again in another event. It doesn't have any functional reason for existing here. I guess we should find out what DepositAction is used for (in the product?) and then decide if we need it or not.

state-chain/pallets/cf-ingress-egress/src/lib.rs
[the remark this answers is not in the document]
Janislav: I think my point is that a Refund is something that is triggered only from within in the system and can never come from outside and with this also never be tainted. If we reached this part of the code for a Refund, then this is a bug, and we should do nothing other than fire an event or anything similar to get informed about it. The cleanest solution from a syntactic point of view would be to just put a `unreachable!` where we derive the refund address and just panic if this should ever be hit, but since we are in the Runtime, I would like to avoid that.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2646,37 +2637,83 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+						broker_fees: broker_fees.clone(),
+						channel_metadata: channel_metadata.clone(),
+						refund_params: refund_params
+							.map_address(|address| address.into_foreign_chain_address()),
+						dca_params: dca_params.clone(),
+					},
+					source_address,
+				),
+				Err(reason) => (
+					ChannelAction::Refund {
```
dandanlen, line 2710: I think we should emit an event here.

dandanlen (approved):
LGTM!

PR #5509, opened 2024-12-18
feat: vault swap rpc call for building EVM call

syan095:
# Pull Request

Closes: PRO-1642

## Checklist

## Summary

Added EVM support for building Vault Swap transactions via broker-api / rpc calls


state-chain/chains/src/cf_parameters.rs
```diff
@@ -28,6 +39,40 @@ pub struct VaultSwapParameters {
...
+			ccm_additional_data: ccm.ccm_additional_data.clone(),
+			vault_swap_parameters,
+		})
+		.encode(),
+		None => VersionedCfParameters::V0(CfParameters {
+			ccm_additional_data: (),
+			vault_swap_parameters,
+		})
+		.encode(),
+	}
```
dandanlen, line 73:
```suggestion
	VersionedCcmCfParameters::V0(CfParameters {
		ccm_additional_data: ccm.map(|ccm| ccm.ccm_additional_data.clone()),
		vault_swap_parameters,
	})
	.encode()
```

syan095: The CfParameters must have either Bytes, or () as its `ccm_additional_data` type (It's not an Option)
so we cannot dynamically map this into different return types within the same statement.

THis is also the reason why we return Encoded Bytes here, instead of the `VersionedCcmCfParameters` type, because the Generic will be different depending on whether CCM is present or not.

state-chain/chains/src/sol/instruction_builder.rs
```diff
@@ -43,27 +36,12 @@ impl SolanaInstructionBuilder {
 		agg_key: SolPubkey,
 		destination_asset: Asset,
 		destination_address: EncodedAddress,
-		broker_id: AccountId,
-		broker_commission: BasisPoints,
-		refund_parameters: ChannelRefundParametersDecoded,
-		boost_fee: BasisPoints,
-		affiliate_fees: BoundedVec<AffiliateAndFee, ConstU32<MAX_AFFILIATES>>,
```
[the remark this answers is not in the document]
syan095: I've rebased the EVM work on top of the base branch, which has Solana work merged already.

In this change I've refactored the code that builds the cf_parameter out so it can be shared across EVM and Solana
Hence this change in Solana's InstructionBuilder

state-chain/chains/src/lib.rs
```diff
@@ -991,6 +991,10 @@ pub enum VaultSwapExtraParameters<Address, Number> {
 		min_output_amount: Number,
 		retry_duration: BlockNumber,
 	},
+	Evm {
```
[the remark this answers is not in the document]
syan095: Split EVM into Ethereum and Arbitrum variants.

state-chain/chains/src/evm/api/vault_swaps/x_call_native.rs
```diff
@@ -0,0 +1,126 @@
...
+				.encode_input(&[
+					dest_chain.tokenize(),
+					dest_address_bytes.tokenize(),
+					(dest_asset as u32).tokenize(),
+					ccm.message.to_vec().tokenize(),
+					U256::from(ccm.gas_budget).tokenize(),
+					dummy_cf_parameter(true).tokenize(),
+				])
+				.unwrap()
+		);
```
albert-llimos, line 124: I'd want to hardcode a encoded payload to these tests to make sure that the output matches a real test. I'll try to do that

state-chain/chains/src/evm/api/vault_swaps/x_call_token.rs
```diff
@@ -0,0 +1,147 @@
...
+					dest_chain.tokenize(),
+					dest_address_bytes.to_vec().tokenize(),
+					(dest_asset as u32).tokenize(),
+					ccm.message.to_vec().tokenize(),
+					U256::from(ccm.gas_budget).tokenize(),
+					eth_token.tokenize(),
+					U256::from(amount).tokenize(),
+					dummy_cf_parameter(true).tokenize(),
+				])
+				.unwrap()
```
albert-llimos, line 144: Same here

albert-llimos (commented):
I think that the implementation of the `abi_encoded` and the `get_function` is quite strange and ends up being quite untuitive. I think the `SigData` should be part of the `function_params` and then encoding should just be generic abi encoding of all the function parameters. Mainly it's due to legacy code, this PR is just building on top of that. We can live with it but I just wanted to mention it.

state-chain/runtime/src/chainflip/vault_swaps.rs
```diff
@@ -108,6 +109,119 @@ pub fn bitcoin_vault_swap(
...
+		_ => Err(DispatchErrorWithMessage::from(
+			"Only EVM chains should execute this branch of logic. This error should never happen",
+		)),
+	}?;
+
+	match source_asset.into() {
+		ForeignChain::Ethereum => Ok(VaultSwapDetails::Ethereum(EvmVaultSwapDetails {
+			calldata,
+			// Only return `amount` for native currently. 0 for Tokens
+			value: (source_asset == Asset::Eth).then_some(U256::from(amount)).unwrap_or_default(),
```
albert-llimos, line 208: If I understand this code correctly, if the `U256::from(amount)` fails it will default to 0 too. If that's true, I think we should return an error in that scenario because something has gone wrong with the value and I'd rather error out.

state-chain/runtime/src/chainflip/vault_swaps.rs
```diff
@@ -108,6 +109,119 @@ pub fn bitcoin_vault_swap(
...
+			to: Environment::key_manager_address(),
+		})),
+		ForeignChain::Arbitrum => Ok(VaultSwapDetails::Arbitrum(EvmVaultSwapDetails {
+			calldata,
+			// Only return `amount` for native currently. 0 for Tokens
+			value: (source_asset == Asset::ArbEth)
+				.then_some(U256::from(amount))
+				.unwrap_or_default(),
```
albert-llimos, line 216: Same here.

state-chain/runtime/src/chainflip/vault_swaps.rs
syan095: Oh i see the confusion. 
the logic goes like this:
For Eth -> true -> evaluate(U256::from(amount)) -> Some(Amount) -> unwrap -> Amount
for Tokens -> false -> evaluate to None -> unwrap to default -> 0

So it's just a 1 liner way of saying: return 0 for tokens, Amount for Eth/ArbEth

state-chain/chains/src/evm/api/vault_swaps/x_call_native.rs
albert-llimos: Added the hardcoded test. Very hard to check that especially the cf_parameters work because it requires an end to end testing. We will test that when it's used in the SDK. This is good enough for now.

dandanlen:
Just FYI - I'm being a bit slow but still reviewing this.

engine/src/witness/btc/vault_swaps.rs
```diff
@@ -204,7 +203,7 @@ mod tests {
 	static MOCK_SWAP_PARAMS: LazyLock<UtxoEncodedData> = LazyLock::new(|| UtxoEncodedData {
 		output_asset: cf_primitives::Asset::Dot,
 		output_address: EncodedAddress::Dot(MOCK_DOT_ADDRESS),
-		parameters: SharedCfParameters {
+		parameters: BtcCfParameters {
```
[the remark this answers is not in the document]
syan095: this may not be a straight forward task. The Btc version uses different Refund parameters:
btc: retry_duration + min output (u128
others: retry_duration + refund_address + price (u256)

While we can technically include the min_output as U256, and leave the refund_address as "default" (empty or invalid address that will never be read), the interfacing doesn't look good.

this is also the reason the `refund parameter` was part of the `vault swap extra parameters`. 
i think it would take some thinking to come up with an all-encompassing clean solution for this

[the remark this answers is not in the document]
dandanlen: Bitcoin also uses slightly different encoding that the others, and for example doesn't include the broker_id.

dandanlen (approved):

albert-llimos:
Seems like the upgrade test fail. Do we maybe need some migration? @syan095

PR #5512, opened 2024-12-19
fix: vault swap - solana improvements

syan095:
## Summary

Per discussion, made some improvements to the Solana version of Vault swap RPC call
* Solana vault swaps uses on-chain-key instead of aggkey
* Added a new SolInstructionRpc type that uses SolAddress so it serializes all addresses into bs58 instead of byte arrays.


albert-llimos (approved):
LGTM. While we could just skip looking up the `agg_key` now it's still reasonable to do it check the blacklist.

dandanlen (approved):

syan095:
This is really strange. All the commits are signed - but when the PR is merged, the commits merged into this branch is unsigned.

I'll redo the commit again manually

PR #5514, opened 2024-12-19
feat: Support backwards compatible vault_swap_requests

Janislav:
# Pull Request

Closes: PRO-1906

## Checklist

## Summary

- Catches decoding errors of cf-parameters
- Provides default parameters
- Handles the case if no refund parameters are available on the protocol side

Defaults values

If we can not decode the cf_parameters we use these default vaules:

- broker_fee = broker_account with 0 address, and 0 bps
- affilate_fees = Default / which should just be an empty Vector
- dca_parameters = None
- refund_params = None 

#### Non-Breaking changes

### DIFF
--- a/bouncer/tests/all_swaps.ts
+++ b/bouncer/tests/all_swaps.ts
@@ -80,20 +80,20 @@ async function main() {
   // We observed this values in the wild and want to ensure that the engine can handle them gracefully.
   allSwaps.push(
     testVaultSwap(
-      "ArbEth",
-      "Eth",
+      'ArbEth',
+      'Eth',
       undefined,
-      newCcmMetadata("Eth", undefined, undefined, "0xdeadbeef"),
+      newCcmMetadata('Eth', undefined, undefined, '0xdeadbeef'),
       testAllSwaps.swapContext,
     ),
   );
 
   allSwaps.push(
     testVaultSwap(
-      "Eth",
-      "ArbEth",
+      'Eth',
+      'ArbEth',
       undefined,
-      newCcmMetadata("ArbEth", undefined, undefined, "0xdeadC0de"),
+      newCcmMetadata('ArbEth', undefined, undefined, '0xdeadC0de'),
       testAllSwaps.swapContext,
     ),
   );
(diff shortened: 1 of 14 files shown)


albert-llimos:
I know it's still in draft but leaving a comment - I think we should have some basic bouncer Vault swaps with different undecodable `cf_parameters` that we've seen on mainnet (e.g. empty, `0xdeadC0de`, `0xdeadbeef`) to make sure it works end to end and we don't break compatibility. That is only necessary for EVM Vault swaps, since Solana Vault swaps are not live.
It should be very easy to add to the bouncer, just a few lines on the `all_swaps.ts` but let me know if you want me to help out.

Janislav:
Good point! I going to add it.

bouncer/tests/all_swaps.ts
```diff
@@ -76,5 +76,38 @@ async function main() {
...
   appendSwap('SolUsdc', 'Flip', testVaultSwap, true);
 
+  // Note: Test vault swap with invalid refund params to ensure the engine decoding fallback works.
+  // We observed this values in the wild and want to ensure that the engine can handle them gracefully.
+  allSwaps.push(
+    testVaultSwap(
+      'ArbEth',
+      'Eth',
+      undefined,
+      newCcmMetadata('Eth', undefined, undefined, '0xdeadbeef'),
```
albert-llimos, line 86: I've just realized that my suggestion on how to test this out doesn't work straight away. A parameter of the `newCcmMetadata` confused me (I will fix it). The problem is that the string passed here doesn't end up as the `cf_parameters` but rather as the `ccmAdditionalData`. Then the SDK encodes that into a valid `cf_parameters` in the contract call.
Getting around this is extremely annoying without manually implementing the call to the smart contract, but we might do that anyway to test the broker API.
I'll think about how to do it but I think these three extra swaps can be removed unfortunately. Sorry @Janislav for the confusion :sweat:

albert-llimos: Yeha pretty sure that's the only way. I added that note to this ticket: https://linear.app/chainflip/issue/PRO-1912/implement-bouncer-evm-vault-swaps-via-broker-api
I'll add those tests there, it will be possible after that ticket is implemented.

dandanlen (approved):
LGTM, I just made some small refactors.

PR #5518, opened 2024-12-20
feat: min network fee

msgmaxim:
# Pull Request

Closes: PRO-1701

## Summary

- New `MinimumNetworkFeePerChunk` storage item, set to 0 by default, but can be updated via governance extrinsic. In practice "min fee per chunk" should effectively be "min fee per swap request" since we also enforce min chunk size (which should be sufficiently large). I considered adding checks ensuring that we never update parameters such that this doesn't hold, but it is a bit tricky since chunk size is in target chain asset amount and we don't know the exchange rate... The best we can do it check it against the min chunk limit for USDC and rely on other limits being in line with that. However, I dicussed this with @kylezs and we agreed that this shouldn't really matter considering that the new min network fee will already be incorporated into the price estimation that's used for quotes (i.e. if the user sets a small chunk size and the protocol somehow allows it, they should see the higher network fee).

- Min fee only applies to "regular" swaps, i.e. it doesn't not apply to swaps for ingress/egress fees (which are expected to be tiny). For this I updated FeeType::NetworkFee to include a boolean `min_fee_enforced`, and added a migration to update storage any already queued swaps.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -147,11 +147,10 @@ impl<T: Config> SwapState<T> {
 	}
 }
 
-#[repr(u8)]
 #[derive(Clone, DebugNoBound, PartialEq, Eq, Encode, Decode, TypeInfo)]
 #[scale_info(skip_type_params(T))]
 pub enum FeeType<T: Config> {
-	NetworkFee = 0,
+	NetworkFee { min_fee_enforced: bool },
 	BrokerFee(Beneficiaries<T::AccountId>),
 }
 
@@ -362,6 +361,8 @@ pub enum PalletConfigUpdate<T: Config> {
 	/// Set the broker bond. This is the amount of FLIP that must be bonded to open a private
 	/// broker channel. The funds are getting freed when the channel is closed.
 	SetBrokerBond { bond: T::Amount },
+	/// Set the minimum fee in USDC paid per chunk
+	SetMinimumNetworkFeePerChunk { min_fee: AssetAmount },
 }
 
 impl_pallet_safe_mode! {
@@ -540,6 +541,12 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type BrokerBond<T: Config> = StorageValue<_, T::Amount, ValueQuery, DefaultBrokerBond<T>>;
 
+	/// Minimum network fee charged per chunk (only applies to regular swaps, i.e. it excludes
+	/// internal swaps like ingress/egress fees). In practice this should also effectively be the
+	/// minimum fee charged per swap request due to us also enforcing minimum chunk size.
+	#[pallet::storage]
+	pub type MinimumNetworkFeePerChunk<T: Config> = StorageValue<_, AssetAmount, ValueQuery>;
+
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
@@ -685,6 +692,9 @@ pub mod pallet {
 		},
 		BrokerBondSet {
 			bond: T::Amount,
+		},
+		MinimumNetworkFeeSet {
+			min_fee: AssetAmount,
 		},
 	}
 	#[pallet::error]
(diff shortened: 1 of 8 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1802,11 +1825,16 @@ pub mod pallet {
...
+		pub(super) fn take_network_fee(input: AssetAmount, min_fee_enforced: bool) -> FeeTaken {
 			if input.is_zero() {
 				return FeeTaken { remaining_amount: 0, fee: 0 };
 			}
-			let (remaining, fee) = utilities::calculate_network_fee(T::NetworkFee::get(), input);
+
+			let min_fee = if min_fee_enforced { MinimumNetworkFeePerChunk::<T>::get() } else { 0 };
+
+			let (remaining, fee) =
+				utilities::calculate_network_fee(T::NetworkFee::get(), min_fee, input);
```
dandanlen, line 1836: Since this is called inside a loop, and the min fee/rate are the same on each iteration, it might make sense to refactor this method such that we pass the fee rate and min fee by value instead of reading storage.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2059,7 +2087,12 @@ pub mod pallet {
 
 			let estimation_output = with_transaction_unchecked(|| {
 				TransactionOutcome::Rollback(
-					Self::swap_with_network_fee(input_asset, output_asset, estimation_input).ok(),
+					Self::swap_with_network_fee_for_gas(
```
dandanlen, line 2090: We've renamed this to `swap_with_network_fee_for_gas` presumably because it's always used to in a context where output_asset is the gas asset. If this is the case, we might as well enforce it by removing the argument and just using the GAS_ASSET constant?

dandanlen (commented):
FWIW the effective minimum fee is higher than the minimum: it will be `min_fee + (network_fee_pct * (ingress_fee + egress_fee))`.

Maybe this is acceptable - it's a small difference after all - but we could avoid this (and make the code a bit leaner) by simply not charging a network fee on internal gas swaps. Thoughts?

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: I believe storage items are cached, and reading the same value a few times in a row should be cheap, no? At the same time, if we move this outside of the loop, then we would be reading the storage even when it is not needed. Seems to me like it is not really worth it?

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: `GAS_ASSET` is still different for different chains, and we are inside the chain-agnostic swapping pallet here... There may be a way to pass "chain" as a paramter here, e.g. by making `swap_with_network_fee_for_gas` generic? Not sure if worth it though.

msgmaxim:
> we could avoid this (and make the code a bit leaner) by simply not charging a network fee on internal gas swaps. Thoughts?

I'm not sure. It does make sense to charge on gas swaps, but I agree that it would simplify many things. Seems like mostly a business desicion.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: In that case I would just not rename the method - it's not gas-specific.

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: They are cached in the storage overlay - ie. it doesn't require a full read/traversal of the underlying trie, but it still needs to hash the key, look it up in the backend k-v store and decode the result. It's wasteful but I agree it's not a big deal, feel free to ignore.

dandanlen:
> > we could avoid this (and make the code a bit leaner) by simply not charging a network fee on internal gas swaps. Thoughts?
> 
> I'm not sure. It does make sense to charge on gas swaps, but I agree that it would simplify many things. Seems like mostly a business desicion.

I don't think it's obviously correct (or incorrect) to charge a network fee on gas swaps, but ok.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: But it *is* gas specific, it is just that gas assets can be different depending on the chain. The reason I renamed is that it wasn't immediately clear how this function was used, and whether or not min network fee should be enforced. Once I ensured that it is only used for gas swaps, I knew not to enforce it and renamed that function to make it clear why.

PR #5519, opened 2024-12-23
feat: add bond to private broker channel events

msgmaxim:
# Pull Request

Closes: PRO-1910

## Summary

- Added `get_bond` to the Bonder trait (until now it only existed in the mock)
- added bond to `PrivateBrokerChannelOpened` and `PrivateBrokerChannelClosed`

fyi @niklasnatter

### DIFF
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -441,8 +441,12 @@ impl<T: Config> Bonding for Bonder<T> {
 	type AccountId = T::AccountId;
 	type Amount = T::Balance;
 
-	fn update_bond(authority: &Self::AccountId, bond: Self::Amount) {
-		Account::<T>::mutate_exists(authority, |maybe_account| {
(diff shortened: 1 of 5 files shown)


state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -441,8 +441,12 @@ impl<T: Config> Bonding for Bonder<T> {
 	type AccountId = T::AccountId;
 	type Amount = T::Balance;
 
-	fn update_bond(authority: &Self::AccountId, bond: Self::Amount) {
-		Account::<T>::mutate_exists(authority, |maybe_account| {
+	fn get_bond(account_id: &Self::AccountId) -> Self::Amount {
+		Account::<T>::get(account_id).bond
+	}
```
dandanlen, line 446: We don't need to add this - the information is already accessible through the AccountInfo trait, which is part of the `Chainflip` base config trait.

`<T::FundingInfo as cf_traits::AccountInfo>::bond();`

dandanlen (commented):
See my comment in the Linear issue. I would like to clarify the need for this before add it.

PR #5524, opened 2025-01-02
ci: handle expired artefacts in force version bump 🐛

ahasna:


kylezs:
We have to check it on the network because otherwise we will never know if we've forgotten to bump the version in OLD_VERSION

PR #5526, opened 2025-01-02
chore: remove unused solana channels

ramizhasan111:
# Pull Request

Closes: PRO-1811

## Checklist

## Summary

#### Non-Breaking changes

### DIFF
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -74,7 +74,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: true
+      run-job: false
   publish:
     needs: [package]
     uses: ./.github/workflows/_30_publish.yml
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -74,7 +74,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
   publish:
     needs: [package]
     uses: ./.github/workflows/_30_publish.yml
--- a/state-chain/runtime/src/migrations/solana_remove_unused_channels_state.rs
+++ b/state-chain/runtime/src/migrations/solana_remove_unused_channels_state.rs
@@ -1,5 +1,5 @@
 use crate::*;
-use frame_support::{pallet_prelude::Weight, traits::UncheckedOnRuntimeUpgrade};
+use frame_support::{pallet_prelude::Weight, traits::OnRuntimeUpgrade};
 
 use codec::{Decode, Encode};
 use pallet_cf_elections::{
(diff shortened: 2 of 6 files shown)


.github/workflows/ci-development.yml
```diff
@@ -74,7 +74,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
```
dandanlen, line 77: We need to remember to remove this.

state-chain/runtime/src/migrations/solana_remove_unused_channels_state.rs
```diff
@@ -0,0 +1,68 @@
...
+			Runtime,
+			SolanaInstance,
+		>::iter_keys()
+		.chain(
+			pallet_cf_ingress_egress::DepositChannelPool::<Runtime, SolanaInstance>::iter_values()
+				.map(|deposit_channel| deposit_channel.address),
+		)
+		.collect();
+
+		ElectoralUnsynchronisedStateMap::<Runtime, SolanaInstance>::iter_keys().for_each(|key| {
```
dandanlen, line 27: Better to collect this into a Bev and then iterate over the Bev to do the deletions. The docs state that updating the map while iterating can have unintended side effects.

dandanlen (commented):
How did you test this? Maybe some print statements in the upgrade would be useful to track progress?

ramizhasan111:
> How did you test this? Maybe some print statements in the upgrade would be useful to track progress?

I tested it locally by running the upgrade on the latest release with the try-runtime checks.

state-chain/runtime/src/migrations/solana_remove_unused_channels_state.rs
ramizhasan111: you mean vec right?

ramizhasan111:
added logs which prints the address list before and after migrating the storage.

state-chain/runtime/src/migrations/solana_remove_unused_channels_state.rs
dandanlen: Sorry, phone autocorrect...

dandanlen:
Will wait before merging this since it sounds like we might not activate Solana address recycling for 1.8 after all.

dandanlen:
Actually maybe this is independent of that? We still might as well delete old unused addresses right?

dandanlen (approved):

dandanlen:
Right - I had a short think about it. I think we can keep this, but put it in the housekeeping migration. It should be ok to run it more than once (it can't remove an account twice), so we don't need it to be part of a storage-version increase, for example.

state-chain/runtime/src/migrations/solana_remove_unused_channels_state.rs
```diff
@@ -0,0 +1,78 @@
...
+		log::info!("🚮 Number of addresses deleted: {:?}", removed);
+		if removed == 0 {
+			log::warn!("❗️ No Solana Addresses deleted - consider removing this migration.");
+		}
+
+		Weight::zero()
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, sp_runtime::TryRuntimeError> {
```
kylezs, line 49: nit, putting pre_upgrade before the `on_runtime_upgrade` makes it read sequentially

kylezs:
> I tested it locally by running the upgrade on the latest release with the try-runtime checks.

Did you test there were actually expired channels in storage before running it? I would also run the try-runtime against some of the live networks that have expired channels too.

ramizhasan111:
> Did you test there were actually expired channels in storage before running it? I would also run the try-runtime against some of the live networks that have expired channels too.

Yes, I did. Made sure that the expired channels do in fact get deleted.

ramizhasan111:
this is ok to merge now right?

dandanlen (approved):

PR #5527, opened 2025-01-03
feat: affiliate account registering and withdrawal functionality

Janislav:
# Pull Request

Closes: PRO-1795, PRO-1796

## Checklist

## Summary

In this PR, we add all functionality for handling affiliate brokers. Here is a short explanation of what the handling of affiliate accounts will look like.

An affiliate is getting created/associated by calling the `register_affiliate` extrinsic by a full broker.

During this process, the following steps are happening:

1) An account ID is getting derived from the broker account ID
2) An **not** changeable withdrawal address is getting set → The funds can only be withdrawn to this address
2) The affiliate account is getting locked against the broker account

This has the following implications:

1) All the funds for an affiliate are getting collected in this account
2) The short ID can only be used once for every broker and is taken after registering → 1 Broker can have 255 affiliates.
3) The broker can trigger a withdrawal to the address of the affiliate

With this, an affiliate doesn't need to interact with the state-chain directly and can fully rely on the broker they trust. After registering an affiliate, it's necessary to check the registered withdrawal address by the affiliate to verify the setup process was not maliciously manipulated.

Apart from that, there is a node RPC (`cf_get_affiliate_account_details`) to query the affiliate details. An affiliate can use this to ensure the withdrawal address is setup correctly without the need of trusting the broker. Alternately, it's also possible to verify this in PolkadotJs.  

> **_NOTE:_**  By this change, we introduce a clear separation between a broker to an affiliate. An affiliate is no longer a broker and should also not interact with the state-chain directly.

#### Non-Breaking changes

This should not require any migration and should also not break any existing interfaces.


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1204,6 +1214,66 @@ pub mod pallet {
...
+				Error::<T>::ExpectedEthereumAddress
+			);
+
+			ensure!(
+				!AffiliateWithdrawalAddress::<T>::contains_key(&affiliate_account),
+				Error::<T>::AddressAlreadyRegistered
+			);
+
+			AffiliateWithdrawalAddress::<T>::insert(&affiliate_account, withdrawal_address.clone());
+
```
ramizhasan111, line 1252: Is it by design that only brokers are allowed to execute the withdrawal for one of their affiliates? Can't the affiliates withdraw the fees accumulated for them themselves?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1287,6 +1357,57 @@ pub mod pallet {
...
+			ensure!(earned_fees != 0, Error::<T>::NoFundsAvailable);
+			T::BalanceApi::try_debit_account(account_id, asset, earned_fees)?;
+
+			let ScheduledEgressDetails { egress_id, egress_amount, fee_withheld } =
+				T::EgressHandler::schedule_egress(
+					asset,
+					earned_fees,
+					destination_address_internal,
+					None,
+				)
```
ramizhasan111, line 1382: Is there is a reason why affiliate withdrawal can only be for Usdc and normal broker withdrawal can be for any asset?

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: We changed it now that any broker, doesn't matter if affiliate or not, handles it for itself.

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: All brokers collect fees in Usdc.

dandanlen (commented):
I think this requires a bit more thought/discussion. For example it would be good to consider how we expect an affiliate to interact with this. 

Currently, to lock the account, the affiliate has multiple requirements:
- They need to create a state chain account:
	- Create the key pair
	- Get some FLIP
	- Submit a tx to the SC Gateway contract
- They need to register the account as a broker
	- Submit an extrinsic. How? Via the UI? CLI?
- Lock the account:
	- Submit an extrinsic. How? (See above - not implemented neither via UI, broker api or cli)
- Withdraw funds: they need to issue another SC call, again not clear if this is via the broker/UI etc.

One of the main reasons for introducing affiliates was to avoid the need to interact with the SC. So I was assuming that that all of the above would somehow be handled by the broker.

Some suggestions:
- We can make this functionality available in the asset_balances pallet. I don't see why it needs to be restricted to brokers only, especially since we no longer require affiliates to register as brokers.
- I also think we can relax the 'only Ethereum addresses' restriction - we could allow this for any chain or asset? 
- We need to expose the call somehow, either via api or cli or UI.
- We need to consider if / how we want to allow brokers to control all of the above on behalf of their affiliates (see PRO-1921 for example).

Janislav:
@dandanlen I rested the PR as discussed yesterday. We are now back to the first implementation that expects an affiliate doesn't interact with the state chain directly.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -537,6 +537,11 @@ pub mod pallet {
 		OptionQuery,
 	>;
 
+	/// An optional withdrawal address for affiliate broker.
+	#[pallet::storage]
+	pub type AffiliateWithdrawalAddress<T: Config> =
+		StorageMap<_, Identity, T::AccountId, EncodedAddress, OptionQuery>;
+
```
dandanlen, line 544: As I mentioned yesterday, I think we can make this contain all the affiliate information, for example: 

```
struct AffiliateDetails {
	short_id: ShortId,
	withdrawal_address: Option<EncodedAddress>, // EncodedAddress means the affiliate can inspect it more easily,
}
```

And then it needs to be in a DoubleMap of `BrokerId -> AffiliateId -> AffiliateDetails`.

This will also make it a bit easier to do a lookup of `AffilliateId -> ShortId`: there are a few places where we do this and currently it requires iterating and reversing the AffiliateIdMapping. 

We need to make sure that we always update both AffiliateIdMapping and AffiliateDetails consistently. We also need to make sure that all the data is removed when the broker deregisters.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1204,6 +1209,66 @@ pub mod pallet {
...
+		/// Registers a withdrawal address for an affiliate. This is a one-time operation. Once the
+		/// address is set up, it's not possible to change it any more.
+		/// Note: This extrinsic is secured by the broker that has registered the affiliate account.
+		///
+		/// ## Events
+		///
+		/// - [PrivateBrokerChannelClosed](Event::PrivateBrokerChannelClosed)
+		#[pallet::call_index(15)]
+		#[pallet::weight(T::WeightInfo::register_affiliate_withdrawal_address())]
+		pub fn register_affiliate_withdrawal_address(
```
dandanlen, line 1222: This can be made part of registering an affiliate (ie. moved into register_affiliate). Using the new map, we can make it illegal to change the ShortId of an Affiliate - this will give the affiliate more assurances that the broker can't modify any of its details.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1365,6 +1481,20 @@ pub mod pallet {
 			Ok(())
 		}
 
+		pub fn trigger_commission_distribution() {
```
dandanlen, line 1484: I don't think we need this - it's not clear that automatic distribution is required, and there are annoying edge cases to consider (ie. do we send dust, what if the amount is lower than the fee, etc, etc).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1204,6 +1209,66 @@ pub mod pallet {
...
+		/// Note: This extrinsic is secured by the broker that has registered the affiliate account.
+		///
+		/// ## Events
+		///
+		/// - [WithdrawalRequested](Event::WithdrawalRequested)
+		#[pallet::call_index(16)]
+		#[pallet::weight(T::WeightInfo::affiliate_withdrawal_request())]
+		pub fn affiliate_withdrawal_request(
+			origin: OriginFor<T>,
+			short_id: AffiliateShortId,
```
dandanlen, line 1266: I think using the regular AccountId is fine here. Using the ShortId requires the broker to store the mapping on their side, or to look up the id with an rpc call. Easier to just pass the id.

state-chain/pallets/cf-swapping/src/weights.rs
```diff
@@ -377,4 +387,12 @@ impl WeightInfo for () {
 			.saturating_add(ParityDbWeight::get().reads(2_u64))
 			.saturating_add(ParityDbWeight::get().writes(1_u64))
 	}
+	/// TODO: needs to get generated ob benchmark hardware
```
dandanlen, line 390: Remove these TODOs - otherwise we will forget to remove them when we actually do generate real benchmarks.

dandanlen (commented):
Ok I think this is most of what we need - just some additional refactors and then we're good.

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: But we don't need the struct in this case, right? BrokerId -> AffiliateId -> WithdrawalAddress should be enough.

dandanlen: I think it will be useful to include the ShortId here too.

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: I think if we store the details (or just the address as proposed) in a double map, we:
a) have to pass both here the full AccountId and the ShortId 
**or**
b) Assume and ensure that there is only one entry for the second key (ShortId)

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: I see the point. But what are we going to do with existing affiliate accounts? How can they set a withdrawal address then?

Janislav: Indeed, I think we even could enforce the withdrawal address... I don't see why it should be an optional. Worst case is they don't provide an address, and then they have to go through the hustle of withdrawal the funds themself again.

dandanlen: I agree, I think we can enforce it as long as we also generate an affiliate account similarly to the FRAME utilities pallet ([here](https://paritytech.github.io/polkadot-sdk/master/src/pallet_utility/lib.rs.html#491-495)), ie. the affiliate id would effectively just be a hash of (broker_id, short_id).

This way, 'legacy' affiliates can still use the old method: they need to register an account and need to withdraw their funds via extrinsic. 

"New" affiliates can be controlled by the broker, but need to register a payout address to prevent the broker from stealing or losing the funds.

state-chain/pallets/cf-swapping/src/weights.rs
Janislav: I would assume the entire weights file is getting overwritten, when we generated the benchmarks

dandanlen: We create a PR, so it just merges the diff.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1180,30 +1192,82 @@ pub mod pallet {
...
-			let previous_affiliate_id = AffiliateIdMapping::<T>::take(&broker_id, short_id);
+			ensure!(
+				!AffiliateIdMapping::<T>::contains_key(&broker_id, short_id),
+				Error::<T>::AffiliateAlreadyRegistered
+			);
+
+			ensure!(
+				T::AddressConverter::decode_and_validate_address_for_asset(
+					withdrawal_address.clone(),
+					Asset::Eth,
```
dandanlen, line 1221: It doesn't make much difference but actually the asset is Usdc. 

We could also require the address to be an `EthereumAddress`, then we don't need to do this check.

api/lib/src/lib.rs
```diff
@@ -518,6 +519,7 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
 		&self,
 		affiliate_id: AccountId32,
```
dandanlen, line 520: The `affiliate_id` is derived now, so we shouldn't have to pass it in the rpc.

api/lib/src/lib.rs
```diff
@@ -518,6 +519,7 @@ pub trait BrokerApi: SignedExtrinsicApi + StorageApi + Sized + Send + Sync + 'st
...
 		short_id: Option<AffiliateShortId>,
+		withdrawal_address: EncodedAddress,
 	) -> Result<AffiliateShortId> {
```
dandanlen, line 523: We will need to return the derived affiliate id.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1287,6 +1351,42 @@ pub mod pallet {
 				.collect()
 		}
 
+		fn trigger_withdrawal(
+			account_id: &T::AccountId,
+			asset: Asset,
+			destination_address: EncodedAddress,
+		) -> DispatchResult {
+			let destination_address_internal =
```
msgmaxim, line 1359: Instead of storing encoded address and having to verify it on every withdraw request, I think it would make more sense to do that once upon registration (which you already do) and then store the internal address? (You would instead need to use `to_encoded_address` to get the encoded version for `WithdrawalRequested`, but it seems better since converting to encoded cannot fail.)

Edit: I just noticed that `trigger_withdrawal` is also used in `withdraw` where the conversion was previously happening, which I imagine why we do it here now. However, I think it just means we need to move `decode_and_validate_address_for_asset` back to `withdraw` (where it makes sense since it receives encoded address as argument) and have `trigger_withdrawal` take internal address.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1180,30 +1192,82 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Associates `short_id` with `affiliate_id` for a given broker. Overwrites the record
-		/// under `short_id` if already taken by another affiliate.
+		/// Registers an affiliate for a broker. Sets the withdrawal address for the affiliate which
+		/// will be used to withdraw the affiliate's fees. The withdrawal address can not be changed
+		/// once set. The broker will be able to trigger a withdrawal
+		/// request to the affiliate's withdrawal address. A short id is taken and can only be used
+		/// at a time. The affiliate account is derived from the broker account id.
```
msgmaxim, line 1199: Is this what you intended?
```suggestion
		/// once. The affiliate account is derived from the broker account id.
```

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1716,9 +1718,15 @@ mod private_channels {
 		new_test_ext().execute_with(|| {
```
msgmaxim, line 1718: This test could also be moved under the new `affiliates` module below.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1771,3 +1772,98 @@ mod private_channels {
...
+				OriginTrait::signed(BROKER),
+				affiliate_account_id
+			));
+
+			assert_event_sequence!(
+				Test,
+				RuntimeEvent::Swapping(Event::AffiliateRegistration { .. }),
+				RuntimeEvent::Swapping(Event::WithdrawalRequested { .. }),
+			);
+		});
```
msgmaxim, line 1811: I think in addition to checking events, we should also check that the state has actually changed (the affiliate's balance and maybe the fact that the egress API was called depending if not too difficult).

msgmaxim (commented):
I left a few minor comments, but I think the overall approach is good.

state-chain/pallets/cf-swapping/src/lib.rs
Janislav: I see the point, but:

- I think the advantage of using an EncodedAddress is that it decodes to hex string in PolkadotJs, so it's easy to verify it on-chain for the affiliates.
- If we store it as ForeaginChainAddress, we still have to check the asset type when we register the address, not really an advantage I think.
-  @dandanlen proposed to store directly an Ethereum address, but then we have to convert it to ForeaginChainAddress again when we trigger the withdrawal, also not ideal
- The conversion between the brokerAPI and the extrinsic would also be pretty ugly in both of these scenarios.
- We're already using EncodedAddress in the other extrinsic, and I think we should stay consistent

I think the most important thing is though that the address should be easy to verify. That's basically the main feature and if the affiliate has to decode a byte array manually I think that's not very handy in my opinion.

Janislav:
I changed one thing: We used to have the short_id as an optional in the broker API and get it autogenerated if it's not provided. Since we are now only deal with the shortId and this is the main identifier, I think it should be provided/managed by the broker explicitly to avoid any mistakes or confusion. I think this is clearer and makes the code easier (no code = good code).

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: > If we store it as ForeaginChainAddress, we still have to check the asset type when we register the address, not really an advantage I think.

You already have to call `decode_and_validate_address_for_asset` in `register_affiliate`, which gives you `ForeignChainAddress`. I'm simply suggesting that we use that result.

> We're already using EncodedAddress in the other extrinsic, and I think we should stay consistent

I'm not talking about changing any extrinsics. My thinking was that we take encoded address in `register_affiliate` (as you do), convert it to `ForeignChainAddress` (again, you already do that), and then store `ForeignChainAddress`, and not have to verify it on every withdraw request.

> but then we have to convert it to ForeaginChainAddress again when we trigger the withdrawal, also not ideal

Do you mean convert it to the encoded address? If so, that's correct, but converting to encoded is simpler as it cannot fail, which is the main advantage (this is the point I was trying to make above). It seems wrong to me that we have to verify that the address is correct on every withdrawal, even though we have already done it upon registration (as we should). It makes sense that we store address internally using our internal representation and convert it to the human-readable format when interfacing externally (e.g. in the `WithdrawalRequested` event), I thought this has been our convention.

> I think the most important thing is though that the address should be easy to verify.

Yeah, this is a valid point. However, I'm not sure that how we want things to be presented should have a (significant) impact on how they are stored. Are we sure that polkadot.js will be used this way by affiliates? It is somewhat awkward to use, e.g. we can't even provide a link to a page showing the correct pallet/storage item, and instead we would have to provide a somewhat elaborate set of instructions explaining how to use it (it is not very intuitive if you haven't done it before).

Could we not just provide an RPC method to fetch this info (which sounds like a good idea anyway)? Then we could just provide a curl command retrieving the address for a given broker/affiliate. I imagine, some affiliates might not even bother, instead requesting a withdrawal after a couple swaps to verify that everything works end-to-end.

I don't want to delay this PR if I'm the only one who wants this change. @dandanlen if you think storing the encoded address is fine, I'm fine with merging this as is.

dandanlen: I suggested `EthereumAddress` since it decodes as hex, and since it's always an Ethereum address. Going from `EthereumAddress` to `ForeignChainAddress` is easy.

In the rpc we just need to use AddressString and then verify that it's a valid Ethereum address.

Janislav: If EthereumAddress decodes as something useful in PolkadotJs then we can go with it. We still have to do a conversion on every withdrawal though, but if it encodes nicely I agree that this is the better solution.

@msgmaxim I don't think introducing an RPC for the verification is a good idea since the broker could manipulate the outcome with a man-in-the-middle-attack. We should incentivise affiliates to do the verification themselves, otherwise there is no mutch point in registering an unchangeable withdrawal address.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -309,6 +315,14 @@ impl RpcServer for RpcServerImpl {
 			.cf_get_affiliates(self.api.state_chain_client.account_id(), None)
 			.await?)
 	}
+
+	async fn affiliate_withdrawal_request(
+		&self,
+		affiliate_short_id: AffiliateShortId,
+	) -> RpcResult<()> {
```
dandanlen, line 322: It would be good to return some useful information here, for example `WithdrawFeesDetail` like in a normal broker withdrawal.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -296,10 +302,10 @@ impl RpcServer for RpcServerImpl {
 
 	async fn register_affiliate(
 		&self,
-		affiliate_id: AccountId32,
-		short_id: Option<AffiliateShortId>,
-	) -> RpcResult<AffiliateShortId> {
-		Ok(self.api.broker_api().register_affiliate(affiliate_id.clone(), short_id).await?)
+		short_id: AffiliateShortId,
+		withdrawal_address: AddressString,
```
dandanlen, line 306: Hmm... not sure we need to provide an id here. Otherwise the broker needs to query the chain first, to find out what ids are taken.

Previously it was possible re-assign an id, so it kinda made sense, but not any more.

We only expect this to be used for new ids, so we could just generate one, as long as the provided withdrawal address hasn't been registered already. Then we return the generated `AffiliateDetails`.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1031,6 +1039,8 @@ pub mod pallet {
 				Error::<T>::EarnedFeesNotWithdrawn,
 			);
 
+			let _ = AffiliateAccountDetails::<T>::clear_prefix(&account_id, u32::MAX, None);
```
dandanlen, line 1042: We can't do this unless all the affiliate balances are zero...

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -537,6 +548,19 @@ pub mod pallet {
...
 	>;
 
+	/// Stores the details of an affiliate account against the account id of a broker and the
+	/// derived affiliate id.
+	#[pallet::storage]
+	pub type AffiliateAccountDetails<T: Config> = StorageDoubleMap<
+		_,
+		Identity,
+		T::AccountId,
+		Twox64Concat,
```
dandanlen, line 558:
```suggestion
		Identity,
```

nit: since we use the blake hasher to generate this already, it can be Identity.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -309,6 +315,14 @@ impl RpcServer for RpcServerImpl {
 			.cf_get_affiliates(self.api.state_chain_client.account_id(), None)
 			.await?)
 	}
+
+	async fn affiliate_withdrawal_request(
+		&self,
+		affiliate_short_id: AffiliateShortId,
```
dandanlen, line 321: In the other methods with affiliates (like `request_swap_parameter_encoding`) we take the affiliate's AccountId. I think it would be better to stay consistent. 


(Even better, we could accept one or the other, ie. `enum Affiliate { ShortId(..), LongId(..) }`. But better to leave this for a future PR I think.)

dandanlen (commented):
I have a few more comments...

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: > I don't think introducing an RPC for the verification is a good idea since the broker could manipulate the outcome with a man-in-the-middle-attack.

What do you mean? The level of trust is exactly the same whether you use polkadot.js or send an RPC to the same node that polkadot.js is connected to. In fact, polkadot.js works by sending RPC requests to the public node (namely `state_getStorage` methods).

Janislav: No definitely not. There is another layer you have to trust in between if you go over a broker, which may even have an incentive to rug you since he could lie to you which address is registered on the chain for your account to later withdraw the collected funds to his own address. PolkadotJS doesn't care about your affiliate funds, a broker maybe does....

dandanlen (approved):

Janislav:
Started to update the docs here: https://github.com/chainflip-io/chainflip-docs/pull/487 .

PR #5528, opened 2025-01-03
chore: bouncer EVM & BTC Vault Swaps to use Broker API

albert-llimos:
# Pull Request

Closes: PRO-1912, PRO-1854

## Checklist

## Summary

- Use broker APi for EVM vault swaps
- Add test for legacy vault swaps (undecodable cfParameters) to make sure it doesn't break backwards compatibility. This is to test the logic in #5514 .
- Add BTC Vault swap test to the regular swapping flow. Affiliate fee testing shall be done separately. I've opened a ticket for that: https://linear.app/chainflip/issue/PRO-1927/add-bouncer-affiliate-test-for-vault-swaps
- Fix bug in broker API - it was using the keyManager address.

### DIFF
--- a/bouncer/shared/evm_vault_swap.ts
+++ b/bouncer/shared/evm_vault_swap.ts
@@ -140,7 +140,7 @@ export async function executeEvmVaultSwap(
   const refundParams: ChannelRefundParameters = {
     retry_duration: fokParams.retryDurationBlocks,
     refund_address: fokParams.refundAddress,
-    min_price: fokParams.minPriceX128,
+    min_price: '0x' + new BigNumber(fokParams.minPriceX128).toString(16),
   };
 
   const extraParameters: EvmVaultSwapExtraParameters = {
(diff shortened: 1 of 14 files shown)


bouncer/shared/evm_vault_swap.ts
```diff
@@ -67,45 +87,103 @@ export async function executeEvmVaultSwap(
...
-      fillOrKillParams: fokParams,
-      dcaParams,
-      affiliateFees: undefined,
-    } as ExecuteSwapParams,
-    networkOptions,
-    txOptions,
-  );
+  const fineAmount = amountToFineAmount(amountToSwap, assetDecimals(sourceAsset));
+
+  if (Math.random() > 0.5) {
```
albert-llimos, line 92: For now I just set it to random whether to use the BrokerAPI or the SDK. This way we can test boths. There might be the case that eventually the SDK stops supporting that. We will remove it if and when the time comes.

msgmaxim (approved):

PR #5529, opened 2025-01-03
feat: check if node is bonded before deregistering

dandanlen:
# Pull Request

Closes: PRO-1831

## Checklist

## Summary

Adds a check to ensure accounts are unbonded before allowing deregistration. We already implicitly required this but it seemed to good to have an explicit check.

### DIFF
--- a/state-chain/traits/src/lib.rs
+++ b/state-chain/traits/src/lib.rs
@@ -833,6 +833,25 @@ pub trait AccountRoleRegistry<T: frame_system::Config> {
 	}
 }
 
+pub trait DeregistrationCheck {
+	type AccountId;
+	type Error: Into<DispatchError>;
+	fn check(account_id: &Self::AccountId) -> Result<(), Self::Error>;
+}
+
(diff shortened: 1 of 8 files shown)


state-chain/traits/src/lib.rs
```diff
@@ -833,6 +833,25 @@ pub trait AccountRoleRegistry<T: frame_system::Config> {
 	}
 }
 
+pub trait DeregistrationCheck {
+	type AccountId;
+	type Error: Into<DispatchError>;
+	fn check(account_id: &Self::AccountId) -> Result<(), Self::Error>;
+}
+
+impl<A: DeregistrationCheck, B: DeregistrationCheck<AccountId = A::AccountId>> DeregistrationCheck
```
kylezs, line 842: This isn't used anywhere? Was the intention to add another Check in this PR?

dandanlen: I don't think so - it was more that the trait doesn't make much sense unless you can compose multiple. I don't mind removing it.

kylezs: Ok just checking, I don't mind leaving it anyway. up to you

PR #5531, opened 2025-01-06
fix: ccm validity check for vault swaps

syan095:
# Pull Request

Closes: PRO-1926 

## Checklist

## Summary

Moved the CCM validity checks so that all ccms with destinations to Solana chains are checked.

### DIFF
--- a/state-chain/runtime/src/lib.rs
+++ b/state-chain/runtime/src/lib.rs
@@ -47,10 +47,14 @@ use cf_chains::{
 	arb::api::ArbitrumApi,
 	assets::any::{AssetMap, ForeignChainAndAsset},
 	btc::{api::BitcoinApi, BitcoinCrypto, BitcoinRetryPolicy, ScriptPubkey},
+	ccm_checker::{
+		check_ccm_for_blacklisted_accounts, CcmValidityCheck, CcmValidityChecker,
+		DecodedCcmAdditionalData,
+	},
 	dot::{self, PolkadotAccountId, PolkadotCrypto},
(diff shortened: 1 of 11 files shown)


state-chain/runtime/src/lib.rs
```diff
@@ -2233,18 +2237,49 @@ impl_runtime_apis! {
...
+							.map(|key| key.into())
+							.unwrap_or_else(|_| agg_key);
+
+						check_ccm_for_blacklisted_accounts(
+							&ccm_accounts,
+							vec![api_environment.token_vault_pda_account.into(), agg_key, on_chain_key],
+						)
+						.map_err(DispatchError::from)?;
+					},
+					Ok(DecodedCcmAdditionalData::NotRequired) => {},
```
albert-llimos, line 2278: I'm considering whether we should check if the `ccmAdditionalData` is empty in this case and error out if it isn't. When a vault swap is witnessed we don't want to error out when processing the swap. But here when building the call it might make sense to error out to signal clearly to the integrator that no ccm_additional_data will be used when crafting the transaction.
I guess the same case could be argued when opening a deposit channel but at least in that case we are not building any transaction for the user.
Not a strong opinion but I think I'd add it. Wdyt @syan095 ?

albert-llimos (approved):
LGTM! Just one small comment for a minor improvement.

state-chain/runtime/src/lib.rs
syan095: Done!

albert-llimos (approved):

PR #5532, opened 2025-01-06
feat: add bond to broker cf_account_info

j4m1ef0rd:
# Pull Request

Closes: PRO-1911

## Checklist

## Summary

Just added the bond to the `RpcAccountInfo` Broker variant. So now the `cf_account_info` RPC will show the bond.
Updated the snapshot for the RPC, but  I'm pretty sure this will not break anything because its just adding a field.

### DIFF
--- a/state-chain/runtime/src/runtime_apis.rs
+++ b/state-chain/runtime/src/runtime_apis.rs
@@ -111,8 +111,8 @@ pub enum ChainflipAccountStateWithPassive {
 
 #[derive(Encode, Decode, Eq, PartialEq, TypeInfo, Serialize, Deserialize)]
 pub struct ValidatorInfo {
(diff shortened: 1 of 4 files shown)


state-chain/runtime/src/runtime_apis.rs
```diff
@@ -202,6 +202,7 @@ pub struct BrokerInfo {
 	pub earned_fees: Vec<(Asset, AssetAmount)>,
 	pub btc_vault_deposit_address: Option<String>,
 	pub affiliates: Vec<(AffiliateShortId, AccountId32)>,
+	pub bond: u128,
```
msgmaxim, line 205: Might be better to use the `AssetAmount` alias?

msgmaxim (approved):

PR #5533, opened 2025-01-06
fix: break out of the loop if we can't refund anymore

marcellorigotti:
# Pull Request
Closes: PRO-1924

## Checklist

## Summary
Currently we try to refund every address, even if we don't have enough balance. This in Ethereum case caused the emissions of a lot of `VaultDeficitDetected` events. 
This event should be emitted once per chain hence the addition of the `amount > total_withheld` condition to break out of the loop if we can't refund any more account for that chain, such that the event will be emitted only once per chain.

### DIFF
--- a/state-chain/pallets/cf-asset-balances/src/lib.rs
+++ b/state-chain/pallets/cf-asset-balances/src/lib.rs
@@ -296,7 +296,7 @@ impl<T: Config> Pallet<T> {
 				for (destination, amount) in owed_assets.iter_mut() {
 					debug_assert!(*amount > 0);
 					let _ = Self::reconcile(chain, destination, amount, total_withheld);
-					if *total_withheld == 0 {
+					if amount > total_withheld || *total_withheld == 0 {
 						break;
 					}
 				}


state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -296,7 +296,7 @@ impl<T: Config> Pallet<T> {
 				for (destination, amount) in owed_assets.iter_mut() {
 					debug_assert!(*amount > 0);
 					let _ = Self::reconcile(chain, destination, amount, total_withheld);
-					if *total_withheld == 0 {
+					if amount > total_withheld || *total_withheld == 0 {
```
dandanlen, line 299: I think the reason for *not* doing this previously was to maximise the number of accounts that get refunded. Imagine a situation like this: 

```
total_witheld = 14;
owed_amounts = [10, 5, 2, 1, 1]
```

Using the old method, we can refund `[10, 2, 1, 1]` but using the new method we will refund only `[10]`.

Maybe we can do something like this instead?

```
if amount < total_withheld {
	let _ = Self::reconcile(chain, destination, amount, total_withheld);
}
if *total_withheld == 0 {
	break;
}
```

marcellorigotti: Ok gotcha, I thought the main goal was to refund the top spenders.
I moved the `VaultDeficitDetected` event out of the `reconcile` function, such that we check and emit it once per chain, at the end of all the refunds.

dandanlen: Much better 👍

state-chain/pallets/cf-asset-balances/src/tests.rs
```diff
@@ -88,8 +88,8 @@ pub fn not_enough_withheld_fees() {
 
 		System::assert_has_event(RuntimeEvent::AssetBalances(Event::VaultDeficitDetected {
 			chain: ForeignChain::Bitcoin,
-			amount_owed: BTC_OWED,
-			available: BTC_AVAILABLE,
+			amount_owed: BTC_OWED - BTC_AVAILABLE,
+			available: 0,
```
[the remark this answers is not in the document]
marcellorigotti: exactly!

PR #5538, opened 2025-01-08
feat: add expires_at to solana vault swap details

nmammeri:
# Pull Request

Closes: PRO-1851

## Checklist

## Summary

* Implements adding expires_at field to solana vault swap details
* Test added to bouncer sol_vault_swap test


dandanlen (approved):

albert-llimos:
Removing this from the merge queue because we are most likely making some changes in the Solana Vault swaps that will change this. PR: #5541

albert-llimos:
> Would it be worth rebasing this on top of that PR then, so we can get this ready in the meantime @albert-llimos ?

It doesn't really need to be rebased since the code is independant but it could be done for correctness I guess. This will need to be changed anyway as discussed with @nmammeri . Not sure what he agreed with Dan but we are probably changing the expiry approach for all chains.

albert-llimos:
> From what I understood from my recent conversation with @dandanlen we are not providing any expiry time at all. We just make it clear in the docs. So no need for this PR

Closing then as it's no longer applicable.

PR #5540, opened 2025-01-09
refactor: Use `EncodedAddress` for `cf_get_open_deposit_channels`

MxmUrw:
# Pull Request

Closes: PRO-1878

## Checklist

## Summary

This PR changes the type used for addresses returned by `cf_get_open_deposit_channels`. Instead of our internal `ScriptPubKey` we now use `EncodedAddress`, making the jsonrpc response easier to work with for clients not depending on the backend repo.

Since `EncodedAddress` can contain addresses for any chain this also means that there's now no requirement to eventually have multiple fields in the `ChainAccounts` struct containing the result.

#### Non-Breaking changes

This PR breaks the format of the return value of `cf_get_open_deposit_channels` in the runtime_api and of `get_open_deposit_channels` in the broker API.


dandanlen (approved):

albert-llimos:
Is this complete @MxmUrw ? can it be merged?

MxmUrw:
Yes!

MxmUrw:
I'll look into why this is failing tomorrow, I guess it's because in the upgrade test I actually have to migrate from the old version of the CFDM to the new version in order to be able to interact first with the old broker api format and then the new.

MxmUrw:
I fixed the upgrade test. Reason it failed was that it would, before the upgrade, already start the deposit monitor. Because the previous version of the engine/node where still running, the deposit monitor would receive addresses encoded in the wrong (old) format, and would thus perpetually return a "false" healthcheck.

Now the deposit monitor is started after the broker/lp apis after the upgrade, and an additional 10s wait is introduced to make sure that it has time to start up.

kylezs (approved):

PR #5549, opened 2025-01-15
chore: added broker id to SwapDepositAddressReady event

j4m1ef0rd:
# Pull Request

Closes: PRO-1948

Just added the broker id to the event.


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -572,6 +572,7 @@ pub mod pallet {
 			source_asset: Asset,
 			destination_asset: Asset,
 			channel_id: ChannelId,
+			broker_id: T::AccountId,
```
j4m1ef0rd, line 575: Some of the details including `broker_id` are not checked in any test. Do you think they should be?

dandanlen: No, I think this is fine.

dandanlen (approved):

PR #5550, opened 2025-01-15
test: bouncer vault swap fee collection test

j4m1ef0rd:
# Pull Request

Closes: PRO-1927

## Checklist

## Summary

- Refactored the btc vault swap code to use the state chain RPC instead of the broker.
- Moved the btc vault swap code out of the test and into a shared file.
	- Refactored `openPrivateBtcChannel` and `registerAffiliate` functions to be more useful.
- Refactored the `performVaultSwap` and subsequent functions to take broker/affiliate fees and use them correctly.
	- Note that the SDK needed short id's while the broker API and statechain RPC needed the account id, so it got a little messy.
	- Removed the `openPrivateBtcChannel` from being called every swap.
- Renamed and refactored the btc vault test to be the new vault swap fee collection test.
	- Testing all 4 assets
	- Added to all concurrent tests
- Created a new `setupBrokerAccount` function so we don't need to add an account for each fee collection test in genesis. (Maybe we want to make the broker fee collection test use this as well and remove `BROKER_FEE_TEST` from genesis)
- Changed the EVM vault swap to no longer use the broker api 50% of the time. I put the code in a separate function so we can use it later.


albert-llimos:
> Refactored the btc vault swap code to use the SDK instead of the broker.

Is that right? Do you mean EVM? Because we don't really have support for BTC vault swaps in the SDK afaik.

bouncer/pnpm-lock.yaml
albert-llimos, line 1: I don't think this file should have changed, you might have some different pnpm version. Restore the old one.

bouncer/shared/perform_swap.ts
```diff
@@ -257,12 +258,27 @@ export async function executeVaultSwap(
...
+    commissionBps: number;
+  }[] = [],
+  allowUseBrokerApi = true,
 ) {
   let sourceAddress: string;
   let transactionId: TransactionOriginId;
 
   const srcChain = chainFromAsset(sourceAsset);
 
+  const usableBrokerFees = brokerFees ?? {
```
albert-llimos, line 277: not sure what the prefix "usable" indicates. Is there some better naming we can use?

bouncer/tests/vault_swap_fee_collection.ts
```diff
@@ -0,0 +1,100 @@
...
+    testVaultSwapFeeCollection.swapContext,
+    testVaultSwapFeeCollection.debug,
+    depositAmount,
+    0, // boostFeeBps
+    undefined, // fillOrKillParams
+    undefined, // dcaParams
+    { account: broker.address, commissionBps },
+    [{ accountAddress: affiliate.address, accountShortId: affiliateShotId, commissionBps }],
+    // Important that we use our oun broker account and not the broker API account.
+    false, // allowUseBrokerApi
```
albert-llimos, line 75: I'm not sure I fully understand this.
This seems to only be used in EVM Vault swaps so we don't use the broker API. However, for Solana Vault swaps the broker API is being used and the test does pass. So it'd seem as this is unnecessary. What am I missing?

j4m1ef0rd: No , the Sol vault swaps are using the statechain directly `cf_get_vault_swap_details`, same with the btc vault swaps now. I just needed to stop the EVM vault swaps from using the broker api. Maybe we should be more consistent and make all vault swaps have that 50% chance to use the broker api, but that's for another PR.

j4m1ef0rd:
> > Refactored the btc vault swap code to use the SDK instead of the broker.
> 
> Is that right? Do you mean EVM? Because we don't really have support for BTC vault swaps in the SDK afaik.

Sorry, i meant to say "use the statechain RPC instead"

bouncer/shared/perform_swap.ts
j4m1ef0rd: I just wanted to turn it from an option to a value, but i was not allowed to reuse the parameter, couldn't think of a better name.

bouncer/pnpm-lock.yaml
j4m1ef0rd: If you look at the changes, it's just updating `@chainflip/rpc` to the latest, so i think it's fine to keep the changes.

bouncer/tests/vault_swap_fee_collection.ts
albert-llimos: Why don't we just not use the broker API in EVM vault swaps when using the `cf_get_vault_swap_details` as we do in Solana and EVM? I would prefer not adding this `allowUseBrokerApi`.

bouncer/shared/perform_swap.ts
albert-llimos: I'd suggest naming it `brokerFeesValue` or something like that. `usable` doesn't say much, especially since the `option` and `value` concept are not really a thing in TS ( I do understand what you mean though).

albert-llimos (approved):
I'd like this comment addressed if you agree but looks good otherwise: https://github.com/chainflip-io/chainflip-backend/pull/5550#discussion_r1917818634
Good work on the refactoring!

bouncer/tests/vault_swap_fee_collection.ts
j4m1ef0rd: Yeah that's fine with me. Maybe we should have a test just for the broker api stuff, similar to the lp api test. Now that the broker api has a fair few rpcs.

albert-llimos: Look good but now with your latest refactor the `executeEvmVaultSwapViaBrokerApi` is not being used anywhere, right? If so, please open a ticket to add a test for that in the bouncer, otherwise it'll just be forgotten.

PR #5553, opened 2025-01-15
fix: only apply minimum deposit check on deposit process if it hasn't been boosted

Janislav:
# Pull Request

Closes: PRO-1846

## Checklist

## Summary

Introduces a check on full deposit that a channel has not been boosted. The reason for this is that we want to catch the edge case where we would ignore the finalization of a boosted channel when the deposit minimum increases between the prewitness and the full witness stage.

#### Non-Breaking changes

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -1915,7 +1915,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 	/// Completes a single deposit request.
 	#[transactional]
-	pub(crate) fn process_channel_deposit_full_witness_inner(
+	fn process_channel_deposit_full_witness_inner(
 		DepositWitness { deposit_address, asset, amount, deposit_details }: &DepositWitness<
 			T::TargetChain,
 		>,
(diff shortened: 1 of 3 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2249,9 +2247,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		if let Some(tx_id) = deposit_details.deposit_id() {
 			// Only consider rejecting a transaction if we haven't already boosted it,
 			// since by boosting the protocol is committing to accept the deposit.
-			if TransactionsMarkedForRejection::<T, I>::take(broker, &tx_id).is_some() &&
-				!matches!(boost_status, BoostStatus::Boosted { .. })
-			{
+			if TransactionsMarkedForRejection::<T, I>::take(broker, &tx_id).is_some() && !boosted {
```
dandanlen, line 2250: We can avoid reading `TransactionsMarkedForRejection` for boosted transactions by reversing this check (`if !boosted && ..`).

In fact I think we can move the !boosted check into an outer if clause and inline:

```
if !matches!(boost_status, BoostStatus::Boosted { .. }) {
	// check min deposit
	// check for rejection
}
```

dandanlen (approved):

PR #5557, opened 2025-01-16
chore: added broker_id to swap origin

j4m1ef0rd:
# Pull Request

Closes: PRO-1951

## Checklist

## Summary

Added broker id to the `SwapOrigin` & `DepositOrigin` enums.

### DIFF
--- a/state-chain/chains/src/lib.rs
+++ b/state-chain/chains/src/lib.rs
@@ -670,14 +670,16 @@ impl IntoTransactionInIdForAnyChain<NoneChainCrypto> for () {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
-pub enum SwapOrigin {
+pub enum SwapOrigin<AccountId> {
 	DepositChannel {
 		deposit_address: address::EncodedAddress,
 		channel_id: ChannelId,
 		deposit_block_height: u64,
+		broker_id: AccountId,
 	},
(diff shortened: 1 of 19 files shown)


state-chain/chains/src/lib.rs
```diff
@@ -670,14 +670,16 @@ impl IntoTransactionInIdForAnyChain<NoneChainCrypto> for () {
...
+pub enum SwapOrigin<AccountId> {
 	DepositChannel {
 		deposit_address: address::EncodedAddress,
 		channel_id: ChannelId,
 		deposit_block_height: u64,
+		broker_id: AccountId,
 	},
 	Vault {
 		tx_id: TransactionInIdForAnyChain,
+		broker_id: AccountId,
```
j4m1ef0rd, line 682: From what I can tell, the vault swap always has a broker, so no need to have it as a option (as stated in the issue).

dandanlen: Hmm yeah, looks like we default it to id 0x000..000 and 0 bps.

dandanlen: Pretty sure this was an oversight - if it's optional i'd prefer we make it explicit rather than using a dummy value. I'll refactor this on top of this branch.

dandanlen:
Created #5559 to address the mentioned issue.

dandanlen (approved):

PR #5558, opened 2025-01-16
chore: add versioning to Solana's ccm additional data

syan095:
# Pull Request

Closes: PRO-1943

## Checklist

## Summary

A versioning wrapper was added to Solana's CCM additional data. 
Use V0 for the current variant. 
Moved the "NotRequired" Variant to the top so it looks nicer when the Enum is expanded later.

WIP: Check the Storage migrations


albert-llimos:
Pretty sure this doesn't need any storage migrations since we are just introducing it in 1.8. Also the `DecodedCcmAdditionalData` is not being stored anyway. So I'm taking the liberty to set this to review.

state-chain/chains/src/ccm_checker.rs
```diff
@@ -41,10 +41,15 @@ pub trait CcmValidityCheck {
...
+#[derive(Clone, Debug, Decode, PartialEq, Eq)]
 pub enum DecodedCcmAdditionalData {
-	Solana(SolCcmAccounts),
 	NotRequired,
+	Solana(VersionedSolanaCcmAdditionalData),
+}
+
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq)]
+pub enum VersionedSolanaCcmAdditionalData {
+	V0(SolCcmAccounts),
```
albert-llimos, line 52: @dandanlen One thing that I've thought about is that in Solana we currently use Legacy Transactions (legacy messages). The most likely upgrade to this would be to start using Solana `v0` transactions (and therefore `v0` messages). In that case it will be confusing that our `v0` is Solana legacy while our `v1` is their `v0`.
I do know that we are not necessarily versioning the transaction here, it's different data, and therefore there is no need to follow their naming but maybe we can improve it a bit.
The obvious proposal would be calling it `Legacy` but then we are tying it to the Solana naming. Otherwise maybe something more neutral? Wdyt?

albert-llimos (approved):
LGTM. Dan can take a look, just leaving a comment.

state-chain/chains/src/ccm_checker.rs
dandanlen: I think `V0` is fine. I think if the next version looks like `V1(SolCcmAccountIndices)`, for example, that will be clear enough. And if not, we can always rename without breaking compatibility (since it's scale-encoded anyway).

dandanlen (approved):

PR #5560, opened 2025-01-16
fix: use chain-specific account type for refund params

dandanlen:
# Pull Request

Closes: PRO-1920

This might cause some bouncer issues - will await CI and fix tomorrow.


state-chain/chains/src/cf_parameters.rs
```diff
@@ -8,21 +8,25 @@ use sp_core::ConstU32;
 use sp_runtime::{BoundedVec, Vec};
 
 #[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
-pub enum VersionedCfParameters<CcmData = ()> {
-	V0(CfParameters<CcmData>),
+pub enum VersionedCfParameters<RefundAddress, CcmData = ()> {
```
albert-llimos, line 11: I'm not gonna lie, I'm not a big fan of adding this to the `cfParameters` :sweat_smile: . It is what it is.

albert-llimos (approved):
LGTM and passes the bouncer. We do test all the decodings and the FoK (refund addresses) so looks like it's all good and nothing broke :ok_hand:

state-chain/chains/src/cf_parameters.rs
dandanlen: Yes I don't like it either... but it's more correct, strictly speaking. The length of the params (which we check) depends on the length of the address.

PR #5561, opened 2025-01-17
feat: bouncer test timeout diagnosis

j4m1ef0rd:
# Pull Request

Help solve: PRO-1950

## Checklist

## Summary

The problem is that when a timeout happens in the bouncer, we don't know why. Thanks to the executable test class, we know what test timed out, but not where in that test it was up to. until now....

I added a new class that allows us to instrument the any functions that a test often waits for. The code will look at the call stack and match it against a list of currently running tests, then add the task to that test as a tracked task. If a timeout happens on a test, it will see if it had any tasks that it was waiting on and print a message with the name and location of the task.

```sh
=== Boosting-For-Asset test failed (13:59:11) ===
Timeout waiting for:
  Observing Event assetBalances:AccountCredited at file:///Users/jamieford/alt-chainflip-backend/bouncer/tests/boost.ts:126
Error: Timed out after 12s.
```    
```sh
=== Boosting-For-Asset test failed (14:00:55) ===
Timeout waiting for:
  Balance Increase Btc at file:///Users/jamieford/alt-chainflip-backend/bouncer/tests/boost.ts:198
Error: Timed out after 12s.
```

I added this task tracker to the main utility functions that I expect timeouts to happen on, eg `observeEvent`, `observeBalanceIncrease`, etc. This will enable diagnosing of timeouts in all tests, without any change to the individual test code.
It's not a perfect solution, but should cover 90% of timeouts.

### DIFF
--- a/bouncer/shared/executable_test_task_tracker.ts
+++ b/bouncer/shared/executable_test_task_tracker.ts
@@ -4,16 +4,13 @@ export const runningTests = new Set<ExecutableTest>();
 
 // Look at the call stack and find a known test file, matching against the list of all running test. Returns the test and the location of the call.
 function getTestFromStack(): { test: ExecutableTest; location: string } | undefined {
-  const originalFunc = Error.prepareStackTrace;
-
   try {
     const fakeError = new Error();
 
     if (fakeError.stack !== undefined) {
       const stack = fakeError.stack.split('\n');
       for (let i = 1; i < stack.length; i++) {
-        const fileName = stack[i].split(':')[0];
-        const lineNumber = stack[i].split(':')[1];
+        const [fileName, lineNumber] = stack[i].split(':');
         for (const test of runningTests) {
           if (fileName?.includes(test.fileName)) {
             return { test, location: `${test.filePath}:${lineNumber}` };
@@ -22,11 +19,9 @@ function getTestFromStack(): { test: ExecutableTest; location: string } | undefi
       }
     }
   } catch (e) {
-    Error.prepareStackTrace = originalFunc;
     console.error(e);
   }
 
-  Error.prepareStackTrace = originalFunc;
   return undefined;
 }
 
--- a/bouncer/shared/executable_test_task_tracker.ts
+++ b/bouncer/shared/executable_test_task_tracker.ts
@@ -1,0 +1,57 @@
+import { ExecutableTest } from './executable_test';
+
+export const runningTests = new Set<ExecutableTest>();
+
+// Look at the call stack and find a known test file, matching against the list of all running test. Returns the test and the location of the call.
+function getTestFromStack(): { test: ExecutableTest; location: string } | undefined {
+  const originalFunc = Error.prepareStackTrace;
+
+  try {
+    const fakeError = new Error();
+
+    if (fakeError.stack !== undefined) {
+      const stack = fakeError.stack.split('\n');
+      for (let i = 1; i < stack.length; i++) {
+        const fileName = stack[i].split(':')[0];
+        const lineNumber = stack[i].split(':')[1];
+        for (const test of runningTests) {
+          if (fileName?.includes(test.fileName)) {
+            return { test, location: `${test.filePath}:${lineNumber}` };
+          }
+        }
+      }
+    }
+  } catch (e) {
+    Error.prepareStackTrace = originalFunc;
+    console.error(e);
+  }
+
+  Error.prepareStackTrace = originalFunc;
+  return undefined;
+}
+
+export class TaskTracker {
+  private test: ExecutableTest | undefined;
+
+  private taskId: number | undefined;
+
+  constructor(name: string) {
+    const testDetails = getTestFromStack();
+    if (testDetails !== undefined) {
+      this.test = testDetails.test;
+      this.taskId = testDetails.test.startAwaiting(name, testDetails.location);
+    }
+  }
+
+  stopAwaiting() {
+    this.test?.stopAwaiting(this.taskId);
+  }
+}
+
+// Use to detect where timeouts are happening in tests.
+// Call at the start of any utility function that is expected to take some time to compete.
+// It returns a TaskTracker object that can be use to stop the tracking when the task is completed.
+// Please remember to stop the tracker at all places the function stops.
+export function startAwaitTask(name: string): TaskTracker {
+  return new TaskTracker(name);
+}
--- a/bouncer/shared/utils.ts
+++ b/bouncer/shared/utils.ts
@@ -33,6 +33,7 @@ import { newSolAddress } from './new_sol_address';
 import { getChainflipApi, observeBadEvent, observeEvent } from './utils/substrate';
 import { execWithLog } from './utils/exec_with_log';
 import { send } from './send';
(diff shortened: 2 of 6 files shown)


bouncer/shared/executable_test_task_tracker.ts
```diff
@@ -0,0 +1,57 @@
...
+  return undefined;
+}
+
+export class TaskTracker {
+  private test: ExecutableTest | undefined;
+
+  private taskId: number | undefined;
+
+  constructor(name: string) {
+    const testDetails = getTestFromStack();
```
kylezs, line 39: Conceptually we have a hierarchy of:
- Test Runner
- Tests
- Tasks in tests

So I'd expect that we already have the test information when we start the task up. 

Is the concern that we don't want to pass in `test` to each of these functions or? I think passing in test is a much more robust way to go about it. Then we actually have the test object and can do whatever we want with it, add further parameters that aren't necessarily gettable by the trace. 

What we have now is:
1. Start test, containing all details about the test
2. start a task, not passing in the test
3. do some stuff
4. pull the test from a global of all the tests and then we get the test back

There's a lot of complexity that doesn't seem required.

kylezs (commented):
Functionally I think this is super cool! 
Just think we should rethink the structure a bit, we're currently dropping state and then later recreating it in the same flow, which I think we can avoid, and will make it much easier to follow.

bouncer/shared/executable_test_task_tracker.ts
j4m1ef0rd: > I think passing in test is a much more robust way to go about it

I did consider this, but it has a major disadvantage of needed to instrument every test case and more.
It's not just one level deep, the `observerEvent` function can be used inside of another util function (ie `depositLiquidity`), or it can be even deeper, so you would have to instrument every function in this path and hand the reference all the way down. 
My solution may seem "hacky" but it is a blanket approach with no need to do extensive work. So we can always remove or change it later easy.

> There's a lot of complexity that doesn't seem required.

I would say that it's less complex for the person writing a test. They don't need to even know about any of this stuff, just need to use the executable test class like usual.
For the complexity of the implientation, I think it's quite simple Just matching file names from a call stack.

bouncer/shared/executable_test_task_tracker.ts
```diff
@@ -0,0 +1,57 @@
+import { ExecutableTest } from './executable_test';
+
+export const runningTests = new Set<ExecutableTest>();
```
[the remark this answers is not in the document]
j4m1ef0rd: I was unable to use the same system as `executable_test` because that would require referencing every test file at initialization time, this created circular dependencies. Best solution was to just look at a global list of test that is built after initialization.

>  recipe for confusion/bugs

I don't see how. Does not matter how " test flows" are done. Any executable test should always use the `run()` function and work fine with this list of running tests.

bouncer/shared/executable_test_task_tracker.ts
```diff
@@ -0,0 +1,57 @@
...
+        const lineNumber = stack[i].split(':')[1];
+        for (const test of runningTests) {
+          if (fileName?.includes(test.fileName)) {
+            return { test, location: `${test.filePath}:${lineNumber}` };
+          }
+        }
+      }
+    }
+  } catch (e) {
+    Error.prepareStackTrace = originalFunc;
```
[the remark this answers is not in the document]
j4m1ef0rd: left over from old code. I was trying to use the better way of looking at the call stack like in the executable test `Error.prepareStackTrace = function stackTrace(_err, stack)`, but the line numbers it gives are all wrong, could not find out why, so had to use string splitting instead. sad.

bouncer/shared/executable_test_task_tracker.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+// Use to detect where timeouts are happening in tests.
+// Call at the start of any utility function that is expected to take some time to compete.
+// It returns a TaskTracker object that can be use to stop the tracking when the task is completed.
+// Please remember to stop the tracker at all places the function stops.
+export function startAwaitTask(name: string): TaskTracker {
+  return new TaskTracker(name);
+}
```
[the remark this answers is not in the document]
j4m1ef0rd: I like having a function to call with a clear name/purpose. Thought it would make it easier for ppl to implement. You can't rename a constructor.

bouncer/shared/executable_test_task_tracker.ts
msgmaxim: > I would say that it's less complex for the person writing a test. They don't need to even know about any of this stuff, just need to use the executable test class like usual.

I think this is a case the distinction between simple/complex and easy/hard is important. It may be *easy* to use since most of the time you don't have to worry about it. At the same time, it is *complex* since it involves concepts like the call stack and it makes more implicit assumptions about code. Complex code is usually harder to maintain. Not necessarily saying that this is wrong, but there are tradeoffs and we should keep them in mind.

bouncer/shared/utils.ts
```diff
@@ -838,28 +855,33 @@ export async function observeSolanaCcmEvent(
...
               if (expectedPubkey !== event.data.remaining_pubkeys[index].toString()) {
+                task.stopAwaiting();
                 throw new Error(
                   `Unexpected additional account pubkey: ${event.data.remaining_pubkeys[index].toString()}, expecting ${expectedPubkey}`,
                 );
               }
             }
 
             if (event.data.remaining_is_signer.some((value: boolean) => value === true)) {
+              task.stopAwaiting();
```
msgmaxim, line 866: If we have to remember to call `stopAwaiting()` every time we return/throw, it seems error prone (not to mention, noisy). Keep in mind that in JS functions can throw internally, without any indication that it may do so on the calllers side. You would have to defencively put try-catch everywhere to prevent this.

bouncer/shared/executable_test_task_tracker.ts
msgmaxim: We do already have a concept of `SwapContext` which keeps track of all ongoing swaps. Maybe we could instead extend it so in can be used for keeping track of all tests/tasks?

Also, when I introduced swap context, I did find myself passing it around deep into many functions, but in the end I don't think it was a big deal. I would be interested to see an attempt at an alternative implementation that does that, instead of relying on the call stack.

kylezs: Jamie and I were chatting in dms about an idea that Andrew had, to use a testing framework like [vitest](https://vitest.dev/), and a logging framework like https://getpino.io/#/ (both of which were Andrew's recommendations). These are likely to give us what we want (and more), but requires some experimentation. Adding a proper logging library to bouncer is something we've been thinking about for a while.

bouncer/shared/utils.ts
j4m1ef0rd: If the thrown error not within a try catch it will stop the test anyway. The only case that is a problem is if an error is thrown but we have a try catch at a higher level that will cause it to not stop the test, leaving a stale task on the awaiting list. Not to much of a problem, just a misleading log, same problem as the ` swapContext.updateStatus` getting missed by this edge case.

j4m1ef0rd:
closing in favor of #5610

PR #5566, opened 2025-01-17
fix: add ingress-egress pallet config metadata

kylezs:
# Pull Request

## Checklist

## Summary

Was missed when we added these variants. We do this manually to workaround a bug in polkadot-js :/

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -346,6 +346,17 @@ where
 									f.ty::<TargetChainAmount<T, I>>().name("minimum_deposit")
 								}),
 						)
+					})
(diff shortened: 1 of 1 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -346,6 +346,17 @@ where
 									f.ty::<TargetChainAmount<T, I>>().name("minimum_deposit")
 								}),
 						)
+					})
+					.variant(append_chain_to_name!(SetDepositChannelLifetime), |v| {
+						v.index(2).fields(
+							Fields::named()
+								.field(|f| f.ty::<TargetChainBlockNumber<T, I>>().name("lifetime")),
```
dandanlen, line 353: Think i mentioned this before - it seems strange that we alter field name rather than the type name... but that's for another time..

dandanlen (approved):

PR #5568, opened 2025-01-17
fix: map node communication port for polkadot localnet outside of chainflip network range

dandanlen:
# Pull Request

Polkadot localnet was advertising on the default port, which is being used by the chainflip node as the bootnode port.


kylezs (approved):
danke dir

PR #5569, opened 2025-01-17
Feature/pro 1937

marcellorigotti:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

### DIFF
--- a/state-chain/runtime/src/chainflip/bitcoin_elections.rs
+++ b/state-chain/runtime/src/chainflip/bitcoin_elections.rs
@@ -1,4 +1,6 @@
-use crate::{BitcoinChainTracking, BitcoinIngressEgress, Runtime};
+use std::cmp::PartialEq;
+use std::collections::BTreeMap;
+use crate::{BitcoinChainTracking, BitcoinIngressEgress, Block, Runtime};
 use cf_chains::{
 	btc::{self, BitcoinFeeInfo, BitcoinTrackedData},
 	Bitcoin,
@@ -10,6 +12,7 @@ use sp_core::Get;
 use cf_chains::instances::BitcoinInstance;
 
 use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::__private::sp_tracing::event;
 use pallet_cf_elections::{
 	electoral_system::ElectoralSystem,
 	electoral_systems::{
@@ -42,6 +45,7 @@ use pallet_cf_ingress_egress::{
 use scale_info::TypeInfo;
 
 use sp_std::vec::Vec;
+use cf_chains::btc::BlockNumber;
 
 pub type BitcoinElectoralSystemRunner = CompositeRunner<
 	(BitcoinBlockHeightTracking, BitcoinDepositChannelWitnessing),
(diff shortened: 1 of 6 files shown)


state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -299,56 +290,279 @@ impl Hook<btc::BlockNumber, Vec<DepositChannelDetails<Runtime, BitcoinInstance>>
...
+	None,
+}
+
+#[derive(
+	Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Encode, Decode, TypeInfo, MaxEncodedLen,
+)]
+pub struct DepositChannelWitessingProcessor<ChainBlockNumber: Ord, BlockData, Event> {
+	pub blocks_data: BTreeMap<ChainBlockNumber, (BlockData, ChainBlockNumber)>,
+	pub reorg_events: BTreeMap<ChainBlockNumber, Vec<Event>>,
+	pub rules: BoundedVec<BtcRuleType, ConstU32<2>>,
```
MxmUrw, line 370: The rules shouldn't change during the runtime, so we should ensure that they don't have to be stored in storage.
We can encode the rules by adding them as another associated type in the `BWProcessorTypes`:
```
...
   type Rules : Hook<(Self::ChainBlockNumber, Age), Self::Event>;
```
We'll still have a `rules` field here, but it will only contain data when running tests and will be empty when building for real.

```suggestion
        pub rules: Rules,
```

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -299,56 +290,279 @@ impl Hook<btc::BlockNumber, Vec<DepositChannelDetails<Runtime, BitcoinInstance>>
...
+							return false;
+						}
+					}
+				}
+				true
+			})
+			.collect::<Vec<_>>()
+	}
+}
+fn execute_events(events: Vec<BtcEvent>) {
```
MxmUrw, line 466: I think this should also be encoded as a hook in the `BWTypes`, the same way as the `rules: Rules`. Because these two things are what users should be able to customize when they want to modify the behaviour of the processor.

MxmUrw (commented):
Looking good, I added comments related to cleaning things up. Also I think we should move the Processor related stuff into its own file.

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -299,56 +290,279 @@ impl Hook<btc::BlockNumber, Vec<DepositChannelDetails<Runtime, BitcoinInstance>>
...
+		self.blocks_data.retain(|key, (data, age)| *age <= BitcoinIngressEgress::witness_safety_margin().unwrap());
+		self.reorg_events.retain(|key, data| *key > n - BUFFER_EVENTS);
+	}
+
+	fn process_rules(&mut self, last_height: BlockNumber) -> Vec<BtcEvent> {
+		warn!("Processing rules... last_height: {last_height:#?}");
+		let mut last_events: Vec<BtcEvent> = vec![];
+		for (block, (data, last_age)) in self.blocks_data.clone() {
+			warn!("Rules for block {block:?}, age: {last_age:?}, data: {data:?}");
+			for age in last_age..=last_height - block {
```
MxmUrw, line 428: We should probably rename `last_age` to `next_age` because "last" makes it sound like it has already been dealt with, which I think, in fact, it has not.

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -299,56 +290,279 @@ impl Hook<btc::BlockNumber, Vec<DepositChannelDetails<Runtime, BitcoinInstance>>
...
+				last_events = last_events
+					.into_iter()
+					.chain(self.process_rules_for_age_and_block(block, age, &data))
+					.collect();
 			}
+			self.blocks_data.insert(block, (data.clone(), last_height - block + 1));
```
MxmUrw, line 434: I might be misreading sth here, but are we reinserting into the `blocks_data` array which already contains this `block` and `data`? So the same data is going to appear multiple times if it was processed multiple times?

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -299,56 +290,279 @@ impl Hook<btc::BlockNumber, Vec<DepositChannelDetails<Runtime, BitcoinInstance>>
...
+		data: &BlockData,
+	) -> Vec<BtcEvent> {
+		let mut events: Vec<BtcEvent> = vec![];
+		for rule in &self.rules {
+			events = events.into_iter().chain(apply_rule(block, rule, age, data)).collect();
+		}
+		events
+			.into_iter()
+			.filter(|last_event| {
+				for (_, events) in &self.reorg_events {
```
MxmUrw, line 454: I think you can do sth along the lines of
```rust
.filter(|last_event| !self.reorg_events.iter().flat_map(|(_, events)| events).contains(last_event))
```

MxmUrw (commented):
Some more comments!

PR #5571, opened 2025-01-20
chore: don't redact local endpoints

j4m1ef0rd:
# Pull Request

Closes: PRO-1908

## Checklist

## Summary

Added a `is_local_url` function that uses the `ipv4.is_private()` to find out if its a local address in the url and skip the redacting. The secret will still get redacted if one is found in the url, regardless of it being local.

### DIFF
--- a/utilities/src/with_std/redact_endpoint_secret.rs
+++ b/utilities/src/with_std/redact_endpoint_secret.rs
@@ -1,3 +1,4 @@
+use core::net::IpAddr;
 use regex::Regex;
 use serde::Deserialize;
 use std::fmt::{Debug, Display};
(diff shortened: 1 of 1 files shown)


utilities/src/with_std/redact_endpoint_secret.rs
```diff
@@ -162,5 +182,27 @@ mod tests {
...
+			"wss://192.168.0.123/ws/v3/d52****"
+		);
+
+		// Local addresses without secrets should not be redacted
+		assert_eq!(
+			format!("{}", SecretUrl("ws://10.0.0.17".to_string())),
+			"ws://10.0.0.17".to_string()
+		);
+		assert_eq!(
+			format!("{}", SecretUrl("wss://127.0.0.1".to_string())),
```
kylezs, line 200: one for localhost string would be nice too, since it's a bit of a special case

kylezs (approved):
Nice 👍

msgmaxim (approved):

kylezs:
Created an issue to address that flaky test: https://linear.app/chainflip/issue/PRO-1971/fix-broker-level-screening-flaky-test

PR #5573, opened 2025-01-20
chore: add vault swap ingress fee

albert-llimos:
# Pull Request

## Checklist

## Summary

Add missing Vault Swap Ingress Fee.


state-chain/runtime/src/lib.rs
```diff
@@ -1683,9 +1683,12 @@ impl_runtime_apis! {
 
 			let include_fee = |fee_type: FeeTypes| !exclude_fees.contains(&fee_type);
 
-			let (amount_to_swap, ingress_fee) = if include_fee(FeeTypes::Ingress) {
-				remove_fees(IngressOrEgress::Ingress, input_asset, input_amount)
-			} else {
+			// Default to using the DepositChannel fee unless specified.
```
albert-llimos, line 1686: I considered adding a check that both the IngressDepositChannel and IngressVaultSwap are not included because it wouldn't make sense to count both:
```
if include_fee(FeeTypes::IngressDepositChannel) && include_fee(FeeTypes::IngressVaultSwap) {
	return Err(DispatchErrorWithMessage::from("Cannot include both IngressDepositChannel and IngressVaultSwap fees."));
}
```
However it'd then be mandatory to exclude one of the two and that seems annoying. Defaulting to using the deposit channel fee unless it's explicilty excluded for a Vault swap seems better.

state-chain/chains/src/sol.rs
```diff
@@ -277,6 +277,17 @@ impl FeeEstimationApi<Solana> for SolTrackedData {
 		self.calculate_transaction_fee(compute_units_per_fetch)
 	}
 
+	fn estimate_ingress_fee_vault_swap(&self) -> Option<<Solana as Chain>::ChainAmount> {
+		use compute_units_costs::*;
+
+		// Some of the fetches might be batches but we need to estimate pessimistically.
+		let compute_units_per_fetch_and_close = compute_limit_with_buffer(
+			COMPUTE_UNITS_PER_FETCH_AND_CLOSE_VAULT_SWAP_ACCOUNTS + COMPUTE_UNITS_PER_CLOSE_ACCOUNT,
```
dandanlen, line 285: In the calculation in the transaction builder we use `COMPUTE_UNITS_PER_CLOSE_ACCOUNT * number_of_accounts`. Does this mean this is an underestimate?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2689,13 +2690,17 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 	) -> AmountAndFeesWithheld<T, I> {
 		if matches!(origin, &DepositOrigin::DepositChannel { .. }) {
-			Self::withhold_ingress_or_egress_fee(IngressOrEgress::Ingress, asset, available_amount)
+			Self::withhold_ingress_or_egress_fee(
+				IngressOrEgress::IngressDepositChannel,
+				asset,
+				available_amount,
+			)
 		} else {
-			// No ingress fee for vault swaps.
```
dandanlen, line 2694: `// No ingress fee for vault swaps.` optimistic 🙈

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2689,13 +2690,17 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		origin: &DepositOrigin<T, I>,
 	) -> AmountAndFeesWithheld<T, I> {
 		if matches!(origin, &DepositOrigin::DepositChannel { .. }) {
-			Self::withhold_ingress_or_egress_fee(IngressOrEgress::Ingress, asset, available_amount)
+			Self::withhold_ingress_or_egress_fee(
```
dandanlen, line 2693: nit: i think it better if we write this using a normal match statement (`match origin {
	Deposit {..} => .., Vault {..} => .., }`) in case we add a new variant and then forget to handle it correctly.

state-chain/runtime/src/lib.rs
dandanlen: Hmm yeah this is a bit annoying: if you want to use vault fee estimation you need to *exclude* deposit channel ingress fees... I guess this is ok, we need to use *some* default.

dandanlen (commented):
Looks fine

state-chain/chains/src/sol.rs
albert-llimos: no. Each Vault swap will pay `COMPUTE_UNITS_PER_FETCH_AND_CLOSE_VAULT_SWAP_ACCOUNTS + COMPUTE_UNITS_PER_CLOSE_ACCOUNT` and then the batch transaction will cost `COMPUTE_UNITS_PER_CLOSE_ACCOUNT * number_of_accounts` while the total amount collected will have been `number_of_accounts * (COMPUTE_UNITS_PER_FETCH_AND_CLOSE_VAULT_SWAP_ACCOUNTS + COMPUTE_UNITS_PER_CLOSE_ACCOUNT)` which will always be larger.

dandanlen (approved):

PR #5575, opened 2025-01-20
fix: retry submission on StateDiscarded error

dandanlen:
This is in response to some keygen issues noticed on our engines recently.

### DIFF
--- a/engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
+++ b/engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
@@ -155,6 +155,7 @@ pub struct SubmissionWatcher<
 
 pub enum SubmissionLogicError {
 	NonceTooLow,
+	StateDiscarded,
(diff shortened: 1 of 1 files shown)


engine/src/state_chain_observer/client/extrinsic_api/signed/submission_watcher.rs
```diff
@@ -282,6 +283,13 @@ impl<'a, 'env, BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static>
 
 							self.runtime_version = new_runtime_version;
 						},
+						ClientError::Call(obj)
+							if obj.code() == 1002 &&
+								obj.message().contains("State already discarded") =>
```
kylezs, line 288: makes it a little less likely this breaks later with updates
```suggestion
								obj.message()
								    .to_lowercase()
						            .contains("state already discarded") =>
```

kylezs (approved):

PR #5577, opened 2025-01-21
fix: use is_vote_needed to filter redundant deposit votes

dandanlen:


state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -96,7 +97,65 @@ where
 
 		Ok(())
 	}
+
+	/// Refreshes the election for the given identifier with updated properties if `state`
+	/// is different from the election properties.
+	///
+	/// NOTE: This assumes that the accounts in the properties are a superset of the accounts in the
+	/// state. This holds because whenever we
```
dandanlen, line 105: comment is incomplete :)

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -96,7 +97,65 @@ where
...
+		ensure!(
+			state.keys().all(|account| properties.contains_key(account)),
+			CorruptStorageError::new()
+		);
+
+		// if the keys in state and channels don't match up, or if the balances don't match up
+		// then we refresh the election
+		if state.iter().any(|(account, state_balance)| match properties.get(account) {
+			Some((_, property_balance)) => state_balance.amount != property_balance.amount,
+			None => true,
```
dandanlen, line 128: Can we ever reach this condition? Isn't it excluded by the `ensure!` check above? (not that it matters much...)

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -267,28 +326,20 @@ where
...
-				}
+			for closed_channel in closed_channels {
+				pending_ingress_totals.remove(&closed_channel);
+				channels.remove(&closed_channel);
+			}
+
+			if channels.is_empty() {
+				election_access.delete();
 			} else {
-				election_access.set_state(pending_ingress_totals)?;
```
dandanlen, line 286: > `election_access.set_state(pending_ingress_totals)?;`

Am I right that this is no longer needed at all? And if we never call set_state, do we need to read the `pending_ingress_totals` from state at all?

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
MxmUrw: True.

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
MxmUrw: Yes, it was supposed to be catched by the case above, and it would have been possible to simply `None => return Err(..)`.

Anyways, I have rewritten quite a bit now.

MxmUrw:
Pushed a fix where I take into account both the unsynchronized state map and the election settings for creating the election properties.

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -174,116 +174,84 @@ where
...
+							new_totals.push((account.clone(), *new_total));
 						},
 						Ordering::Greater => {
 							Sink::on_ingress_reverted(
 								account.clone(),
-								details.asset,
-								ingress_total.amount - previous_amount,
+								*asset,
+								*ingress_amount - *initial_amount,
 							);
```
MxmUrw, line 223:
```suggestion
							new_totals.push((account.clone(), *new_total));
```
I think we also want to update the election properties when the amount decreased, since this is the value we had consensus on.

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -159,10 +157,7 @@ where
 			<Self::Vote as VoteStorage>::Vote,
 		),
 	) -> bool {
-		proposed_partial_vote.into_iter().any(|(account, new_balance)| {
-			new_balance.amount !=
-				current_partial_vote.get(&account).map(|total| total.amount).unwrap_or_default()
-		})
+		current_partial_vote != proposed_partial_vote
```
dandanlen: Was this intentionally removed or was it a side-effect of the merge/patch?

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
dandanlen: I'm actually not all sure if/how we can come to consensus on a decrease...

state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
dandanlen: As discussed - this might actually make sense :)

dandanlen:
Replaced with #5585

PR #5578, opened 2025-01-21
fix: various small fixes for 1.7.7 to help with reliability and scalability

kylezs:
- Set search transaction history to true. This ensures the engines will witness Solana broadcast success even if the broadcast was a while ago. 
- Catch state discarded error that can occur when the SC has a reorg. This was previously causing some engines that saw the reorg to shutdown. 
- Fix Solana deposit vote filtering by refreshing the election properties with latest channel balances
- Filter out duplicate liveness votes
- Bump version 1.7.7

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
@@ -53,6 +53,7 @@ where
 	Sink: IngressSink<DepositDetails = ()> + 'static,
 	Settings: Parameter + Member + MaybeSerializeDeserialize + Eq,
 	<Sink as IngressSink>::Account: Ord,
+	<Sink as IngressSink>::Amount: Default,
 	ValidatorId: Member + Parameter + Ord + MaybeSerializeDeserialize,
 {
 	pub fn open_channel<ElectoralAccess: ElectoralWriteAccess<ElectoralSystem = Self>>(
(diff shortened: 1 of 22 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/blockchain/delta_based_ingress.rs
```diff
@@ -184,13 +184,21 @@ where
 			};
 
 			let mut closed_channels = Vec::new();
-			for (account, (details, _)) in &channels {
+			let mut new_properties: Self::ElectionProperties = BTreeMap::new();
+			let old_properties = channels.clone();
+			for (account, (details, old_consensus_ingress_total)) in &channels {
```
ramizhasan111, line 189: I think we can just mutate the ```channels``` instead of creating the ```new_properties```. we could have mutable references to the destructured components of the ```channels``` (which is properties) in this for loop. And then we would only update ```consensus_ingess_total``` value whereever needed (only in the Some()) and then later check ```channels``` != ```old_properties```. It looks like the ```new_properties``` variable is actually redundant.

dandanlen: For sure there are efficiency gains to be made but for now I think we can keep what we have - most of this will be re-written anyway to remove the tracking of pending_ingress_totals.

MxmUrw: Yes you're right, it seems that `new_properties` is actually redundant. I talked with @dandanlen about this, and his opinion was to keep things for the moment as-is in order to not "over-optimize". At least in the current way we treat "channels" exactly the same way we did before, and the behavioural changes are restricted to new variables.

PR #5586, opened 2025-01-25
fix: only set new properties if action dispatched

kylezs:


dandanlen:
Closing - we can easily recover from this is if required. Will fix on main (see #5585 )

PR #5587, opened 2025-01-27
chore: Preparation for transaction v0 work.

syan095:
No logical changes.

This PR moved some code around to prepare the codebase for Solana's Versioned Transaction V0 implementation.

Renamed SolMessage -> SolLegacyMessage, SolTransaction -> SolLegacyTransaction.

Moved primitive types other than Message and Transaction to "primitives.rs".
Moved Messge and Transaction into `legacy` mod.
Moved tests to dedicated `test.rs` mod.


albert-llimos:
I think it definitely makes sense to do these changes. Only two comments:
- In the SDK the "LegacyTransaction" is just called "Transaction" but I'm fine renaming it to LegacyTransaction for clarity. Same with the message. I actually think this is more clear than the SDK, just pointing it out when you pull more code from there.
- I think now that we are reshuffling this we should move all the primitives to `foreign-chains/solana/sol-prim`, not sure why it's not there already. Basically the part of the code that is copied from Solana (e.g. Transaction and Message declarations, bpf_uploader instructions...)

syan095:
I've renamed the Transaction and Message in the Legacy module so they are more clear.
I've moved a few modules into sol-prim folder, but left all the "instruction" related mods inside the cf-chains - I feel like these are constructing Instructions to be used and should be part of the "Chain" instead of "primitives". 

The Message and Transaction are also left in the cf-chains for a few reasons:
* as they are tightly coupled through feature (runtime-integration-test feature)
* they are not copy-pasted from Solana 100%, but modified
* They make use of basic "primitives" types to make more complicated structs specific to the Solana chain- so it feels right that they belong to the cf-chains crate.
* Message and Transaction are more complex structs, not simple, basic "primitive" structs.

albert-llimos (approved):
I think this looks much cleaner than before. Also you're right it makes sense to leave Message and Transaction in cf-chains.

PR #5589, opened 2025-01-27
feat: `VoterApi::vote` now returns an Option.

MxmUrw:
# Pull Request

Closes: PRO-1979

## Checklist

## Summary

The voter api can now signal that it doesn't have anything to vote for by returning `Ok(None)`. Previously, in such cases we had to return `Err(..)`, which lead to the RetrierClient scheduling an exponential backoff.

### DIFF
--- a/engine/src/elections.rs
+++ b/engine/src/elections.rs
@@ -22,7 +22,7 @@ use std::{
 	collections::{BTreeMap, HashMap},
 	sync::Arc,
 };
-use tracing::{error, info, warn};
+use tracing::{debug, error, info, warn};
 use voter_api::CompositeVoterApi;
 
 const MAXIMUM_CONCURRENT_FILTER_REQUESTS: usize = 16;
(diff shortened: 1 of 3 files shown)


engine/src/elections.rs
```diff
@@ -176,6 +176,9 @@ where
 
 						pending_submissions.insert(election_identifier,	(partial_vote, vote));
 					},
+					Ok(None) => {
+						info!("Voting task for election '{:?}' returned 'None' (nothing to submit).", election_identifier);
```
dandanlen, line 180: nit: I feel like debug would be more appropriate here otherwise we'll see a lot of spam.

dandanlen:
```suggestion
						debug!("Voting task for election '{:?}' returned 'None' (nothing to submit).", election_identifier);
```

kylezs (approved):
Agree re: the log level otherwise looks good 👍

PR #5590, opened 2025-01-27
refactor: always search history for signature statuses

kylezs:
# Pull Request

Closes: PRO-1976

## Checklist


albert-llimos (approved):
Any reason why you removed that everywhere instead of leaving it programmable? Is it because we believe this call is problematic with the flag set to false to we hardcode it always to true?

kylezs:
> Any reason why you removed that everywhere instead of leaving it programmable? Is it because we believe this call is problematic with the flag set to false to we hardcode it always to true?

We don't use it anywhere else, so no point having an extra parameter that is always true.

dandanlen (approved):

PR #5596, opened 2025-01-30
fix: bouncer sending too many eth txs at once priority issue

kylezs:
# Pull Request

Closes: PRO-1990

## Checklist

## Summary


dandanlen:
I think the issue with this solution is that the mutex doesn't apply between processes. So when we kick off a new test in a new process, the nonce value can conflict with the previous test. 

Apparently we can get around this by using a file descriptor as a cross-process mutex...

PR #5599, opened 2025-02-03
feat: Add base for versioned solana transaction

syan095:
## Summary

Added versioned transaction and versioned message support
no actual logic change in Solana transaction building

### DIFF
--- a/foreign-chains/solana/sol-prim/src/consts.rs
+++ b/foreign-chains/solana/sol-prim/src/consts.rs
@@ -18,8 +18,6 @@ pub const MAX_BASE58_LEN: usize = 44;
 /// Bit mask that indicates whether a serialized message is versioned.
 pub const MESSAGE_VERSION_PREFIX: u8 = 0x80;
 
-pub const SIGNATURE_BYTES: usize = 64;
-
 pub const fn const_address(s: &'static str) -> Address {
 	Address(bs58_array(s))
 }
--- a/foreign-chains/solana/sol-prim/src/consts.rs
+++ b/foreign-chains/solana/sol-prim/src/consts.rs
@@ -10,6 +10,16 @@ pub const SOLANA_PDA_MAX_SEEDS: u8 = 16;
 pub const SOLANA_PDA_MAX_SEED_LEN: usize = 32;
 pub const SOLANA_PDA_MARKER: &[u8; 21] = b"ProgramDerivedAddress";
 
+pub const HASH_BYTES: usize = 32;
+
+/// Maximum string length of a base58 encoded pubkey
+pub const MAX_BASE58_LEN: usize = 44;
+
+/// Bit mask that indicates whether a serialized message is versioned.
+pub const MESSAGE_VERSION_PREFIX: u8 = 0x80;
+
+pub const SIGNATURE_BYTES: usize = 64;
+
 pub const fn const_address(s: &'static str) -> Address {
 	Address(bs58_array(s))
 }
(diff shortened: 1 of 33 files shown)


dandanlen:
FYI a change to Cargo.lock snuck into that last commit. I removed it and replaced the commit.

state-chain/chains/src/sol/sol_tx_core/tests.rs
```diff
@@ -40,6 +40,102 @@ enum BankInstruction {
...
+	};
+
+	use crate::sol::sol_tx_core::transaction::{
+		v0::VersionedMessageV0, VersionedMessage, VersionedTransaction,
+	};
+
+	use super::*;
+
+	#[test]
+	fn create_simple_tx() {
```
dandanlen, line 57: What is this testing? There aren't any assertions. Just checking we don't panic?

state-chain/chains/src/sol/sol_tx_core/transaction.rs
dandanlen, line 1: What's the difference between sol-prim and sol_tx_core? Like why not just put all this in the sol-prim crate?

dandanlen: It would make it clearer which of the code is purely ours, and what (if any) adjustments have been made to the original sol libs.

albert-llimos: See this previous comment: https://github.com/chainflip-io/chainflip-backend/pull/5587#issuecomment-2617793647

dandanlen: Thanks - tbh. I think it would have been more consistent to put all of the solana core stuff like Transaction, Message etc. in the `sol-prim` crate (or if we don't consider them 'primitive' enough, then it could be an adjacent a `sol-tx` crate). 

Right now it's not really clear what the criteria are for one or the other. Some of what is in sol-prim also has modifications (for no-std compatibility). So the distinction seems a bit arbitrary.

Having said that, I think it's ok to leave it as is, I understand it might be a major pain to change. It would be good to add some comments explaining if/where the code has been modified and why.

albert-llimos: Yes I totally agree. Actually that explanation from Roy is because I made pretty much the same comment, saying we should pull all that code into `sol-prim`.
Then we leave it as is and maybe we do the refactor at some point.

albert-llimos (approved):
@syan095 Good work on the PR and thanks for the refactoring, I think it's a lot better like this.
LGTM but we can wait for Dan to take one last look at it.

state-chain/chains/src/sol/sol_tx_core/tests.rs
syan095: Removed

foreign-chains/solana/sol-prim/src/consts.rs
```diff
@@ -10,6 +10,16 @@ pub const SOLANA_PDA_MAX_SEEDS: u8 = 16;
...
 
+pub const HASH_BYTES: usize = 32;
+
+/// Maximum string length of a base58 encoded pubkey
+pub const MAX_BASE58_LEN: usize = 44;
+
+/// Bit mask that indicates whether a serialized message is versioned.
+pub const MESSAGE_VERSION_PREFIX: u8 = 0x80;
+
+pub const SIGNATURE_BYTES: usize = 64;
```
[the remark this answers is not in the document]
syan095: Removed

syan095:
OK looked like everyone wants everything moved into sol-prim. so I tried and it's actually not too bad, everything just all have to be moved together. 

Now all the Solana SDK stuff is in the Sol-prim crate, and all our types are in the sol_tx_core. 

All sol-prim types are re-exported via the sol_tx_core, so most of the codebase is unaffected

syan095:
And also i moved the ALT stuff into its own file.

foreign-chains/solana/sol-prim/src/lib.rs
```diff
@@ -568,3 +625,188 @@ impl FromStr for Hash {
 		}
 	}
 }
+
```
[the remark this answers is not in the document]
syan095: got it!

dandanlen (approved):
Thanks Roy, I realise it was a bit of a chore to shift everything around again, but I think it's clearer.

PR #5607, opened 2025-02-04
fix: recover boost funds

dandanlen:
Does 2 things:

1. Reads boost pool state, reverses the effects of losing the funds (basically, restores the bool to prewitnessed state as of SC block `0xffb441c0d6c45c79b88dd2e31b7ed6355cffae14761bf8c45fdc351ccad35f43`). Then applies the same effect as confirming the deposit, meaning the balances are distributed to the boosters.

2. Inserts the missing utxo into the available utxo set. 

The try-runtime checks show that exactly 1BTC is recovered:

```
[2025-02-04T20:05:44Z INFO  state_chain_runtime] ✅ Boost contributions recovered
[2025-02-04T20:05:44Z INFO  state_chain_runtime] Recovered boost liquidity: 100000000
```

There is also a check to ensure the utxo is in the utxo set post-migration.


state-chain/runtime/src/lib.rs
```diff
@@ -1285,116 +1287,109 @@ type PalletMigrations = (
...
+		if !Self::should_run() {
+			return Weight::zero();
 		}
+
+		pallet_cf_ingress_egress::Pallet::<Runtime, BitcoinInstance>::recover_boost_deposit(
+			BtcAsset::Btc,
+			POOL_TIER,
+			migrations::boost_recovery::PREWITNESSED_DEPOSIT_ID,
+			migrations::boost_recovery::scaled_boost_contributions_from_raw(),
+			Default::default(),
```
msgmaxim, line 1324: Does this mean we had no withdrawals at the time the deposit was considered "lost"?

state-chain/pallets/cf-ingress-egress/src/boost_pool.rs
```diff
@@ -404,6 +404,34 @@ where
...
+		if self
+			.pending_boosts
+			.insert(prewitnessed_deposit_id, booster_contributions)
+			.is_some()
+		{
+			log::warn!("Can't recover a deposit if it already exists. Aborting.");
+			return Default::default();
+		}
+
+		for booster_id in pending_withdrawals {
```
msgmaxim, line 425: We need to be sure that `pending_withdrawals` only include accounts who contributed to the "lost" deposit (i.e. not just accounts who were withdrawing at the time, though it is possible that they are the same sets). Otherwise these pending_withdrawals entries will never be removed. I think we can check this by looking at `booster_contributions`.

However, I see that we are providing an empty set in the migration, so this won't matter in that case.

state-chain/runtime/src/lib.rs
dandanlen: Yes, it was empty.

dandanlen: But I suppose it would have been more correct to take it from the snapshot.

martin-chainflip (approved):

PR #5609, opened 2025-02-05
fix: no gaps in root btc source

kylezs:
# Pull Request

## Checklist

## Summary

Fixes an extreme BTC witnessing edge case that's only currently possible when a sub 5 second bitcoin block occurs. In most cases this is handled. But can occur under very specific rotation conditions. This will be more simply resolved with the upcoming larger witnessing refactor.

### DIFF
--- a/engine/src/witness/btc/source.rs
+++ b/engine/src/witness/btc/source.rs
@@ -89,9 +89,11 @@ where
 
 						let yield_new_best_header: bool = match &mut stream_state {
 							Some(state)
-								if state.last_block_yielded_index.saturating_add(1) ==
-									best_block_header.height || state.last_block_yielded_hash !=
-									best_block_header.hash =>
+								// We want to immediately yield the new best header if we've reorged on the same block
+								// or it's the next block we expect
+								if (state.last_block_yielded_index == best_block_header.height &&
+									state.last_block_yielded_hash != best_block_header.hash) ||
+									state.last_block_yielded_index.saturating_add(1) == best_block_header.height =>
 								true,
 							// If we don't yet have state (we're initialising), then we want to
 							// yield the new best header immediately
--- a/engine/src/witness/btc/source.rs
+++ b/engine/src/witness/btc/source.rs
@@ -101,9 +101,9 @@ where
 								if state.last_block_yielded_index < best_block_header.height =>
 							{
 								// Update the state for the next iteration to backfill
-								stream_state.as_mut().map(|state| {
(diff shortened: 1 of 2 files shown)


engine/src/witness/common/chain_source/strictly_monotonic.rs
```diff
@@ -83,14 +83,15 @@ mod test {
 					Header { index: 4u32, hash: (), parent_hash: Some(()), data: () },
 					Header { index: 3u32, hash: (), parent_hash: Some(()), data: () },
 					Header { index: 2u32, hash: (), parent_hash: Some(()), data: () },
+					Header { index: 10u32, hash: (), parent_hash: Some(()), data: () },
```
dandanlen, line 86: Is the addition of this to test case meaningful? How? (Would it have previously not passed?)

engine/src/witness/btc/source.rs
```diff
@@ -40,22 +47,88 @@ where
...
+						}
+
 						tick.tick().await;
 
 						let best_block_header = client.best_block_header().await;
-						if last_block_hash_yielded != Some(best_block_header.hash) {
+
+						let yield_new_best_header: bool = match &mut stream_state {
+							Some(state)
+								if state.last_block_yielded_index.saturating_add(1) ==
```
MxmUrw, line 92: What purpose does the `state.last_block_yielded_index.saturating_add(1) == best_block_header.height` serve? It seems that the hash check should be enough, no?

engine/src/witness/common/chain_source/strictly_monotonic.rs
kylezs: It's not meaningful, it's just not obvious that it allows skipping blocks without that. Which informed me a little on how the rest of this would/could work

engine/src/witness/btc/source.rs
```diff
@@ -40,22 +47,90 @@ where
...
+							// yield the new best header immediately
+							None => true,
+							// If we've progressed by more than one block, then we need to backfill
+							Some(state)
+								if state.last_block_yielded_index < best_block_header.height =>
+							{
+								// Update the state for the next iteration to backfill
+								if let Some(state) = stream_state.as_mut() {
+									state.best_known_block_index = best_block_header.height;
+								}
```
MxmUrw, line 108: We already have a mutable reference to `state`, why are we creating a new one? I'm actually surprised that this works, because it looks like we have two mut references to the same thing?

MxmUrw (approved):
LGTM!

PR #5610, opened 2025-02-06
feat: use vitest to run bouncer

j4m1ef0rd:
# Pull Request

Closes: PRO-1995

## Checklist

## Summary

Please read the `README.md` file for instructions on how to use `vitest` with the bouncer.

- Removed the `ExecutableTest` class.
	- Refactored all tests to use `vitest` instead.
- Removed most of the old test command files.
- Replaced the `all_concurrent.ts` file with `fast_bouncer.test.ts`.
- Replaced the `run_test.ts` file with `full_bouncer.test.ts`.
- Changed the `full_bouncer.sh` & `run.sh` scripts to use `vitest`.
- Added pino as a logger.
	- Set pino to output to the cli at `info` level and to a file `bouncer.log` at `trace` level. The output is not very useful yet, until all the console logs's are replaced in PRO-1996.
	- Refactored the tests to use the logger. (only top level for now).
- Made `vitest` print the report of the `SwapContext` after each test.


bouncer/tests/boost.ts
```diff
@@ -128,7 +132,7 @@ async function testBoostingForAsset(asset: Asset, boostFee: number, lpUri: strin
   // Do a swap
   const swapAsset = asset === Assets.Usdc ? Assets.Flip : Assets.Usdc;
   const destAddress = await newAddress(swapAsset, 'LP_BOOST');
-  testBoostingSwap.debugLog(`Swap destination address: ${destAddress}`);
+  console.log(`Swap destination address: ${destAddress}`);
```
msgmaxim, line 135: Should this be `logger.log`?

msgmaxim (approved):

bouncer/tests/boost.ts
j4m1ef0rd: Because these functions are used outside of the test, I left them as console log for now to avoid work that will be covered in the next task PRO-1996.

bouncer/README.md
```diff
@@ -48,5 +48,57 @@ The following commands should be executed from the bouncer directory.
   `pnpm prettier:write`
 - Check linting:<br>
   `pnpm eslint:check`
+  `pnpm tsc --noEmit`
```
kylezs, line 51: technically this is a compiler check rather than linting

bouncer/README.md
```diff
@@ -48,5 +48,57 @@ The following commands should be executed from the bouncer directory.
...
+- only use the given logger. ie do not use any `console.log()`.
+
+### Running the test
+
+To run the test you must add it to one of the `.test.ts` files. We have 3 options. Depending on where you add the test, depends on how it will be ran by the ci.
+
+- Run on the ci = add to the `full_bouncer.test.ts` file.
+  - Run concurrently with the other tests = add to `ConcurrentTests` section.
+  - Run by its self = add to `SerialTests` section.
+- Not ran by the ci, just ran manually = add to the `other.tests.ts` file.
```
kylezs, line 86: Not sure what this is referring to? ci-development will run `run.sh` by default, which is what is run on a normal branch pipeline. But ci-main-merge queue runs `full_bouncer` which includes these tests that you've put in "OtherTests"

bouncer/README.md
```diff
@@ -48,5 +48,57 @@ The following commands should be executed from the bouncer directory.
...
+```
+
+Now that the test is added, we can run it using the `vitest` command.
+
+```sh copy
+pnpm vitest run -t "myNewTest"
+```
+
+If your test uses the `SwapContext` within the `TestContext`, then the report will be automatically logged when the test finishes.
+If you would like to run your test with custom arguments, then you will have to create a test command file. See the `test_commands` folder for examples.
```
kylezs, line 104: Why do we need test_commands at all? Don't necessarily have to do in this PR, but this implies that it's not possible to get rid of it? Seems like we should be able to, by refactoring the tests, for example, so stress_test is a ts test instead of a bash script.

bouncer/full_bouncer.sh
```diff
@@ -1,17 +1,4 @@
 set -e
 echo "Running full bouncer 🧪"
 ./setup_for_test.sh
-./commands/run_test.ts "Gas-Limit-Ccm-Swaps"
-./tests/all_concurrent_tests.ts $1
-./commands/run_test.ts "Rotates-Through-BTC-Swap"
-./commands/run_test.ts "BTC-UTXO-Consolidation"
-./commands/run_test.ts "Rotation-Barrier"
-./commands/run_test.ts "Minimum-Deposit"
-./commands/run_test.ts "Solana-Vault-Settings-Governance"
```
kylezs, line 10: yeah, as a above this is actually changing the current behaviour of what tests are run on CI and when. We should still run these tests in CI as part of full bouncer, but not as part of `run.sh` (which we could rename to be fast_bouncer or something)

bouncer/tests/full_bouncer.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import { testCancelOrdersBatch } from './create_and_delete_multiple_orders';
+import { depositChannelCreation } from './request_swap_deposit_address_with_affiliates';
+import { testBrokerLevelScreening } from './broker_level_screening';
+import { legacyEvmVaultSwaps } from './legacy_vault_swap';
+import { testFundRedeem } from './fund_redeem';
+import { testSwapAfterDisconnection } from './swap_after_temp_disconnecting_chains';
+import { concurrentTest, serialTest } from '../shared/utils/vitest';
+
+// Tests that will run in parallel
+describe('ConcurrentTests', () => {
```
kylezs, line 26: same as above^ concurrent tests shouldn't be the same as full bouncer

bouncer/tests/gaslimit_ccm.ts
```diff
@@ -368,9 +373,10 @@ async function spamChain(chain: Chain) {
   }
 }
 
-export async function main() {
+export async function testGasLimitCcmSwaps(testContext: TestContext) {
+  const logger = testContext.logger;
```
kylezs, line 377: rather than do this everywhere, we should just create `info`, `debug`, `warn` etc. functions *on* testContext that pass through to the underlying logger. Calling code shouldn't need to know / care that there's a logger field, only that the testContext can log

bouncer/tests/other.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe } from 'vitest';
+import { testBtcUtxoConsolidation } from './btc_utxo_consolidation';
+import { testDeltaBasedIngress } from './delta_based_ingress';
+import { testDoubleDeposit } from './double_deposit';
+import { testRotatesThroughBtcSwap } from './rotates_through_btc_swap';
+import { testRotateAndSwap } from './rotation_barrier';
+import { testSolanaVaultSettingsGovernance } from './solana_vault_settings_governance';
+import { serialTest } from '../shared/utils/vitest';
+
+// Tests that are not run by the ci ("full bouncer") but can still be ran manually by using the command:
```
kylezs, line 10: as above^, they should be run

bouncer/README.md
j4m1ef0rd: no, the full_bouncer.sh script only runs the tests in `./tests/full_bouncer.test.ts`.
But yeah, maybe i should make it more clear about what ci. Maybe a table will be better.

bouncer/README.md
j4m1ef0rd: unfortunately I have not found a way of passing arguments to a test when it is ran via `vitest`, so if you want to use a test like a command and run it with custom params given via the cli, you will have to create a script to do that.
At least with a command script we can make it user friendly using `yargs`, unlike when we were running the test with extra args with the executable test class. I think this is fine since its quite rare.

bouncer/tests/gaslimit_ccm.ts
j4m1ef0rd: Already done.

bouncer/full_bouncer.sh
j4m1ef0rd: Yes, my bad. I like that `fast_bouncer` it is.

bouncer/run.sh
kylezs, line 1: we want to also rename this I think - and ofc requires the change in localnet manage

bouncer/shared/utils/vitest.ts
```diff
@@ -0,0 +1,53 @@
...
+      // Run the test with the test context
+      await testFunction(context.testContext).catch((error) => {
+        // We must catch the error here to be able to log it
+        context.testContext.error(error);
+        throw error;
+      });
+    },
+    timeoutSeconds * 1000,
+  );
+}
```
kylezs, line 53: Something like:

```suggestion
function createTestFunction(
  name: string,
  testFunction: (context: TestContext) => Promise<void>,
) {
  return async (context: { testContext: TestContext }) => {
    // Attach the test name to the logger
    context.testContext.logger = context.testContext.logger.child({ test: name });
    // Run the test with the test context
    await testFunction(context.testContext).catch((error) => {
      // We must catch the error here to be able to log it
      context.testContext.error(error);
      throw error;
    });
  };
}

export function concurrentTest(
  name: string,
  testFunction: (context: TestContext) => Promise<void>,
  timeoutSeconds: number,
) {
  it.concurrent<{ testContext: TestContext }>(
    name,
    createTestFunction(name, testFunction),
    timeoutSeconds * 1000,
  );
}

export function serialTest(
  name: string,
  testFunction: (context: TestContext) => Promise<void>,
  timeoutSeconds: number,
) {
  it<{ testContext: TestContext }>(
    name,
    createTestFunction(name, testFunction),
    timeoutSeconds * 1000,
  );
}
```

bouncer/tests/full_bouncer.test.ts
```diff
@@ -0,0 +1,18 @@
...
+import { testBtcUtxoConsolidation } from './btc_utxo_consolidation';
+import { testDeltaBasedIngress } from './delta_based_ingress';
+import { testDoubleDeposit } from './double_deposit';
+import { testRotatesThroughBtcSwap } from './rotates_through_btc_swap';
+import { testRotateAndSwap } from './rotation_barrier';
+import { testSolanaVaultSettingsGovernance } from './solana_vault_settings_governance';
+import { serialTest } from '../shared/utils/vitest';
+
+// Tests that are run by the ci-main-merge one at a time
+describe('SlowTests', () => {
```
kylezs, line 11: It's not necessarily that they're slow - but that they can't be run in parallel, perhaps SerialTests is better

```suggestion
describe('SlowTests', () => {
```

bouncer/tests/fast_bouncer.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // Post test checks
+  serialTest('CheckSolEventAccountsClosure', checkSolEventAccountsClosure, 150);
+  serialTest('CheckAvailabilityAllSolanaNonces', checkAvailabilityAllSolanaNonces, 50);
+});
+
+// Tests that will run one at a time
+describe('SerialTests', () => {
+  serialTest('GasLimitCcmSwaps', testGasLimitCcmSwaps, 1800);
```
kylezs, line 63: This test should be run first, as it was before - this is a hard test to not make flaky afaik (CC: @albert-llimos )

bouncer/full_bouncer.sh
```diff
@@ -1,17 +1,4 @@
...
-
-if [[ $LOCALNET == false ]]; then
-  echo "🤫 Skipping tests that require localnet"
-else
-  echo "🚀 Running tests that require localnet"
-  ./commands/run_test.ts "Swap-After-Disconnection"
-fi
+NODE_COUNT=$1 LOCALNET=$LOCALNET pnpm vitest run
```
kylezs, line 4: What is actually happening here? What is the test execution order and how do we know that? I think it would be nicer to have this explicit like we do for fast_bouncer

bouncer/tests/full_bouncer.test.ts
j4m1ef0rd: Already have a group called `SerialTests` in the fast bouncer file, so I didn't want to call it that. Maybe `ExtraSerialTests`?

bouncer/tests/fast_bouncer.test.ts
j4m1ef0rd: Dang, I was hoping we would not have to do that, because I don't have much control over the order of execution. I'll do some more research and see what i can do.

bouncer/full_bouncer.sh
j4m1ef0rd: Thought it would be nice to have all the tests to run in a single command as vitest intends. But we can break it up to make sure the order is correct.

kylezs (approved):

PR #5612, opened 2025-02-06
chore: bump try-runtime version to v0.8.0

kylezs:
# Pull Request

Closes: PRO-2013

## Checklist

## Summary

Builds off improvements in: https://github.com/chainflip-io/chainflip-backend/pull/5680

1. Bumps the try-runtime version to 0.8.0 (current latest).
2. Displays the error that the underlying `try-runtime` returns when erroring, instead of the "check node logs" message.
3. Uses HTTP instead of WS for the calls across runtime boundaries, decreasing flakiness.
4. Uses HTTP retries to catch flaky connection issues which have been a source of flakiness in the past.
5. Adds some extra logging.


nmammeri:
One thing I'm not sure about is hardcoding the blocktime to 6000. Apart from that, all looks good to me.

kylezs:
> One thing I'm not sure about is hardcoding the blocktime to 6000. Apart from that, all looks good to me.

Hmm, not sure where we would fetch it from exactly.

Based on the PR it was added in: https://github.com/paritytech/try-runtime-cli/pull/86/files it seems to be related to multi-block migrations (MBMs) - so it seems to just check that the migration can complete within that time, so not super important really (in fact is a little bit annoying we have to provide this as an argument to `try-runtime` at all tbh).

dandanlen:
The upgrade test ran for much longer than usual and then timed out:

```
Error: Timed out after 1800s.
    at runWithTimeout (/opt/actions-runner/_work/chainflip-backend/chainflip-backend/bouncer/shared/utils.ts:304:27)
    at runWithTimeoutAndExit (/opt/actions-runner/_work/chainflip-backend/chainflip-backend/bouncer/shared/utils.ts:317:9)
    at <anonymous> (/opt/actions-runner/_work/chainflip-backend/chainflip-backend/bouncer/commands/upgrade_network.ts:125:7)
```

Could be related to this block time argument?

kylezs:
@coderabbit review

PR #5615, opened 2025-02-06
fix: make naming more consistent for request_swap_parameter_encoding

dandanlen:
# Pull Request

Closes: PRO-2012


state-chain/custom-rpc/src/lib.rs
```diff
@@ -994,7 +994,7 @@ pub trait CustomApi {
 	) -> RpcResult<()>;
 
 	#[method(name = "get_vault_swap_details")]
```
kylezs, line 996: we're not changing the method name here?

dandanlen, line 996:
```suggestion
	#[method(name = "request_swap_parameter_encoding")]
```

PR #5616, opened 2025-02-07
fix: swap after disconnect polkadot bouncer issue

kylezs:
# Pull Request

Closes: PRO-2017

## Checklist

## Summary

The issue stems from now having 2 nodes in the polkadot network, and disconnecting it from the docker network causes it to stop finalising blocks (only sometimes). Here I'm just disabling the test, and we'll enable it for asset hub which has just one node and tests the same code. 

Also included some other small things I found while looking into it.

### DIFF
--- a/bouncer/tests/swap_after_temp_disconnecting_chains.ts
+++ b/bouncer/tests/swap_after_temp_disconnecting_chains.ts
@@ -27,7 +27,6 @@ async function main() {
   );
 
   await Promise.all([
-    testSwap('Dot', 'Btc', undefined, undefined, testSwapAfterDisconnection.swapContext),
     testSwap('Btc', 'Flip', undefined, undefined, testSwapAfterDisconnection.swapContext),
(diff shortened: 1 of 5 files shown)


martin-chainflip (approved):

dandanlen (commented):
The test is disabled on main - will need to be reenabled. (See full_bouncer.sh)

bouncer/tests/swap_after_temp_disconnecting_chains.ts
```diff
@@ -27,7 +27,6 @@ async function main() {
   );
 
   await Promise.all([
-    testSwap('Dot', 'Btc', undefined, undefined, testSwapAfterDisconnection.swapContext),
```
dandanlen, line 30: We still have polkadot1 in `allExternalNodes`.

dandanlen: (I removed it)

PR #5617, opened 2025-02-07
feat: BitcoinLiveness election

marcellorigotti:
# Pull Request

Closes: PRO-1865

## Checklist

## Summary

Instantiation of the Liveness ES for bitcoin

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
@@ -34,17 +34,19 @@ pub trait OnCheckComplete<ValidatorId> {
 
 #[macro_export]
 macro_rules! create_on_check_complete_hook {
-	($name:ident, $chain:expr) => {
-		pub struct $name;
+	($chain:ident) => {
+		sp_core::paste::paste! {
+			pub struct [<$chain OnCheckCompleteHook>];
 
-		impl OnCheckComplete<<Runtime as Chainflip>::ValidatorId> for $name {
-			fn on_check_complete(validator_ids: BTreeSet<<Runtime as Chainflip>::ValidatorId>) {
-				const CHAIN: ForeignChain = $chain; // Enforce ForeignChain at compile time
+			impl OnCheckComplete<<Runtime as Chainflip>::ValidatorId> for [<$chain OnCheckCompleteHook>]{
+				fn on_check_complete(validator_ids: BTreeSet<<Runtime as Chainflip>::ValidatorId>) {
+					const CHAIN: ForeignChain = ForeignChain::$chain; // Enforce ForeignChain at compile time
 
-				<Reputation as OffenceReporter>::report_many(
-					Offence::FailedLivenessCheck(CHAIN),
-					validator_ids,
-				);
+					<Reputation as OffenceReporter>::report_many(
+						Offence::FailedLivenessCheck(CHAIN),
+						validator_ids,
+					);
+				}
 			}
 		}
 	};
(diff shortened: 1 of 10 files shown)


state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -302,11 +311,32 @@ impl Hook<btc::BlockNumber, Vec<DepositChannelDetails<Runtime, BitcoinInstance>>
...
+pub struct OnCheckCompleteHook;
+
+impl OnCheckComplete<<Runtime as Chainflip>::ValidatorId> for OnCheckCompleteHook {
+	fn on_check_complete(validator_ids: BTreeSet<<Runtime as Chainflip>::ValidatorId>) {
+		<Reputation as OffenceReporter>::report_many(
+			Offence::FailedLivenessCheck(ForeignChain::Bitcoin),
+			validator_ids,
+		);
+	}
+}
```
kylezs, line 331: we can probably macrofy this out with the solana one - we're going to need it for all the other chains too

kylezs (approved):
Would also just enable the liveness penalties for bitcoin on a localnet, and stop and start the engine to ensure you see the penalties etc as an extra check.

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -337,6 +374,18 @@ impl Hooks<BitcoinBlockHeightTracking, BitcoinDepositChannelWitnessing> for Bitc
 			>,
 		>(deposit_channel_witnessing_identifiers.clone(), &chain_progress)?;
 
+		// I think we should always use this as our upper limit to avoid not reaching consensus in
+		// case there is a reorg, using this block means safety margin will be kept into account for
+		// this election
```
kylezs, line 379: what do you mean here?

kylezs, line 379:
```suggestion
		// We use `ProcessedUpTo` as our upper limit to avoid not reaching consensus in
		// case there is a reorg, using this block means safety margin will be kept into account for
		// this election, and thus are much less likely to ask nodes to query for a block they don't have.
```

state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
```diff
@@ -32,6 +32,24 @@ pub trait OnCheckComplete<ValidatorId> {
 	fn on_check_complete(validator_ids: BTreeSet<ValidatorId>);
 }
 
+#[macro_export]
+macro_rules! create_on_check_complete_hook {
+	($name:ident, $chain:expr) => {
```
kylezs, line 37: Could just provide the chain name, and then derive `ForeignChain::<chain_name>` and the name of the hook?

state-chain/pallets/cf-elections/src/electoral_systems/liveness.rs
```diff
@@ -32,6 +32,24 @@ pub trait OnCheckComplete<ValidatorId> {
 	fn on_check_complete(validator_ids: BTreeSet<ValidatorId>);
 }
 
+#[macro_export]
+macro_rules! create_on_check_complete_hook {
```
dandanlen, line 36: This looks like it could just be a generic struct. Does something like this work?

```rust
struct ReportFailedLivenessCheck<C>;

impl<C: Get<ForeignChain>> OnCheckComplete<<Runtime as Chainflip>::ValidatorId> for ReportFailedLivenessCheck<C> {
	fn on_check_complete(validator_ids: BTreeSet<<Runtime as Chainflip>::ValidatorId>) {
		<Reputation as OffenceReporter>::report_many(
			Offence::FailedLivenessCheck(C::get()),
			validator_ids,
		);
	}
}

// then in bitcoin_elections.rs:
pub type ChainLiveness = Liveness<
	BlockNumber,
	Hash,
	cf_primitives::BlockNumber,
	ReportFailedLivenessCheck<Bitcoin>,
	<Runtime as Chainflip>::ValidatorId,
>;
```

kylezs (approved):

PR #5622, opened 2025-02-10
fix: delete Solana channel state when closed

dandanlen:
This prevents Solana channels state from accumulating on-chain. We can revert this change if/when we activate Solana channel recycling.


kylezs (approved):
can probably just delete the tests, esp. if on release branch

PR #5641, opened 2025-02-14
feat: returning detailed account information for affiliate in the cf_account_info rpc for brokers

Janislav:
# Pull Request

Closes: PRO-2038

## Checklist

## Summary

- Return affiliate details + usdc balance of affiliates for a broker account
- Deprecated earned_fees field

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -241,6 +241,7 @@ pub enum RpcAccountInfo {
 	Broker {
 		flip_balance: NumberOrHex,
 		bond: NumberOrHex,
+		#[deprecated(note = "This field is deprecated and will be replaced in a future release")]
 		earned_fees: any::AssetMap<NumberOrHex>,
 		affiliates: Vec<(AffiliateShortId, AccountId)>,
 		btc_vault_deposit_address: Option<String>,
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -297,9 +297,9 @@ impl RpcAccountInfo {
 				.affiliates
 				.into_iter()
 				.map(|(short_id, account_id)| {
-					let withdrawal_address = affiliate_details.get(&account_id).map(
-						|AffiliateDetails { withdrawal_address, .. }| withdrawal_address.clone(),
-					);
+					let withdrawal_address = affiliate_details
+						.get(&account_id)
+						.map(|AffiliateDetails { withdrawal_address, .. }| *withdrawal_address);
 					(short_id, account_id, withdrawal_address)
 				})
 				.collect(),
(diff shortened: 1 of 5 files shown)


state-chain/custom-rpc/src/lib.rs
```diff
@@ -240,9 +247,11 @@ pub enum RpcAccountInfo {
 	Broker {
 		flip_balance: NumberOrHex,
 		bond: NumberOrHex,
+		#[deprecated(note = "This field is deprecated and will be replaced in a future release")]
 		earned_fees: any::AssetMap<NumberOrHex>,
 		affiliates: Vec<(AffiliateShortId, AccountId)>,
```
dandanlen, line 252: Might as well deprecate this too and add the account_id into RpcAffiliateDetails.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -209,6 +209,13 @@ pub struct RpcLiquidityProviderBoostPoolInfo {
 	pub is_withdrawing: bool,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct RpcAffiliateDetails {
+	pub short_id: AffiliateShortId,
+	pub withdrawal_address: EthereumAddress,
+	pub earned_fees: NumberOrHex,
```
dandanlen, line 216:
```suggestion
	pub balance: NumberOrHex,
```

Feedback was that `earned_fees` was misleading because it didn't include 'all time' fees.

dandanlen:
I think the idea was to also add account balances for all account types, so that it would also show if the account is an unregistered affiliate.

Something like this:

```rust
pub struct RpcAccountInfoWithBalances {
	asset_balances: any::AssetMap<NumberOrHex>,
	#[serde(flatten)]
	account_info: RpcAccountInfo,
}
```

And then change the rpc return type to `RpcAccountInfoWithBalances`.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -197,6 +197,7 @@ pub struct BrokerInfo {
 	pub btc_vault_deposit_address: Option<String>,
 	pub affiliates: Vec<(AffiliateShortId, AccountId32)>,
 	pub bond: AssetAmount,
+	pub affiliate_details: Vec<(AffiliateDetails, AssetAmount)>,
```
dandanlen, line 200: Unfortunately this is a breaking change to the runtime API. We can do this but we should manage it using runtime api versioning (see the doc comment on the CustomRuntimeApi trait).

Alternatively maybe we can get the data we need via `get_affiliates`.

state-chain/custom-rpc/src/lib.rs
Janislav: I think balance is misleading since it is not the on chain balance but the collected fees in usdc

GabrielBuragev (approved):

PR #5643, opened 2025-02-17
feat: Pino logging in bouncer

j4m1ef0rd:
Address PR comments# Pull Request

Closes: PRO-1996

## Checklist

## Summary

- Replaced all `console.log` calls in all tests, setup scripts and any shared functions that are used in tests. This involved giving every function the logger as an argument. A lot of small changes :(
- Replaced some `console.log`s in  command scripts.
- Set all log messages to `debug` by default. Set important logs to `info`. Set some unnecessary ones to `trace`.
- Added to the readme file with information about logging.
- Added trace logs to some util functions like `observeEvent` to help with debugging.
- Removed the `log` argument from functions. (was used to turn logging on/off).
- Started using my new `throwError` function in some util functions, to help keep relevant info when throwing an error.
- Added some info to the logger in some test where it seemed helpful.
- Removed the `tag` from the start of log messages and instead added it to the logger as a binding.
- Output the new `bouncer.log` file as an artifact on the CI.


bouncer/commands/setup_vaults.ts
```diff
@@ -162,11 +176,11 @@ async function main(): Promise<void> {
...
 
   // Confirmation
-  console.log('Waiting for new epoch...');
-  await observeEvent('validator:NewEpoch').event;
+  logger.info('Waiting for new epoch...');
+  await observeEvent(logger, 'validator:NewEpoch').event;
 
-  console.log('=== New Epoch ===');
-  console.log('=== Vault Setup completed ===');
+  logger.info('New Epoch');
```
msgmaxim, line 182: Any particular reason to remove `===`? 😄

bouncer/shared/new_btc_address.ts
```diff
@@ -61,14 +61,8 @@ export async function newBtcAddress(seed: string, type: BtcAddressType): Promise
...
 
-  let walletDescriptor;
-  try {
-    walletDescriptor = (await axios.post(btcEndpoint, getDescriptorData, axiosConfig)).data.result
-      .descriptor;
-  } catch (err) {
-    console.log(err);
-    process.exit(-1);
-  }
+  const walletDescriptor = (await axios.post(btcEndpoint, getDescriptorData, axiosConfig)).data
```
msgmaxim, line 64: How do we know that removing try/catch is safe?

bouncer/shared/perform_swap.ts
```diff
@@ -196,61 +200,63 @@ export async function performSwap(
...
+    }`,
+  );
 
   const swapParams = await requestNewSwap(
+    logger,
     sourceAsset,
     destAsset,
     destAddress,
-    tag,
+    undefined,
```
msgmaxim, line 223: If we always provide logger, does it mean that tag should never be provided and can be removed as a parameter?

bouncer/commands/setup_vaults.ts
j4m1ef0rd: No, just didn't seem needed anymore. Just made them `info` level instead, that should make them stand out enough.

bouncer/shared/new_btc_address.ts
j4m1ef0rd: This was is how we do it in other utils, I don't see why this would be different. `vitest` will catch it.

bouncer/shared/perform_swap.ts
j4m1ef0rd: As discussed in person. The swap can be started at each of the functions, so each needs to take the tag and add it to the logger if needed.
I do think that this "perform swap" system needs to be cleaned up. It's quite messy and confusing.

msgmaxim (approved):

albert-llimos:
I don't have much context on how much better this logging is but I'm not a big fan of adding an additional required parameter to so many functions for logging purposes. For example, requiring a logger to be passed to a function that sends sol/eth seems unnecessary. We need to pass these loggers all the way down to all functions which makes the bouncer even more verbose than it already is.
Is there no better way to do this?

kylezs:
> I don't have much context on how much better this logging is but I'm not a big fan of adding an additional required parameter to so many functions for logging purposes. For example, requiring a logger to be passed to a function that sends sol/eth seems unnecessary. We need to pass these loggers all the way down to all functions which makes the bouncer even more verbose than it already is.
Is there no better way to do this?

There is, and it already exists here, you can import the global logger. However, in doing so we would lose the ability to construct and inherit child loggers which allow the parent to pass down context that will now exist on the logger for it's lifetime. e.g. attaching "swap direction and amount" to a particular swap process, now this means we will have this context even for RPC calls which we previously didn't have. So it seems messier, but we gain a lot more granularity as a result, and we gain it accidentally almost. Future developers don't need to think "oh have I attached the correct context" etc. They just use the logger and any upstream context is already there. 

This is something we were already doing, but in an inconsistent way, by having special "context" objects for different test types etc. Now it can all be done in the logger. And in cases where we don't care about context, the globalLogger can be used.

bouncer/README.md
```diff
@@ -122,3 +128,88 @@ pnpm vitest run -t "ConcurrentTests"
...
+Another option for adding information to the logger is using the custom `loggerChild` function.
+This will append the given string to the `module` binding.
+
+```ts
+import { loggerChild } from '../shared/utils/logger';
+
+const logger1 = loggerChild(parentLogger, `myTestCase`);
+const logger2 = loggerChild(logger1, `setupFunction`);
+const logger3 = loggerChild(logger2, `myFunction`);
+// {"module": `myTestCase::setupFunction::myFunction`}
```
kylezs, line 182: ?

bouncer/README.md
```diff
@@ -122,3 +128,88 @@ pnpm vitest run -t "ConcurrentTests"
...
+```
+
+### Logging outside of a test
+
+If you need use the logger in a command or any other non-test code, you can use the `globalLogger`.
```
kylezs, line 187:
```suggestion
If you need to use the logger in a command or any other non-test code, you can use the `globalLogger`.
```

bouncer/README.md
```diff
@@ -122,3 +128,88 @@ pnpm vitest run -t "ConcurrentTests"
...
+### Debugging
+
+To debug a failed test you can use the `bouncer.log` file. It will have logs at `Trace` level of all test that where ran.
+To filter for the test you are debugging, you can use the `test` value that should be on all log messages (Excluding logs from commands, setup scripts and thrown errors).
+
+An example of using `jq` to filter the logs for an individual test and put them in another file:
+
+```sh copy
+jq 'select(.test=="BoostingForAsset")' /tmp/chainflip/bouncer.log > /tmp/chainflip/failed_test.log
+```
```
kylezs, line 207: 👌

bouncer/shared/perform_swap.ts
```diff
@@ -54,18 +55,19 @@ export type SwapParams = {
...
   tag = '',
   messageMetadata?: CcmDepositMetadata,
   brokerCommissionBps?: number,
-  log = true,
   boostFeeBps = 0,
   fillOrKillParams?: FillOrKillParamsX128,
   dcaParams?: DcaParams,
 ): Promise<SwapParams> {
-  const addressPromise = observeEvent('swapping:SwapDepositAddressReady', {
+  const logger = tag ? parentLogger.child({ tag }) : parentLogger;
```
kylezs, line 69: This is a little strange, why do we need to take a tag and a logger? Why not just attach this to the logger outside this function. If there's a tag, then we add it to the logger and pass that logger down. I think the logger should replace this notion of "tag" now.

bouncer/shared/perform_swap.ts
```diff
@@ -196,61 +200,63 @@ export async function performSwap(
...
 // function to create a swap and track it until we detect the corresponding broadcast success
 export async function performAndTrackSwap(
+  parentLogger: Logger,
   sourceAsset: Asset,
   destAsset: Asset,
   destAddress: string,
   amount?: string,
   tag?: string,
 ) {
+  const logger = tag ? parentLogger.child({ tag }) : parentLogger;
```
kylezs, line 242: same here and all of them that have a tag ofc

bouncer/shared/perform_swap.ts
```diff
@@ -362,17 +374,18 @@ export async function performVaultSwap(
   }[] = [],
 ): Promise<VaultSwapParams> {
   const tag = swapTag ?? '';
```
kylezs, line 376: also the swapTag should be the same, should be part of the logger that's passed in

bouncer/shared/perform_swap.ts
j4m1ef0rd: It was to avoid changing the code too much. Plus we use the tag in the swap context to identify a swap. But I was able to find a work around for this. We might want to think about using a better id method for the swaps in the future.

acdibble:
> I don't have much context on how much better this logging is but I'm not a big fan of adding an additional required parameter to so many functions for logging purposes. For example, requiring a logger to be passed to a function that sends sol/eth seems unnecessary. We need to pass these loggers all the way down to all functions which makes the bouncer even more verbose than it already is. Is there no better way to do this?

There is an API that basically provides storage based around the callstack
https://nodejs.org/api/async_context.html#class-asynclocalstorage

.github/workflows/upgrade-test.yml
```diff
@@ -285,6 +285,15 @@ jobs:
           path: |
             /tmp/chainflip/logs/*.log
 
+      - name: Upload Bouncer Logs 💾
+        if: always()
+        continue-on-error: true
+        uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874
+        with:
+          name: bouncer-logs
```
kylezs, line 293: Needs a different name, otherwise it'll overwrite the logs from post_check

PR #5651, opened 2025-02-19
chore: add assert_matches! utility function for testing

syan095:
# Pull Request
## Checklist

## Summary

Added a new "assert_matches" utility macro that can be used to replace `assert!(matches!())`
Just adds a little syntactic sugar.


syan095:
I noticed that we have alot of "assert!(matches!(a, b))"s in our code. I thought why not just create a macro that combines the two, so we just need 1 layer of macro to do the same logic?

dandanlen:
I think we used to have something like this and then removed it again... 

My (not very strong) opinion is that I don't see a big benefit of this: it doesn't really save much typing, and doesn't add any extra functionality (for example a better error message). It's basically just an alias, except slightly less convenient since you need to remember to import cf_utilities.

If people prefer this way don't have any objections to merging it.

dandanlen (approved):

PR #5653, opened 2025-02-19
feat: BTC egress witnessing ES

marcellorigotti:
# Pull Request

Closes: PRO-2040

## Checklist

## Summary

Egress witnessing ES for btc

### DIFF
--- a/engine/src/witness/btc/deposits.rs
+++ b/engine/src/witness/btc/deposits.rs
@@ -29,6 +29,7 @@ use cf_chains::{
 	btc::{deposit_address::DepositAddress, Utxo, UtxoId},
 	Bitcoin,
 };
+use pallet_cf_broadcast::TransactionConfirmation;
 
 impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 	pub fn btc_deposits<ProcessCall, ProcessingFut>(
@@ -184,6 +185,34 @@ pub fn deposit_witnesses(
 				.collect::<Vec<_>>()
 		})
 		.collect()
+}
+
+pub fn egress_witnessing(
+	txs: &[VerboseTransaction],
+	monitored_tx_hashes: Vec<cf_chains::btc::Hash>,
+) -> Vec<TransactionConfirmation<state_chain_runtime::Runtime, BitcoinInstance>> {
+	monitored_tx_hashes
+		.into_iter()
+		.filter_map(|hash| {
+			for tx in txs {
+				let tx_hash: cf_chains::btc::Hash = tx.txid.to_byte_array().into();
+				if hash == tx_hash {
+					return Some(TransactionConfirmation {
+						tx_out_id: hash,
+						// we don't care about this value for btc (we don't refund out aggkey)
+						// the correct value to put here should be the aggKey which signed the tx
+						// (either current or previous) I guess we can derive it from
+						// VerobseTransaction, if yes how?
+						signer_id: DepositAddress::new([0; 32], 0).script_pubkey(),
+						tx_fee: tx.fee.unwrap_or_default().to_sat(),
+						tx_metadata: (),
+						transaction_ref: hash,
+					});
+				}
+			}
+			None
+		})
+		.collect::<Vec<_>>()
 }
 
 pub fn map_script_addresses(
--- a/state-chain/chains/src/dot.rs
+++ b/state-chain/chains/src/dot.rs
@@ -27,7 +27,7 @@ use scale_info::TypeInfo;
 use sp_core::{sr25519, ConstBool, H256};
 
 #[cfg_attr(feature = "std", derive(Hash))]
-#[derive(Debug, Encode, Decode, TypeInfo, Eq, PartialEq, Clone)]
+#[derive(Debug, Encode, Decode, TypeInfo, Eq, PartialEq, Clone, Serialize, Deserialize)]
 pub struct PolkadotSignature(sr25519::Signature);
 impl PolkadotSignature {
 	fn verify(&self, payload: &EncodedPolkadotPayload, signer: &PolkadotPublicKey) -> bool {
--- a/state-chain/chains/src/evm.rs
+++ b/state-chain/chains/src/evm.rs
@@ -369,7 +369,9 @@ impl Tokenizable for AggKey {
 	}
 }
 
-#[derive(Encode, Decode, TypeInfo, Copy, Clone, RuntimeDebug, PartialEq, Eq, Serialize)]
+#[derive(
+	Encode, Decode, TypeInfo, Copy, Clone, RuntimeDebug, PartialEq, Eq, Serialize, Deserialize,
+)]
 pub struct SchnorrVerificationComponents {
 	/// Scalar component
 	pub s: [u8; 32],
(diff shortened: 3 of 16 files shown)


state-chain/chains/src/dot.rs
```diff
@@ -27,7 +27,7 @@ use scale_info::TypeInfo;
 use sp_core::{sr25519, ConstBool, H256};
 
 #[cfg_attr(feature = "std", derive(Hash))]
-#[derive(Debug, Encode, Decode, TypeInfo, Eq, PartialEq, Clone)]
+#[derive(Debug, Encode, Decode, TypeInfo, Eq, PartialEq, Clone, Serialize, Deserialize)]
```
MxmUrw, line 30: Are these still required? Or could we revert all dot and evm related to their previous state?

state-chain/chains/src/evm.rs
```diff
@@ -369,7 +369,9 @@ impl Tokenizable for AggKey {
 	}
 }
 
-#[derive(Encode, Decode, TypeInfo, Copy, Clone, RuntimeDebug, PartialEq, Eq, Serialize)]
+#[derive(
+	Encode, Decode, TypeInfo, Copy, Clone, RuntimeDebug, PartialEq, Eq, Serialize, Deserialize,
+)]
```
MxmUrw, line 374: Same.

state-chain/chains/src/evm.rs
```diff
@@ -658,7 +660,19 @@ impl From<CheckedTransactionParameter> for TransactionVerificationError {
...
+	PartialEq,
+	Eq,
+	Encode,
+	Decode,
+	TypeInfo,
+	MaxEncodedLen,
+	Copy,
+	Serialize,
+	Deserialize,
+)]
```
MxmUrw, line 675: And again.

state-chain/chains/src/lib.rs
MxmUrw, line 1: Also I would hope that with our current approach we don't need these changes as well? Correct me if I'm wrong though...

state-chain/chains/src/mocks.rs
```diff
@@ -56,7 +56,9 @@ impl Get<ChainChoice> for MockBroadcastBarriers {
 	}
 }
 
-#[derive(Debug, Clone, Default, PartialEq, Eq, Encode, Decode, TypeInfo)]
+#[derive(
+	Debug, Clone, Default, PartialEq, Eq, Encode, Decode, TypeInfo, Serialize, Deserialize,
+)]
```
MxmUrw, line 61: Same?

state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
```diff
@@ -204,6 +242,46 @@ impl Hook<HookTypeFor<TypesVaultDepositWitnessing, DedupEventsHook>>
 	}
 }
 
+impl Hook<HookTypeFor<TypesEgressWitnessing, DedupEventsHook>> for TypesEgressWitnessing {
+	fn run(
+		&mut self,
+		events: Vec<(BlockNumber, BtcEvent<TransactionConfirmation<Runtime, BitcoinInstance>>)>,
+	) -> Vec<(BlockNumber, BtcEvent<TransactionConfirmation<Runtime, BitcoinInstance>>)> {
+		// Map: deposit_witness -> chosen BtcEvent
+		// todo! this is annoying, it require us to implement Ord down to the Chain type
```
MxmUrw, line 251: I think this todo isn't true anymore, right?

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -357,26 +456,32 @@ impl
...
 			Vec<
 				ElectionIdentifier<
-					<BitcoinBlockHeightTrackingES as ElectoralSystemTypes>::ElectionIdentifierExtra,
+					<BitcoinDepositChannelWitnessingES as ElectoralSystemTypes>::ElectionIdentifierExtra,
 				>,
 			>,
 			Vec<
 				ElectionIdentifier<
-					<BitcoinDepositChannelWitnessingES as ElectoralSystemTypes>::ElectionIdentifierExtra,
+					<BitcoinVaultDepositWitnessingES as ElectoralSystemTypes>::ElectionIdentifierExtra,
```
MxmUrw, line 472: It's interesting that this works apparently, but it seems like we now have two `BitcoinVaultDepositWitnessingES` and no `DepositChannelWitnessingES`.

engine/src/witness/btc/deposits.rs
```diff
@@ -186,6 +187,34 @@ pub fn deposit_witnesses(
...
+		.filter_map(|hash| {
+			for tx in txs {
+				let tx_hash: cf_chains::btc::Hash = tx.txid.to_byte_array().into();
+				if hash == tx_hash {
+					return Some(TransactionConfirmation {
+						tx_out_id: hash,
+						// we don't care about this value for btc (we don't refund out aggkey)
+						// the correct value to put here should be the aggKey which signed the tx
+						// (either current or previous) I guess we can derive it from
+						// VerobseTransaction, if yes how?
```
MxmUrw, line 205: Is this a TODO?

state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -1046,6 +1074,19 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		Self::egress_success(
+			OriginFor::<T>::none(),
+			egress.tx_out_id,
+			egress.signer_id,
+			egress.tx_fee,
+			egress.tx_metadata,
+			egress.transaction_ref,
+		)
+		.expect("REASON");
+		//todo! handle possible error, expect will cause panic! -> when can this error?
```
MxmUrw, line 1088: This is an interesting case, the `egress_success` function returns an error if the `tx_out_id` is unknown to us. This certainly can happen if nodes start voting for wrong txs, but what should we do in that case? Punish them?

Also it feels like we might want to check this already during consensus, such that we don't count votes if they contain invalid txs? But votes are for full blocks, so we would discard the whole vote?

state-chain/chains/src/lib.rs
marcellorigotti: I don't think we can get rid of these changes that easily.. as long as we have this in the ES trait:
`type ElectoralUnsynchronisedState: Parameter + Member + **MaybeSerializeDeserialize**;
`
Then everything that goes into the BlockProcessor has to to impl Serialize/Deserialize

MxmUrw: If we want, we can do it the same way we did for the Encode/Decode bounds, by adding annotations where we derive Serialize/Deserialize. Here: https://github.com/chainflip-io/chainflip-backend/pull/5657

marcellorigotti: Yeah ofc that's the way to do it, was just having a look at the PR now!

MxmUrw: OTOH it might be fine to simply add these new bounds, the PR above is more of an FYI...

state-chain/pallets/cf-broadcast/src/lib.rs
marcellorigotti: This should be impossible since in the voter impl we vote only if the hashes match one of our tx, meaning we can only vote for `tx_out_id` we know.
I'll log in case it errors, like we do for solana

MxmUrw: I think it's worth thinking about the case where the engines don't behave as we expect them to do, to cover all possible cases. Anyways, logging should be fine since nothing of consequence would happen in such an error case.

PR #5654, opened 2025-02-19
chore: minor state machine es clean up items

kylezs:


MxmUrw (approved):
Thanks, let's merge this! 

I'll have to update this to describe how multiple elections are handled now, and how the results of applying the step function multiple times are collected.

PR #5655, opened 2025-02-19
chore: re-enable manual blockwitnesser tests

MxmUrw:
# Pull Request

Closes: PRO-2058

## Checklist

## Summary

@kylezs had written many tests for manually testing the BW behaviour. This PR re-enables them for the new SM-based BW ES.

Changes in behaviour of the new SM-based BW compared to its previous incarnation:
 - The `generate_election_properties` hook is called for each election which is open after running the `step` function on all inputs. Previously it was only called when elections were created. The new behaviour might be more costly in terms of decoding storage (since the hook uses the deposit channel lookup storage map), but it is more correct: In block chains with long block times, we might open an election for a given block height `n`, and while it is open, we might be opening new deposit channels. Then, even though there was no consensus, the next time we run the `on_finalize`, we want to actually recreate that election to include the new deposit channels. The BW tests had to be updated to reflect this new behaviour

Other changes in the BW tests file:
 - [**checks**] The `generate_election_properties_called_n_times` check now counts the number of times the hook was called during *this* `on_finalize`, not the total number of calls.
 - [**architecture**] The BW tests file now uses hook-internal storage to store e.g. how often a hook was called. This means that we no longer require `thread_local` variables.

Additional changes which got included in this PR:
 - rename `BWSettings` to `BlockWitnesserSettings`, `BWState` to `BlockWitnesserState`
 - fix indexing in ElectionsTracker in the BW state machine. Previously we had `highest_election`, but this forced us to do `.saturating_sub(1)` in some places, which lead to misbehavior for the edge case were `highest_election=0`. Now, instead, we have `next_election` which is meant to be `highest_election+1`.


marcellorigotti:
In the `tests/block_witnesser` file there are a bunch of checks commented out, do we want to keep them and fix them later or can those comments be removed?
Other than that LGTM!

MxmUrw:
> In the `tests/block_witnesser` file there are a bunch of checks commented out, do we want to keep them and fix them later or can those comments be removed? Other than that LGTM!

Yes good catch - these tests are all related to the block processor, whose design changed a lot since introducing the state machines. My plan was to fix them up later after I looked at fuzzy testing the block processor.

PR #5661, opened 2025-02-21
feat: refund fee

j4m1ef0rd:
# Pull Request

Closes: PRO-2000

## Checklist

## Summary

- Reusing the minimum network fee as a refund fee.
- Made the gas fee calculation more general so I could reuse the code for the refund fee.
	- Add fee estimation basis for the rest of the assets, not just gas ones anymore.
- Updated an existing refund test to take the refund fee into account and check that it operates correctly.
- Added some more init swap request functions for the different types.
- Removed a few duplicate swapping unit tests.

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced swap functionality with dedicated fee management options, allowing for network, ingress/egress, and internal fee processing.
	- Added improved asset estimation for determining the necessary input to achieve a desired output, including clearer refund fee calculations.

- **Refactor**
	- Streamlined swap request processing by reducing complexity in request parameters.
	- Improved error reporting during conversions, leading to more precise handling of fee-related operations.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -1994,19 +1994,20 @@ pub mod pallet {
 			from: Asset,
 			to: Asset,
 			input_amount: AssetAmount,
+			min_fee_enforced: bool,
 		) -> Result<SwapOutput, DispatchError> {
 			Ok(match (from, to) {
 				(_, STABLE_ASSET) => {
 					let FeeTaken { remaining_amount: output, fee } = Self::take_network_fee(
 						T::SwappingApi::swap_single_leg(from, to, input_amount)?,
-						false,
+						min_fee_enforced,
 					);
 
 					SwapOutput { intermediary: None, output, network_fee: fee }
 				},
 				(STABLE_ASSET, _) => {
 					let FeeTaken { remaining_amount: input_amount, fee } =
-						Self::take_network_fee(input_amount, false);
+						Self::take_network_fee(input_amount, min_fee_enforced);
 
 					SwapOutput {
 						intermediary: None,
@@ -2017,7 +2018,7 @@ pub mod pallet {
 				_ => {
 					let FeeTaken { remaining_amount: intermediary, fee } = Self::take_network_fee(
 						T::SwappingApi::swap_single_leg(from, STABLE_ASSET, input_amount)?,
-						false,
+						min_fee_enforced,
 					);
 
 					SwapOutput {
@@ -2315,7 +2316,7 @@ pub mod pallet {
 
 			let estimation_output = with_transaction_unchecked(|| {
 				TransactionOutcome::Rollback(if with_network_fee {
-					Self::swap_with_network_fee(input_asset, output_asset, estimation_input)
+					Self::swap_with_network_fee(input_asset, output_asset, estimation_input, false)
 						.map(|swap| swap.output)
 				} else {
 					T::SwappingApi::swap_single_leg(input_asset, output_asset, estimation_input)
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -566,9 +566,9 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type BrokerBond<T: Config> = StorageValue<_, T::Amount, ValueQuery, DefaultBrokerBond<T>>;
 
-	/// Minimum network fee charged per chunk (only applies to regular swaps, i.e. it excludes
-	/// internal swaps like ingress/egress fees). In practice this should also effectively be the
-	/// minimum fee charged per swap request due to us also enforcing minimum chunk size.
+	/// Minimum network fee in USDC, charged per chunk (only applies to regular swaps, i.e. it
+	/// excludes internal swaps like ingress/egress fees). In practice this should also effectively
+	/// be the minimum fee charged per swap request due to us also enforcing minimum chunk size.
 	#[pallet::storage]
 	pub type MinimumNetworkFeePerChunk<T: Config> = StorageValue<_, AssetAmount, ValueQuery>;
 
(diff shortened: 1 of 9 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2093,6 +2105,50 @@ pub mod pallet {
...
+
+			let (remaining_amount_to_refund, refund_fee) =
+				if required_refund_fee_as_input_asset < total_input_amount {
+					(
+						total_input_amount - required_refund_fee_as_input_asset,
+						required_refund_fee_as_input_asset,
+					)
+				} else {
+					(0, total_input_amount)
+				};
```
kylezs, line 2135:
```suggestion
			let refund_fee =
				sp_std::cmp::min(required_refund_fee_as_input_asset, total_input_amount);
			let remaining_amount_to_refund = total_input_amount.saturating_sub(refund_fee);
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2093,6 +2105,50 @@ pub mod pallet {
...
+
+			if refund_fee > 0 {
+				Self::init_swap_request(
+					input_asset,
+					refund_fee,
+					Asset::Flip,
+					SwapRequestType::NetworkFee,
```
kylezs, line 2142: It's strange that we allow the specification of a non-FLIP asset for NetworkFee swaps at the interface level? Not strictly this PR related, but wanted to point that out

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1619,33 +1619,45 @@ pub mod pallet {
...
-								account_id,
-								request.input_asset,
-								amount_to_refund,
-							);
-						},
+								total_input_remaining
+							},
+						};
+
+					if amount_to_refund > 0 {
```
kylezs, line 1634: We probably should emit an event in the case we're not above 0, else it's hard to see where the flow ended.

dandanlen: We have `SwapRequestCompleted` a bit further down - seems like enough?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1619,33 +1619,45 @@ pub mod pallet {
...
-								amount: amount_to_refund,
-							});
-
-							T::BalanceApi::credit_account(
-								account_id,
-								request.input_asset,
-								amount_to_refund,
-							);
-						},
+								total_input_remaining
```
dandanlen, line 1630: nit: indentation is a bit off here.

Also: maybe we can push this log message into the take_refund_fee function?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2093,6 +2105,50 @@ pub mod pallet {
...
+		pub(super) fn take_refund_fee(
+			total_input_amount: AssetAmount,
+			input_asset: Asset,
+		) -> Result<AssetAmount, DispatchError> {
+			let refund_fee_usdc = MinimumNetworkFeePerChunk::<T>::get();
+
+			let required_refund_fee_as_input_asset = if input_asset == STABLE_ASSET {
+				refund_fee_usdc
+			} else {
+				with_transaction_unchecked(|| {
```
dandanlen, line 2118: Any particular reason for `unchecked`? According to the docs, it only exists for backwards compatibility.

dandanlen: Btw. i thought we already had a conversion api for this somewhere? AFAIK it's more accurate since it uses the conversion in the 'right' direction. Here, you'r converting from usdc to the input asset and assuming that the opposite exchange rate is equivalent. This is not necessarily true.

msgmaxim: Do you mean `calculate_input_for_gas_output`? Seems like currently it is designed for gas specifically, but assuming that gas amounts aren't too different from what we are going to charge here, it should work fine (might need to rename a few things though).

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: I was initially sceptical about making SwapRequestHandler more complex by adding more specialised methods (I think @dandanlen was suggesting something similar for on-chian swaps), but I'm now warming up to this idea, especially since we can provide a default implementation (for e.g `init_network_fee_swap_request`) that calls `init_swap_request` internally. Then it would be slightly easier to find where network fee is charged, and we can make sure that we convert into the right asset and that other parameters are set correctly (e.g. empty broker fees, SwapOrigin::Internal etc).

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1791,21 +1799,34 @@ mod on_chain_swapping {
...
-				SwapRate::set(DEFAULT_SWAP_RATE as f64 / 2.0);
+				SwapRate::set(NEW_SWAP_RATE);
 			})
 			.then_process_blocks_until_block(CHUNK_2_BLOCK)
 			.then_execute_with(|_| {
+				const REFUND_FEE: AssetAmount = MIN_NETWORK_FEE * NEW_SWAP_RATE as u128;
 				// Only one chunk is expected to be swapped:
 				const EXPECTED_OUTPUT_AMOUNT: AssetAmount =
-					CHUNK_AMOUNT * DEFAULT_SWAP_RATE * DEFAULT_SWAP_RATE;
+					CHUNK_AMOUNT * DEFAULT_SWAP_RATE * DEFAULT_SWAP_RATE -
```
msgmaxim, line 1809: Nit: I would group like this to make it clear that network fee is deducted between leg 1 and leg 2:

`(CHUNK_AMOUNT * DEFAULT_SWAP_RATE - MIN_NETWORK_FEE) * DEFAULT_SWAP_RATE`.

Also, I think it is a little subtle why we charge min network fee rather than the usual 10bps. I'm assuimg it is because the chunk amounts are too small and/or that network fee is currently set to 0? I think it is woth adding a comment clarifying this.

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: But then the log message will not contain the swap request id.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2249,44 +2282,53 @@ pub mod pallet {
...
 			input_asset: C::ChainAsset,
 			required_gas: C::ChainAmount,
 		) -> Option<C::ChainAmount> {
+			Self::calculate_input_for_desired_output(
+				input_asset.into(),
+				C::GAS_ASSET.into(),
+				required_gas.into(),
+				true,
+			)
+			.map(|amount| C::ChainAmount::try_from(amount).expect("Asset amount is for this chain"))
```
kylezs, line 2296: Is it really true this can't panic? The asset amount is specifically not for this chain before this point right? It's a general asset amount that could be outside the bounds of a particular ChainAmount when we get to here. For example at a time where the liquidity is way off, it could mean we require a huge amount of input asset to get the output asset, which might fit AssetAmount, but not ChainAmount.

kylezs: Would be nice to have a test for this case.

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -636,3 +636,23 @@ fn test_refund_fee_calculation() {
 		assert_eq!(Swapping::take_refund_fee(3, Asset::Eth), Ok(0));
 	});
 }
+
+#[test]
+fn gas_calculation_can_handle_invalid_swap_rate() {
```
kylezs, line 641: these aren't really "invalid", they are "extreme"

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -636,3 +636,23 @@ fn test_refund_fee_calculation() {
...
+	new_test_ext().execute_with(|| {
+		fn test_invalid_swap_rate(swap_rate: f64) {
+			SwapRate::set(swap_rate);
+			assert_eq!(
+				Swapping::calculate_input_for_gas_output::<Ethereum>(
```
kylezs, line 646: I think this is a reasonable test, but it does not test the panic case I mentioned above. Reverting the change, this test still passes. This is because in all these cases, we get a None from `calculate_input_for_desired_output`. But we want to get a value that is > u64::MAX and then try to convert it to u64

j4m1ef0rd:
@coderabbitai review

kylezs (approved):

msgmaxim (approved):

PR #5662, opened 2025-02-21
feat: fee tracker ES for BTC

marcellorigotti:
# Pull Request

Closes: PRO-1322

## Checklist

## Summary

Added the fee electoral system for btc

### DIFF
--- a/engine/src/witness/btc_e.rs
+++ b/engine/src/witness/btc_e.rs
@@ -43,7 +43,7 @@ use anyhow::Result;
 
 use sp_core::H256;
 use state_chain_runtime::chainflip::bitcoin_elections::{
-	BitcoinEgressWitnessingES, BitcoinVaultDepositWitnessingES,
+	BitcoinEgressWitnessingES, BitcoinFeeTracking, BitcoinVaultDepositWitnessingES,
 };
 use std::sync::Arc;
 
@@ -255,6 +255,26 @@ impl VoterApi<BitcoinEgressWitnessingES> for BitcoinEgressWitnessingVoter {
 }
 
 #[derive(Clone)]
+pub struct BitcoinFeeVoter {
+	client: BtcRetryRpcClient,
+}
+
+#[async_trait::async_trait]
+impl VoterApi<BitcoinFeeTracking> for crate::witness::btc_e::BitcoinFeeVoter {
+	async fn vote(
+		&self,
+		_settings: <BitcoinFeeTracking as ElectoralSystemTypes>::ElectoralSettings,
+		_properties: <BitcoinFeeTracking as ElectoralSystemTypes>::ElectionProperties,
+	) -> Result<Option<VoteOf<BitcoinFeeTracking>>, anyhow::Error> {
+		if let Some(fee) = self.client.next_block_fee_rate().await {
+			Ok(Some(fee))
+		} else {
+			Ok(None)
+		}
+	}
+}
+
+#[derive(Clone)]
 pub struct BitcoinLivenessVoter {
 	client: BtcRetryRpcClient,
 }
@@ -296,6 +316,7 @@ where
 						BitcoinDepositChannelWitnessingVoter { client: client.clone() },
 						BitcoinVaultDepositWitnessingVoter { client: client.clone() },
 						BitcoinEgressWitnessingVoter { client: client.clone() },
+						BitcoinFeeVoter { client: client.clone() },
 						BitcoinLivenessVoter { client },
 					)),
 				)
--- a/state-chain/pallets/cf-chain-tracking/src/lib.rs
+++ b/state-chain/pallets/cf-chain-tracking/src/lib.rs
@@ -178,11 +178,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 	}
 
 	pub fn inner_update_fee(new_fee: <T::TargetChain as Chain>::TrackedData) -> DispatchResult {
-		CurrentChainState::<T, I>::try_mutate::<_, Error<T, I>, _>(|previous_chain_state| {
+		CurrentChainState::<T, I>::mutate(|previous_chain_state| {
 			previous_chain_state.as_mut().expect(NO_CHAIN_STATE).tracked_data = new_fee;
-
-			Ok(())
-		})?;
+		});
 
 		Ok(())
 	}
--- a/state-chain/pallets/cf-chain-tracking/src/lib.rs
+++ b/state-chain/pallets/cf-chain-tracking/src/lib.rs
@@ -176,6 +176,21 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		Ok(())
 	}
+
+	pub fn inner_update_fee(new_fee: <T::TargetChain as Chain>::TrackedData) -> DispatchResult {
+		CurrentChainState::<T, I>::try_mutate::<_, Error<T, I>, _>(|previous_chain_state| {
+			previous_chain_state
+				.as_mut()
+				.expect(
+					"Impossible to not have a value for chain tracking, check storage item docs",
+				)
+				.tracked_data = new_fee;
+
+			Ok(())
+		})?;
+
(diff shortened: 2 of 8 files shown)


engine/src/witness/btc_e.rs
```diff
@@ -254,6 +255,28 @@ impl VoterApi<BitcoinEgressWitnessingES> for BitcoinEgressWitnessingVoter {
...
+}
+
+#[async_trait::async_trait]
+impl VoterApi<BitcoinFeeTracking> for crate::witness::btc_e::BitcoinFeeVoter {
+	async fn vote(
+		&self,
+		_settings: <BitcoinFeeTracking as ElectoralSystemTypes>::ElectoralSettings,
+		_properties: <BitcoinFeeTracking as ElectoralSystemTypes>::ElectionProperties,
+	) -> Result<Option<VoteOf<BitcoinFeeTracking>>, anyhow::Error> {
+		if let Some(fee) = self.client.next_block_fee_rate().await {
```
kylezs, line 270: We don't want to use an infinite request here (at least make a comment to be addressed in the other issue)

engine/src/witness/btc_e.rs
```diff
@@ -254,6 +255,28 @@ impl VoterApi<BitcoinEgressWitnessingES> for BitcoinEgressWitnessingVoter {
...
+			warn!("Fee are: {fee:#?}");
```
kylezs, line 271: warns can be removed?

engine/src/witness/btc_e.rs
kylezs: This is actually different to the current implementation.

If the estimate_smart_fee returns None (which is possible at times during low number of txs in blocks / mempool), then we should fallback to useing `average_block_fee_rate`

Here's current implementation:

```rust
async fn get_tracked_data(
		&self,
		header: &Header<<cf_chains::Bitcoin as cf_chains::Chain>::ChainBlockNumber, BlockHash, ()>,
	) -> Result<<cf_chains::Bitcoin as cf_chains::Chain>::TrackedData, anyhow::Error> {
		let fee_rate = if let Some(next_block_fee_rate) = self.next_block_fee_rate().await {
			next_block_fee_rate
		} else {
			self.average_block_fee_rate(header.hash).await
		};

		Ok(BitcoinTrackedData { btc_fee_info: BitcoinFeeInfo::new(fee_rate) })
	}
```

This means we should never return None, since we always have an average block fee rate (assuming no errors ofc).

state-chain/pallets/cf-elections/src/electoral_systems/unsafe_median.rs
```diff
@@ -77,7 +87,9 @@ impl<
 			let election_access = ElectoralAccess::election_mut(election_identifier);
 			if let Some(consensus) = election_access.check_consensus()?.has_consensus() {
 				election_access.delete();
-				ElectoralAccess::set_unsynchronised_state(consensus)?;
+				ElectoralAccess::set_unsynchronised_state(consensus.clone())?;
+				warn!("Consensus reached: {consensus:#?}");
```
kylezs, line 91: remove warn?

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -449,6 +451,24 @@ pub type BitcoinLiveness = Liveness<
 	<Runtime as Chainflip>::ValidatorId,
 >;
 
+pub struct BitcoinFeeUpdateHook;
+impl UpdateFeeHook<BtcAmount> for BitcoinFeeUpdateHook {
+	fn update_fee(fee: BtcAmount) {
+		if let Err(err) = BitcoinChainTracking::inner_update_fee(BitcoinTrackedData {
+			btc_fee_info: BitcoinFeeInfo::new(fee),
+		}) {
+			log::error!("Failed to update chain state: {:?}", err);
```
kylezs, line 460: worth including the state that was attempted to update to as well

state-chain/pallets/cf-chain-tracking/src/lib.rs
```diff
@@ -176,6 +176,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		Ok(())
 	}
+
+	pub fn inner_update_fee(new_fee: <T::TargetChain as Chain>::TrackedData) -> DispatchResult {
+		CurrentChainState::<T, I>::try_mutate::<_, Error<T, I>, _>(|previous_chain_state| {
+			previous_chain_state.as_mut().unwrap().tracked_data = new_fee;
```
kylezs, line 182: 👀  - please no unwraps where possible

engine/src/witness/btc_e.rs
marcellorigotti: I didn't implement it like this because we don't have access to the blockhash here.
My thought was: since we now keep querying for this value as soon as the previous election reach consensus, we can afford to keep the old value until a new value can be calculated (when enough txs are present in the mempool).

kylezs: Using the average_block_fee_rate would mean we can also remove the option, and guarantee the response which would be nice, and guarantee correctness. 

Still think we should do it, as it is more correct, it's especially important in testnet and localnet land where we don't necessarily have as many txs. 

We can create a ticket to add the fallback - because this would be made easier with the client cache - as part of the client cache that will be shared across all the voters, one of the cache items can be "latest header" or something, that would allow us to do this easily.

marcellorigotti: https://linear.app/chainflip/issue/PRO-2069/btc-fee-tracking-election

state-chain/pallets/cf-chain-tracking/src/lib.rs
```diff
@@ -176,6 +176,21 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		Ok(())
 	}
+
+	pub fn inner_update_fee(new_fee: <T::TargetChain as Chain>::TrackedData) -> DispatchResult {
+		CurrentChainState::<T, I>::try_mutate::<_, Error<T, I>, _>(|previous_chain_state| {
+			previous_chain_state
+				.as_mut()
+				.expect(
```
kylezs, line 184: by no unwraps() I meant no unwraps() or expects(). Ik it should be impossible to hit this, but we're already returning an error from this function, so we might as well throw here instead

marcellorigotti: we use this: `CurrentChainState::<T, I>::get().expect(NO_CHAIN_STATE)` all over the place, IMO think it makes sense to keep it consinstent since we know it cannot panic.

kylezs: Yeah I think that's a mistake tbh (and I wrote that code)

marcellorigotti: For now I did the same here, will create a ticket to change the sorage item to be storageQuery, that's the best solution IMO.

state-chain/pallets/cf-chain-tracking/src/lib.rs
```diff
@@ -176,6 +176,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 
 		Ok(())
 	}
+
+	pub fn inner_update_fee(new_fee: <T::TargetChain as Chain>::TrackedData) -> DispatchResult {
+		CurrentChainState::<T, I>::try_mutate::<_, Error<T, I>, _>(|previous_chain_state| {
+			previous_chain_state.as_mut().expect(NO_CHAIN_STATE).tracked_data = new_fee;
+
+			Ok(())
+		})?;
```
kylezs, line 185: We might as well use mutate in this case now, we never throw an error.
```rust
		CurrentChainState::<T, I>::mutate(|previous_chain_state| {
			previous_chain_state.as_mut().expect(NO_CHAIN_STATE).tracked_data = new_fee;
		});

		Ok(())
```

PR #5664, opened 2025-02-23
chore: fix cancel_orders_batch benchmarking

syan095:
# Pull Request

Closes: PRO-2050

## Checklist

## Summary
rewrote the benchmarking for cancel_orders_batch


dandanlen (approved):
Thanks Roy - this still looks very expensive per order... let's merge but we should continue investigating. I'll update the issue.

state-chain/pallets/cf-pools/src/benchmarking.rs
```diff
@@ -280,28 +279,25 @@ mod benchmarks {
...
-			orders_to_delete
-				.try_push(CloseOrder::Range {
-					base_asset: Asset::Eth,
-					quote_asset: Asset::Usdc,
-					id: i as u64,
-				})
-				.expect("cannot fail");
-		}
+
+		let orders_to_delete = (1i32..n as i32)
```
msgmaxim, line 283: Shouldn't this be `1i32..=n`? @syan095

PR #5665, opened 2025-02-24
feat: improve age loop

marcellorigotti:
# Pull Request

Closes: PRO-2051

## Checklist

## Summary

Instead of looping through every age, check if specific ages we are interested in are included in the range.


state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -142,21 +142,20 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
...
+							.into_iter()
+							.map(|(_, event)| event)
+							.collect::<Vec<_>>();
+						match self.reorg_events.get_mut(&n) {
+							None => {
+								self.reorg_events.insert(n, events);
+							},
+							Some(previous_events) => {
+								previous_events.extend(events.into_iter());
+							},
```
MxmUrw, line 158: I think there's a slight change of behaviour here: if `next_age == 0`, then previously the loop wouldn't even run. Now we are calling `process_rules_for_ages_and_block(n, 0..=0, &data)` since `0.saturating_sub(1) == 0`.

This probably wouldn't lead to any failing tests since I think this codepath always runs with `next_age >= 1`. But nevertheless I would try to make this more self-consistent.

One option would be to let the `process_rules_for_ages_and_block` take a `Range` instead of a `RangeInclusive`.

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -190,14 +189,15 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
...
+		for (block_height, (data, next_age_to_process)) in self.blocks_data.clone() {
 			let current_age = T::ChainBlockNumber::steps_between(&block_height, &last_height).0;
-			for age in next_age..=current_age as u32 {
-				last_events = last_events
-					.into_iter()
-					.chain(self.process_rules_for_age_and_block(block_height, age, &data))
-					.collect();
-			}
+			let age_range: RangeInclusive<u32> =
+				RangeInclusive::new(next_age_to_process, current_age as u32);
```
MxmUrw, line 195: nit: I think you can use the `a..=b` notation to construct `RangeInclusive` objects. (And `a..b` for `Range`s of course).

MxmUrw (commented):
Looks good! There's one small change of behaviour I think which we should discuss.

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
marcellorigotti: _This probably wouldn't lead to any failing tests since I think this codepath always runs with next_age >= 1._

This was my thought, but I agree better to make it consistent (I had to go through the code again now to realize it is not a bug)

MxmUrw (approved):
Nice, LGTM!

PR #5667, opened 2025-02-25
feat: retry ccms egress until auxiliary data is ready or expiry

syan095:
## Checklist

## Summary

Ccm message now contains a "lookup key", and will retry for egress until its aux data is ready.
The ccm will be failed and refunded after waiting past the max wait time.
Changed the current election expiry time (actual value TBD) to a smaller number to reduce the integration test wait time.

## Summary by CodeRabbit

- **New Features**
	- Enhanced cross-chain messaging by introducing an auxiliary data lookup mechanism.
	- Implemented a fallback refund process that triggers when messages exceed the maximum wait time for auxiliary data.
  
- **Refactor**
	- Updated API signatures and internal parameter naming for consistent cross-chain operations.
	- Upgraded migration and error handling processes as part of a core version update.


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1732,12 +1731,30 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		// A CCM is ready to be egressed when:
+		// A. No aux data lookup is required. OR
+		// B. Lookup is required:
+		// 		i. Its Auxiliary data is ready to be looked up, or
+		// 		ii. the max wait time has passed
+		let ccm_ready_for_egress = |aux_lookup: &CcmAuxDataLookupKey<BlockNumberFor<T>>| -> bool {
+			match aux_lookup {
+				CcmAuxDataLookupKey::Alt(swap_id, block_started) =>
+					T::NetworkEnvironment::ccm_auxiliary_data_ready(*swap_id) ||
+						(current_block.saturating_sub(*block_started)) >= ccm_max_wait_time,
```
albert-llimos, line 1746: I'm not sure we want to attempt to egress  the ccm normally if we have timed out here. For Solana if then the transaction is too long because we don't have lookup tables it will be invalid and it will do a fallback egress, which is fine. However, since this logic is generic, it feels like we should be failing to egress if it times out.
What about making it so when it times out it executes the same logic as when it fails to build? Emiting the `CcmEgressInvalid` and then proceeding to check if there is a fallback egress we can do.

ramizhasan111:
Why cant we just keep it as is where when ccm tries to egress, it looks up the alts while building the ccm transfer apicall. And when if cant lookup (the alt witnessing hasnt completed), it fails to build which will then cause it to re-enter the ccm egress queue, until it is tried again at a later point. It will then keep on trying until the alt witnessing completes and it succeeds.

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: OK that makes sense. I've changed the logic so we go straight to "refund" when we wait too long

syan095:
> it fails to build which will then cause it to re-enter the ccm egress queue

I think if we know that the build will fail, we can avoid taking the ccm out of storage, then putting failed ones back in (we use a "filter" to remove ccms from storage, then iterate to egress). 

Also, in most cases, when a CCM fails to build, we go straight to a refund. I think it's cleaner if we keep this logic and do not make exceptions for the "aux data not ready" error.

This is also why I wanted to make the ALT lookup infallible (i.e. if we can't lookup, just return vec![]) so we will try once to build the transaction, and if it fails, then it fails and we refund. Feels like a clean solution. I believe this approach will be slightly more efficient as well.

syan095:
> Besides my previous comment, the main thing to review is how it fits into the full architecture and if there are nicer ways to do it

Yeah, I'm happy to just wait until the base branch is ready.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1732,28 +1734,62 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		};
+
+		let ccms_to_send: Vec<CrossChainMessage<T::TargetChain, BlockNumberFor<T>>> =
 			ScheduledEgressCcm::<T, I>::mutate(|ccms: &mut Vec<_>| {
 				// Filter out disabled assets, and take up to batch_size requests to be sent.
 				ccms.extract_if(.., |ccm| {
 					!DisabledEgressAssets::<T, I>::contains_key(ccm.asset()) &&
-						Self::should_fetch_or_transfer(&mut maybe_no_of_transfers_remaining)
+						Self::should_fetch_or_transfer(&mut maybe_no_of_transfers_remaining) &&
+						ccm_ready_for_egress(&ccm.aux_data_lookup_key)
```
ramizhasan111, line 1764: there is a sneaky little bug here (similar to the one we have seen before): ```should_fetch_or_transfer``` always decrements the number of transfers remaining so even if ```ccm_ready_for_egress``` returns false and we dont include the ccm, we would wrongly decrement the number of transfers remaining.

syan095: ohh good spotting!

syan095:
@coderabbitai review

state-chain/pallets/cf-ingress-egress/src/mocks/mock_btc.rs
```diff
@@ -75,7 +75,7 @@ impl AddressDerivationApi<Bitcoin> for MockAddressDerivation {
 
 pub struct MockNetworkEnvironmentProvider {}
 
-impl NetworkEnvironmentProvider for MockNetworkEnvironmentProvider {
+impl NetworkEnvironmentProvider<BlockNumberFor<Test>> for MockNetworkEnvironmentProvider {
```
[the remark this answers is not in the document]
syan095: @coderabbitai 
we can ignore these because the trait contains default implementations for these fields

syan095:
@coderabbitai summary

ramizhasan111:
> I think if we know that the build will fail, we can avoid taking the ccm out of storage, then putting failed ones back in (we use a "filter" to remove ccms from storage, then iterate to egress).
> 
> Also, in most cases, when a CCM fails to build, we go straight to a refund. I think it's cleaner if we keep this logic and do not make exceptions for the "aux data not ready" error.
> 
> This is also why I wanted to make the ALT lookup infallible (i.e. if we can't lookup, just return vec![]) so we will try once to build the transaction, and if it fails, then it fails and we refund. Feels like a clean solution. I believe this approach will be slightly more efficient as well.

hm..I am not sure I agree. I think this introduces a lot of custom solana ccm specific logic (with solana specific terms like alts) in the ingress egress pallet (which is generic for all chains). Also, the part of the expiry logic is also replicated here which is being dealt with as part of alt elections (see latest on my branch). I think we should aim to keep as much logic outside ingress egress pallet as possible, in solana specific code impls. I implemented an alternative to this (basically what I was suggesting) in my branch. It makes minimal changes in the pallet and the changes (introducing an error type for waiting) feels more generic over chains and extends better to other chain instances of the pallet. Please have a look and let me know what you think.

As far as efficiency is concerned, I think efficiency is not a big concern here as the differences are minimal. Taking ccm out of storage and putting back should be a big overhead since the storage is anyways cached during extrinsic execution and only committed to the database once extrinsic (or the whole block building?) completes. Even so, the other approach could also be made so that we just take out of storage once. And the failing of ccm tx building can also be made more effecient by checking the alt election in the first line of the ccm apicall building function (in new_unsigned of ccm apicall) and returning immediately if the elections have not completed.

One thing that is missing from my PR is the fallback transfer if alt witnessing doesnt succeed but I think It could be added to the election expiry mechanism that I have, in a clean way. 

Please have a look at my PR. Happy to discuss this further if you dont agree.

syan095:
>Please have a look at my PR. 

Ok I can see the merits of your approach - basically we try to build every block, then fail early because the ALT lookup fails. When we get this specific error (TryAgainLater) then we re-add the ccm back into storage to be retried next block.

Then, on expiry, just remove the stale CCM and trigger the refund logic. 

With this approach, minimal code change is needed. I agree that the extra overhead is not significant. 

Ok I think we can go with your approach. Sorry this PR didn't help you that much

But I think the part about a AuxLookupKey that is expandable can be used - ALT variant contains info that can be used for lookup and handle expiry, and we can add more lookup variant in the future.

albert-llimos:
Alright, then let's go for Ramiz's approach to it if you both agree :ok_hand: 
I agree with Roy that the AuxLookupKey approach in this PR is better both to handle expiry and also is more usable in the future for other use cases. @ramizhasan111 @syan095

ramizhasan111:
> Alright, then let's go for Ramiz's approach to it if you both agree 👌 I agree with Roy that the AuxLookupKey approach in this PR is better though. Both to handle expiry and also is more usable in the future for other use cases. I think we should port that to the Ramiz's PR. @ramizhasan111 @syan095

I agree. I think we can keep that part of this PR. I initially thought we could trigger the fallback tx from the election expiry. But its not simple to do so since the election doesnt have access to the ccm. So I think we can keep this. @syan095

syan095:
@coderabbitai summary

syan095:
@coderabbitai review

syan095:
OK i've rebased this branch onto the latest from Alt branch and have everything compiles and tests pass. 
@ramizhasan111 Please give it another review and merge if you are happy.

Cheers!

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1772,39 +1789,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-									amount: egress_details.egress_amount,
-									destination_address: fallback_address,
-								}),
-								Err(e) => log::warn!("Cannot refund failed Ccm: failed to Egress. swap_request_id: {:?}, Error: {:?}", ccm.swap_request_id, e),
-							};
-						}
-					} else {
-						log::warn!("Cannot refund failed Ccm: failed to decode `ccm_additional_data`. swap_request_id: {:?}", ccm.swap_request_id);
-					}
+					log::warn!("Failed to construct CCM. Fund will be refunded to the fallback refund address. egress_id: {:?}, aux_data_lookup_key: {:?}, Error: {:?}", ccm.egress_id, ccm.aux_data_lookup_key, error);
```
kylezs, line 1792:
```suggestion
					log::warn!("Failed to construct CCM. Funds will be refunded to the fallback refund address. egress_id: {:?}, aux_data_lookup_key: {:?}, Error: {:?}", ccm.egress_id, ccm.aux_data_lookup_key, error);
```

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2838,15 +2824,43 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		) {
+			if let Some(fallback_address) = decoded_data.refund_address::<T::TargetChain>() {
+				match Self::schedule_egress(
+					ccm.asset,
+					ccm.amount,
+					fallback_address.clone(),
+					None,
+					None
+				) {
+					Ok(egress_details) => Self::deposit_event(Event::<T, I>::InvalidCcmRefunded {
```
kylezs, line 2844: I'm not sure this will be enough information in the event for frontends to link things together easily. May need to include the egress_id too. CC: @acdibble

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2046,12 +2046,26 @@ pub mod pallet {
...
+					T::CcmValidityChecker::decode_unchecked(
+						ccm.channel_metadata.ccm_additional_data.clone(),
+						asset.into(),
+					) {
+					// Only pass in `swap_request_id` if Ccm requires auxiliary data lookup.
+					aux_lookup_key = (!sol_ccm_data.address_lookup_tables().is_empty())
+						.then_some(swap_request_id);
+				}
+			}
+
```
kylezs, line 2062:
```suggestion
			let aux_lookup_key = maybe_ccm_metadata.as_ref().and_then(|ccm| {
				T::CcmValidityChecker::decode_unchecked(
					ccm.channel_metadata.ccm_additional_data.clone(),
					asset.into(),
				)
				.ok()
				.and_then(|decoded| match decoded {
					DecodedCcmAdditionalData::Solana(sol_ccm_data)
						if !sol_ccm_data.address_lookup_tables().is_empty() =>
						Some(swap_request_id),
					_ => None,
				})
			});
```

state-chain/primitives/src/lib.rs
```diff
@@ -457,3 +457,31 @@ pub struct DcaParameters {
...
+
+#[derive(RuntimeDebug, PartialEq, Eq, Encode, Decode, TypeInfo, Clone, Default)]
+pub enum CcmAuxDataLookupKey<BlockNumber: Clone> {
+	#[default]
+	NotRequired,
+	Alt {
+		// Used to lookup ALT query result
+		swap_request_id: SwapRequestId,
+		// Used to calculate when to expire and refund ccm
+		block_created: BlockNumber,
```
kylezs, line 469: nit: The type already tells what type it is (plus `block_created` sounds like a function that returns a bool)
```suggestion
		created_at: BlockNumber,
```
would be same for the function too ofc

state-chain/pallets/cf-ingress-egress/src/migrations/add_ccm_aux_data_lookup_key.rs
```diff
@@ -0,0 +1,56 @@
...
+		// Where funds might be returned to if the message fails.
+		pub ccm_additional_data: CcmAdditionalData,
+		pub gas_budget: GasAmount,
+	}
+}
+
+pub struct AddCcmAuxDataLookupKeyMigration<T, I>(PhantomData<(T, I)>);
+
+impl<T: Config<I>, I: 'static> UncheckedOnRuntimeUpgrade for AddCcmAuxDataLookupKeyMigration<T, I> {
+	fn on_runtime_upgrade() -> Weight {
```
kylezs, line 30: Should have some pre/post upgrade hooks

state-chain/primitives/src/lib.rs
```diff
@@ -457,3 +457,31 @@ pub struct DcaParameters {
...
+		}
+	}
+
+	pub fn block_created(&self) -> Option<BlockNumber> {
+		match self {
+			CcmAuxDataLookupKey::NotRequired => None,
+			CcmAuxDataLookupKey::Alt { block_created, .. } => Some(block_created.clone()),
+		}
+	}
+}
```
kylezs, line 487: We only call these functions once, and they can never really be true without the other on. i.e. block_created() can't return Some() if swap_request_id returns None. This suggests the structure can be improved in `do_egress_scheduled_ccm` to eliminate this separation, and make these two things more tightly coupled.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2838,15 +2824,43 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+						asset: ccm.asset,
+						amount: egress_details.egress_amount,
+						destination_address: fallback_address,
+					}),
+					Err(e) => log::warn!("Cannot refund failed Ccm: failed to Egress. aux_data_lookup_key: {:?}, Error: {:?}", ccm.aux_data_lookup_key, e),
+				};
+			}
+		} else {
+			log::warn!("Cannot refund failed Ccm: failed to decode `ccm_additional_data`. aux_data_lookup_key: {:?}", ccm.aux_data_lookup_key);
+		}
```
kylezs, line 2854: This seems like a pretty bad case? How can this happen in reality?

state-chain/pallets/cf-swapping/src/lib.rs
syan095: Nice!

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: This should never happen. We guard against invalid CCM message in Vault swaps and when opening deposit channel. This is just there just in case.

state-chain/primitives/src/lib.rs
syan095: yeah but the request ID and created_at block are used for different purposes at different places. I feel like it's cleaner to be able to extract the exact information when it's needed. 

Alternatively the extracted data have to be saved as Option<(RequestId, BlockNumberFor)>, or (Option<RequestId>, Option<BlockNumberFor>) - nether feel like very nice

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: I've added the egress_id in anyway

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: > We guard against invalid CCM message in Vault swaps and when opening deposit channel

Where is that code?

syan095: Yeah the check is done with the `CcmValidityChecker::check_and_decode` function. It's done at the following places:
`process_vault_swap_request_prewitness()` (IngressEgress::vault_swap_request)
`process_vault_swap_request_full_witness()` (IngressEgress::vault_swap_request)
`pallet_cf_swapping::request_swap_deposit_address_with_affiliates()`
`cf_request_swap_parameter_encoding()` in our RuntimeApi for requesting Vault Swap encodings

state-chain/primitives/src/lib.rs
kylezs: I was just pointing out a symptom hoping you'd dive a bit further, sorry if it wasn't super clear.  I'm suggesting this is a "code smell", and that it's symptomatic of something deeper in the structure that could be improved here.

There a number of problems I can see including the above comment that are tied together
1. We take a swap_request_id for every execute swap and call, despite it only being used by one instantiation.
2. The swap_request_id is later used in an `if let Some()` to just append to something that's always passed in (this suggest to me that if we could do this somewhere else, we could remove the Option entirely)
3. Solana specific stuff is now scattered around the ingress-egress pallet. It's likely we'll need *some* modifications to the ingress-egress pallet ofc. But I think we can do better to hide it around abstractions like Chain::AuxData or something similar.
4. The same aux data type is used for all chains - I would question why that's necessary.  The other chains don't need it at all. If we needed something else for AssetHub CCM, would we add another parameter, or would it be another enum variant? I think it makes more sense to make this generic per chain in some way as in 3.

Didn't have too much time to go further (could find some time tomorrow possibly) but perhaps the above will spark some ideas? 

Also to be clear, not necessarily saying you introduced the structural problems in this PR either, it could be that the current code isn't flexible enough.

foreign-chains/solana/sol-prim/src/consts.rs
```diff
@@ -61,4 +61,7 @@ pub const ACCOUNT_KEY_LENGTH_IN_TRANSACTION: usize = 32usize;
 pub const ACCOUNT_REFERENCE_LENGTH_IN_TRANSACTION: usize = 1usize;
 
 pub const MAX_CCM_USER_ALTS: u8 = 3u8;
-pub const EXPIRY_TIME_FOR_ALT_ELECTIONS: BlockNumber = 43200u32;
+
+/// The maximum number of blocks we wait for external Aux data elections to finish
+/// before we fail to egress CCM messages. TODO Albert: come up with a reasonable number.
+pub const EXPIRY_TIME_FOR_ALT_ELECTIONS: BlockNumber = 600u32;
```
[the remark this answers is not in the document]
dandanlen: Re. these constants - the aren't really Solana primitives, these are protocol-specific config items. For example `EXPIRY_TIME_FOR_ALT_ELECTIONS` only seems to be used in solana_elections.rs, we might as well define it where it is used. 

(Similar thing goes for MAX_CCM_USER_ALTS)

albert-llimos:
Seems like it was just rebased but it still needs to be rebased properly again :thinking:

kylezs:
> Seems like it was just rebased but it still needs to be rebased properly again 🤔

Just based on this comment, I'm guessing the issue is that it requires an interactive rebase, and on this branch we just need to drop the commits that are already on the target branch of this PR (git can't interpret them as the same because it's base has changed commit hash with the squash merge).

foreign-chains/solana/sol-prim/src/consts.rs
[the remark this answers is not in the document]
syan095: This is now moved to cf_chians::sol::sol_tx_core, where our state_chain logic is. 
Also the number has been updated to 100 (10 min)

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: Got it. I've added the new egress ID as well - so now both IDs are there in the event.

state-chain/primitives/src/lib.rs
syan095: I did a large refactor for this. 
Now that the aux data is part of the Chain trait, everything is done through generic traits, it feels more correct to separate the idea of "expiry time" which will most likely be common across all types of aux data (I know we only have ALT now, but we may add more in the future), with the concept of "key". 

Therefore, I still kept the "created_at" and the "swap_request_id" functions separate - it just makes more sense this way. Also the code is now more "generic" and contains less hard-coded, chain specific types

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -259,16 +259,17 @@ pub struct CrossChainMessage<C: Chain> {
 	// Where funds might be returned to if the message fails.
 	pub ccm_additional_data: CcmAdditionalData,
 	pub gas_budget: GasAmount,
-	pub swap_request_id: SwapRequestId,
+	// Contains information used to lookup for auxiliary data.
+	pub aux_data_lookup_key: CcmAuxDataLookupKey<BlockNumber>,
```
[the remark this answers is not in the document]
syan095: I've now incorporated the Aux data into the Chain trait. All conversions are now done through traits.
Please have a look to see if the new design makes sense

state-chain/pallets/cf-ingress-egress/src/migrations.rs
```diff
@@ -1,4 +1,15 @@
...
-pub type PalletMigration<T, I> = (PlaceholderMigration<20, Pallet<T, I>>,);
+pub type PalletMigration<T, I> = (
+	VersionedMigration<
+		20,
+		21,
+		add_ccm_aux_data_lookup_key::AddCcmAuxDataLookupKeyMigration<T, I>,
+		Pallet<T, I>,
+		<T as frame_system::Config>::DbWeight,
+	>,
+	PlaceholderMigration<21, Pallet<T, I>>,
```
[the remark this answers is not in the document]
syan095: I'll do a rebase for the entire feature base-branch. Then I can fix it in the base branch

syan095:
@coderabbitai review

syan095:
Ok everything's factored and rebased and ready to be reviewed and merged

syan095:
https://github.com/coderabbitai summary

state-chain/traits/src/lib.rs
```diff
@@ -1180,5 +1181,10 @@ pub trait InitiateSolanaAltWitnessing {
 	fn initiate_alt_witnessing(
 		_ccm_channel_metadata: CcmChannelMetadata,
 		_swap_request_id: SwapRequestId,
-	);
+	) {
+	}
+
+	fn max_wait_time_for_alt_witnessing() -> BlockNumber {
```
albert-llimos, line 1187: Now that we have renamed the `SolanaAltWitnessingHandler` to `CcmAuxDataWitnessingHandler` can't we also rename this to `max_wait_time_for_ccm_aux_data_witnessing` or something like that? Same for the trait itself (`InitiateSolanaAltWitnessing`) and the other method, `initiate_alt_witnessing`.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1746,6 +1752,24 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				{
+					Self::refund_invalid_ccm(
+						ccm,
+						ExecutexSwapAndCallError::DispatchError(
+							Error::<T, I>::ExceededMaxCcmAuxDataWaitTime.into(),
+						),
+					);
+					continue;
+				}
+			};
```
albert-llimos, line 1771: I think there is an edge case with this approach.
It could be the case that the election has finished at this point but the we decide it has timed out and proceed to refund. While this sounds very unlikely, it could well happen in Solana since the number of nonces we have is limited, even if it will be increased to 50 in this release. A ccm egress could be waiting for several minutes and not being tried because the `maybe_no_of_transfers_remaining` (nonces) above is preventing it.
I don't think there is an easy way to solve this looking at the code (I could be wrong). I think this is just minor improvements, similar to another edge case we want to address. I'd be happy merging this and I can open a ticket to improve it. The only painful thing is if it requires a migration if we don't address it before 1.9 is released, but I think that's fine.
Wdyt @kylezs @syan095 ?

albert-llimos: Actually, there might be an easy way to implement this 🤔
Can we just do this check of expiry in the ExecutexSwapAndCallError::TryAgainLater below and refund if we deem it as expired, otherwise reappend it as done currently. It's true that then we might build a transaction past the max_wait_time_for_alt_witnessing time, but I think it's better to at least try once since this is not a hard timeout.

syan095:
@coderabbitai review

state-chain/traits/src/lib.rs
syan095: Ok hear me out on this one. The design I had in mind is to have this generic "CcmAuxDataWitnessingHandler", that can be used to handle all specific types of Aux data (when we have more than 1 in the future). 
The design I had in mind is - say we now have another Aux data - maybe we need to witness data X for Ethereum ccm now. 

THen we will have another trait called `HandleEthereumXDataWitnessing`, and that the `CcmAuxDataWitnessingHandler` will need to impl both traits. 

This way, each specific trait is used to manage 1 thing/type of aux data. But the over-arching trait that connects them all with the pallet `CcmAuxDataWitnessingHandler` is generic.

This is the design I had in mind - since we only have 1 type of aux data it may be unclear. What do you think?

albert-llimos: Gotcha. Alright, I'm fine with that :+1:

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: (we could also write a test to catch this case)

state-chain/traits/src/lib.rs
kylezs: Mmmm if we do this way it will mean we will have many different branches in `egress_scheduled_ccm` where we do. 
```
if let Some(ccm_channel_metadata) = channel_metadata {
					if ForeignChain::Solana == destination_asset.into() {
						T::CcmAuxDataWitnessingHandler::initiate_alt_witnessing(
							ccm_channel_metadata,
							swap_request_id,
						);
					}
				}
```
 
which is what we want to avoid. 

I don't think it's too difficult to make some of this general now after the refactoring done so far.

All you'd need to is push the checks that are specific for solana into the *implementations* rather than the generic code like we have here.

So instead of this:
```rust
			if let Some(created_at) =
				ccm.aux_data_lookup_key.clone().and_then(|key| key.created_at())
			{
				if current_block >=
					created_at +
						T::CcmAuxDataWitnessingHandler::max_wait_time_for_alt_witnessing()
				{
					Self::refund_invalid_ccm(
						ccm,
						ExecutexSwapAndCallError::DispatchError(
							Error::<T, I>::ExceededMaxCcmAuxDataWaitTime.into(),
						),
					);
					continue;
				}
			};
```

We would have something like:
```
If AuxDataHandler::should_refund(ccm.aux_data_lookup_key) {
   Self::refund_invalid_ccm(
		ccm,
		ExecutexSwapAndCallError::DispatchError(
			Error::<T, I>::ExceededMaxCcmAuxDataWaitTime.into(),
		),
	);
	continue;

}
```

I agree that it's a little difficult to generalise it now with only one chain that really needs this. But I think with current information we can still do a little better.

The `initiate_alt_witnessing` wouldn't check the chain for example, it would just always be called, and it would be a noop in the implementations that don't need it.

```rust
if let Some(ccm_channel_metadata) = channel_metadata {
					T::CcmAuxDataWitnessingHandler::initiate_aux_data_witnessing(
						ccm_channel_metadata,
						swap_request_id,
					);
				}
```

Now we don't have anything solana specific in this layer. The *type* of data passed through is solana specific ofc, because we don't know what we need for the others, so would say that's fair. But at least we've removed chain branching inside the generic code.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2046,12 +2047,29 @@ pub mod pallet {
 		) {
 			let is_ccm_swap = maybe_ccm_metadata.is_some();
 
+			let aux_lookup_key = maybe_ccm_metadata.as_ref().and_then(|ccm| {
+				T::CcmValidityChecker::decode_unchecked(
```
kylezs, line 2051: We only test decode_unchecked returns Ok() - we should probably check the data is correct too in some of the tests?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2046,12 +2047,29 @@ pub mod pallet {
...
+					ccm.channel_metadata.ccm_additional_data.clone(),
+					asset.into(),
+				)
+				.ok()
+				.and_then(|decoded| match decoded {
+					DecodedCcmAdditionalData::Solana(sol_ccm_data)
```
kylezs, line 2057: This is also similar to the above comment on the ingress-egress pallet. It's less straightforward in the swapping pallet - given it's not instantiable. However In the same way, EgressHandler: EgressApi trait dispatches to each of the chains, hiding the details and keeping the swapping code very tidy, we could do the same here for AuxData. It should be the Any type's responsibility to do this, rather than the swapping pallet's

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: Have a look at the latest commit! 
I've passed the entire key to the API call building, and the error returns the "created_at" value to calculate expiry. This gets around the edge case, AND also also solves the problem @kylezs mentioned before, where the "created_at" and "swap_id" are fragmented.

syan095: see if the deisgn looks OK - I can add more tests once we have the logic locked down

state-chain/traits/src/lib.rs
syan095: I just read this comment. I'll think about this a bit more! thanks!

state-chain/runtime/src/lib.rs
```diff
@@ -421,7 +418,7 @@ impl pallet_cf_ingress_egress::Config<Instance2> for Runtime {
 	type AssetConverter = Swapping;
 	type FeePayment = Flip;
 	type SwapRequestHandler = Swapping;
-	type CcmAuxDataWitnessingHandler = SolanaAltWitnessingHandler;
+	type CcmAuxDataWitnessingHandler = CfSolanaAltWitnessingHandler;
```
kylezs, line 421: 👀  - this is strange having solana in non-solana instances, my comment above would resolve this (by having an instance for each pallet - where all but solana would be noops).

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: Yeah definitely a step in the right direction! (would still rename `max_wait_time_for_alt_witnessing` to `min_wait_time_for_alt_witnessing`. However, this is likely not required if implementing my other suggestions to put this stuff into the generic implementations instead of in the pallet.

albert-llimos: The change looks good to me :+1: 
Agree with Kyle that if we leave the implementation as is it makes sense to rename that.

state-chain/traits/src/lib.rs
syan095: Ok - this is now changed. the trait handles expiry checks.

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: I've added the check to after we have tried to build so we will try to build 1 more time after expiry - this should cover the special edge case. I've also added an integration test for this.

syan095: @kylezs I thought about this and still feel like "Max" is the correct name here. Max means we will wait for this long, but the wait can be shorter (when the data is ready sonner). "Min" would imply that we will always wait for at least this amount of time, even when the aux data is ready earlier, which is not the case here.

state-chain/pallets/cf-swapping/src/lib.rs
syan095: Yeah it is safe to decode without checks here because the check would have been done for all cases when the swap was initiated. We do this check when deposit channels are open and when Vault swaps are processed. So by the time we are here - we can be sure that the metadata is valid.

kylezs: Sorry, wasn't super clear, I mean to add actual unit tests for it, nothing to change in this code here :)

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: We can wait longer than the max, which makes it not a max. 
Min wait time can be exited in the case where we are no longer waiting if that makes sense. Basically we're saying "we'll definitely wait at least this long, unless it succeeds" so it's the "min wait time" but not the "min success time"

Though I do agree, it's still a bit confusing.... so we should just delete the function, which we can do now :). Then we can just change the tests to move forward the constant.

state-chain/runtime/src/lib.rs
syan095: Yeah I've refactored this. Now only the Solana uses the Solana ALT handler, while the rest uses a "noop" version.

state-chain/pallets/cf-ingress-egress/src/lib.rs
kylezs: We still have SolanaTransactionBuildingError in the ingress-egress pallet which suggests we can still go further here.

What we can do is make the ExecutexSwapAndCallTrait more general. The fact we have `FailedToBuildCcmForSolana` as a variant suggests we can do more here, and that's much clearer now after doing the changes so far.

What are we trying to do here? We're saying: When we build a ccm call, we can error for it to be rescheduled, OR we can error for it to be refunded.

Note in the above there's nothing about solana specifically. We can could this in the `new_unsigned`. Where new_unsigned returns back:
- the api_call -> Ok() as is currently the case
- ExecutexSwapAndCallError::Reschedule(ccm)
- ExecutexSwapAndCallError::RefundInvalidCcm(ccm)

Then:
- Everything about the created_at etc. can be hidden away
- There's no need for Solana specific variants on the generic error type
- We can simplify the `CcmAuxDataWitnessingHandler` by removing `should_expire`
- We don't have two calls to `refund_invalid_ccm` (that fact we have two also indicates we can improve this)

Thoughts? 
CC: @dandanlen too

state-chain/pallets/cf-swapping/src/lib.rs
syan095: I've refactored some logic into the `AnyChainCcmAuxDataLookupKey`. I'm not sure if this is good enough but I really struggle to make this more generic.

state-chain/pallets/cf-swapping/src/lib.rs
syan095: Sorry I misread your comment. I've now added a unit test for this

state-chain/pallets/cf-ingress-egress/src/lib.rs
syan095: I feel like what you are suggesting is a much larger refactoring - not necessarily in terms of lines changed but more of a change in the general design. 

e.g. you mentioned `SolanaTransactionBuildingError` should be removed from the IngressEgress - I think it's just there for convenience. In the past, specific errors are often swallowed and turned into a more generic error (like "ccm_failed") with no information on the exact reason something has failed. This is why I thought it would be an improvement to pass the error all the way into the Event, so when something does fail, it provides the user with as much information as possible. Otherwise why would we even have all these detailed errors in the code anyways? they all get lost in the conversions and just is never useful. 
Another point I would like to ask for clarification is exactly how "generic" does the pallet need to be? because from my point of view, I'm jumping through hoops and going around in circle with the use of traits just to remove a variable or function name that has "solana" in its name. Even then, it cannot be exactly completely "generic". E.g. here we need to create a `AltLookup` that needs a SwapRequestId, and a block number. How do we make it more generic? Only the pallet have access to block number and the swap request ID, there's no way to delegate everything completely to the Chain implementation - we have to, at some point pass in the SwapRequestId and get the current block number no matter how we use the traits. Since we only have 1 type of Aux data now, we cannot distinguish what is generic and what is not. Maybe the next aux data also uses SwapRequestID, maybe it uses something else - we cannot know. Where do we draw the line here then? is it generic enough to pass in the SwapRequestID, because it is used specifically by Solana? if it's not generic enough how do we pass in the SwapRequestId in generically enough? if it is just about changing a name and use a trait (like calling if "into_aux_key" instead of "solana_alt_lookup", then this feels superfical. I think I want to know more about your thought on this.

I think the confusion rises exactly because we only have this 1 use case for Solana, so this logic feels not generic enough. e.g. for calculating expiry, if we do the calculation in the ingress-egress pallet, this feels like Solana specific logic, because it is, and only Solana uses this logic. What if the next aux data also expires using the same mechanism? then this code will be generic enough and is shared between the different aux witnessing. I feel like we are just over-engineering here. THere's no way for us to know what the next thing will be, and we are just spending lots of time (I've already spent a week on this trying to make everything generic enough) trying to make everything fits the "overall design" when in my opinion I just think we do not have enough dots to even draw a line. 

For the suggestion of improving the `ExecutexSwapAndCallTrait` trait - I really don't have a strong opinion on this. In the past, I've just trying to keep everything the same as what was previously written. I have not thought this needs changing. If we agree we want to change this, then I have no strong opinion eitherway. 

But even in this new design "Everything about the created_at etc. can be hidden away" is still not exactly true. To know if something can expire, you will still need to know the created_at block, the current block and the max/min waiting time(whatever we want to call it). These parameter still need to be passed into the ApiCallBuilder, it's just you do it in the call builder instead of at the pallet.  Does this make it more generic? that we do the expiry calculation outside the ingress egress pallet but still passes these data into a trait function? I really don't know.

Finally I'm not sure about how soon we want to get this versioned transaction feature out. I feel like this is feature complete for a long time already and we are just fluffing it up. If a refactor is potentially quite large and can be done more independently I would suggest that we do it as a separate ticket. This PR is specially about expiring ccms, and I feel like there's a scope creep and it's not really focused on this anymore. 

This is just my thoughts though. If you and Dan think we should do this refactor here then I'd be happy to do it. I'm taking next week and half off and I was hoping to finish the whole versioned transaction feature before I go. If we need to make more changes as part of the versioned transaction feature, then someone else will need to take over, or wait until I come next later next week. I 'm working through the weekend hoping I'd get everything finished in time but if it can't happen then so be it.

I feel like my understanding of our design philosophy is still lacking and I know I still need to learn more.

syan095: This feels like a bigger refactor that can be split off into a separate PR/ticket. 

I have a few question that needs clarification:
1. Regarding the concept of the pallet being generic, I think I kind of understand the goal, but unclear on where we should draw the line. for example - in creating the lookup key for ALT, we will need the "swap request ID" and current block number - these MAY be Solana specific (It is impossible to know until we need more types of Aux data). How is it that if we create the key in a trait impl that's OK, but it's not generic if we do it in the pallet. The decision seems a bit artificial if it is just because the code has the word "Solana" in it. Because no matter how we generalize the code, we still need to pass these Solana specific variable off somewhere, so we can never get rid of "solana specific" logic.

2. Another thing I'm unsure about is we are adding complications when trying to keep everything "generic". e.g. the code for lookup key used to be in just 2 places, 1 in swapping for key creation, and 1 in ingress-egress for key consumption. 
But once the AuxData type is introduced to the Chain trait, the lookup key logic is now spread into multiple places scattered across the code base (we have the chain trait, Auxdata key trait, 3 types of AuxDataLookupKey - 1 for Solana, 1 for non-solana and 1 for AnyChain, and Anychain's EgressApi impl etc.). To keep the code "generic", the code is more far more complicated and much harder to understand and manage. I'm just unsure in what case we should generalize things into traits and when we can just use concrete types to keep things simple.

Anyways I was hoping to finish this up before I go on a break, that's why I was suggesting to do this as a separate ticket. But it's better to get it right the first time, then I'm happy to just wait and finish this when I come back. It would be good to lock down the design/spec so avoid scope creep.

state-chain/runtime/src/lib.rs
```diff
@@ -389,7 +389,7 @@ impl pallet_cf_ingress_egress::Config<Instance1> for Runtime {
 	type AssetConverter = Swapping;
 	type FeePayment = Flip;
 	type SwapRequestHandler = Swapping;
-	type CcmAuxDataWitnessingHandler = NoOpAuxDataWitnessingHandler;
+	type CcmAuxDataWitnessingHandler = SolanaCcmAuxDataWitnessingHandler;
```
dandanlen, line 392: This doesn't seem right?

albert-llimos: I think Github might be displaying a funny diff (maybe your comment was done before?). The diff seems fine to me.

-  type SolanaAltWitnessingHandler = SolanaAltWitnessingHandler;
+ type CcmAuxDataWitnessingHandler = SolanaCcmAuxDataWitnessingHandler;

albert-llimos: I think you're right that my change might be wrong but then the fix needs to be applied in some other way :thinking:

dandanlen: FWIW I don't think it was wrong, just the naming was quite confusing. (see my commit below).

syan095:
This PR is now redundant. Refactors will be done in separate PR

PR #5671, opened 2025-02-25
feat: remove dedup hook

marcellorigotti:
# Pull Request

Closes: PRO-2054

## Checklist

## Summary

Remove dedup hook and perform the deduplication (which is common to all the ES) inside the execute hook -> this can now easily be generalized into a function.

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
@@ -28,8 +28,7 @@ use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, marker::PhantomData, 
 /// - The block data type.
 /// - The event type produced during block processing.
 /// - The rules to generate events (for example, pre-witness and full witness rules).
-/// - The logic for executing events.
-/// - The logic for deduplicating events.
+/// - The logic for executing and deduplicating events.
 ///
 /// These are defined via the [`BWProcessorTypes`] trait, which is a generic parameter for this
 /// processor.
@@ -42,7 +41,6 @@ use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, marker::PhantomData, 
 ///     - `Event`: The type of event generated from processing blocks.
 ///     - `Rules`: A hook to process block data and generate events.
 ///     - `Execute`: A hook to execute generated events.
-///     - `DedupEvents`: A hook to deduplicate events.
 ///     - `SafetyMargin`: A hook to retrieve the chain specific safety-margin
 #[derive_where(Debug, Clone, PartialEq, Eq;
 	T::ChainBlockNumber: Debug + Clone + Eq,
@@ -50,7 +48,6 @@ use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, marker::PhantomData, 
 	T::Event: Debug + Clone + Eq,
 	T::Rules: Debug + Clone + Eq,
 	T::Execute: Debug + Clone + Eq,
-	T::DedupEvents: Debug + Clone + Eq,
 	T::SafetyMargin: Debug + Clone + Eq,
 )]
 #[derive(Encode, Decode, TypeInfo, Deserialize, Serialize)]
@@ -60,7 +57,6 @@ use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, marker::PhantomData, 
 	T::Event: Encode,
 	T::Rules: Encode,
 	T::Execute: Encode,
-	T::DedupEvents: Encode,
 	T::SafetyMargin: Encode,
 ))]
 pub struct BlockProcessor<T: BWProcessorTypes> {
@@ -72,7 +68,6 @@ pub struct BlockProcessor<T: BWProcessorTypes> {
 	pub reorg_events: BTreeMap<T::ChainBlockNumber, Vec<T::Event>>,
 	pub rules: T::Rules,
 	pub execute: T::Execute,
-	pub dedup_events: T::DedupEvents,
 	pub safety_margin: T::SafetyMargin,
 }
 impl<BlockWitnessingProcessorDefinition: BWProcessorTypes> Default
@@ -84,7 +79,6 @@ impl<BlockWitnessingProcessorDefinition: BWProcessorTypes> Default
 			reorg_events: Default::default(),
 			rules: Default::default(),
 			execute: Default::default(),
-			dedup_events: Default::default(),
 			safety_margin: Default::default(),
 		}
 	}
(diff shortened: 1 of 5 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -323,17 +313,21 @@ pub(crate) mod test {
 		}
 	}
 
-	impl Hook<HookTypeFor<Types, DedupEventsHook>> for Types {
+	impl Hook<HookTypeFor<Types, SafetyMarginHook>> for Types {
+		fn run(&mut self, _input: ()) -> u32 {
+			3
```
kylezs, line 318: 👀 We should be fetching this value from ElectoralSettings?

state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
```diff
@@ -43,46 +41,106 @@ type TypesVaultDepositWitnessing = TypesFor<BitcoinVaultDepositWitnessing>;
...
+						chosen.insert(deposit, (block, event));
+					},
+					// This should be impossible to reach!
+					(_, _) => (),
+				}
+			},
+		}
+	}
+	chosen.into_values().collect()
+}
```
kylezs, line 75:
```suggestion
fn dedup_events<T: Ord + Clone>(
	events: Vec<(BlockNumber, BtcEvent<T>)>,
) -> Vec<(BlockNumber, BtcEvent<T>)> {
	let mut chosen: BTreeMap<T, (BlockNumber, BtcEvent<T>)> = BTreeMap::new();

	for (block, event) in events {
		let deposit = event.deposit_witness().clone();

		// Only insert if no event exists yet, or if we're upgrading from PreWitness to Witness
		if !chosen.contains_key(&deposit) ||
			(matches!(chosen.get(&deposit), Some((_, BtcEvent::PreWitness(_)))) &&
				matches!(event, BtcEvent::Witness(_)))
		{
			chosen.insert(deposit, (block, event));
		}
	}

	chosen.into_values().collect()
}
```

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
marcellorigotti: This is the test, in the actual impl we read directly from storage:
```rust
fn run(&mut self, _input: ()) -> u32 {
		u64::steps_between(&0, &BitcoinIngressEgress::witness_safety_margin().unwrap_or(0)).0 as u32
	}
``` 
Is there any advantage to put this in the settings?

kylezs: Yes, it can be changed dynamically, it's not hardcoded in there, and will be correctly synchronised between elections.

marcellorigotti: Will create a ticket to update this 👍

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/state_machine.rs
```diff
@@ -68,12 +68,6 @@ impl<T: BWProcessorTypes> HookType for HookTypeFor<T, RulesHook> {
...
 pub struct ExecuteHook;
 impl<T: BWProcessorTypes> HookType for HookTypeFor<T, ExecuteHook> {
-	type Input = (T::ChainBlockNumber, T::Event);
-	type Output = ();
-}
-
-pub struct DedupEventsHook;
-impl<T: BWProcessorTypes> HookType for HookTypeFor<T, DedupEventsHook> {
 	type Input = Vec<(T::ChainBlockNumber, T::Event)>;
 	type Output = Vec<(T::ChainBlockNumber, T::Event)>;
```
MxmUrw, line 72:
```rust
    type Input = Vec<(T::ChainBlockNumber, T::Event)>;
	type Output = ();
```
The output type should be `()` since we don't have to care about which event where actually executed, that's entirely up to the discretion of the ExecuteHook.

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -355,13 +349,7 @@ pub(crate) mod test {
 					},
 				}
 			}
-			chosen.into_values().collect()
```
MxmUrw, line 358: I think this hook implementation can now be empty since we don't want to return anything from this hook. Instead we could unit-test the the `deduplicate_events()` function (if we don't do that already...).

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -163,12 +157,9 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
...
 		}
 		let events = self.process_rules(last_block);
-		let last_events = self.dedup_events.run(events);
-		for event in &last_events {
-			self.execute.run(event.clone());
-		}
+		let executed_events = self.execute.run(events.clone());
 		self.clean_old(last_block);
-		last_events
+		executed_events
```
MxmUrw, line 162: Here we actually want to return `events`.

state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
```diff
@@ -43,46 +41,95 @@ type TypesVaultDepositWitnessing = TypesFor<BitcoinVaultDepositWitnessing>;
...
+				},
+				BtcEvent::Witness(deposit) => {
+					BitcoinIngressEgress::process_channel_deposit_full_witness(
+						deposit.clone(),
+						*block,
+					);
+				},
+			}
 		}
+		deduped_events
```
MxmUrw, line 89: We could deduplicate the logic here by creating a generic `DedupExecHook<Number, Blockdata, Deposit, Prewitness: Hook<>, Witness: Hook<>>`

Then we could even instantiate this hook for testing.

state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
```diff
@@ -43,46 +41,95 @@ type TypesVaultDepositWitnessing = TypesFor<BitcoinVaultDepositWitnessing>;
...
+		for (_, event) in &deduped_events {
+			match event {
+				BtcEvent::PreWitness(_) => { /* We don't care about pre-witnessing an egress*/ },
+				BtcEvent::Witness(egress) => {
+					BitcoinBroadcaster::broadcast_success(egress.clone());
+				},
+			}
 		}
+		deduped_events
 	}
```
MxmUrw, line 133: Could be deduplicated, see above.

PR #5672, opened 2025-02-25
fix: eth broker level screening, push empty events in broker api

MxmUrw:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

We had an "optimization" in place in 1.7 where we wouldn't push events if they vec was empty, on main we already removed it. This PR also removes it here.

In particular, since we now also have eth events this would have meant that we won't push any events if there aren't any btc events.

### DIFF
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -240,13 +240,6 @@ impl RpcServer for RpcServerImpl {
 					.await
 				{
 					Ok(events) => {
-						// We only want to send a notification if there have been events.
-						// If other chains are added, they have to be considered here as
-						// well.
-						if events.btc_events.is_empty() {
-							continue;
-						}
-
 						let block_update = BlockUpdate {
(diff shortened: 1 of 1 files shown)


api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -240,13 +240,6 @@ impl RpcServer for RpcServerImpl {
 					.await
 				{
 					Ok(events) => {
-						// We only want to send a notification if there have been events.
-						// If other chains are added, they have to be considered here as
-						// well.
-						if events.btc_events.is_empty() {
```
Janislav, line 246: We could explicit check for chains instead of suddenly allow all chains. Not sure what this could have for side effects - was just wondering why not just add Ethereum for e.x: `events.btc_events.is_empty() || events.eth_events.is_empty()`

Janislav: `if events.btc_events.is_empty() && events.eth_events.is_empty()`

PR #5675, opened 2025-02-26
feat: bouncer run test command

j4m1ef0rd:
# Pull Request

## Checklist

## Summary

Welcome back the `run_test.ts` command.
Solves the problem of having to know the name of the test to run it. Now you can just give the command a file path and it will run the test using `vitest`.
It works by exploiting the `vitest list` command to execute the `.test` files without running the tests. I use this to save the test names and function names to a file that I can then be cross referenced with the given test file to find the correct test name to run. A bit hacky, but it works great.

Usage: 
```sh copy
./commands/run_test.ts ./tests/boost.ts
```

### DIFF
--- a/bouncer/commands/run_test.ts
+++ b/bouncer/commands/run_test.ts
@@ -9,9 +9,7 @@
 
 import { execSync } from 'child_process';
 import { existsSync, readFileSync } from 'fs';
-
-// Note: This must be the same path as the one in shared/utils/vitest.ts. Importing it here breaks vitest.
-export const testInfoFile = '/tmp/chainflip/test_info.csv';
+import { testInfoFile } from '../shared/utils';
 
 // Check that a test file was provided as an argument
 const testFile = process.argv[2];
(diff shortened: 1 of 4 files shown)


bouncer/commands/run_test.ts
```diff
@@ -0,0 +1,87 @@
...
+// This command takes one argument.
+// It will find and run the test function (using vitest) in the test file provided as an argument.
+//
+// For example: ./commands/run_test.ts ./tests/boost.ts
+// This is the equivalent of running `pnpm vitest run -t "BoostingForAsset"`
+
+import { execSync } from 'child_process';
+import { existsSync, readFileSync } from 'fs';
+
+// Note: This must be the same path as the one in shared/utils/vitest.ts. Importing it here breaks vitest.
```
kylezs, line 13: This is why we dupe it? Sounds strange that importing would break something? even if so, could we just import it the other way. Define here and import in vitest or the opposite to whichever way didn't work.

bouncer/commands/run_test.ts
```diff
@@ -0,0 +1,87 @@
...
+      testNamesAndFunctions.push({
+        testName: columns[0].trim(),
+        functionName: columns[1].trim(),
+      });
+    }
+  }
+} catch (err) {
+  console.error(`Error reading the ${testInfoFile} file:`, err);
+  process.exit(1);
+}
```
kylezs, line 59:
```suggestion
let testNamesAndFunctions;
try {
  testNamesAndFunctions = readFileSync(testInfoFile, 'utf8')
    .split('\n')
    .filter(row => row.includes(','))
    .map(row => {
      const [testName, functionName] = row.split(',').map(col => col.trim());
      return { testName, functionName };
    });
} catch (err) {
  console.error(`Error reading the ${testInfoFile} file:`, err);
  process.exit(1);
}
```

bouncer/shared/utils/vitest.ts
```diff
@@ -1,6 +1,26 @@
...
 
+export const testInfoFile = '/tmp/chainflip/test_info.csv';
+
+// Write the test name and function name to a file to be used by the `run_test.ts` command
+function writeTestInfoFile(name: string, testFunction: (context: TestContext) => Promise<void>) {
+  try {
+    let existingContent = '';
+    if (existsSync(testInfoFile)) {
+      existingContent = readFileSync(testInfoFile, 'utf-8');
+    }
```
kylezs, line 13:
```suggestion
const existingContent = existsSync(testInfoFile) ? readFileSync(testInfoFile, 'utf-8') : '';
```

bouncer/commands/run_test.ts
j4m1ef0rd: would have to put it in another file that both can import. Didnt think it was worth creating a file for. Or i guess i could just put it in utils.

kylezs (approved):

msgmaxim (approved):

PR #5683, opened 2025-02-27
fix: spec version checker 🐛

ahasna:


ci/scripts/check_runtime_spec_version.sh
```diff
@@ -4,8 +4,10 @@ network=$1
 
 runtime_spec_version_file="state-chain/runtime/src/lib.rs"
 
-# Extract the spec_version from the Rust file
-spec_version=$(grep -F '#[sp_version::runtime_version]' -A 10 $runtime_spec_version_file | grep 'spec_version' | awk '{print $2}' | tr -d ',')
+# Extract the spec_version from the Rust file and convert it to a comparable number
+raw_version=$(grep -F '#[sp_version::runtime_version]' -A 10 $runtime_spec_version_file | grep 'spec_version' | awk '{print $2}' | tr -d ',')
+# Convert version like 1_07_10 to proper semver by removing leading zeros
```
dandanlen, line 9: Just a nit, but it's not really semver, it's just a number.

ci/scripts/check_runtime_spec_version.sh
```diff
@@ -25,10 +27,13 @@ fi
 
 live_runtime_version=$(curl -s -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "state_getRuntimeVersion", "params":[]}' $RPC_ENDPOINT | jq .result.specVersion)
```
dandanlen, line 28: Does the live runtime version not also need the same conversion?

PR #5687, opened 2025-02-28
feat: enable bls for arb

Janislav:
# Pull Request

Closes: PRO-2079

## Checklist

## Summary

- Adds the Arbitrum support to the runtime config
- Adds a migration for Arb saftey margin
- Extends the broker API to support Arb

## Summary by CodeRabbit

- **New Features**
	- Introduced support for Arbitrum transactions with updated handling for transaction rejections.
	- Enabled transaction reporting to improve oversight.
	- Added a runtime upgrade process that adjusts the system’s safety margin for Arbitrum operations.


Janislav:
Have to fix something. Dont merge

Janislav:
Done

Janislav:
@coderabbitai review

state-chain/runtime/src/migrations/increase_arb_safety_margin.rs
```diff
@@ -0,0 +1,40 @@
...
+use sp_runtime::DispatchError;
+#[cfg(feature = "try-runtime")]
+use sp_std::vec::Vec;
+
+use codec::{Decode, Encode};
+
+pub struct Migration;
+
+// Keeping it a multiple of 24 to match the witness period of Arbitrum
+const NEW_ARB_SAFETY_MARGIN: u64 = 672;
```
dandanlen, line 15: Is this correct? The current value is 1...

cc @kylezs

kylezs: Well, this is what we would have if we were also to increase ETH to 2+ mins - this is the value I had here: https://github.com/chainflip-io/chainflip-backend/pull/5686

We can probably lower it to closer to what ETH is though I think: 6 (eth margin) * 14 (eth block time in secs) * 4(arb blocks per second) = 336

dandanlen: Ok but I still don't understand why it's currently 1, particularly in the context of the comment that says we're "keeping" it a multiple of 24.

dandanlen (approved):

PR #5690, opened 2025-03-03
feat: update ccm checker and test Solana CCM Versioned transactions

albert-llimos:
# Pull Request

Closes: PRO-2046

## Checklist


state-chain/chains/src/sol/transaction_builder.rs
```diff
@@ -876,11 +877,82 @@ pub mod test {
...
+			agg_key(),
+			durable_nonce(),
+			compute_price(),
+			TEST_COMPUTE_LIMIT,
+			vec![chainflip_alt()],
+		)
+		.unwrap();
+
+		let serialized_tx = sign_and_serialize(transaction);
+		assert_eq!(serialized_tx.len(), MAX_TRANSACTION_LENGTH - MAX_CCM_BYTES_SOL);
```
syan095, line 915: It's a bit unclear why this number is used to check the length. 
Is it that the `MAX_CCM_BYTES_SOL` is calculated using the MAX - Len(transfer_tx)? Might be good to leave a comment here to elaborate

syan095 (approved):
Overall looking good. I've fixed up some minor typos (I have a spell checker, so I just fixed some type in the same file even if they are not part of your work per se) I hope this is OK with you.

PR #5693, opened 2025-03-04
refactor: initialise when necessary

kylezs:
Minor refactor to initialise values only when necessary - some of these values are initialised very far away from where they're used, making it harder to read. 
Also, they're often never needed in the case of early returns.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -2511,18 +2511,20 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			}
 		}
 
+		let destination_address = match T::AddressConverter::decode_and_validate_address_for_asset(
+			destination_address,
+			destination_asset,
+		) {
+			Ok(address) => address,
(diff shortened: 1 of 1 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2546,7 +2527,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 			destination_asset,
-			destination_address: destination_address_internal,
+			destination_address: match T::AddressConverter::decode_and_validate_address_for_asset(
+				destination_address,
+				destination_asset,
+			) {
+				Ok(address) => address,
+				Err(_) => {
+					emit_deposit_failed_event(DepositFailedReason::InvalidDestinationAddress);
+					return;
```
msgmaxim, line 2537: I don't mind reordering code, but IMO inlining control flow doesn't really help readability.

msgmaxim (approved):

dandanlen (approved):
This will conflict with #5507 (FYI @Janislav )

PR #5703, opened 2025-03-05
fix: ensure boosted txs can be rejected

dandanlen:


MxmUrw (commented):
Ok, so it looks like the only behavioural change is that we now *always* check whether reports from *either* the `ScreeningBrokerId` or the channel owner exist, even if these are the same account ids.

In the previous logic it should have been probably
```
if ScreeningBrokerId::get() == owner { vec![owner] } else { vec![owner, ScreeningBrokerId::get()]}
```

dandanlen:
> Ok, so it looks like the only behavioural change is that we now _always_ check whether reports from _either_ the `ScreeningBrokerId` or the channel owner exist, even if these are the same account ids.
> 
> In the previous logic it should have been probably
> 
> ```rust
> if ScreeningBrokerId::get() == owner { vec![owner] } else { vec![owner, ScreeningBrokerId::get()]}
> ```

Yes exactly - however I figured we might as well always check both.

Janislav:
Somehow the pipeline seem to be stuck 🤔.

dandanlen:
Closing this, will be released with 1.8 (see #5704)

PR #5630, opened 2025-02-12
refactor: make refund_params mandatory

Janislav:
# Pull Request

Closes: PRO-1702

## Checklist

Please conduct a thorough self-review before opening the PR.

- [x] I am confident that the code works.
- [x] I have written sufficient tests.
- [x] I have written and tested required migrations.
- [ ] I have updated documentation where appropriate.

## Summary

*Please include a succinct description of the purpose and content of the PR. What problem does it solve, and how? Link issues, discussions, other PRs, and anything else that will help the reviewer.*

## Summary by CodeRabbit

- **Refactor**
  - Refund parameters for swap deposit and vault swap operations are now required, ensuring consistent and robust handling of refund data.
  - Parameter usage has been simplified by removing optional wrappers, reducing complexity.

- **Tests**
  - Test cases have been updated to align with the new required refund parameters.
  - Legacy test cases related to deprecated vault swap functionality have been removed.

- **New Features**
  - Default fallback mechanisms for refund and fill-or-kill parameters have been introduced for improved operational robustness.

PR #5704, opened 2025-03-05
fix: ensure boosted txs can be rejected

dandanlen:
Another fix for BLS for boosted txs.

Same as #5703 but some differences due to structural differences between release branches.

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -1585,19 +1585,17 @@ impl<T: Config<I>, I: 'static> IngressSink for Pallet<T, I> {
 }
 
 impl<T: Config<I>, I: 'static> Pallet<T, I> {
-	fn is_whitelisted_broker_or(broker_id: T::AccountId) -> T::AccountId {
-		if WhitelistedBrokers::<T, I>::contains_key(&broker_id) {
-			T::ScreeningBrokerId::get()
-		} else {
-			broker_id
-		}
-	}
-
 	fn mark_transaction_for_rejection_inner(
 		account_id: T::AccountId,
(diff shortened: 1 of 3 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2414,10 +2412,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				return Err(DepositFailedReason::BelowMinimumDeposit);
 			}
 			if T::AllowTransactionReports::get() {
-				if let (Some(tx_ids), Some(broker_id)) = (
-					deposit_details.deposit_ids(),
-					origin.broker_id().map(|id| Self::is_whitelisted_broker_or(id.clone())),
-				) {
+				if let (Some(tx_ids), Some(broker_id)) =
+					(deposit_details.deposit_ids(), origin.broker_id())
+				{
```
MxmUrw, line 2417: The only thing that stands out to me is that we don't run this code if `origin.broker_id() == None`, even though the ScreeningBrokerId might have requested rejection.

Though the behaviour before this commit was the same, so I assume that's correct?

MxmUrw (commented):
Apart from that, it looks correct to me.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Yeah - now that we enforce a broker, this will always return Some. It's not yet explicit in the storage types, however. This will follow in the 1.9 release.

dandanlen: (See [here](https://github.com/chainflip-io/chainflip-backend/pull/5630#issuecomment-2696950439))

PR #5708, opened 2025-03-07
feat: dynamically scale transaction fees for LP calls

syan095:
# Pull Request

Closes: PRO-2096

## Checklist

## Summary

Adds fee scaling for LPs per pool. There is a governance configurable scaling configuration:
```rust
pub struct ExponentBufferFeeConfig {
	// Number of calls before fee scaling is applied.
	pub buffer: u32,
	// Base for the fee scaling exponent.
	pub exp_base: FixedU64,
}
```

This is then applied as effectively:
`exp_base^(number_of_calls_counted_in_block.saturating_sub(buffer)`

## Summary by CodeRabbit

- **New Features**
  - Enhanced transaction fee calculations with dynamic fee scaling that adjusts based on liquidity pool activity and transaction context, leading to more predictable and fair fee assessments.
  - Introduced dynamic configuration updates for fee and slashing rate parameters, improving overall transaction reliability and performance.
  
- **Tests**
  - Expanded integration tests to validate consistent fee behavior and ensure robust network transaction processing under varied scenarios.


state-chain/primitives/src/lib.rs
```diff
@@ -457,3 +458,20 @@ pub struct DcaParameters {
 }
 
 pub type ShortId = u8;
+pub type OrderId = u64;
+pub type Multiplier = FixedU64;
+
+#[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, Default, RuntimeDebug)]
+pub enum ScalableFeeCallInfo<AccountId> {
```
dandanlen, line 465: nit: trait-specific types can just be defined in the same place as the trait.

state-chain/primitives/src/lib.rs
```diff
@@ -457,3 +458,20 @@ pub struct DcaParameters {
...
+	// The call is normal, its transaction fee is not scalable.
+	#[default]
+	NoScaling,
```
dandanlen, line 468: nit: we can express this using `Option`.

state-chain/primitives/src/lib.rs
```diff
@@ -457,3 +458,20 @@ pub struct DcaParameters {
...
+	// Transactions about updating Liquidity Pool positions.
+	PoolPositionUpdates(AccountId, OrderId),
```
dandanlen, line 470: I think we want to use Asset instead of OrderId.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -340,10 +337,18 @@ pub mod pallet {
 	pub type HistoricalEarnedFees<T: Config> =
 		StorageDoubleMap<_, Identity, T::AccountId, Twox64Concat, Asset, AssetAmount, ValueQuery>;
 
+	#[pallet::storage]
+	pub type OrderModified<T: Config> =
+		StorageDoubleMap<_, Identity, T::AccountId, Twox64Concat, OrderId, u32, ValueQuery>;
```
dandanlen, line 342: I had an idea for a better approach that storing and updating this in the pools pallet: We can store the counts in the Flip pallet instead. 

We need an enum that 'indexes' the call info, to be used as a key:

```rust
pub enum CallInfo {
	PoolOrderUpdate(AccountId, Asset),
	// Maybe other things that we want to throttle
}
```

Then a storage item in the Flip pallet:

```rust
pub type CallCounter<T: Config> =
	StorageMap<_, TwoX64Concat, CallInfo, u32, ValueQuery>
```

And we can update this storage in `correct_and_deposit_fee`.

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
```diff
@@ -67,6 +88,10 @@ impl<T: TxConfig + FlipConfig + Config> OnChargeTransaction<T> for FlipTransacti
 				surplus.peek()
 			};
 
+			let to_burn =
+				T::TransactionFeeScaler::get_fee_multiplier(escrow.scalable_fee_call_info)
+					.saturating_mul_int(to_burn);
```
dandanlen, line 93: Better than just returning a multiple would be to have the scaler implement some scaling function and separate the state access from the scaling implementation:

```suggestion
			// (See my other comment re. CallCounter<T>)
			let scale_factor = CallCounter<T: Config>::mutate(|c| { c += 1; c });
			let to_burn =
				T::TransactionFeeScaler::scale_fee(to_burn, scale_factor);
```

dandanlen (commented):
General approach looks promising! I have some suggestions that I think can simplify things.

state-chain/runtime/src/chainflip.rs
```diff
@@ -1004,3 +1004,32 @@ impl cf_traits::MinimumDeposit for MinimumDepositProvider {
...
+			ScalableFeeCallInfo::PoolPositionUpdates(caller, order_id) =>
+				Multiplier::from_rational(
+					sp_std::cmp::max(
+						pallet_cf_pools::OrderModified::<Runtime>::get(caller, order_id),
+						1u32,
+					) as u128,
+					1u128,
+				),
+		}
+	}
```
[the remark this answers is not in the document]
kylezs: Some quick investigation: 

set_limit_order is: 0.000003884142 FLIP... multiply this by 10 and it is still not very large, even by 100 it's still small. ... multiply by 2^10, and we're getting somewhere 2^20 and we've maxed out to the ceil of 1 FLIP that we deducted in pre_dispatch.

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
```diff
@@ -67,6 +88,10 @@ impl<T: TxConfig + FlipConfig + Config> OnChargeTransaction<T> for FlipTransacti
 				surplus.peek()
 			};
 
+			let to_burn =
```
kylezs, line 91: This seems very dangerous now. If we have an exponential fee multiplier, then the to_burn can get *very* large very quickly... and this to_burn is passed directly into Flip::burn... meaning we could burn basically everything, just from one person spamming.
1. We should explicitly ceil the to_burn to the pre_dispatch max.
2. We should have a test for this.

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: > How do the calls get pushed into the FLIP pallet?

The method that takes the fees into escrow has access to the call. We can map this to an internal `CallInfo` type, which contains just enough information to index the calls by 'category' that we want to restrict (for example `CallInfo::LimitOrder(AccountId, Asset)`) and then store that in the pallet.

It would require a trait similar to `WaiveFees::should_waive_fees` that is implemented at the runtime level and does the mapping.

dandanlen: > We definitely want to include set_limit_order and set_range_order here too - not just update, I can only see set_limit_orders on mainnet, checked about 20 blocks for an update extrinsic, and couldn't find one.

Sure. The naming wasn't meant to imply that it would only include update_limit_order calls. The LimitOrderUpdate (or CallInfo::LimitOrder) could be used to index both set_limit_order and update_limit_order under the same 'bucket' of things.

state-chain/runtime/src/lib.rs
```diff
@@ -682,6 +682,8 @@ impl pallet_cf_flip::Config for Runtime {
 	type OnAccountFunded = pallet_cf_validator::UpdateBackupMapping<Self>;
 	type WeightInfo = pallet_cf_flip::weights::PalletWeight<Runtime>;
 	type WaivedFees = chainflip::WaivedFees;
+	type TransactionFeeScaler = chainflip::CfTransactionFeeScaler;
+	type SpamPreventionUpfrontFee = sp_core::ConstU128<FLIPPERINOS_PER_FLIP>; // 1 Flip. TODO: Adjust to a reasonable number
```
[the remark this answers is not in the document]
dandanlen: > If someone has 10 FLIP then they can submit a maximum of 10 order sets/updates across all pools, which at first thought doesn't sound too good.

I'm not sure this is true. The fees aren't all deducted at the start: the extrinsics are processed one by one, and after each one the fee is reduced.

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
dandanlen: You can't burn more than is in the account, the Surplus/Deficit types make sure of this. (see imbalances.rs)

state-chain/runtime/src/lib.rs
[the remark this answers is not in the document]
kylezs: Isn't withdraw_fee executed at tx validation, not execution? So we would have to validate this? Unless it's just checking them statelessly?
e.g.
User has 1FLIP
1. Submit tx cost = 1 FLIP
2. withdraw_fee ✅ , we have 1 FLIP
3. Submit another tx, cost = 1 FLIP
4. User has 1 FLIP ✅  

So basically we're not checking any state... however it looks like we're calling try_debit, which would change the state, meaning we would actually need all the FLIP up front... worth making sure on this one with a test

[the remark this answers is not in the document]
dandanlen: Makes sense, I see your point. I'm *fairly* sure that the resultant Surplus would be dropped in the case where 'validate' is called in the tx pool and therefore would have no persistent effect on storage (ie. validating one tx shouldn't change the storage for another tx). From the tx-pool README: 

```
1. We periodically validate all transactions in the pool in batches.

1. To minimize runtime calls, we introduce the batch-verify call. Note it should
   reset the state (overlay) after every verification.
```

I did a little digging and found [this recent PR](https://github.com/paritytech/polkadot-sdk//issues/3685) - it introduces a new trait method `can_withdraw_fee`, it seems this would allow us to separate the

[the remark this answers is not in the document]
kylezs: Based on that PR it seems:
pre-dispatch -> prepare, this is executed at block creation, and is where the fee would be taken, so this isn't run in the validate function, but when the block is being run - so this would be fine. It would work like:
1. Extrinsic1 being prepared
2. Predispatch fee taken
3. Extrinsic logic/state change executed
4. post_dispatch fee refund granted
5. Extrinsic2 being prepared... and so on.

Meaning at step 5 the refund has from Extrinsic1 is already back in the balance.

tl;dr the current method should work.

kylezs:
One idea I had too, is we may want to have some "cliff" before scaling the fee. This would allow LPs to set multiple orders in a block at different spreads (as is quite common) without worrying about the fee increase. The cliff could still be quite tight, to 2-3 txs per pool for example

state-chain/runtime/src/chainflip.rs
```diff
@@ -1021,3 +1022,42 @@ impl cf_traits::MinimumDeposit for MinimumDepositProvider {
...
+				..
+			}) |
+			RuntimeCall::LiquidityPools(pallet_cf_pools::Call::set_range_order {
+				base_asset,
+				..
+			}) |
+			RuntimeCall::LiquidityPools(pallet_cf_pools::Call::update_range_order {
+				base_asset,
+				..
+			}) => Some(CallInfoId::Pool(PoolTouched {
```
dandanlen, line 1047: Why not differentiate between limit orders and range orders?

state-chain/runtime/src/chainflip.rs
```diff
@@ -1021,3 +1022,42 @@ impl cf_traits::MinimumDeposit for MinimumDepositProvider {
...
+		}
+	}
+
+	fn scale_fee(
+		pre_scaled_fee: FlipBalance,
+		scale_factor: u16,
+		exp_base: FixedU64,
+	) -> FlipBalance {
+		let multiplier = exp_base.saturating_pow(scale_factor.into());
+		multiplier.saturating_mul_int(pre_scaled_fee)
```
dandanlen, line 1061: I'm not sure this works as intended - if the scale factor is zero, the multiplier will be one. But we want to scale it down, right? Or am I misreading something?

dandanlen: I think it's just the fact that this is implemented in terms of a `scale_factor` rather than, say, `call_count`?

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -132,6 +177,14 @@ pub mod pallet {
 	#[pallet::getter(fn offchain_funds)]
 	pub type OffchainFunds<T: Config> = StorageValue<_, T::Balance, ValueQuery>;
 
+	// Counts the number of calls within a block against a particular call info id.
+	#[pallet::storage]
+	pub type CallCounter<T: Config> =
+		StorageMap<_, Identity, CallInfoId<T::AccountId>, u16, ValueQuery>;
+
+	#[pallet::storage]
+	pub type FeeScalingRateConfig<T: Config> = StorageValue<_, ExponentBufferFeeConfig, ValueQuery>;
```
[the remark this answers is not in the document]
kylezs: Yeah that would be the next way to extend it, left it out for now, though I could add it

state-chain/runtime/src/chainflip.rs
kylezs: This was primarily to avoid needing to have an extra parameter on the trait, only to do the substraction inside. I had it inside before, and then took it out.

state-chain/runtime/src/chainflip.rs
kylezs: Why would we?

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
```diff
@@ -74,15 +91,36 @@ impl<T: TxConfig + FlipConfig + Config> OnChargeTransaction<T> for FlipTransacti
...
 			// case, we shouldn't refund anything, we can just burn all fees in escrow.
-			let to_burn = if frame_system::Pallet::<T>::account_exists(who) {
+			let pre_scaled_fee_to_burn = if frame_system::Pallet::<T>::account_exists(who) {
 				corrected_fee
 			} else {
 				surplus.peek()
 			};
 
+			let to_burn = if let Some(call_info_id) = escrow.call_info_id {
+				let before_count = crate::CallCounter::<T>::mutate(&call_info_id, |count| {
```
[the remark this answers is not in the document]
kylezs: Call counter is 0 to start with. So, the first time this is called, it is the 0th call. Therefore we should use the before_count. If we used the after count. And had a buffer of 0, then we'd immediately scale 2^1 which would be 2. And so the very first call would have a fee 2x than expected.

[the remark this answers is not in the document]
dandanlen: You're arguing based on a specific implementation here. I'm just suggesting it might easier in general to think in terms of "this is the first call" rather than "this is the zeroth" call.

state-chain/runtime/src/chainflip.rs
dandanlen: Because it's valid to update a range order and limit order in the same block (they are different calls) whereas with multiple limit order updates, all but the last are effectively no-ops and just consume block space for no reason.

state-chain/runtime/src/chainflip.rs
```diff
@@ -1021,3 +1021,49 @@ impl cf_traits::MinimumDeposit for MinimumDepositProvider {
...
+		pre_scaled_fee: FlipBalance,
+		call_count: u16,
+	) -> FlipBalance {
+		match fee_scaling_config {
+			FeeScalingRateConfig::ExponentBuffer { buffer, exp_base } => {
+				let multiplier = exp_base.saturating_pow(call_count.saturating_sub(buffer).into());
+				multiplier.saturating_mul_int(pre_scaled_fee)
+			},
+			FeeScalingRateConfig::NoScaling => pre_scaled_fee,
+		}
```
dandanlen, line 1067: I think this is wrong. We want to scale the fee *down*.

On the first (or zeroth) call, the exponent is always zero, so the multiplier will always be one. 

Also, it feels a bit weird having the fee scaling implemented in this trait, IMO would be more natural to have the impl co-located with the enum definition and then just call a method here: 

```rust
fn scale_fee(
		fee_scaling_config: FeeScalingRateConfig,
		pre_scaled_fee: FlipBalance,
		call_count: u16,
	) -> FlipBalance {
		fee_scaling_config.scale(pre_scaled_fee, call_count)
	}
```

state-chain/runtime/src/chainflip.rs
kylezs: There could be a strategy that uses both range order and limit orders, updating ranges between their limit order ticks to catch volume between ticks. This would be a rational strategy. 
Anyone using range orders as is typically used isn't going to hit rate limiting anyway and so this wouldn't apply.

state-chain/runtime/src/chainflip.rs
kylezs: the tests seem to work as expected, not sure what you mean

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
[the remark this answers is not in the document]
dandanlen: IMO that's fine, all the ugliness should be in the implementation, where it can be tested in context, rather than in the surrounding structure, where it will tested against a mock.

state-chain/runtime/src/chainflip.rs
dandanlen: Not really sure what the argument is here. Why would we rate-limit both together? ie. Why should an LP pay more for a limit order if they already placed a range order?

state-chain/runtime/src/chainflip.rs
dandanlen: I mean that we deduct 1 FLIP before execution, we can't burn more than that, we can only *reduce* the fee, we can't increase it. This function takes the already deducted fee and then scales it up - or am I misreading it? 

But we can't burn more than we have already deducted. For one, we can't guarantee that the account has sufficient balance.

state-chain/traits/src/lib.rs
```diff
@@ -1223,6 +1223,53 @@ pub trait MinimumDeposit {
...
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Copy, PartialEq, Eq, RuntimeDebug)]
+pub struct PoolTouched<AccountId> {
+	pub account: AccountId,
+	pub base_asset: Asset,
+}
+
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Copy, PartialEq, Eq, RuntimeDebug)]
+pub enum CallInfoId<AccountId> {
+	Pool(PoolTouched<AccountId>),
+}
```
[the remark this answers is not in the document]
kylezs: ah yes, forgot to remove that one

state-chain/cf-integration-tests/src/fee_scaling.rs
```diff
@@ -0,0 +1,143 @@
...
+			FeeScalingRate::<Runtime>::set(FeeScalingRateConfig::ExponentBuffer {
+				buffer: 0,
+				exp_base: FixedU64::from_rational(2, 1u128),
+			});
+
+			let mut can_pay = true;
+			(1..50).for_each(|_| {
+				match apply_extrinsic_and_calculate_gas_fee(lp, call.clone()) {
+					Ok((gas, remaining_balance)) => {
+						assert!(gas >= last_gas);
```
dandanlen, line 73: If the fee is always the MAX (because it doesn't scale) then this test will always pass.

dandanlen: (changing the check to `gas > last_gas` causes it to fail on the first iteration)

kylezs:
@coderabbitai full review

state-chain/cf-integration-tests/src/fee_scaling.rs
```diff
@@ -0,0 +1,152 @@
...
+							TransactionValidityError::Invalid(InvalidTransaction::Payment)
+						);
+						// No balance change if transaction fails
+						assert_eq!(Flip::total_balance_of(&lp_account_id), last_remaining_balance);
+					},
+				}
+			});
+
+			// We should have run out of balance by the end.
+			assert!(!can_pay);
```
dandanlen, line 104: I don't think this is a necessary test condition. It also means that the test could succeed eg. because fees are applied incorrectly and immediately deduct the full balance.

state-chain/cf-integration-tests/src/network.rs
```diff
@@ -883,3 +888,38 @@ pub fn register_refund_addresses(account_id: &AccountId) {
 		));
 	}
 }
+
+pub fn apply_extrinsic_and_calculate_gas_fee(
```
dandanlen, line 892: Not sure why this function is here - it doesn't have much to do with `network` and is only used by the fee scaling test.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -32,29 +32,74 @@ use scale_info::TypeInfo;
...
+		&self,
+		pre_scaled_fee: Balance,
+		call_count: u16,
+	) -> Balance {
+		match self {
+			FeeScalingRateConfig::ExponentBuffer { buffer, exp_base } => {
+				let multiplier = (*exp_base)
+					// we subtract an extra 1 to account for the fact that the first call passes
+					// through a count of 1, but we don't want to scale it.
+					.saturating_pow(call_count.saturating_sub(*buffer).saturating_sub(1).into());
```
dandanlen, line 95: I dislike the `.saturating_sub(1)`. It means that if the buffer is 1, 2 calls are scaling-free. 

IMO a buffer of zero means no buffer: you scale from the first call. 
A buffer of one means: 1 call is free (not scaled), scaled after that. 
Etc. 

Maybe naming is something other than `buffer` would make it clearer.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -151,6 +210,9 @@ pub mod pallet {
 		SlashingRateUpdated {
 			slashing_rate: Permill,
 		},
+		FeeScalingRateUpdated {
+			fee_scaling_rate: FeeScalingRateConfig,
+		},
```
dandanlen, line 215: Similar to the extrinsic, we could merge these into a single event `PalletConfigUpdated { update: PalletConfigUpdate }`

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
```diff
@@ -74,11 +83,22 @@ impl<T: TxConfig + FlipConfig + Config> OnChargeTransaction<T> for FlipTransacti
...
-				corrected_fee
+				if let Some(call_info_id) = call_info_id {
+					let call_count = crate::CallCounter::<T>::mutate(&call_info_id, |count| {
+						*count += 1;
+						*count
+					});
+					match call_info_id {
+						CallInfoId::Pool { .. } =>
+							crate::FeeScalingRate::<T>::get().scale_fee(corrected_fee, call_count),
+					}
```
dandanlen, line 98: The match statement is unnecessary? If there is *any* call info, we want to scale the fee according to the call count.

state-chain/pallets/cf-flip/src/tests.rs
```diff
@@ -649,11 +649,81 @@ mod test_tx_payments {
...
+
+		let large_buffer_config = FeeScalingRateConfig::ExponentBuffer {
+			buffer: BUFFER,
+			exp_base: FixedU64::from_rational(2, 1),
+		};
+
+		const BUFFER: u16 = 10;
+		// In reality the call count will be 1 (hence the +1), but we test 0 behaves reasonably in
+		// case.
+		for call_count in 0..=BUFFER + 1 {
```
dandanlen, line 715: We only need the +1 because we subtract 1 in the implementation...

state-chain/traits/src/lib.rs
```diff
@@ -1223,6 +1223,28 @@ pub trait MinimumDeposit {
...
+// Creates an identifier that is used to group calls that should be scaled together.
+// Returns None if the call should not be scaled.
+pub trait FeeScalingCallInfoIdentifier<Call, AccountId, Balance> {
+	fn call_info_and_spam_prevention_upfront_fee(
+		_call: &Call,
+		_caller: &AccountId,
+	) -> Option<(CallInfoId<AccountId>, Balance)> {
+		None
+	}
+}
```
dandanlen, line 1240: I'm actually not sure this needs to care about the balance and account id at all. The fee is constant, it doesn't depend on the call. The account is simply passed through. 

All we need is function that decides which calls need to be scaled. Even the identifier can be totally opaque. 

```suggestion
pub trait CallIndexer<Call> {
	type CallIndex: Encode;
	fn call_index(
		_call: &Call,
	) -> Option<CallIndex> {
		None
	}
	
	fn opaque_call_index(
		call: &Call,
	) -> {
		self.call_index(call).map(Encode::encode)
	}
}
```

dandanlen (commented):
Still more comments 🙈 

I will address these myself if i have time.

state-chain/cf-integration-tests/src/fee_scaling.rs
```diff
@@ -0,0 +1,152 @@
...
+			assert_eq!(FeeScalingRate::<Runtime>::get(), FeeScalingRateConfig::NoScaling);
+
+			let (mut last_gas, mut last_remaining_balance) =
+				apply_extrinsic_and_calculate_gas_fee(lp, call.clone()).unwrap();
+			(0..50).for_each(|_| {
+				let (gas, remaining_balance) =
+					apply_extrinsic_and_calculate_gas_fee(lp, call.clone()).unwrap();
+				assert_ne!(gas, 0);
+				assert_eq!(gas, last_gas);
+				last_gas = gas;
```
syan095, line 60: small nitpick - if we expect no scaling and the gas fee to stay constant, then you won't need this assignment here

state-chain/cf-integration-tests/src/fee_scaling.rs
```diff
@@ -0,0 +1,152 @@
...
+			// Again, once each on each pool.
+			let (gas_pool1_again, remaining_balance_pool1_again) =
+				apply_extrinsic_and_calculate_gas_fee(lp, call.clone()).unwrap();
+			assert!(remaining_balance_pool1_again < remaining_balance_pool2);
+			let (gas_pool2_again, remaining_balance_pool2_again) =
+				apply_extrinsic_and_calculate_gas_fee(lp, call2.clone()).unwrap();
+			assert!(gas_pool1_again > gas_pool1);
+			assert!(gas_pool2_again > gas_pool2);
+			assert_eq!(gas_pool1_again, gas_pool2_again);
+			assert!(remaining_balance_pool2_again < remaining_balance_pool2);
```
syan095, line 150: maybe we should call once more and observe the 2 pools' gas fee start to scale independently after 2 calls

state-chain/pallets/cf-flip/src/benchmarking.rs
```diff
@@ -28,9 +30,36 @@ mod benchmarks {
 	use super::*;
 
 	#[benchmark]
-	fn set_slashing_rate() {
+	fn on_initialize() {
+		let whitelisted_callers = T::AccountRoleRegistry::generate_whitelisted_callers_with_role(
```
syan095, line 34: I think it would be better to benchmark the weight it takes to remove 1 storage, then calculate the weight using the actual number of accounts to remove, than to use an arbitrary number of 50

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -32,29 +32,74 @@ use scale_info::TypeInfo;
...
+
+impl FeeScalingRateConfig {
+	pub fn scale_fee<Balance: AtLeast32BitUnsigned + Copy>(
+		&self,
+		pre_scaled_fee: Balance,
+		call_count: u16,
+	) -> Balance {
+		match self {
+			FeeScalingRateConfig::ExponentBuffer { buffer, exp_base } => {
+				let multiplier = (*exp_base)
```
syan095, line 92: Just to note here - if you use y = k^x, when k = 1, the result will always be 1
otherwise y=k^x and y = x^k produces similar exponential curves.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -165,31 +227,46 @@ pub mod pallet {
...
-		///
-		/// ## Dependencies
-		///
-		/// - [EnsureGovernance]
+		/// Requires Governance.
 		#[pallet::call_index(0)]
-		#[pallet::weight(T::WeightInfo::set_slashing_rate())]
-		pub fn set_slashing_rate(origin: OriginFor<T>, slashing_rate: Permill) -> DispatchResult {
+		#[pallet::weight(T::WeightInfo::update_pallet_config())]
+		pub fn update_pallet_config(
```
syan095, line 247: Do we still prefer to deprecate and create new extrinsic calls over changing existing extrinsics? (but this is used only for Governance, so I suppose it's fine to just change it here)

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -165,31 +227,46 @@ pub mod pallet {
...
+				match update {
+					PalletConfigUpdate::SetSlashingRate(slashing_rate) => {
+						SlashingRate::<T>::set(slashing_rate);
+						Self::deposit_event(Event::SlashingRateUpdated { slashing_rate });
+					},
+					// You can used FixedU64::from_rational or from_float to convert the input
+					// number to FixedU64.
+					// The range is: [0.000000000, 18446744073.709551615]
+					// i.e. there are 9 decimal places.
+					PalletConfigUpdate::SetFeeScalingRate(fee_scaling_rate) => {
```
syan095, line 264: do we want to set like a hardcoded "safety margin" for this scaling rate? e.g. 1 < scaling_rate < 100 or something?

state-chain/traits/src/lib.rs
syan095: The caller ID is provided in both the "pre-" and "post-" dispatch calls, so you won't need to pass it through.
But the "Balance" here is supposed to be the "upfront" fee we charge.

I would suggest that the upfront fee be made configurable, together with the scaling rate - since this acts also as a "ceiling" - the maximum amount fo gas fee can be changed to the user (due to how the `Imbalance` mechanism works) so this is an important parameter.

dandanlen: I agree, but unfortunately the 'up-front' fee can't be made configurable if we want to avoid storage access in the pre-dispatch part...

state-chain/pallets/cf-flip/src/lib.rs
dandanlen: Yeah I think it makes sense for the exponent to be constant (ie. 1: linear scaling, 2: quadratic scaling, etc), instead of the reverse.

dandanlen: k^x means we don't need to use fixed point types at all.

The multiplier is just `call_count ^ x` where is x is: `0` for constant, `1` for linear, `2` for quadratic, etc. I think this is easier to visualize/reason about.

state-chain/pallets/cf-flip/src/lib.rs
kylezs: yeah exactly, will be in a minor version release, and is just for gov so it's fine

state-chain/pallets/cf-flip/src/lib.rs
kylezs: It's governance applied, so don't think it's necessary

state-chain/pallets/cf-flip/src/lib.rs
kylezs: this is consistent with the other pallets, though I agree, that makes sense

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
```diff
@@ -89,3 +117,71 @@ impl<T: TxConfig + FlipConfig + Config> OnChargeTransaction<T> for FlipTransacti
...
+	pub fn scale_fee<Balance: AtLeast32BitUnsigned + Copy>(
+		&self,
+		pre_scaled_fee: Balance,
+		call_count: u16,
+	) -> Balance {
+		match self {
+			FeeScalingRateConfig::DelayedExponential { threshold, exponent } => {
+				let multiplier =
+					1 + call_count.saturating_sub(*threshold).saturating_pow(*exponent as u32);
+				pre_scaled_fee.saturating_mul(multiplier.into())
```
kylezs, line 156: In the base case, exponent is 1. It's the first call, so we pass through call_count = 1. Threshold is 0. So now on the first call we do 1 + 1^1 so the multiplier is 2 not one. The first call should have a multiplier of 1.

state-chain/runtime/src/chainflip.rs
```diff
@@ -1024,3 +1024,31 @@ impl cf_traits::MinimumDeposit for MinimumDepositProvider {
 		}
 	}
 }
+
+pub struct LpOrderCallIndexer;
+impl CallIndexer<RuntimeCall> for LpOrderCallIndexer {
```
kylezs, line 1029: I'm not sure this way of abstracting it improves things. We always have to pass in the indexer to the flip pallet. If we were to index other things for fee scaling, would we put in a separate CallIndex instance, with a separate associated type in on the pallet?

I think it makes more sense to have the more globally named and instantiated call indexer, that's for all calls, rather than an instance for each.

state-chain/pallets/cf-flip/src/tests.rs
```diff
@@ -649,11 +649,73 @@ mod test_tx_payments {
...
+
+		let large_buffer_config =
+			FeeScalingRateConfig::DelayedExponential { threshold: BUFFER, exponent: 2 };
+
+		// In reality the call count will be 1 (hence the +1), but we test 0 behaves reasonably in
+		// case.
+		for call_count in 0..=BUFFER {
+			assert_eq!(large_buffer_config.scale_fee(pre_scaled_fee, call_count), pre_scaled_fee);
+		}
+		for i in 0..BUFFER {
```
kylezs, line 714: kinda confusing having the buffer here, it's meaningless that we go to the buffer in this case, better to separate them - also as above, this test should catch the issue I mentioned

kylezs: Not sure if we need this test as well as the tests in `on_charge_transaction`

state-chain/pallets/cf-flip/src/on_charge_transaction.rs
dandanlen: Yeah I think I confused myself there... will fix.

state-chain/runtime/src/chainflip.rs
dandanlen: For now it only applies to LP orders, and I wanted to avoid naming everything `CallIndexer`. The intention wasn't necessarily that we would have multiple of these.

kylezs: Fair

state-chain/pallets/cf-flip/src/tests.rs
dandanlen: Yeah, forgot to remove these.

kylezs (approved):
👌

syan095:
Something to note: Since we removed the "off by one" adjustment Kyle initially had, the fee scaling is delayed by 1
i.e. at the threshold, scaling factor is 0^exp, which is 1 and therefore no scalling happens.

syan095:
LGTM! 👍

PR #5711, opened 2025-03-07
feat: add RPC to return all current vault addresses

Janislav:
# Pull Request

Ticket: PRO-2109

## Checklist

## Summary

Adds a new endpoint to return:
- The Ethereum vault address
- The Arbitrum vault address
- All (pervious current addresses) for private broker channels in a vector (AccountId, EncodedAddress)


Janislav:
@coderabbitai review

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -76,6 +76,13 @@ pub enum GetOpenDepositChannelsQuery {
 	Mine,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct VaultAddresses {
+	ethereum: String,
+	arbitrum: String,
+	bitcoin: String,
```
dandanlen, line 83: For bitcoin there should be multiple addresses: each broker can have one private channel and each private channel needs two addresses: one the current and one for the previous epoch's key.

state-chain/runtime/src/chainflip/address_derivation/btc.rs
```diff
@@ -49,7 +49,7 @@ impl AddressDerivationApi<Bitcoin> for AddressDerivation {
 /// Derives the BTC vault deposit address from the private channel id.
 /// Note: This function will **panic** if the private channel id is out of bounds or if there is
 /// no active epoch key for Bitcoin.
-pub(crate) fn derive_btc_vault_deposit_address(private_channel_id: u64) -> String {
+pub fn derive_btc_vault_deposit_address(private_channel_id: u64) -> String {
```
dandanlen, line 52: We might have to alter this method to allow us to generate the previous key too. The deposit monitor will need to track this.

state-chain/runtime/src/lib.rs
```diff
@@ -2529,6 +2529,13 @@ impl_runtime_apis! {
 				pallet_cf_swapping::AffiliateAccountDetails::<Runtime>::iter_prefix(&broker).collect()
 			}
 		}
+
+		fn cf_vault_addresses() -> (EncodedAddress, EncodedAddress, EncodedAddress) {
+			let eth_vault_address = EncodedAddress::Eth(Environment::eth_vault_address().into());
+			let arb_vault_address = EncodedAddress::Arb(Environment::arb_vault_address().into());
+			let btc_vault_address = EncodedAddress::Btc(derive_btc_vault_deposit_address(0).into());
```
dandanlen, line 2536: Here: we can't just use 0, we need to use the private channels of each broker.

state-chain/runtime/src/chainflip/vault_swaps.rs
```diff
@@ -95,7 +95,7 @@ pub fn bitcoin_vault_swap(
 
 	Ok(VaultSwapDetails::Bitcoin {
 		nulldata_payload: encode_swap_params_in_nulldata_payload(params),
-		deposit_address: derive_btc_vault_deposit_address(private_channel_id),
+		deposit_address: derive_btc_vault_deposit_addresses(private_channel_id).1,
```
dandanlen, line 98: Maybe there's a better way to ensure that this is correct? Right now I have to read the implementation of `derive_btc_vault_deposit_addresses` to know what the `.1` represents. 

*edit: see my other comment re. returning a struct instead of a tuple*

state-chain/runtime/src/lib.rs
```diff
@@ -2529,6 +2529,23 @@ impl_runtime_apis! {
 				pallet_cf_swapping::AffiliateAccountDetails::<Runtime>::iter_prefix(&broker).collect()
 			}
 		}
+
+		fn cf_vault_addresses() -> VaultAddresses<EncodedAddress> {
+			// By passing 0 we get the acutal vault address.
+			let (previous, current) = derive_btc_vault_deposit_addresses(0);
```
dandanlen, line 2535: I'm pretty sure the vault (index 0) isn't monitored by the engines.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -344,6 +344,47 @@ pub struct TransactionScreeningEvents {
...
+	pub previous: AddressType,
+	pub current: AddressType,
+}
+
+#[derive(Encode, Decode, TypeInfo, Serialize, Deserialize, Clone)]
+pub struct VaultAddresses<AddressType> {
+	pub ethereum_vault: AddressType,
+	pub arbitrum_vault: AddressType,
+	pub bitcoin_vault: BtcDepositAddresses<AddressType>,
+	pub bitcoin_deposit_addresses: BTreeMap<AccountId32, BtcDepositAddresses<AddressType>>,
```
dandanlen, line 359: As noted above, there is no `bitcoin_vault`, we only witness the broker private channel addresses. I don't think the deposit monitor cares about who owns them either, or which is previous/current (maybe @MxmUrw can confirm).

I think we can simplify to this:

```
struct VaultAddresses<AddressType> {
	pub ethereum: AddressType,
	pub arbitrum: AddressType,
	pub bitcoin: Vec<AddressType>,
}
```

(also, nit: we don't need the `Type` suffix. We already know it's a type.)

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -156,6 +157,9 @@ pub trait Rpc {
 		&self,
 		affiliate_account_id: AccountId32,
 	) -> RpcResult<WithdrawFeesDetail>;
+
+	#[method(name = "vault_addresses", aliases = ["broker_getVaultAddresses"])]
```
dandanlen, line 161: Would suggest either this:

```
#[method(name = "vault_addresses", aliases = ["broker_VaultAddresses"])]
```

Or this:

```
#[method(name = "get_vault_addresses", aliases = ["broker_getVaultAddresses"])]
```

dandanlen (commented):
One other thing: we return EncodedAddress from the custom rpc layer, but AddressString from the Broker API. It would be better to be consistent and use the same in both APIs. Since (afaik) the deposit monitor currently uses `EncodedAddress` so let's stick with that.

state-chain/runtime/src/runtime_apis.rs
MxmUrw: Yep, the deposit monitor needs to know the owner in order to be able to filter the addresses according to it's configuration.

Janislav:
@dandanlen I added all of your points.

dandanlen (approved):
LGTM, thanks. I'm going to make some small refactors before merging.

dandanlen:
@MxmUrw does the interface look ok to you?

MxmUrw:
> @MxmUrw does the interface look ok to you?

Yes, looks good!

PR #5712, opened 2025-03-07
chore(bouncer): Update sdk to official 1.8.0-beta.2 [WEB-1995]

niklasnatter:


bouncer/shared/evm_vault_swap.ts
```diff
@@ -67,11 +68,12 @@ export async function executeEvmVaultSwap(
 
   if (erc20Assets.includes(sourceAsset)) {
     // Doing effectively infinite approvals to make sure it doesn't fail.
-    // eslint-disable-next-line @typescript-eslint/no-use-before-define
-    await approveEvmTokenVault(
+
+    await approveErc20(
```
albert-llimos, line 72: I don't think this will work. The `approveErc20` will be done from the wallet address, instead of from the new `evmWallet`. We should keep it as it was.

bouncer/shared/send_evm.ts
```diff
@@ -42,10 +38,7 @@ export async function getNextEvmNonce(
       const txCount = await web3.eth.getTransactionCount(address);
       nextEvmNonce[chain] = txCount;
     }
-    // The SDK returns null if no transaction is sent
-    if (callback && (await callback(nextEvmNonce[chain]!)) === null) {
-      return nextEvmNonce[chain]!;
-    }
```
albert-llimos, line 48: Why did you remove this? Has the behaviour of the SDK change? Touching this nonce logic is very prone to error.

bouncer/shared/new_swap.ts
```diff
@@ -51,9 +51,9 @@ export async function newSwap(
           destAddress: destinationAddress,
           destChain: chainFromAsset(destAsset) as Chain,
           ccmParams: messageMetadata && {
-            message: messageMetadata.message as `0x${string}`,
+            message: messageMetadata.message,
             gasBudget: messageMetadata.gasBudget.toString(),
-            ccmAdditionalData: messageMetadata.ccmAdditionalData as `0x${string}`,
+            additionalData: messageMetadata.additionalData,
```
albert-llimos, line 56: Why changing the `ccmAdditionalData` to `additionalData`? I guess it's because it's within the `ccmParams` so it's a bit redundant to call it `ccmAdditionalData` but I'd rather keep it as is to match exactly the naming in the protocol.

bouncer/shared/evm_vault_swap.ts
niklasnatter: makes sense! changed it because the sdk does not have a `approveVault` method anymore and this looked like a replacement

bouncer/shared/send_evm.ts
niklasnatter: the `getNextEvmNonce` is not invoked with a `callback` anywhere, so the condition will never be true and this would never be executed. i guess this is a recent change then?

bouncer/shared/new_swap.ts
niklasnatter: will check!

bouncer/shared/send_evm.ts
albert-llimos: Oh true, we changed some of this logic and ended up using different wallets instead of relying on this. Makes sense to remove it then.

bouncer/shared/evm_vault_swap.ts
albert-llimos: Oh okay, you removed that from the SDK, makes sense.
Then we should probable do that approval differently with some of the existing approval code but it might need a small refactoring to not add too much spaghetti code. I can do it if you want, should be very quick @niklasnatter

bouncer/shared/new_dot_address.ts
```diff
@@ -2,7 +2,7 @@ import Keyring from '../polkadot/keyring';
 
 export async function newDotAddress(seed: string): Promise<string> {
   const keyring = new Keyring({ type: 'sr25519' });
-  const { address } = keyring.createFromUri('//' + seed);
+  const { addressRaw } = keyring.createFromUri('//' + seed);
 
-  return address;
+  return keyring.encodeAddress(addressRaw, 0);
```
niklasnatter, line 7: return dot address (1...) instead of generic substrate address (5...)

albert-llimos (approved):

PR #5713, opened 2025-03-07
chore: Add Apache 2.0 license to backend

martin-chainflip:
# Pull Request

Closes: PRO-1421

Here's a summary of possible licenses:
Using "GPL" would discourage integrators to use chainflip, since anything that uses our code could no longer be used commercially
Using "BSD" or "MIT" would basically mean "do whatever you want with the code, as long as you don't pretend it's yours". The two licenses have different styles, but are largely identical in their effect
"Apache 2.0" allows integrators to use our code in their commercial applications.  But it has some additional features:
- Anyone contributing code to our repo grants us a perpetual, free license to use any software patents that might cover their code or the code resulting from merging their changes. This is to prevent patent trolls from submitting some cool feature and then demanding royalties because our code touches their patent
- In turn, we "forward" these patents to anyone using our code. This again makes it easier for third parties to integrate chainflip code without worrying about patents and so on.
- We even grant third parties the right to patent their code if it uses ours or is based on ours. However, should they ever use these resulting patents in order to come after us, this grant is void, so that they don't try anything funny
- Any trademarks, logos etc. that are part of the project are NOT licensed. However, they are free to use them in a reasonable way in their documentation/marketing (like saying that their system is based on our code)

We're going for Apache 2.0, also because almost all of the Substrate and Parity code is already under Apache 2.0, which makes it easier for us to just "continue" this license as opposed to changing the license.

The only area of concern is where we took existing code and modified it. In that case we need to add attributions, for example as happened in the "chainflip-backend/foreign-chains/solana/sol-prim" folder


dandanlen (approved):
I can't see any attributions (ie. in sol-prim), do we have a follow-up issue for this?

martin-chainflip:
There are attributions in the comment of each source file of sol-prim

dandanlen:
> There are attributions in the comment of each source file of sol-prim

OK! I missed the `and Anza Maintainers <maintainers@anza.xyz>`, didn't realise it was as simple as this.

PR #5728, opened 2025-03-12
fix: bump stale egress

dandanlen:


.github/workflows/_21_build_m2.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       - name: Configure Git 🛠️
         run: git config --global --add safe.directory "${GITHUB_WORKSPACE}"
 
+      - name: Install Rust Toolchain 💿
+        run: rustup toolchain install
```
martin-chainflip, line 45: make it "rustup toolchain install nightly"

PR #5733, opened 2025-03-12
feat: Bitcoin ES: change ExecuteHook to return ()

marcellorigotti:
# Pull Request

Closes: PRO-2102

## Checklist

## Summary

Changed the return type of ExecuteHook to ()

### DIFF
--- a/state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
+++ b/state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
@@ -65,10 +65,7 @@ fn dedup_events<T: Ord + Clone>(
 impl Hook<HookTypeFor<TypesDepositChannelWitnessing, ExecuteHook>>
 	for TypesDepositChannelWitnessing
 {
-	fn run(
-		&mut self,
-		events: Vec<(BlockNumber, BtcEvent<DepositWitness<Bitcoin>>)>,
-	) -> Vec<(BlockNumber, BtcEvent<DepositWitness<Bitcoin>>)> {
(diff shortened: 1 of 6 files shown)


marcellorigotti:
TODO: Update tests to reflect this change

state-chain/runtime/src/chainflip/bitcoin_block_processor.rs
```diff
@@ -235,6 +229,23 @@ impl Hook<HookTypeFor<TypesEgressWitnessing, SafetyMarginHook>> for TypesEgressW
...
+			(8, BtcEvent::<u8>::PreWitness(9)),
+			(10, BtcEvent::<u8>::Witness(10)),
+			(10, BtcEvent::<u8>::Witness(11)),
+			(8, BtcEvent::<u8>::PreWitness(11)),
+			(10, BtcEvent::<u8>::PreWitness(12)),
+		];
+		let deduped_events = dedup_events(events);
+		assert_eq!(deduped_events.len(), 4);
+		assert!(!deduped_events.contains(&(8, BtcEvent::<u8>::PreWitness(9))));
+		assert!(!deduped_events.contains(&(8, BtcEvent::<u8>::PreWitness(11))));
```
MxmUrw, line 247: I think it would be useful to check against the full result vector here (that we expect), then the expected behavior is clearer on a quick glance & we're actually checking more.

MxmUrw (approved):
Thanks! Looks good! I would update the test case a bit, after that feel free to merge.

PR #5737, opened 2025-03-13
refactor: include expire block in liquidity deposit details

j4m1ef0rd:
# Pull Request

Closes: PRO-2039

## Checklist

## Summary

Added a new RPC `lp_request_liquidity_deposit_address`, It returns a struct with the address and expire block.
~~This is a breaking change for this RPC. @acdibble~~

### DIFF
--- a/api/bin/chainflip-cli/src/main.rs
+++ b/api/bin/chainflip-cli/src/main.rs
@@ -30,6 +30,7 @@ use cf_chains::eth::Address as EthereumAddress;
 use cf_utilities::{clean_hex_address, round_f64, task_scope::task_scope};
 use chainflip_api::{
 	self as api,
+	lp::LiquidityDepositDetails,
 	primitives::{state_chain_runtime, FLIPPERINOS_PER_FLIP},
 	rpc_types::RedemptionAmount,
 	BrokerApi,
@@ -83,7 +84,7 @@ async fn run_cli() -> Result<()> {
 							.broker_api()
 							.withdraw_fees(params.asset, params.destination_address)
 							.await?;
-						println!("Withdrawal request successfull submitted: {}", withdraw_details);
+						println!("Withdrawal request successful submitted: {}", withdraw_details);
 					},
 					BrokerSubcommands::RegisterAccount => {
 						api.broker_api().register_account().await?;
@@ -97,7 +98,7 @@ async fn run_cli() -> Result<()> {
 						asset,
 						boost_fee,
 					} => {
-						let address = api
+						let LiquidityDepositDetails { deposit_address, deposit_chain_expiry_block } = api
 							.lp_api()
 							.request_liquidity_deposit_address(
 								asset,
@@ -106,7 +107,7 @@ async fn run_cli() -> Result<()> {
 							)
 							.await?
 							.unwrap_details();
-						println!("Deposit Address: {address}");
+						println!("Deposit Address: {deposit_address}\nDeposit chain expire block: {deposit_chain_expiry_block}");
 					},
 
 					LiquidityProviderSubcommands::RegisterLiquidityRefundAddress {
--- a/api/bin/chainflip-cli/src/main.rs
+++ b/api/bin/chainflip-cli/src/main.rs
@@ -84,7 +84,7 @@ async fn run_cli() -> Result<()> {
 							.broker_api()
 							.withdraw_fees(params.asset, params.destination_address)
 							.await?;
-						println!("Withdrawal request successful submitted: {}", withdraw_details);
+						println!("Withdrawal request successfully submitted: {}", withdraw_details);
 					},
 					BrokerSubcommands::RegisterAccount => {
 						api.broker_api().register_account().await?;
--- a/api/bin/chainflip-cli/src/main.rs
+++ b/api/bin/chainflip-cli/src/main.rs
@@ -30,7 +30,7 @@ use cf_chains::eth::Address as EthereumAddress;
 use cf_utilities::{clean_hex_address, round_f64, task_scope::task_scope};
 use chainflip_api::{
 	self as api,
-	lp::LiquidityDepositDetails,
+	lp::LiquidityDepositChannelDetails,
 	primitives::{state_chain_runtime, FLIPPERINOS_PER_FLIP},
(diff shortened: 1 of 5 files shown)


dandanlen:
This change makes sense, but is a breaking change for LPs in a core rpc call... they will have to be notified of this. I'll create a companion issue to update the docs.

api/bin/chainflip-cli/src/main.rs
```diff
@@ -106,7 +107,7 @@ async fn run_cli() -> Result<()> {
 							)
 							.await?
 							.unwrap_details();
-						println!("Deposit Address: {address}");
+						println!("Deposit Address: {deposit_address}\nDeposit chain expire block: {deposit_chain_expiry_block}");
```
kylezs, line 110:
```suggestion
						println!("Deposit Address: {deposit_address}\nDeposit chain expiry block: {deposit_chain_expiry_block}");
```

api/cf-rpc-types/src/lp.rs
```diff
@@ -135,3 +135,9 @@ impl TryFrom<CloseOrderJson> for CloseOrder {
 		})
 	}
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct LiquidityDepositDetails {
+	pub deposit_address: AddressString,
+	pub deposit_chain_expiry_block: u64,
```
kylezs, line 142:
```suggestion
	pub deposit_chain_expiry_block: <AnyChain as Chain>::ChainBlockNumber,
```

api/cf-rpc-types/src/lp.rs
```diff
@@ -135,3 +135,9 @@ impl TryFrom<CloseOrderJson> for CloseOrder {
 		})
 	}
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct LiquidityDepositDetails {
```
kylezs, line 140: Technically details of a channel that will be deposited into, rather than a deposit itself
```suggestion
pub struct LiquidityDepositChannelDetails {
```

kylezs (commented):
As @dandanlen says, it's breaking. I think we can just have this be an entirely separate call and deprecate the old/current one.

kylezs (approved):
Nice 👍

kylezs:
> I have updated documentation where appropriate.

Definitely should do this too (saying that this is active from v1.9 onwards)

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -63,13 +63,22 @@ pub trait Rpc {
...
 	async fn request_liquidity_deposit_address(
 		&self,
 		asset: Asset,
 		wait_for: Option<WaitFor>,
 		boost_fee: Option<BasisPoints>,
-	) -> RpcResult<ApiWaitForResult<String>>;
+	) -> RpcResult<ApiWaitForResult<AddressString>>;
+
+	#[method(name = "liquidity_deposit_v2")]
+	async fn request_liquidity_deposit_address_v2(
```
dandanlen, line 76: nit: It's not clear why we have a different rpc name than the method name.

We could take advantage of this and just use the more explicit naming for the new method (which is a little more consistent with other methods, ie. `{verb}_{thing}`):

 ```suggestion
	#[method(name = "request_liquidity_deposit_address")]
	async fn request_liquidity_deposit_address(
```

(And rename the old method as `request_liquidity_deposit_address_legacy` or something).

File: api/bin/chainflip-lp-api/src/main.rs [hunk 5]
```diff
@@ -231,19 +240,35 @@ impl From<LpApiError> for ErrorObjectOwned {
...
 			.api
 			.lp_api()
 			.request_liquidity_deposit_address(asset, wait_for.unwrap_or_default(), boost_fee)
 			.await
-			.map(|result| result.map_details(|address| address.to_string()))?)
+			.map(|wait_for_result| match wait_for_result {
+				ApiWaitForResult::TxDetails { tx_hash, response } =>
+					ApiWaitForResult::TxDetails { tx_hash, response: response.deposit_address },
+				ApiWaitForResult::TxHash(tx_hash) => ApiWaitForResult::TxHash(tx_hash),
+			})?)
```
dandanlen, line 258:
```suggestion
			.map(|wait_for_result| {
				wait_for_result.map_details(|response| response.deposit_address)
			})?)
```

api/cf-rpc-types/src/lp.rs
```diff
@@ -135,3 +135,9 @@ impl TryFrom<CloseOrderJson> for CloseOrder {
 		})
 	}
 }
+
+#[derive(Serialize, Deserialize, Clone)]
```
dandanlen, line 139:
```suggestion
#[derive(Serialize, Deserialize, Clone, Debug)]
```

This is nice to have...

api/bin/chainflip-cli/src/main.rs
```diff
@@ -83,7 +84,7 @@ async fn run_cli() -> Result<()> {
 							.broker_api()
 							.withdraw_fees(params.asset, params.destination_address)
 							.await?;
-						println!("Withdrawal request successfull submitted: {}", withdraw_details);
+						println!("Withdrawal request successful submitted: {}", withdraw_details);
```
dandanlen, line 87:
```suggestion
						println!("Withdrawal request successfully submitted: {}", withdraw_details);
```

bouncer/tests/lp_api_test.ts
```diff
@@ -98,8 +98,8 @@ async function testLiquidityDeposit(logger: Logger) {
   ).event;
 
   const liquidityDepositAddress = (
-    await lpApiRpc(logger, `lp_liquidity_deposit`, [testRpcAsset, 'InBlock'])
-  ).tx_details.response;
+    await lpApiRpc(logger, `lp_liquidity_deposit_v2`, [testRpcAsset, 'InBlock'])
```
msgmaxim, line 101: If we are not immediately deprecating the previous call, would it make sense to test both v1 and v2?

dandanlen (approved):
Thanks Jamie

msgmaxim (approved):

PR #5743, opened 2025-03-14
refactor: make custom rpc logic shareable

nmammeri:
# Pull Request

Closes: PRO-2126

## Checklist

## Summary

* This PR moves common logic (interact with the runtime API, manage subscriptions, and handle block
notifications) to a separate backend module.
* The main goal is that this new `CustomRpcBackend` struct can be instantiated and the logic is shared by different CustomApi providers such as: the custom rpc, lp_api and broker


dandanlen (approved):
👌 Nice

PR #5744, opened 2025-03-17
fix: more robust cancellation of strategy orders

msgmaxim:
# Pull Request

## Summary

Now instead of having to remember which orders a strategy creates (which given the old code would result in an error if the strategy was closed in the same block as its creation), the strategy simply asks the pool api to cancel all of orders recorded to belong to that strategy. (I think this is a better solution since it will probably be more error prone keeping track of open orders in the strategy pallet especially once we add more complex strategies.) More context: https://github.com/chainflip-io/chainflip-backend/pull/5715#issuecomment-2724769776

### DIFF
--- a/state-chain/pallets/cf-pools/src/lib.rs
+++ b/state-chain/pallets/cf-pools/src/lib.rs
@@ -1147,6 +1147,31 @@ impl<T: Config> PoolApi for Pallet<T> {
 		Pools::<T>::iter_keys().map(|asset_pair| asset_pair.assets()).collect()
 	}
 
+	fn cancel_all_limit_orders(account: &Self::AccountId) -> DispatchResult {
+		// Collect to avoid undefined behaviour (See StorageMap::iter_keys documentation)
+		for asset_pair in Pools::<T>::iter_keys().collect::<Vec<_>>() {
+			let pool = Pools::<T>::get(asset_pair).unwrap();
+
+			for (asset, orders) in
+				pool.limit_orders_cache.as_ref().into_iter().filter_map(|(asset, cache)| {
+					cache.get(account).cloned().map(|orders| (asset, orders))
+				}) {
+				for (id, tick) in orders {
+					Self::cancel_limit_order(
+						account,
+						asset_pair.assets().base,
+						asset_pair.assets().quote,
+						asset.sell_order(),
+						id,
+						tick,
+					)?;
+				}
+			}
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1147,6 +1147,31 @@ impl<T: Config> PoolApi for Pallet<T> {
 		Pools::<T>::iter_keys().map(|asset_pair| asset_pair.assets()).collect()
 	}
 
+	fn cancel_all_limit_orders(account: &Self::AccountId) -> DispatchResult {
+		// Collect to avoid undefined behaviour (See StorageMap::iter_keys documentation)
+		for asset_pair in Pools::<T>::iter_keys().collect::<Vec<_>>() {
+			let pool = Pools::<T>::get(asset_pair).unwrap();
```
kylezs, line 1153:
```suggestion
for (asset_pair, pool) in Pools::<T>::iter().collect::<Vec<_>>() {
```

kylezs: I see you copied that code... guess we can change it at the source too.

kylezs (approved):
Might be worth thinking about if there's anything that can be simplified in the lp-api now, since we have cancel_all_orders RPC there. Would require some rethinking interfaces of cancel_batch for example... don't necessarily think that's necessary, but worth mentioning that these share concepts.

dandanlen (approved):

state-chain/pallets/cf-pools/src/lib.rs
kylezs: The comment is about collecting, which this still does so avoids what the iter_keys() doc is talking about, but having all pools in memory is a fair reason not to - we could comment that here.

msgmaxim: > having all pools in memory is a fair reason not to

This seems to be like a case of premature optimisation though? It is not clear to me that having a tiny memory spike (we are likely talking about well under 1MB here, right?) is worse than the overhead from looking up entries individually. When it is not clear, IMO we should prefer the cleaner solution (e.g. one that doesn't require extra comments) like what Kyle suggested.

I will add a comment and leave the code as is for now (so we can merge it now), but if we don't mind either way, I might change this at some point later.

PR #5748, opened 2025-03-18
Feat: trading strategy benchmarks

msgmaxim:
# Pull Request

Closes: PRO-2113, PRO-2114

## Summary

Benchmarks are straightforward, but note that I had to extract `do_create_pool` from the `new_pool` extrinsic (without changes) to make it callable from another pallet through a trait.


dandanlen (approved):
👌

PR #5752, opened 2025-03-20
test: trading strategies integration test

j4m1ef0rd:
# Pull Request

Closes: PRO-2115

## Checklist

## Summary

- Added an integration test that just test the happy path for trading strategies.
- Added a check when deploying a strategy to make sure the pool exists first.


j4m1ef0rd:
Any ideas why it can find `deploy_trading_strategy` on the CI?

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -71,9 +71,12 @@ impl TradingStrategy {
 		}
 	}
 	pub fn supported_assets(&self) -> Vec<Asset> {
+		self.supported_asset_pairs().into_iter().flat_map(|(a, b)| vec![a, b]).collect()
```
dandanlen, line 74: This could result in a list with duplicates, like if the supported pairs are ETH-USDC/BTC-USDC we will get [ETH, USDC, BTC, USDC].

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -250,6 +251,14 @@ pub mod pallet {
 		) -> DispatchResult {
 			let lp = &T::AccountRoleRegistry::ensure_liquidity_provider(origin)?;
 
+			// Check that the pools exist
+			for (base, quote) in strategy.supported_asset_pairs() {
+				if !T::PoolApi::pool_exists(base, quote) {
+					return Err(Error::<T>::PoolDoesNotExist.into());
```
dandanlen, line 257: Do we have to verify this here?

Isn't it fine if we create a strategy that then just can't add orders because the pool doesn't exist? 

What if the pool exists when the strategy is created, but is deleted after? (Unlikely I agree, but so is creating a strategy for non-existent pools).

state-chain/cf-integration-tests/src/swapping.rs
```diff
@@ -116,6 +118,147 @@ fn credit_account(account_id: &AccountId, asset: Asset, amount: AssetAmount) {
...
+	state_chain_runtime::AllPalletsWithoutSystem::on_finalize(3);
+	state_chain_runtime::AllPalletsWithoutSystem::on_idle(
+		4,
+		Weight::from_parts(1_000_000_000_000, u64::MAX),
+	);
+
+	// Check for swap events
+	if from_asset == STABLE_ASSET || to_asset == STABLE_ASSET {
+		let (.., (input_amount, output_amount), egress_id) = assert_events_match!(
+		Runtime,
```
msgmaxim, line 200: Nit: could improve indentation

state-chain/cf-integration-tests/src/trading_strategy.rs
```diff
@@ -0,0 +1,110 @@
...
+
+            // Start trading strategy
+            register_refund_addresses(&DORIS);
+            credit_account(&DORIS, BASE_ASSET, AMOUNT);
+            const STRATEGY: TradingStrategy =
+                TradingStrategy::SellAndBuyAtTicks { sell_tick: 1, buy_tick: -1, base_asset: BASE_ASSET };
+            assert_ok!(TradingStrategyPallet::deploy_trading_strategy(
+                RuntimeOrigin::signed(DORIS),
+                STRATEGY.clone(),
+                BTreeMap::from_iter(vec![(BASE_ASSET, AMOUNT)]),
```
msgmaxim, line 65: Nit: you don't have to create a Vec, `from_iter` works on slices too. (Doesn't really matter here in a test, but it is useful to know.)

```suggestion
                BTreeMap::from_iter([(BASE_ASSET, AMOUNT)]),
```

j4m1ef0rd: Yeah i did realise this later and used just a slice but then forgot to come back and change this one 😅

state-chain/pallets/cf-trading-strategy/src/lib.rs
j4m1ef0rd: I put it here because I was hitting the `log_or_panic` in the idle function when trying to create the limit orders.
But you are right, the pool could be deleted, so we should change it from a `log_or_panic` to an expected error case.
I still think that we should check that the pool exists here and give early feedback to the user.

msgmaxim (approved):

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -189,19 +192,16 @@ pub mod pallet {
 							if balance >= threshold {
 								weight_used += limit_order_update_weight;
 
-								if let Err(e) = T::PoolApi::update_limit_order(
+								// We expect an error here if the pool does not exist
+								let _error = T::PoolApi::update_limit_order(
```
dandanlen, line 196: nit: it's not necessarily an `_error`...

state-chain/pallets/cf-trading-strategy/src/lib.rs
dandanlen: I feel like we should return an error when it's an error, not just as a convenience. I see little added value in checking this her since it's (a) very unlikely to happen and (b) has no real consequence except the strategy might not do anything and needs to be canceled. 

I don't think the benefit justifies the added trait method/error case.

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -247,6 +246,8 @@ pub mod pallet {
 		LpHasActiveStrategies,
 		/// Strategies are disabled due to safe mode
 		TradingStrategiesDisabled,
+		/// One or more of the supported pools does not exist.
+		PoolDoesNotExist,
```
dandanlen, line 250: This error is unused.

state-chain/pallets/cf-trading-strategy/src/lib.rs
dandanlen: This can still result in duplicates?

j4m1ef0rd: I changed it to a `BTreeSet` in a previous commit. But i might as well just remove `supported_asset_pairs` now that it is unused.

dandanlen: True, missed that.

PR #5755, opened 2025-03-21
feat: safe mode for trading strategies

msgmaxim:
# Pull Request

Closes: PRO-2116

## Summary

Added a single safe mode flag that disables:
- deployment of new strategies
- adding funds to strategies
- closing strategies
- automatic updates of limit orders (in `on_idle`)

Added unit tests to check all 4.

### DIFF
--- a/state-chain/pallets/cf-trading-strategy/src/lib.rs
+++ b/state-chain/pallets/cf-trading-strategy/src/lib.rs
@@ -157,8 +157,9 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_idle(_current_block: BlockNumberFor<T>, remaining_weight: Weight) -> Weight {
-			let mut weight_used: Weight = T::DbWeight::get().reads(1);
+			let mut weight_used: Weight = Weight::zero();
 
+			// We assume this consumes 0 weight since safe mode is likely in cache
 			if !T::SafeMode::get().strategy_execution_enabled {
 				return weight_used
 			}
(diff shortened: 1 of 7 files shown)


state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -50,6 +50,8 @@ pub const PALLET_VERSION: StorageVersion = StorageVersion::new(1);
 // have a fixed order id (at least until we develop more advanced strategies).
 const STRATEGY_ORDER_ID: OrderId = 0;
 
+impl_pallet_safe_mode!(PalletSafeMode; trading_strategies_enabled);
```
kylezs, line 53: probably makes sense to have this more granular, so add / deploy and close are separate options in safe mode.

dandanlen: I agree I think it makes sense to differentiate between: 
- add/update
- close
- execution (`on_idle`)

I can see why we might want to pause updates, but allow closing of strategies (in fact it might be nice to be able to force-close all strategies, but that's another topic). An I can also see why we might want to allow strategy updates but temporarily pause order updates.

msgmaxim: OK, split into three options:
- strategy_updates_enabled
- strategy_closure_enabled
- strategy_execution_enabled

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -154,8 +159,11 @@ pub mod pallet {
 		fn on_idle(_current_block: BlockNumberFor<T>, remaining_weight: Weight) -> Weight {
 			let mut weight_used: Weight = T::DbWeight::get().reads(1);
 
-			// TODO: use safe mode here checking if auto strategies are enabled
+			if !T::SafeMode::get().strategy_execution_enabled {
+				return weight_used
+			}
 
+			weight_used += T::DbWeight::get().reads(1);
```
dandanlen, line 166: FWIW I think we can assume the read access for safe mode is effectively free: because it's read all over the place, it is likely to be in the storage cache. 

(This is what is meant by 'whiltelisted storage' in the benchmarks: storage that is frequently accessed and therefore (almost) guaranteed to be cached.)

dandanlen (approved):

PR #5756, opened 2025-03-21
chore: refactor and update SM docs

MxmUrw:
# Pull Request

Closes: PRO-2061

## Checklist

## Summary

This is a clean-up PR, it does:
 - restructure how the "input index" type of state machines is defined: Previously the `Input` type had an `Indexed` bound, with an associated type `Input::Index`. Now this associated type is directly part of the definition of a statemachine itself. And the link between the input type and the index type is a parent trait of the state machine trait. This allows us to do away with all the `ConstantIndex` and `MultiIndexAndValue` type wrappers!
 - Remove both `ConstantHook` and `IncreasingHook`, instead use the new `MockHook` everywhere.
 - Rename `StateMachineESInstance` to `StatemachineElectoralSystem` and `StateMachineES` to `StatemachineElectoralSystemTypes`.
 - Update the comments of the core `Statemachine` trait and of the `StatemachineElectoralSystem`.


engine/src/witness/btc_e.rs
```diff
@@ -103,7 +106,11 @@ impl VoterApi<BitcoinVaultDepositWitnessingES> for BitcoinVaultDepositWitnessing
 		_settings: <BitcoinVaultDepositWitnessingES as ElectoralSystemTypes>::ElectoralSettings,
 		properties: <BitcoinVaultDepositWitnessingES as ElectoralSystemTypes>::ElectionProperties,
 	) -> Result<Option<VoteOf<BitcoinVaultDepositWitnessingES>>, anyhow::Error> {
-		let (witness_range, vaults, _extra) = properties;
+		let BWElectionProperties {
+			block_height: witness_range,
+			properties: vaults,
+			reorg_id: _,
```
kylezs, line 112: nit, we can just use `..`
```suggestion
			properties: vaults,
            ..,
```

kylezs: same below too ofc

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -111,19 +111,14 @@ impls! {
...
 
-		// restating types since we have to prove that they have the correct bounds
-		type Consensus2 = InputHeaders<Self>;
-		type Vote2 = InputHeaders<Self>;
-		type VoteStorage2 = vote_storage::bitmap::Bitmap<InputHeaders<Self>>;
-
 		// the actual state machine and consensus mechanisms of this ES
 		type ConsensusMechanism = BlockHeightTrackingConsensus<Self>;
-		type StateMachine = BlockHeightTrackingSM<Self>;
+		type Statemachine = BlockHeightTrackingSM<Self>;
```
kylezs, line 121: We still have `BwStateMachine` with capital M. (Don't mind which one we use, but yeah, should be consistent)

kylezs: also `StateMachineES` and the assoc type `StateMachine`

kylezs (approved):

state-chain/runtime/src/chainflip/bitcoin_elections.rs
MxmUrw: Yeah thanks, should be all now.

engine/src/witness/btc_e.rs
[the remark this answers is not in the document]
MxmUrw: I assume so.

PR #5757, opened 2025-03-21
feat: Caching request

marcellorigotti:
# Pull Request

Closes: PRO-2059, PRO-2052, PRO-2069

## Checklist

## Summary

Added caching_request, which allows to cache results to avoid performing the same request multiple times. The cache needs to be 'manually' cleared.

- Implemented a BtcCachingClient, which wraps a CachingRequest around each BitcoinRpcApi using the BtcRetryRpcClient.
- BtcRetryRpcClient updated to retry all the request twice, one with primary and one with backup endpoint.
- Removed BtcRetryRpcApi and impl BtcRpcApi for all the btc clients (BtcCachingClient and BtcRetryRpcClient) such that they make better use of the cache.

## Summary by CodeRabbit

- **New Features**
	- Introduced an enhanced caching client that improves Bitcoin data retrieval performance.
- **Bug Fixes**
	- Improved error handling in Bitcoin RPC interactions to prevent potential underflow issues.
- **Refactor & Improvements**
	- Standardized asynchronous caching across components and streamlined API interactions.
	- Enhanced election-related notifications for updated cache management.
- **Chores**
	- Added a new testing dependency and removed obsolete modules for a cleaner codebase.


engine/src/caching_client.rs
```diff
@@ -0,0 +1,163 @@
...
+								if let Some(value) = cache.get(&request)  {
+										let _ = sender.send(value.clone());
+								} else if let Some(result_senders) = in_flight.get_mut(&request) {
+									result_senders.push(sender);
+								} else {
+									let client = client.clone();
+									let sender_result = cache_sender.clone();
+									in_flight.insert(request.clone(), vec![sender]);
+									scope.spawn(async move {
+										let result = future_any_fn(client).await?;
```
kylezs, line 59: If one particular future errors, then it will propagate through the task scope system, shutting down the whole engine. 

We probably want to send the error down the channel. The errors currently in get here:
```rust
let result = timeout(Duration::from_secs(30), rx).await.map_err(anyhow::Error::new)??;
```
are: 
1. Timeout
2. Sender dropped
3. What we should add, the underlying result.

engine/src/caching_client.rs
```diff
@@ -0,0 +1,163 @@
...
+		let future_any_fn: FutureAnyGenerator<Client> = Box::pin(move |client| {
+			let future = future(client);
+			Box::pin(async move {
+				let result = future.await?;
+				let result: ArcAny = Arc::new(result);
+				Ok(result)
+			})
+		});
+		self.sender.send((key, future_any_fn, tx)).unwrap();
+		let result = timeout(Duration::from_secs(30), rx).await.map_err(anyhow::Error::new)??;
```
kylezs, line 91: nit: const for duration

engine/src/caching_client.rs
```diff
@@ -0,0 +1,163 @@
...
+		let (tx, rx) = oneshot::channel::<ArcAny>();
+		let future_any_fn: FutureAnyGenerator<Client> = Box::pin(move |client| {
+			let future = future(client);
+			Box::pin(async move {
+				let result = future.await?;
+				let result: ArcAny = Arc::new(result);
+				Ok(result)
+			})
+		});
+		self.sender.send((key, future_any_fn, tx)).unwrap();
```
kylezs, line 90: we should just do a `let _result = self.sender.send((key, future_any_fn, tx));` here to ignore the result instead of unwrap. 

If the receiver has been dropped, then we are already propagating the error, there's no need to panic again, potentially creating a misleading stack trace (this is how we do it in other areas of the code too).

kylezs: Looking a bit further though, if we resolve the issue mentioned above, then we never expect the inner loop to exit, and so in that case, panicking would be acceptable imo

kylezs (commented):
Will look more tomorrow, but the general structure of this is a lot better 👌

engine/src/caching_client.rs
```diff
@@ -0,0 +1,163 @@
...
+	pub cache: mpsc::Sender<()>,
+}
+
+impl<Key: Eq + Hash + Clone + Send + 'static, Client: Clone + Send + Sync + 'static>
+	CachingClient<Key, Client>
+{
+	pub fn new(scope: &Scope<'_, anyhow::Error>, client: Client) -> Self {
+		let (sender_request, mut receiver_request) =
+			mpsc::unbounded_channel::<(Key, FutureAnyGenerator<Client>, oneshot::Sender<ArcAny>)>();
+		let (cache_invalidation_sender, mut cache_invalidation_receiver) = mpsc::channel::<()>(1);
```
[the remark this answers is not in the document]
marcellorigotti: wdym?

engine/src/caching_client.rs
marcellorigotti: To avoid propagating the Result<_> is it acceptable to simply drop the sender?

engine/src/btc/cached_rpc.rs
```diff
@@ -0,0 +1,172 @@
...
+	send_raw_transaction: CachingClient<Vec<u8>, Txid, BtcRetryRpcClient>,
+	next_block_fee: CachingClient<(), Option<BtcAmount>, BtcRetryRpcClient>,
+	avg_fee_rate: CachingClient<BlockHash, BtcAmount, BtcRetryRpcClient>,
+	block_header: CachingClient<BlockHash, BlockHeader, BtcRetryRpcClient>,
+	best_block_hash: CachingClient<(), BlockHash, BtcRetryRpcClient>,
+
+	pub cache_invalidation_sender: mpsc::Sender<()>,
+}
+
+impl BtcCachingClient {
```
MxmUrw, line 27: I have a feeling that we could/should deduplicate this code. Maybe https://crates.io/crates/duplicate would be useful? Up to discussion.

MxmUrw (commented):
Looking very nice already! Left a few talking points about some last things to discuss.

engine/src/caching_client.rs
marcellorigotti: NVM in the end I forwarded the error

engine/src/caching_client.rs
```diff
@@ -0,0 +1,219 @@
...
+				.boxed()
+			})
+		});
+
+		Self { request_sender, cache_invalidation_sender, client }
+	}
+
+	pub(crate) async fn get(
+		&self,
+		future: TypedFutureGenerator<ResultType, Client>,
```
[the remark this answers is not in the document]
marcellorigotti: As discussed irl this is not feasible

engine/src/elections.rs
```diff
@@ -61,7 +61,7 @@ where
 		scope: &Scope<'_, anyhow::Error>,
 		state_chain_client: Arc<StateChainClient>,
 		voter: VoterClient,
-		cache_invalidation_sender: Option<mpsc::Sender<()>>,
+		cache_invalidation_sender: Option<Vec<mpsc::Sender<()>>>,
```
MxmUrw, line 64: nit: this could be `cache_invalidation_senders`.

MxmUrw (approved):
LGTM! Added a few naming nits, otherwise feel free to merge! Of course @kylezs might want to have a last look as well?

engine/src/caching_client.rs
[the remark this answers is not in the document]
kylezs: looks used now

engine/src/caching_client.rs
```diff
@@ -0,0 +1,224 @@
...
+							Some((future, request_key, sender)) = request_receiver.recv() => {
+								if let Some(value) = cache.get(&request_key)  {
+										let _ = sender.send(Ok(value.clone()));
+								} else if let Some(result_senders) = in_flight.get_mut(&request_key) {
+									result_senders.push(sender);
+								} else {
+									let sender_result = cache_sender.clone();
+									in_flight.insert(request_key.clone(), vec![sender]);
+									scope.spawn(async move {
+										let result = future.await;
```
kylezs, line 78: nit: would inline

engine/src/caching_client.rs
```diff
@@ -0,0 +1,224 @@
...
+					}
+				}
+				.boxed()
+			})
+		});
+
+		(Self { request_sender, client }, cache_invalidation_sender)
+	}
+
+	pub(crate) async fn get(
```
kylezs, line 95: nit: we could probably use a slightly more descriptive name like `get_or_fetch` to imply there are two modes that could be happening when this function is called

engine/src/caching_client.rs
```diff
@@ -0,0 +1,224 @@
...
+						Vec<oneshot::Sender<Result<ResultType, anyhow::Error>>>,
+					> = HashMap::default();
+					let (cache_sender, mut cache_receiver) =
+						mpsc::unbounded_channel::<(Key, Result<ResultType, anyhow::Error>)>();
+					loop {
+						tokio::select! {
+							biased;
+							Some(()) = cache_invalidation_receiver.recv() => {
+								// We only clear the cache, we don't want to cancel in_flight requests, this way we are sure that every request will eventually complete
+								// regardless of the time it takes to complete
```
kylezs, line 52: Would also note that in practice we wrap each request in a timeout to ensure no request runs forever

engine/src/caching_client.rs
```diff
@@ -0,0 +1,224 @@
...
+	impl Rpc for Client {
+		async fn a(&self) -> anyhow::Result<u32> {
+			sleep(Duration::new(1, 0)).await;
+			A.fetch_add(1, Ordering::Relaxed);
+			Ok(random::<u32>())
+		}
+	}
+
+	#[tokio::test]
+	async fn internal_client_called_once() {
```
kylezs, line 138: A bit harder to test, but I think we should ensure that we clearly test 2 cases of:
1. Request for something already cached only results in 1 request (pulled from cache)
2. Request for something already *in flight* results in 1 request (gets a clone of receiver)


Overall, they are nice tests btw 👌

engine/src/lib.rs
```diff
@@ -18,6 +18,7 @@ pub mod witness;
 
 // Blockchains
 pub mod btc;
+mod caching_client;
```
kylezs, line 21: nit: this isn't a blockchain (should be moved to a diff place)

engine/src/witness/btc/chain_tracking.rs
kylezs, line 1: 🚀

engine/src/witness/btc/source.rs
```diff
@@ -45,8 +45,12 @@ where
 					loop {
 						tick.tick().await;
 
+						let best_block_hash = client
+							.best_block_hash()
+							.await
+							.expect("TODO: This whole source will be deleted ");
```
kylezs, line 51: This is just waiting on the ingress-egress tracker it seems. @MxmUrw you were looking into that right?

MxmUrw: Yes it's on my list, but haven't specifically started yet.

engine/src/caching_client.rs
marcellorigotti: do we?

kylezs (approved):

kylezs:
@coderabbitai full review

engine/src/caching_request.rs
```diff
@@ -0,0 +1,267 @@
...
+	) -> Result<ResultType, anyhow::Error> {
+		let (tx, rx) = oneshot::channel::<Result<ResultType, anyhow::Error>>();
+		let client = self.client.clone();
+		let future = future(client);
+		self.request_sender.send((future, key, tx)).unwrap();
+		let result = timeout(MAX_WAIT_TIME_FOR_REQUEST, rx).await???;
+		Ok(result)
+	}
+}
+
```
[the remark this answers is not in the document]
kylezs: @marcellorigotti for this one, we probably should at least have an expect to explain why it's ok to unwrap here.

engine/src/caching_request.rs
```diff
@@ -98,7 +98,9 @@ impl<
 		let (tx, rx) = oneshot::channel::<Result<ResultType, anyhow::Error>>();
 		let client = self.client.clone();
 		let future = future(client);
-		self.request_sender.send((future, key, tx)).unwrap();
+		self.request_sender.send((future, key, tx)).expect(
+			"Inner loop containing the receiver should never exits, engine is shutting down",
```
kylezs, line 102:
```suggestion
			"Inner loop containing the receiver should never exit, engine is shutting down",
```

kylezs (approved):

PR #5758, opened 2025-03-24
feat: trading strategy limits

j4m1ef0rd:
# Pull Request

Closes: PRO-2120 and PRO-2122

## Checklist

## Summary

- New minimum amount that can be added to a strategy after its deployed.
- Added defaults to all 4 thresholds. Turned off except for stable coins.
- Added pallet config updates to set all thresholds
- Added tests to cover pallet updates and new minimum

### DIFF
--- a/state-chain/pallets/cf-pools/src/lib.rs
+++ b/state-chain/pallets/cf-pools/src/lib.rs
@@ -43,10 +43,10 @@ use frame_support::{
 use cf_traits::HistoricalFeeMigration;
 
 use cf_traits::LpRegistration;
-use frame_system::pallet_prelude::OriginFor;
+use frame_system::{pallet_prelude::OriginFor, WeightInfo as SystemWeightInfo};
 use serde::{Deserialize, Serialize};
 use sp_arithmetic::traits::{SaturatedConversion, Zero};
-use sp_std::{boxed::Box, vec::Vec};
+use sp_std::{boxed::Box, collections::btree_map::BTreeMap, vec::Vec};
 
 pub use pallet::*;
 
@@ -62,6 +62,28 @@ mod mock;
 mod tests;
 
 impl_pallet_safe_mode!(PalletSafeMode; range_order_update_enabled, limit_order_update_enabled);
+
+type SweepingThresholds = BoundedBTreeMap<Asset, AssetAmount, ConstU32<1000>>;
+pub struct DefaultLimitOrderAutoSweepingThresholds<T> {
+	_phantom: PhantomData<T>,
+}
+impl<T: Config> Get<SweepingThresholds> for DefaultLimitOrderAutoSweepingThresholds<T> {
+	fn get() -> SweepingThresholds {
+		Asset::all()
+			.map(|asset| {
+				(
+					asset,
+					match asset {
+						Asset::Usdc | Asset::Usdt => 1_000 * 10u128.pow(6),
+						_ => u128::MAX, // Turned off by default for all other assets
+					},
+				)
+			})
+			.collect::<BTreeMap<Asset, AssetAmount>>()
+			.try_into()
+			.expect("Asset list will fit")
+	}
+}
 
 pub const MAX_ORDERS_DELETE: u32 = 100;
 #[derive(
@@ -173,6 +195,11 @@ impl<T> AskBidMap<T> {
 	pub fn map<S, F: FnMut(T) -> S>(self, mut f: F) -> AskBidMap<S> {
 		AskBidMap { asks: f(self.asks), bids: f(self.bids) }
 	}
+}
+
+#[derive(Clone, RuntimeDebugNoBound, PartialEq, Eq, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub enum PalletConfigUpdate {
+	LimitOrderAutoSweepingThresholds { asset: Asset, amount: AssetAmount },
 }
 
 pub const PALLET_VERSION: StorageVersion = StorageVersion::new(5);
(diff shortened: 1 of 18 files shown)


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -962,6 +993,31 @@ pub mod pallet {
...
 		}
+
+		/// Apply a list of configuration updates to the pallet.
+		///
+		/// Requires Governance.
+		#[pallet::call_index(11)]
+		#[pallet::weight(<T as frame_system::Config>::SystemWeightInfo::set_storage(updates.len() as u32))]
+		pub fn update_pallet_config(
+			origin: OriginFor<T>,
+			updates: BoundedVec<PalletConfigUpdate, ConstU32<10>>,
```
msgmaxim, line 1004: As discussed, I don't think there is a huge disadvantage to slightly overestimating the number of assets here. I think we want to change it to support, say 100 elements, so we are able to update all assets in a single extrinsic in the forseeable future.

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -343,8 +416,52 @@ pub mod pallet {
...
 		}
+
+		/// Apply a list of configuration updates to the pallet.
+		///
+		/// Requires Governance.
+		#[pallet::call_index(4)]
+		#[pallet::weight(<T as frame_system::Config>::SystemWeightInfo::set_storage(updates.len() as u32))]
+		pub fn update_pallet_config(
+			origin: OriginFor<T>,
+			updates: BoundedVec<PalletConfigUpdate, ConstU32<10>>,
```
msgmaxim, line 438: Same here: probably better to slightly overestimate the nubmer of updates.

state-chain/pallets/cf-trading-strategy/src/tests.rs
```diff
@@ -15,18 +15,20 @@
...
 		pool_api::{MockLimitOrder, MockPoolApi},
 	},
 	BalanceApi, Side,
 };
-use frame_support::{assert_err, assert_ok};
+use frame_support::{assert_err, assert_noop, assert_ok, sp_runtime};
 
 use crate::{mock::*, *};
 
+pub const ALICE: <Test as frame_system::Config>::AccountId = 123u64;
```
msgmaxim, line 30: 123 is already taken by `LP` 😄 (see mock.rs). You can either just use `LP`, or use a different id here in case you explicitly don't want them to have any role. Also, this is only used in a single place, so I would just move it there instead of giving it a wide scope (and it probably doesn't need `pub`).

state-chain/pallets/cf-trading-strategy/src/tests.rs
```diff
@@ -399,3 +484,174 @@ fn deregistration_check() {
 			assert_ok!(TradingStrategyDeregistrationCheck::<Test>::check(&LP));
 		});
 }
+
+#[test]
+fn can_update_all_config_items() {
+	new_test_ext().execute_with(|| {
+		const NEW_MIN_DEPLOY_AMOUNT_USDC: AssetAmount = 50_000 * 10u128.pow(6);
```
msgmaxim, line 491: Nit: might make sense to define a const:
```rust
const ONE_USD: AssetAmount = 10u128.pow(6)
```

msgmaxim (approved):

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -52,11 +52,79 @@ const STRATEGY_ORDER_ID: OrderId = 0;
...
+impl<T: Config> Get<BTreeMap<Asset, AssetAmount>> for DefaultLimitOrderUpdateThresholds<T> {
+	fn get() -> BTreeMap<Asset, AssetAmount> {
+		Asset::all()
+			.map(|asset| {
+				(
+					asset,
+					match asset {
+						Asset::Usdc | Asset::Usdt | Asset::ArbUsdc | Asset::SolUsdc =>
+							1_000_000_000, // $1,000 USD
+						_ => u128::MAX, // Turned off by default all other assets
```
dandanlen, line 67: Same comment: we don't need to duplicate the default behaviour in our default config.

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -52,11 +52,79 @@ const STRATEGY_ORDER_ID: OrderId = 0;
 
 impl_pallet_safe_mode!(PalletSafeMode; strategy_updates_enabled, strategy_closure_enabled, strategy_execution_enabled);
 
+pub struct DefaultLimitOrderUpdateThresholds<T> {
```
dandanlen, line 55: The `T` parameter doesn't seem to be necessary?

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -52,11 +52,79 @@ const STRATEGY_ORDER_ID: OrderId = 0;
...
+	_phantom: PhantomData<T>,
+}
+impl<T: Config> Get<BTreeMap<Asset, AssetAmount>> for DefaultLimitOrderUpdateThresholds<T> {
+	fn get() -> BTreeMap<Asset, AssetAmount> {
+		Asset::all()
+			.map(|asset| {
```
dandanlen, line 61: There's a lot of repetition here. 

We could use generics to make this easier to read. This seems to work (compile at least):

```rust
pub struct StablecoinDefaults<const N: u128>();

impl<const N: u128> Get<BTreeMap<Asset, AssetAmount>> for StablecoinDefaults<N> {
	fn get() -> BTreeMap<Asset, AssetAmount> {
		[Asset::Usdc, Asset::Usdt, Asset::ArbUsdc, Asset::SolUsdc]
			.into_iter()
			.map(|asset| (asset, N))
			.collect()
	}
}
```

j4m1ef0rd: This cleans things up, but its not clear how to expand this later if we want to add non-stable assets.

PR #5760, opened 2025-03-25
fix: prevent subscriptions dropping by properly unpinning blocks

nmammeri:
# Pull Request

Closes: PRO-2152

## Checklist

## Summary

* This PR adds a subscription garbage collector where it tracks the all pinned blocks reported by `chainhead_v1_follow` and starts unpinning old blocks once the its queue reaches a certain capacity (default 64)
* Manual tests were carried out on a local node with MAX_PINNED_BLOCKS set to 32 to speed up seeing subscriptions dropping effect.
* A full test with 100 subscriptions was run on a local node with MAX_PINNED_BLOCKS set to the default 512.

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -23,6 +23,7 @@ use futures::{stream, stream::StreamExt, FutureExt};
 use jsonrpsee::{
 	core::async_trait,
 	proc_macros::rpc,
+	tokio::sync::Mutex,
 	types::{
 		error::{ErrorObject, ErrorObjectOwned},
 		ErrorCode,
@@ -70,7 +71,7 @@ use state_chain_runtime::{
 	Hash, NetworkFee, SolanaInstance,
 };
 use std::{
-	collections::{BTreeMap, BTreeSet, HashMap},
+	collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
 	marker::PhantomData,
 	sync::Arc,
 };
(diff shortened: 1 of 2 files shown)


state-chain/custom-rpc/src/lib.rs
```diff
@@ -1969,6 +1970,80 @@ pub enum NotificationBehaviour {
...
+			if !pinned_blocks.contains(new_hash) {
+				// Unpin the oldest block if we are at capacity
+				if pinned_blocks.len() + 1 >= pinned_blocks.capacity() {
+					if let Some(old_hash) = pinned_blocks.pop_front() {
+						if let Err(e) = self
+							.chain_head_client
+							.call::<_, ()>(
+								"chainHead_v1_unpin",
+								[&self.sub_id, &format!("{:?}", old_hash)],
+							)
```
dandanlen, line 2028: There is a jsonrpsee macro `rpc_params!` that makes this nicer to write:

```suggestion
							.call::<_, ()>(
								"chainHead_v1_unpin",
								jsonrpsee::rpc_params![self.sub_id, old_hash],
							)
```

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1969,6 +1970,80 @@ pub enum NotificationBehaviour {
 	New,
 }
 
+/// Number of pinned blocks before starting to unpin old blocks. Must be lower than
+/// `ChainHeadConfig` maximum pinned blocks across all connections (ChainHeadConfig default is 512).
+const MAX_RETAINED_PINNED_BLOCKS: usize = 64;
```
dandanlen, line 1975: Why 64, why not just set it to the be the same as the chain head config value?

dandanlen (commented):
Thanks for this @nmammeri.

I feel like this solution works, but the garbage collection approach seems a bit unnecessary - we know when we no longer need that has (after we call the passed function `f`), so we could just call the `unpin` rpc after we call `f`? 

I had a quick look at this and it appears that lifetimes might be a bit tricky. Did you try this?

nmammeri:
> Thanks for this @nmammeri.
> 
> I feel like this solution works, but the garbage collection approach seems a bit unnecessary - we know when we no longer need that has (after we call the passed function `f`), so we could just call the `unpin` rpc after we call `f`?
> 
> I had a quick look at this and it appears that lifetimes might be a bit tricky. Did you try this?

I thought of doing this first but haven't tried it because I think it won't work in the we have 2 subscriptions:
* Subscription1 -> New blocks 
* Subscription1 -> Finalized blocks 
* Since the pinned blocks cache is shared between all subscription, If we unpin a block received from Subscription1 after we process `f` of Subscription1, the block won't be available for Subscription2.

state-chain/custom-rpc/src/lib.rs
nmammeri: I think this should be as small as possible, all this doing is offering is a buffer time for how long we need the block for. Making it small will also be faster since we search the queue whenever a new hash is added. I think 64 is too conservative, may be we should set it lower value?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -2074,28 +2168,47 @@ where
...
 						..
 					}),
-				) => Some(vec![finalized_block_hashes
-					.pop()
-					.expect("Guaranteed to have at least one element.")]),
+				) => {
+					let initial_hash_vec = vec![finalized_block_hashes
+						.pop()
+						.expect("Guaranteed to have at least one element.")];
+					sub_gc.add(&initial_hash_vec).await;
```
dandanlen, line 2175: I think we need to unpin all the hashes returned in `finalized_block_hashes`...

dandanlen:
> * Since the pinned blocks cache is shared between all subscription

Not sure this is true. According to the spec: 

> A block is pinned only in the context of a specific subscription. If multiple chainHead_v1_follow subscriptions exist, then each (subscription, block) tuple must be unpinned individually.

Btw. I also read this: 

> The finalized blocks reported in the initialized event, and each subsequent block reported with a newBlock event, are automatically considered by the JSON-RPC server as pinned.

So I guess we need to unpin anything that is reported in the initialized event, and anything we see in the newBlock event. 

*However*... since we never use any of the other chainhead apis (like `chainhead_v1_header` etc.) I think we also don't benefit from the 'pinned blocks' cache, so maybe we can just immediately unpin every block as soon as we see the newBlock event?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -2074,28 +2168,47 @@ where
...
+				},
 				(
 					NotificationBehaviour::Best,
 					FollowEvent::BestBlockChanged(sc_rpc_spec_v2::chain_head::BestBlockChanged {
 						best_block_hash,
 					}),
-				) => Some(vec![best_block_hash]),
+				) => {
+					let best_block_vec = vec![best_block_hash];
+					sub_gc.add(&best_block_vec).await;
```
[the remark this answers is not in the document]
dandanlen: See my other comment... having re-read the spec, I think we'll need to unpin `new` blocks even if we don't enter this branch by default because of the notification behaviour. Otherwise blocks that are new but never finalised won't be unpinned and we will eventually drop the subscription.

dandanlen:
As discussed IRL: Let's keep the garbage-collecting approach and just change the match statement so that we always add new/finalized/intitialized event hashes to the GC, irrespective of the configured NotificationBehaviour.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1969,6 +1970,80 @@ pub enum NotificationBehaviour {
...
+			pinned_hashes: Arc::new(Mutex::new(VecDeque::with_capacity(capacity))),
+		}
+	}
+
+	pub async fn add(&self, new_hashes: &[Hash]) {
+		let mut pinned_blocks = self.pinned_hashes.lock().await;
+
+		for new_hash in new_hashes.iter() {
+			// don't add duplicates
+			if !pinned_blocks.contains(new_hash) {
```
dandanlen, line 2019: It might we worth monitoring how this performs and considering a more efficient caching method, like LRU cache used by the underlying server. (in a future PR)

dandanlen (approved):

PR #5761, opened 2025-03-25
feat: ignore redemption tax when redeem all

Janislav:
# Pull Request

Closes: PRO-2035

## Checklist

## Summary

- We check if we try to redeem max and if we try to redeem the total balance of the account. If yes, we don't burn any tax and ignore the tax aside for the rest of the calculations
- Updated all tests where necessary and tried to understand if the change of the test makes sense

### DIFF
--- a/bouncer/tests/fund_redeem.ts
+++ b/bouncer/tests/fund_redeem.ts
@@ -83,7 +83,7 @@ async function main(logger: Logger, providedSeed?: string) {
   logger.debug(`Testing redeem all`);
   const redeemedAll = await redeemAndObserve(logger, seed, redeemEthAddress as HexString, 'Max');
   // We expect to redeem the entire amount minus the exact amount redeemed above + tax & gas for both redemptions
-  const expectedRedeemAllAmount = fundAmount - redeemedExact;
+  const expectedRedeemAllAmount = fundAmount - redeemedExact - redemptionTaxAmount;
   assert(
     redeemedAll >= expectedRedeemAllAmount - gasErrorMargin &&
       redeemedAll <= expectedRedeemAllAmount,
--- a/bouncer/tests/fund_redeem.ts
+++ b/bouncer/tests/fund_redeem.ts
@@ -83,7 +83,7 @@ async function main(logger: Logger, providedSeed?: string) {
   logger.debug(`Testing redeem all`);
   const redeemedAll = await redeemAndObserve(logger, seed, redeemEthAddress as HexString, 'Max');
   // We expect to redeem the entire amount minus the exact amount redeemed above + tax & gas for both redemptions
-  const expectedRedeemAllAmount = fundAmount - redeemedExact - redemptionTaxAmount * 2;
+  const expectedRedeemAllAmount = fundAmount - redeemedExact;
   assert(
     redeemedAll >= expectedRedeemAllAmount - gasErrorMargin &&
       redeemedAll <= expectedRedeemAllAmount,
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -380,13 +380,6 @@ pub mod pallet {
 				}
 			}
 
-			// If we are redeeming the max amount, we don't charge a fee.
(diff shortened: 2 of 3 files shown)


bouncer/tests/fund_redeem.ts
```diff
@@ -83,7 +83,7 @@ async function main(logger: Logger, providedSeed?: string) {
   logger.debug(`Testing redeem all`);
   const redeemedAll = await redeemAndObserve(logger, seed, redeemEthAddress as HexString, 'Max');
   // We expect to redeem the entire amount minus the exact amount redeemed above + tax & gas for both redemptions
-  const expectedRedeemAllAmount = fundAmount - redeemedExact - redemptionTaxAmount * 2;
+  const expectedRedeemAllAmount = fundAmount - redeemedExact - redemptionTaxAmount;
```
Janislav, line 86: This is only the redemption tax once now because we just take it for the exact redeem and not also for the all.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -409,8 +414,7 @@ pub mod pallet {
 			));
 
 			let (debit_amount, redeem_amount) = match amount {
-				RedemptionAmount::Max =>
-					(liquid_balance, liquid_balance.saturating_sub(redemption_fee)),
+				RedemptionAmount::Max => (liquid_balance, liquid_balance),
```
dandanlen, line 417: I think what we need to do is something like this: 

```suggestion
				RedemptionAmount::Max => {
					if liquid_balance == T::Flip::balance(&account_id) {
						// Withdrawing all -> no tax
						liquid_balance
					} else {
						liquid_balance.saturating_sub(redemption_fee)
					}
				},
```

state-chain/pallets/cf-funding/src/tests.rs
```diff
@@ -1561,7 +1568,6 @@ fn skip_redemption_of_zero_flip() {
 	}
 
 	inner_test(100, RedemptionAmount::Exact(0));
-	inner_test(REDEMPTION_TAX, RedemptionAmount::Max);
```
Janislav, line 1564: I think this doesent make sense anymore since we would now just allow the redemption instead of throwing an error.

state-chain/pallets/cf-funding/src/tests.rs
```diff
@@ -267,7 +267,7 @@ fn cannot_double_redeem() {
 fn redemption_cannot_occur_without_funding_first() {
 	new_test_ext().execute_with(|| {
 		const FUNDING_AMOUNT: u128 = 45;
-		const REDEEMED_AMOUNT: u128 = FUNDING_AMOUNT - REDEMPTION_TAX;
+		const REDEEMED_AMOUNT: u128 = FUNDING_AMOUNT;
```
Janislav, line 270: Since we are redeeming max in this test, we don't need to subtract the redemption tax.

state-chain/pallets/cf-funding/src/tests.rs
```diff
@@ -594,17 +594,23 @@ fn restore_restricted_balance_when_redemption_expires() {
 				Default::default()
 			));
 
+			// We ignore the redemption tax if we claim the max amount.
+			let (total_funds, restricted_amount) = if redeem_amount == RedemptionAmount::Max {
+				(TOTAL_FUNDS, RESTRICTED_AMOUNT - REDEMPTION_TAX)
```
Janislav, line 599: I think this makes sense since we try to redeem all, but we can't so we burn the fee from the restricted amount. What confuses me though is why we don't seem to update the total balance, but that is true for all kind of these tests and not only to the one to which the changes apply 🤔.

dandanlen: I think this is a bug: what it means is that we restore all the funds, but not the full restricted amount. 

For example if you have 100FLIP, all restricted, and redeem_all, then let it expire, you finish with 100FLIP, 95 restricted (assuming 5 FLIP tax).

dandanlen (approved):
I'll fix this

PR #5762, opened 2025-03-25
feat: type-safe ccm data

dandanlen:
- reduces the need to re-check ccm adata
- allows us to remove the CcmChecker trait
- allows to expose the actual ccm data

This touches a *lot* of files, but the core changes are in `ccm_checker.rs` and `cf-chains/lib.rs`. Everything else flows from that.


syan095 (approved):
I think this is ready to be merged into the base branch

PR #5767, opened 2025-03-27
feat: single tick trading strategy

j4m1ef0rd:
# Pull Request

Closes: PRO-2154

## Checklist

## Summary

- Changed the trading strategy to just have one param for the spread. 
- Added a new check to the deployment of a strat that makes sure the tick is within the valid range that the aam needs.

We can just add a separate variant in the future when we want a more complex strategy.


state-chain/pallets/cf-trading-strategy/src/tests.rs
```diff
@@ -32,16 +32,12 @@ const QUOTE_ASSET: Asset = cf_primitives::STABLE_ASSET;
 const BASE_AMOUNT: AssetAmount = 100_000;
 const QUOTE_AMOUNT: AssetAmount = 50_000;
 
-const BUY_TICK: Tick = -1;
-const SELL_TICK: Tick = 1;
+const SPREAD: Tick = 1;
```
dandanlen, line 35:
```suggestion
const SPREAD_TICK: Tick = 1;
```

Otherwise it's easy to get confused between the ticks `(-1, 1)` and the spread (`2` ticks).

dandanlen (approved):

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -56,7 +56,7 @@ impl_pallet_safe_mode!(PalletSafeMode; strategy_updates_enabled, strategy_closur
 	Clone, Debug, Encode, Decode, TypeInfo, serde::Serialize, serde::Deserialize, PartialEq, Eq,
 )]
 pub enum TradingStrategy {
-	SellAndBuyAtTicks { sell_tick: Tick, buy_tick: Tick, base_asset: Asset },
+	SellAndBuyAtTicks { spread_tick: Tick, base_asset: Asset },
```
msgmaxim, line 59: Maybe we should reserve this name for a potential future strategy where you can configure sell and buy ticks individually (as before)? How about `TickZeroCentered` for this one?

msgmaxim (approved):

PR #5770, opened 2025-03-28
feat: remove pool fee for limit orders

msgmaxim:
# Pull Request

Closes: PRO-2034

## Summary

- As discussed, when comparing limit and range order prices I chose to "inverse adjust for fee" the limit order price since that led to much cleaner and less error prone code (even though this seems backwards).
- Added a test for fee adjustment math, and also a test to check that we can correctly transition between range and limit orders when executing a swap. Previously I don't think this was sufficiently tested: I almost missed an infinite loop because of this.
- Limit orders no longer charge pool fee; the pool fee field still exists in limit order's PoolState, but (as discussed) I think it would be easier to change the funcitonality first and then clean up later.
- Now setting pool fee only sets it for range orders.
- Added a migration to set limit order pool fee to 0 which is necessary to correctly "collect" fees from any existing limit orders.
- Updates to existing tests:
  - `limit_orders::historics` <- removed as it was focused on fee collection
  - `limit_orders::swap` <- removed a test case that checked fee collection
  - `cf_pools::can_update_pool_liquidity_fee_and_collect_for_limit_order` <- fee related, removed
  - `pallet_limit_order_is_in_sync_with_pool` and `test_cancel_orders_batch` <- minor updates related to fees
  - `cf_pools::fees_are_recorded` -> `range_order_fees_are_recorded`

### DIFF
--- a/state-chain/amm/src/lib.rs
+++ b/state-chain/amm/src/lib.rs
@@ -186,10 +186,16 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
 	>(
 		&mut self,
 	) -> Option<SqrtPriceQ64F96> {
-		match (
-			self.limit_orders.current_sqrt_price::<SD>(),
-			self.range_orders.current_sqrt_price::<SD>(),
-		) {
+		let limit_orders_sqrt_price = self.limit_orders.current_sqrt_price::<SD>();
+		let range_orders_sqrt_price =
+			self.range_orders.current_sqrt_price::<SD>().map(|sqrt_price| {
+				sqrt_price_adjusted_by_pool_fee::<SD>(
+					sqrt_price,
+					self.range_orders.fee_hundredth_pips,
+				)
+			});
+
+		match (limit_orders_sqrt_price, range_orders_sqrt_price) {
 			(Some(limit_order_sqrt_price), Some(range_order_sqrt_price)) =>
 				if SD::sqrt_price_op_more_than(limit_order_sqrt_price, range_order_sqrt_price) {
 					Some(range_order_sqrt_price)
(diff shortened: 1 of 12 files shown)


albert-llimos:
> current_price/current_sqrt_price/range_order_depth (used in various RPCs) now take the fee into account, which seems more correct
I don't think I agree with that given that this is based on the Uniswap V3 pool and users/LPs are used to that. Adjusting the current price for the fee makes sense in our case because we are comparing those values when trying to assess whether we should swap a LO or a RO. However, I don't think it make sense outside of that and I'd argue it's more confusing for someone trying to understand that state of a pool/order.

state-chain/amm/src/range_orders.rs
```diff
@@ -795,9 +810,17 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
 	) -> (Amount, Amount) {
 		let mut total_output_amount = Amount::zero();
 
+		// For historical reasons the code below does not take pool fee into account
+		// when comparing against the price limit, so we adjust here to compensate
+		// for that:
+		let sqrt_price_limit = sqrt_price_limit.map(|price| {
+			sqrt_price_adjusted_by_pool_fee_inverse::<SD>(price, self.fee_hundredth_pips)
+		});
+
```
albert-llimos, line 819: It seems a bit counterintuitive to me that we would pass a squared limit price to this swap function but then we would modify it internally before doing the swap. That's not the behavior I'd expect when using it.
I see how having it one layer above in the AMM swap function might also be strange so maybe this is okay.

state-chain/amm/src/range_orders.rs
```diff
@@ -1287,3 +1306,121 @@ fn one_amount_delta_ceil(
...
+	// fee_factor isn't price, but we can use the same code as for the price
+	// because they use the same fixed point representation:
+	let sqrt_fee_factor = price_to_sqrt_price(fee_factor);
+
+	mul_div_floor(sqrt_price, sqrt_fee_factor, U256::one() << SQRT_PRICE_FRACTIONAL_BITS)
+}
+
+/// Take adjusted sqrt_price and compenate for that adjustment
+fn sqrt_price_adjusted_by_pool_fee_inverse<SD: SwapDirection>(
+	adjusted_sqrt_price: SqrtPriceQ64F96,
```
albert-llimos, line 1354: Shouldn't this just be `sqrt_price`? It's not inverse adjusted yet.

albert-llimos (commented):
The idea and implementation to compare the squared root prices when starting a swap looks good to me :ok_hand:

state-chain/amm/src/range_orders.rs
msgmaxim: You need "adjusted" price to "inverse adjust" it so it becomes `sqrt_price` 🙂 Anyway, I'm removing this code, this function turned out not to be necessary in my final implementation.

state-chain/amm/src/range_orders.rs
msgmaxim: This is also now removed. The price is now adjusted only in the outer loop and inside the limit order logic. This still isn't perfect, but I'm happy to discuss (on discord/over a call) the alternatives I considered and why I chose this approach.

albert-llimos (approved):
Fyi Dan, me and Maxim had a in-depth chat this morning about this.
This final approach makes sense to me and the implementation looks good.
I'm approving it but @dandanlen should also review it (and CI tests need to be updated ofc).

state-chain/amm/src/lib.rs
```diff
@@ -519,3 +552,44 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
+	fee_hundredth_pips: u32,
+) -> SqrtPriceQ64F96 {
+	use cf_amm_math::SQRT_PRICE_FRACTIONAL_BITS;
+
+	let fee_factor = pool_fee_scaling_factor(SD::INPUT_SIDE.sell_order(), fee_hundredth_pips);
+	// fee_factor isn't price, but we can use the same code as for the price
+	// because they use the same fixed point representation:
+	let sqrt_fee_factor = price_to_sqrt_price(fee_factor);
+
+	mul_div_floor(sqrt_price, sqrt_fee_factor, U256::one() << SQRT_PRICE_FRACTIONAL_BITS)
```
dandanlen, line 594: Was it a conscious decision to do it this way (converting the factor) rather than the more obvious (converting the price)? 

```
let price = sqrt_price_to_price(sqrt_price);
let adjusted_price = match side {
	Side::Buy => grow_by_pool_fee(price, fee_hundredth_pips),
	Side::Sell => reduce_by_pool_fee(price, fee_hundredth_pips),
};
price_to_sqrt_price(adjusted_price)
```

dandanlen (approved):

dandanlen:
One other question - how does this show up in rpcs/subscriptions? I guess we don't compensate for the fee in those? ie. if the RO price is better than the LO price, we return the RO price regardless? (I think this is fine if so).

msgmaxim:
> One other question - how does this show up in rpcs/subscriptions? I guess we don't compensate for the fee in those? ie. if the RO price is better than the LO price, we return the RO price regardless? (I think this is fine if so).

I think we should similary compensate for the RO's price in `current_sqrt_price` where the LO and RO prices are directly compared, so I will change it. I'm not sure about `range_order_depth` where the behaviour is exactly as it was before (pool fee was never taken into account and LO fee is irrelevant), so I will leave it as is for now.

PR #5771, opened 2025-03-28
feat: combine eth/btc mocks for ingress-egress pallet

dandanlen:
Allows us to re-use utilities for both sets of tests, at the expense of some additional instance annotations.


msgmaxim (approved):

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -2172,7 +2277,7 @@ fn rejects_invalid_swap_by_witnesser() {
 		// Invalid BTC address:
 		assert_ok!(submit_vault_swap_request(
 			Asset::Eth,
-			Asset::Btc,
+			Asset::Dot,
```
Janislav, line 2280: Why is this Dot now? 🤔

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -2172,7 +2277,7 @@ fn rejects_invalid_swap_by_witnesser() {
 		// Invalid BTC address:
```
Janislav, line 2277: I think this should be be `Invalid DOT address`

Janislav (approved):
👍

state-chain/pallets/cf-ingress-egress/src/tests.rs
dandanlen, line 2277:
```suggestion
		// Invalid combination of Asset/Destination.
```

PR #5775, opened 2025-04-01
feat: Generic typeinfo

martin-chainflip:
Closes: [PRO-1587](https://linear.app/chainflip/issue/PRO-1587)

Summary
This change introduces a new derive macro "GenericTypeInfo" that can replace the existing "TypeInfo" macro for structs or enums in generic pallets that have different types depending on the chain the pallet is implemented for.

Leaking memory for the string is ok, because the Wasm runtime is re-instantiated for each method call.


state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+
+fn derive_fields(d: Fields, chain_ident: Ident) -> TokenStream2 {
+	match d {
+		syn::Fields::Named(fields_named) => {
+			let fields: Vec<TokenStream2> = fields_named.named.iter().map(|field| {
+                let ty = field.clone().ty;
+                let typename = clean_type_string(&quote!(#ty).to_string());
+                let name = field.clone().ident.unwrap();
+                quote!(.field(|f| {
+					let full_typename = scale_info::prelude::format!("{}For{}", #typename, #chain_ident::NAME).leak();
```
MxmUrw, line 56: What consistency requirements does the metadata generation process have? What I mean is:

Here we give names for the types of all fields. Should the name match up with the name that is generated for the field type *in its own* `TypeInfo` implementation? For example, if we have a field of type `RequestId` (without any type parameters), here we're gonna say the type is (e.g.) `RequestIdForBitcoin`. But since `RequestId` itself doesn't have any type parameters, we're gonna generate its type-info with the normal substrate derive macro which does not add any `ForChain`-suffix - this seems like there's gonna be a name mismatch in the generated metadata, no?

state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+			let full_pathname = scale_info::prelude::format!("{}For{}", ::core::stringify!(#ident), #chain_ident::NAME).leak();
+			scale_info::Type::builder()
+			.path(scale_info::Path::new(full_pathname, ::core::module_path!()))
+			.variant(
+				scale_info::build::Variants::new() #( #variants )* )
+		}
+	})
+}
+
+fn clean_type_string(input: &str) -> String {
```
MxmUrw, line 129: It feels like it would be nice to have a little comment about what the purpose of this cleaning is? Is there some target parser whose formatting requirements we're trying to match up with?

state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+		},
+		syn::Fields::Unnamed(fields_unnamed) => {
+			let fields: Vec<TokenStream2> = fields_unnamed
+				.unnamed
+				.iter()
+				.map(|field| {
+					let ty = field.clone().ty;
+					let typename = clean_type_string(&quote!(#ty).to_string());
+					quote!(.field(|f| {
+						let full_typename = scale_info::prelude::format!("{}For{}", #typename, #chain_ident::NAME).leak();
```
MxmUrw, line 69: A comment why the `leak()` is fine would be useful probably.

MxmUrw (commented):
Looking nice, I have a question about how it's working in the case of a recursive lookup of types though.

kylezs (commented):
This will also need to be applied to the elections pallet types 🙏

state-chain/generic-typeinfo-derive/src/lib.rs
martin-chainflip: The types remain the same. The "name" is supposed to be purely for human interpretation. The bug that this PR fixes is that the PolkadotJS library relies on the uniqueness of this name to decode types that it should decode according to their unique ID.

MxmUrw: Yes, but since PolkadotJS relies on the names to decode, shouldn't we generate the same names for types that use the standard `TypeInfo` macro as the standard macro does?

state-chain/generic-typeinfo-derive/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+use syn::{Data, DataEnum, DataStruct, DeriveInput, Fields, Ident, ImplGenerics, TypeGenerics};
+
+#[proc_macro_derive(GenericTypeInfo)]
+pub fn derive(item: TokenStream) -> TokenStream {
+	let item2: TokenStream2 = item.into();
+	let ast: DeriveInput = syn::parse2(item2).unwrap();
+	let ident = ast.ident;
+	let chain_generic_ident = ast.generics.type_params().next().unwrap().ident.clone();
```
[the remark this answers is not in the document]
martin-chainflip: good point. done

state-chain/generic-typeinfo-derive/src/lib.rs
martin-chainflip: added a comment to explain. it's essentially to align with what the original typeinfo derive macro does

state-chain/generic-typeinfo-derive/src/lib.rs
martin-chainflip: added

Cargo.toml
```diff
@@ -305,46 +306,13 @@ sol-prim = { path = "foreign-chains/solana/sol-prim", default-features = false }
 # This list is ordered alphabetically.
 [profile.dev.package]
 blake2 = { opt-level = 3 }
-blake2b_simd = { opt-level = 3 }
-chacha20poly1305 = { opt-level = 3 }
-cranelift-codegen = { opt-level = 3 }
-cranelift-wasm = { opt-level = 3 }
-crc32fast = { opt-level = 3 }
-crossbeam-deque = { opt-level = 3 }
-crypto-mac = { opt-level = 3 }
```
dandanlen, line 314: Any particular reason for deleting these? I think the reason for these entries is to speed up tests by ensuring certain (transitive) dependencies are always compiled with full optimisations.

All the remove packages seem to be needed (nothing is removed from `Cargo.lock`).

martin-chainflip: this was done automatically when cherry-picking. i'll add them back

dandanlen (commented):
There are still a few issues that I think I mentioned before: 
- We call it 'generic' but it's very much specific to our use case: we just assume that the first type param has a `NAME` trait constant. There are no constraints that enforce this, so I can imagine there might be some confusing error messages if this is used in the wrong context. 
- Internally, we call use this `#chain_ident::NAME`, but in fact it's just some type param that implements a trait (any trait) that has a `NAME` associated constant.
- We're duplicating the `Chain::NAME` constant in each of the configs that use this. 


I think one simple (?) improvement would be to allow passing the 'discriminator' suffix as a param to the derive macro. For example I can imagine it being used like this: 

```rust
#[derive(GenericTypeInfo(T::TargetChain::NAME))]
struct MyConfigGeneric<T: Config<I>, I: 'static> {
	call: T::RuntimeCall,
}


#[derive(GenericTypeInfo(T::NAME))]
struct MyChainGeneric<T: Chain> {
	thing: T::ChainAccount,
}

```

It would also be nice to have some tests to convince ourselves the output is as expected.

dandanlen:
Also, am I right that this will generate custom type names for all types in the struct, even primitives? For example we might end up with `u32ForBitcoin` etc?

MxmUrw:
> Also, am I right that this will generate custom type names for all types in the struct, even primitives? For example we might end up with `u32ForBitcoin` etc?

Yeah, it looks like it. We have for example `ForeignChainAddress` whose typeinfo is generated with the substrate macro:
```
{
  id: 274
  type: {
    path: [
      cf_chains
      address
      ForeignChainAddress
    ]
    params: []
    def: {
      ...
      }
  }
}
```
and then it referenced in the `TransactionRejectionDetails` typeinfo (generated by the new macro) as `ForeignChainAddressForEthereum`.
```
{
  id: 737
  type: {
    path: [
      pallet_cf_ingress_egress
      TransactionRejectionDetailsForEthereum
    ]
    params: []
    def: {
      Composite: {
        fields: [
          {
            name: refund_address
            type: 274
            typeName: ForeignChainAddressForEthereum
            docs: []
          }
          {
            name: asset
            type: 30
            typeName: TargetChainAsset<T, I>ForEthereum
            docs: []
          }
          ...
        ]
      }
    }
    docs: []
  }
}
```

martin-chainflip:
To address some of the comments:
If T doesn't have the required NAME property, the error message is very clear. However, I made the parameter explicit, so that we can set whatever we want as the name.
I also added a way to skip expanding names, for example for primitives, even though that is not really necessary. It is perfectly fine to have the same type show up under multiple names. This happens already without this change, for example we have several names for u16. If the polkajs library can't find the type via name, it will use the unique ID. The problematic thing is the other way around, where different types have the same name.

dandanlen:
As discussed - i added some changes to enable taking the suffix from the chain instance definitions, which should enforce some consistency. Note this doesn't work for the elections pallet because the types in that pallet are not explicitly dependent on the chain, so I left the existing config item (but renamed it from NAME to the more explicit TYPE_INFO_SUFFIX).

dandanlen (approved):
I think this is good to go :)

state-chain/pallets/cf-threshold-signature/src/lib.rs
```diff
@@ -157,6 +160,7 @@ pub enum KeyRotationStatus<T: Config<I>, I: 'static = ()> {
 		ceremony_id: CeremonyId,
 		response_status: KeyHandoverResponseStatus<T, I>,
 		receiving_participants: BTreeSet<T::ValidatorId>,
+		#[skip_name_expansion]
```
kylezs, line 163: is there a reason we do this on some like this but not on others like `CeremonyId` above?

PR #5778, opened 2025-04-02
feat: change minimum network fee to be per request

j4m1ef0rd:
# Pull Request

Closes: PRO-2151

## Checklist

## Summary

- Renamed `MinimumNetworkFeePerChunk` -> `MinimumNetworkFee`
	- Added a migration for this rename
- Added 2 new fields to the DcaState storage.
	- Added a migration for these new fields.
- Change the way the network fee is calculated to take into account the amount collected on previous chunks and the minimum.
	- Removed the `calculate_network_fee` function as it did not seem needed anymore.
- Reworked the network fee calculation test
- Added a test for dca swap with minimum network fee

TODO: test migrations


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2053,20 +2068,45 @@ pub mod pallet {
 			})
 		}
 
-		pub(super) fn take_network_fee(input: AssetAmount, min_fee_enforced: bool) -> FeeTaken {
+		/// Take network fee from the input amount and return the remaining amount and the fee.
+		/// If `swap_request_id` is provided, the minimum network fee is enforced.
+		pub(super) fn take_network_fee(
+			input: AssetAmount,
+			swap_request_id: Option<SwapRequestId>,
```
msgmaxim, line 2075: Without reading the comment it is not very clear that min fee is applied differently depending on whether swap request id is provided. Would it not be better to introduce a new enum? Something like:

```rust
enum MinFeePolicy {
  Enforced { swap_request_id: SwapRequestId },
  NotEnforced
}
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2053,20 +2068,45 @@ pub mod pallet {
...
 			if input.is_zero() {
 				return FeeTaken { remaining_amount: 0, fee: 0 };
 			}
 
-			let min_fee = if min_fee_enforced { MinimumNetworkFeePerChunk::<T>::get() } else { 0 };
+			// Check if minimum network fee still needs to be filled for this swap request.
+			let network_fee = T::NetworkFee::get();
+			let fee = if let Some(swap_request_id) = swap_request_id {
+				let swap_request =
+					SwapRequests::<T>::get(swap_request_id).expect("Swap request should exist");
```
msgmaxim, line 2085: I think if it is not locally clear that this won't panic (like if you just checked it a line above), we should wirte code more defensively: we could simply fallback to charging 0 if you can't find the information you need.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1773,6 +1783,12 @@ pub mod pallet {
 						);
 
 						dca_state.status = DcaStatus::ChunkScheduled(swap_id);
+						dca_state
+							.network_fee_collected
+							.saturating_accrue(swap.network_fee_taken.unwrap_or_default());
+						dca_state
+							.accumulated_input_amount
+							.saturating_accrue(swap.swap.input_amount);
```
msgmaxim, line 1791: As discussed, this needs to be the amout of the stable asset. Should probably change the tests so they would catch this.

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: Yeah i was debating whether to do this.

msgmaxim (approved):

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -127,45 +127,97 @@ fn test_buy_back_flip() {
 
 #[test]
 fn test_network_fee_calculation() {
```
kylezs, line 129: A couple of things I think we can improve on in this test:
1. Test the `CollectedNetworkFee` storage is as we expect - this is possibly the most important and it's not tested here.
2. Rather than insert a swap request over the top of another request, why not just have the function take an optional swap request id rather than this `with_swap_request: bool` ?  Then the same id can be used for multiple tests and it's clear that the intention is that it's the same swap request, rather than each invocation actually overwriting the swap request, making it look like each one is new (which it actually kinda is at the moment).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1495,12 +1507,21 @@ pub mod pallet {
...
+							let FeeTaken { remaining_amount, fee } = Self::take_network_fee(
+								stable_amount,
+								if *min_fee_enforced {
+									MinFeePolicy::Enforced {
+										swap_request_id: swap.swap.swap_request_id,
+									}
+								} else {
+									MinFeePolicy::NotEnforced
+								},
+							);
```
kylezs, line 1524: not related to this PR specifically, but there doesn't seem to be a need for the `let mut stable_amount` variable. We only assign to it to later assign to swap... we could just assign to swap directly?

kylezs (approved):

state-chain/pallets/cf-swapping/src/migrations.rs
```diff
@@ -28,5 +31,19 @@ pub type PalletMigration<T> = (
...
 	>,
-	PlaceholderMigration<8, Pallet<T>>,
+	VersionedMigration<
+		8,
+		9,
+		dca_state_migration::Migration<T>,
+		Pallet<T>,
+		<T as frame_system::Config>::DbWeight,
+	>,
+	VersionedMigration<
```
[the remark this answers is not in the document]
j4m1ef0rd: They are 2 separate operations, so i thought it would be cleaner to keep them seperate.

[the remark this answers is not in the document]
j4m1ef0rd: Turns out I broke the existing migration by changing the DCA struct that it used. So I had to merge the 2 swap request migrations into one anyway.
Once again, i'm glad i took the effort to actually run the try-migrations.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1506,14 +1506,13 @@ pub mod pallet {
 					"All swaps should have Stable amount set here"
 				);
 
-				let mut stable_amount = swap.stable_amount.unwrap_or_default();
-				swap.stable_amount_before_fees = Some(stable_amount);
```
kylezs: This is actually subtly different to before. If there are no fees in `swap.swap.fees`, and swap.stable amount is set to None then `stable_amount_before_fees` might stay as None.

In the previous implementation this was guaranteed to be Some after a call to this function.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1494,30 +1506,38 @@ pub mod pallet {
 					"All swaps should have Stable amount set here"
 				);
 
-				let mut stable_amount = swap.stable_amount.unwrap_or_default();
+				swap.stable_amount_before_fees = swap.stable_amount;
```
kylezs, line 1509: This is actually subtly different to before. 

If `swap.swap.fees` is empty. Then in this new implementation it's possible that `stable_amount_before_fees` is None. 
Note that in the old implementation that `stable_amount_before_fees` is always Some after a call to this function.

j4m1ef0rd: I did see that, but we have a `debug_assert!(swap.stable_amount.is_some())` earlier in the function, so I thought it would be fine. I think it's more correct to leave `stable_amount_before_fees` as None in that case and make sure we handle it later when the value is used elsewhere.

PR #5780, opened 2025-04-02
Moved beginning of election to Egress (instead of ingress).

syan095:
Removed Swap ID from the Election workflow

# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

### DIFF
--- a/state-chain/chains/src/ccm_checker.rs
+++ b/state-chain/chains/src/ccm_checker.rs
@@ -30,7 +30,7 @@ use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
 use sp_runtime::DispatchError;
-use sp_std::{collections::btree_set::BTreeSet, vec, vec::Vec};
+use sp_std::{collections::btree_set::BTreeSet, vec::Vec};
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
 pub enum CcmValidityError {
@@ -118,10 +118,11 @@ impl VersionedSolanaCcmAdditionalData {
 		}
 	}
 
-	pub fn address_lookup_tables(&self) -> Vec<SolAddress> {
+	pub fn address_lookup_tables(&self) -> Option<Vec<SolAddress>> {
 		match self {
-			VersionedSolanaCcmAdditionalData::V0(..) => vec![],
-			VersionedSolanaCcmAdditionalData::V1 { alts, .. } => alts.clone(),
+			VersionedSolanaCcmAdditionalData::V0(..) => None,
+			VersionedSolanaCcmAdditionalData::V1 { alts, .. } =>
+				(!alts.is_empty()).then_some(alts.clone()),
 		}
 	}
 }
@@ -154,7 +155,7 @@ impl CcmValidityChecker {
 			.map_err(|_| CcmValidityError::CannotDecodeCcmAdditionalData)?;
 
 			let ccm_accounts = decoded_data.ccm_accounts();
-			let address_lookup_tables = decoded_data.address_lookup_tables();
+			let address_lookup_tables = decoded_data.address_lookup_tables().unwrap_or_default();
 			let num_address_lookup_tables = address_lookup_tables.len();
 
 			if num_address_lookup_tables > MAX_CCM_USER_ALTS as usize {
--- a/state-chain/chains/src/sol/api.rs
+++ b/state-chain/chains/src/sol/api.rs
@@ -41,7 +41,7 @@ use crate::{
 	TransferFallbackError,
 };
 
-use cf_primitives::{EgressId, ForeignChain, GasAmount, SwapRequestId};
+use cf_primitives::{EgressId, ForeignChain, GasAmount};
 
 #[derive(Clone, Encode, Decode, PartialEq, RuntimeDebug, TypeInfo)]
 pub struct ComputePrice;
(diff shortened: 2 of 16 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2899,6 +2888,15 @@ impl<T: Config<I>, I: 'static> EgressApi<T::TargetChain> for Pallet<T, I> {
 							amount,
 						);
 
+					// Initiate LUT witnessing if required.
```
dandanlen, line 2891: nit: Let's use just one acronym, ALT or LUT :)

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1973,14 +1971,6 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 					dca_params.clone(),
 					origin.into(),
 				);
-				if let Some(ccm_channel_metadata) = channel_metadata {
```
kylezs, line 1976: 👌

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -702,47 +700,27 @@ impl FromSolOrNot for SolanaVaultSwapDetails {
...
-			_ => {
-				log_or_panic!("Invalid CCM data type for Solana alt witnessing");
-			},
-		}
+pub(crate) fn initiate_solana_alt_election(alts: Vec<SolAddress>) {
+	if alts.is_empty() {
+		return
 	}
+
+	// TODO roy: fix this.
```
kylezs, line 708: fix what?

state-chain/chains/src/sol/api.rs
```diff
@@ -383,13 +383,20 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
 		ccm_additional_data: VersionedSolanaCcmAdditionalData,
 	) -> Result<Self, SolanaTransactionBuildingError> {
 		let ccm_accounts = ccm_additional_data.ccm_accounts();
+		let alts = ccm_additional_data
+			.address_lookup_tables()
```
kylezs, line 387: I think renaming this to be something like `address_lookup_table_addresses` makes it a bit clearer, since here they are just the addresses of the tables, rather than the tables themselves.

state-chain/chains/src/sol/api.rs
```diff
@@ -383,13 +383,20 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
 		ccm_additional_data: VersionedSolanaCcmAdditionalData,
 	) -> Result<Self, SolanaTransactionBuildingError> {
 		let ccm_accounts = ccm_additional_data.ccm_accounts();
+		let alts = ccm_additional_data
```
kylezs, line 386: nit: would inline this (or at the very least move it to right before it's used

kylezs (commented):
This is looking quite nice 👌

state-chain/chains/src/ccm_checker.rs
```diff
@@ -93,6 +93,13 @@ impl DecodedCcmAdditionalData {
 			_ => None,
 		}
 	}
+
+	pub fn address_lookup_tables(&self) -> Option<Vec<SolAddress>> {
+		match self {
+			DecodedCcmAdditionalData::Solana(sol_data) => sol_data.address_lookup_tables(),
```
dandanlen, line 99: I don't think we need this method - it can be a bit misleading since a return value of `None` can indicate the wrong chain (not Solana) or valid but no lookup tables. Better to just inline this.

dandanlen (commented):
I agree with Kyle's comments, otherwise LGTM 👌

state-chain/chains/src/sol/api.rs
```diff
@@ -401,6 +399,20 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
 		let durable_nonce = Environment::nonce_account()?;
 
+		// Get the Address lookup tables. Chainflip's ALT is proceeded with the User's.
+		let external_alts = ccm_additional_data
+			.alt_addresses()
+			.map(|alt_addresses| Environment::get_address_lookup_tables(alt_addresses))
+			.transpose()
+			.inspect_err(|_| {
+				Environment::recover_durable_nonce(durable_nonce.0);
+			})?
```
dandanlen, line 408: Note here, we have to remember to revert any state if we exit with an error. 

So I moved the ALT lookup (which *takes* the data) after all the `?` points and added a call to recover the nonce if it errors. 

We should not have to rely on remembering this... I think ideally all these tx construction methods should be transactional - then we won't need to remember to eg. `recover_durable_nonce`. I'll raise an issue.

state-chain/runtime/src/chainflip/solana_elections.rs
dandanlen: @syan095 ☝️ do you know what this refers to ?

state-chain/chains/src/sol/api.rs
dandanlen: (https://linear.app/chainflip/issue/PRO-2238/make-transaction-building-transactional)

dandanlen:
> @syan095 I merged the other PR and addressed remaining comments on this one. I think it makes sense to merge this into the base branch now, rebase everything and then do another full review of the feature.

Update - I rebased the base PR - now this one needs to be rebased on top of the base...

PR #5781, opened 2025-04-02
feat: rpc 2.0 move lp and broker apis to node

nmammeri:
# Pull Request

Closes: PRO-2083

## Checklist

## Summary

This PR moves the broker and LP rpcs to the node:
* Implements a `SignedPoolClient` that interacts with the transaction_pool to submit extrinsics and listen for status updates.
* Implements RuntimeDecoder that is created off a runtime metadata and can dynamically decode errors and events
* Moves all Broker and LP api rpcs to the node with all the necessary translations to use the `SignedPoolClient` and the new dynamic events decoding
* Injects the broker and LP keys into the node
* Endpoint updates to bouncer and localnet scripts 
* Includes all review fixes of initial [draft PR](https://github.com/chainflip-io/chainflip-backend/pull/5689) 

Followup issues: 
* [PRO-2104](https://linear.app/chainflip/issue/PRO-2104)
* [PRO-2105](https://linear.app/chainflip/issue/PRO-2105)
* [PRO-2090](https://linear.app/chainflip/issue/PRO-2090))


state-chain/custom-rpc/src/pool_client.rs
```diff
@@ -0,0 +1,610 @@
...
+		+ sp_transaction_pool::runtime_api::TaggedTransactionQueue<B>
+		+ frame_system_rpc_runtime_api::AccountNonceApi<B, AccountId, Nonce>,
+{
+	pub fn new(client: Arc<C>, pool: Arc<FullPool<B, C>>, pair: sp_core::sr25519::Pair) -> Self {
+		Self {
+			_phantom: Default::default(),
+			system_api: System::new(client.clone(), pool.clone(), sc_rpc_api::DenyUnsafe::Yes),
+			pair_signer: PairSigner::new(pair),
+			runtime_decoders: Arc::new(RwLock::new(HashMap::new())),
+			pool_semaphore: Arc::new(Semaphore::new(1)), // Allow only 1 pool access at a time
```
dandanlen, line 137: How is this is different from a mutex? (Or what's the benefit of using a semaphore over a mutex for this?)

api/cf-node-client/src/subxt_state_chain_config.rs
```diff
@@ -42,3 +43,170 @@ impl Config for StateChainConfig {
...
+	}
+}
+
+impl From<cf_static_runtime::runtime_types::sp_runtime::DispatchError>
+	for sp_runtime::DispatchError
+{
+	fn from(error: cf_static_runtime::runtime_types::sp_runtime::DispatchError) -> Self {
+		match error {
+			// TODO: investigate why the types are not symmetrical. may be subxt-cli version
+			// mismatch
```
dandanlen, line 138: I think this might be because some of the information is dropped during encoding. 

For example the `Other` variant is encoded without the inner string...:

```rust
pub enum DispatchError {
	/// Some error occurred.
	Other(
		#[codec(skip)]
		#[cfg_attr(feature = "serde", serde(skip_deserializing))]
		&'static str,
	),
    //...
}
```

nmammeri: shall I remove the TODO then?

dandanlen: Yes I think we can. Or maybe we don't need the extra type at all? In general for DispatchErrors we need to use the metadata to decode the correct information anyway.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1130,6 +1140,9 @@ impl From<CfApiError> for ErrorObjectOwned {
 			},
 			CfApiError::ErrorObject(object) => object,
 			CfApiError::OtherError(error) => internal_error(error),
+			CfApiError::SubstrateClientError(error) => call_error(error),
+			CfApiError::PoolClientError(error) => call_error(error),
+			CfApiError::DynamicEventsError(error) => call_error(error),
```
dandanlen, line 1145: I'm not sure these should all be 'call_error'. CallError is meant to denote that it was some issue while processing the call and that it wasn't applied. 

However, for example, a DynamicEventsError could indicate that the call was successful, but that we couldn't decode the result for some reason.

dandanlen: Actually, looking at the `call_error`, it seems that it maps the error to an InternalError... not sure if this is right. (I know this predates this PR)

api/cf-node-client/src/events_decoder.rs
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	pub fn as_event<E: StaticEvent>(&self) -> Result<Option<E>, DynamicEventError> {
+		self.event.as_event::<E>().map_err(DynamicEventError::from)
+	}
+
+	pub fn as_event_strict<E: StaticEvent>(&self) -> Result<Option<E>, DynamicEventError> {
+		let ev_metadata = self.event.event_metadata();
+		if ev_metadata.pallet.name() == E::PALLET && ev_metadata.variant.name == E::EVENT {
+			let ev_static_metadata = self.event_static_metadata::<E>()?;
```
dandanlen, line 74: What's the difference between `self.metadata` and `self.event.event_metadata()`? Do we ever expect them to be in conflict?

state-chain/custom-rpc/src/lib.rs
nmammeri: I might have got it wrong but the only difference I see between a `call_error` and `internal_error` is that for the internal_error case we log the message as ERROR and return jsonrpc InternalError code. In the call_error case we log the message on the node as DEBUG and return the same jsonrpc InternalError code. 

Shall we  introduce another error function besides `call_error` and `internal_error`? I don't these qualify to be logged as ERRORs in the logs

api/cf-node-client/src/events_decoder.rs
dandanlen: Is one queried at runtime and another compiled in statically?

nmammeri: I'm not sure which `self.metadata` you're referring to but If you meant `self.event_static_metadata`
`self.event_static_metadata`: is the events EventMetadataDetails (pallet and variant) decoded using the runtime the binary was compiled with hence the name **static**
`self.event.event_metadata()`:  is the events EventMetadataDetails (pallet and variant) decoded using the current runtime.

Yes they can be different

dandanlen (approved):
LGTM 🚀 

I made some refactors / did some renaming / documentation as I went. If you disagree with anything @nmammeri let me know, otherwise let's get this merged!

api/cf-node-client/src/events_decoder.rs
```diff
@@ -0,0 +1,263 @@
...
+	type Item = DynamicEvent;
+
+	fn next(&mut self) -> Option<Self::Item> {
+		self.events.pop()
+	}
+}
+
+impl DynamicEvents {
+	/// Find an event by pallet and event name
+	pub fn find_event(&self, pallet_name: &str, event_name: &str) -> Option<&DynamicEvent> {
```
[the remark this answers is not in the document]
nmammeri: yes, will remove it.

nmammeri:
> LGTM 🚀
> 
> I made some refactors / did some renaming / documentation as I went. If you disagree with anything @nmammeri let me know, otherwise let's get this merged!

Awsome changes code looks more cleaner and easier to read  🚀

PR #5782, opened 2025-04-03
fix: correctly estimate input amount in scheduled swaps

msgmaxim:
# Pull Request

Closes: PRO-1929

## Summary

- In the fallback case (no liquidity) we now use the price of the correct asset when estimating input
- For this I added a `PoolPriceApi` trait so that the prices are available from the swapping pallet (previously we were passing the price as a parameter, but now we would need prices for multiple/all assets)
- The use of `PoolPriceApi` means we can write a better test that would catch this bug


dandanlen (approved):
This seems like something that should be cherry-picked to release...

PR #5793, opened 2025-04-08
feat: different network fee for internal swaps

j4m1ef0rd:
# Pull Request

Closes: PRO-2194 & PRO-2173

## Checklist

## Summary

- Added `InternalSwapNetworkFee` & `InternalSwapMinimumNetworkFee` storage items
	- Added them to the config update extrinsic.
  	- Changed the existing migration to set the values of the new items to be the same as the normal network fees
- Changed the fee logic to use the internal fee values when the swap is an internal swap.
- Expanded the unit test to cover the new cases.


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1057,6 +1078,16 @@ pub mod pallet {
...
+					PalletConfigUpdate::SetInternalSwapNetworkFee { fee } => {
+						InternalSwapNetworkFee::<T>::set(fee);
+						Self::deposit_event(Event::<T>::InternalSwapNetworkFeeSet { fee });
+					},
+					PalletConfigUpdate::SetInternalSwapMinimumNetworkFee { min_fee } => {
+						InternalSwapMinimumNetworkFee::<T>::set(min_fee);
+						Self::deposit_event(Event::<T>::InternalSwapMinimumNetworkFeeSet {
+							min_fee,
+						});
+					},
```
dandanlen, line 1090: Maybe it's time to consolidate these events into a single one like we did in other PRs?

dandanlen (commented):
Looking at this and the previous network fee PR (fees per swap instead of per chunk), I think it might be worth considering a refactor of how we process fees. The logic is becoming a bit scattered. Some in the SwapRequestState, some in DcaState, some FeeType, some in storage... 

For example in `take_network_fee` we have a mixture of approaches:
- `MinFeePolicy` is passed as an argument but doesn't apply to all swap types (it's invalid for internal swaps).
- We read other information like fee amounts from storage during execution (makes testing awkward and opens up weird edge cases like what if the fee changes during swap execution?).


I also noticed we attached the fee information to the DCA state, it's not clear why this would be DCA-related. Maybe we can capture this information in the fee type (for example `NetworkFee { min: X, rate: Y }`) or add it to the `SwapState` when we break of a new DCA chunk instead of relying on reading the `SwapRequestState`?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2127,10 +2169,10 @@ pub mod pallet {
 					} else {
 						// No swap request found, so ignore fee collection history and do simple fee
```
kylezs, line 2170: What are the cases where we wouldn't have a swap request here?

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1616,7 +1616,12 @@ mod internal_swaps {
 
 		new_test_ext()
 			.execute_with(|| {
+				// Internal swaps use a different network fee minimum than the regular swaps
+				InternalSwapMinimumNetworkFee::<Test>::set(MIN_NETWORK_FEE);
+				// The refund fee is equal to the minimum network fee, so it needs to be set as
```
kylezs, line 1621: 👀 
Should it be? The argument for potentially having lower internal swap fees is around it not being as expensive to the network. If the refund is for an internal swap, then why wouldn't it pay the internal swap fee rate instead of the standard network fee rate?

msgmaxim: I was going to say the same. If we refund an internal swap, we should charge `InternalSwapMinimumNetworkFee`.

j4m1ef0rd: I didnt want to complicate things more, so i left it. But yeah I agree.

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: None at the moment. But wanted to handle it safely. But I can add a log or panic if you like?

msgmaxim:
To address some of your comments:

> We read other information like fee amounts from storage during execution.

Some information needs to be shared between chunks, and passing the information (like from previous chunk to the next doesn't seem like an obvious improvement. Currently all info stored on `SwapState` is chunk specific (might be worth renaming it).

> I also noticed we attached the fee information to the DCA state, it's not clear why this would be DCA-related.

This just happens to be where we store state shared between chunks. I don't think it is too inappropriate (considering that that all our swaps are technically DCA and DCA implies chunks), but we can rename it if it helps. Maybe a slightly better approach would be to store this on SwapRequestState, but it is an enum so this would need to be reworked, so I'm not sure if it is worth it at this stage (it would be if some of this info was needed for NetworkFee, for example, but it is not the case currently).

> what if the fee changes during swap execution?

There seems to be a test to check that this works as expected (the new fee will be enforced), but I agree that locking in these parameters on swap request would be more correct. I believe @j4m1ef0rd is looking into making this change.

dandanlen:
> To address some of your comments @dandanlen :
> 
> > We read other information like fee amounts from storage during execution.
> 
> Some information needs to be shared between chunks, and passing the information (like the total collected fee) previous chunk to the next doesn't seem like an obvious improvement. Currently all info stored on `SwapState` is chunk specific (might be worth renaming it).
> 
> > I also noticed we attached the fee information to the DCA state, it's not clear why this would be DCA-related.
> 
> This just happens to be where we store state shared between chunks. I don't think it is too inappropriate (considering that that all our swaps are technically DCA and DCA implies chunks), but we can rename it if it helps. Maybe a slightly better approach would be to store this on SwapRequestState, but it is an enum so this would need to be reworked, so I'm not sure if it is worth it at this stage (it would be if some of this info was needed for NetworkFee, for example, but it is not the case currently).
> 
> > what if the fee changes during swap execution?
> 
> There seems to be a test to check that this works as expected (the new fee will be enforced), but I agree that locking in these parameters on swap request would be more correct. I believe @j4m1ef0rd is looking into making this change.

Just to be clear, I wasn't suggesting all of this needs to be addressed in this PR. I just wanted to point out some code smells. I can see how it has evolved like this, but I think we're at a point where it's worth trying to consolidate things into a design that it bit easier to reason about. I do think some small changes like renaming or otherwise clarifying the purpose of some of the data types would be a step in the right direction.

state-chain/pallets/cf-swapping/src/lib.rs
kylezs: Yes, think that's better, currently it reads as though it's a potential case.

j4m1ef0rd:
>  I wasn't suggesting all of this needs to be addressed in this PR.

I have some good ideas for a refactor, but will do it in another PR. For now, can we merge this? or do you have some small changes in mind?

dandanlen:
> > I wasn't suggesting all of this needs to be addressed in this PR.
> 
> I have some good ideas for a refactor, but will do it in another PR. For now, can we merge this? or do you have some small changes in mind? @dandanlen

Will give it another pass and confirm. In general it would be nice if we refactor/rename things as we go, unless it's a major undertaking or implies migrations. But my suggestions above re. fees were intended more as a vague let's-think-about-this than a please-do-this. Sorry if that wasn't clear.

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -187,42 +197,47 @@ fn test_network_fee_calculation() {
 	}
 
 	new_test_ext().execute_with(|| {
-		const ID: Option<SwapRequestId> = Some(SWAP_REQUEST_ID);
+		const ID: SwapRequestId = SWAP_REQUEST_ID;
```
dandanlen, line 200: nit: if the ID is constant/irrelevant, why do we need to pass it as a parameter?

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -187,42 +197,47 @@ fn test_network_fee_calculation() {
...
 		// Normal network fee
-		assert_eq!(take_fees_from_swap(10, 20, 1000, 0, 0, ID), (900, 100));
-		assert_eq!(take_fees_from_swap(10, 20, 1000, 1000, 10_000, ID), (900, 100));
+		assert_eq!(take_fees_from_swap(10, 20, 1000, 0, 0, ID, false), (900, 100));
```
dandanlen, line 202: nit: Some named constants would make these tests more readable (also for example the expected values could be expressed in terms of the inputs)

For example:

```
assert_eq!(
	take_fees_from_swap(RATE, MIN, INPUT, 0, 0, ID, false),
	(INPUT - EXPECTED_FEE, EXPECTED_FEE)
);
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2099,18 +2090,29 @@ pub mod pallet {
...
+									output_action,
+									SwapOutputAction::CreditOnChain { .. }
+								) {
+									// Use alternate network fee for internal swaps
+									(
+										InternalSwapNetworkFee::<T>::get(),
+										InternalSwapMinimumNetworkFee::<T>::get(),
+									)
+								} else {
+									(T::NetworkFee::get(), MinimumNetworkFee::<T>::get())
```
dandanlen, line 2108: I think this can be bundled under 'future improvements' (but maybe worth doing now to avoid additional migrations)... 

Since we always access these together it might make sense to define the network fee as `NetworkFee { rate, minimum }` then we only need to access/update one thing.

dandanlen (approved):
Some final nits but I don't mind merging as is. 

I'll open an issue with some notes from this review and we can think about how to improve things.

dandanlen:
(Created issue PRO-2203 to track improvements)

state-chain/pallets/cf-swapping/src/tests/fees.rs
j4m1ef0rd: Left over from when the swap request was optional.

state-chain/pallets/cf-swapping/src/tests/fees.rs
j4m1ef0rd: personally  I like the raw numbers, but I understand how its hard to read. Filling it out with constants makes it so messy but a little easier to reason with.

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: Seems like a good idea. I think we should leave that for the refactor.

msgmaxim (approved):

state-chain/pallets/cf-swapping/src/tests/fees.rs
dandanlen: I think that where something is part of the repeated test 'furniture', selective use of constants can help make it clear to the reader what is and isn't relevant to each particular test case, and which number are expected to be linked and how.

PR #5794, opened 2025-04-09
refactor: store solana alt election results in unsynchronised state map

syan095:
# Pull Request

Closes: PRO-2167, PRO-2169

## Checklist

## Summary
ALT election results are now stored into the unsynchronised state map with Reference counting
ALT election results are now taken from the state map on ApiCall construction, managed with RC
Removed the expiry system from ALT witnessing elections
Added appropriate integration tests, and fixed a test with a TODO

### DIFF
--- a/state-chain/cf-integration-tests/src/solana.rs
+++ b/state-chain/cf-integration-tests/src/solana.rs
@@ -25,8 +25,8 @@ use cf_chains::{
 	ccm_checker::{CcmValidityError, DecodedCcmAdditionalData, VersionedSolanaCcmAdditionalData},
 	sol::{
 		api::{
-			AltConsensusResult, SolanaApi, SolanaEnvironment, SolanaTransactionBuildingError,
-			SolanaTransactionType,
+			AltWitnessingConsensusResult, SolanaApi, SolanaEnvironment,
+			SolanaTransactionBuildingError, SolanaTransactionType,
 		},
 		sol_tx_core::sol_test_values::{self, user_alt},
 		transaction_builder::SolanaTransactionBuilder,
@@ -182,7 +182,7 @@ fn schedule_deposit_to_swap(
 
 fn vote_for_alt_election(
 	election_identifier: u64,
-	res: AltConsensusResult<Vec<SolAddressLookupTableAccount>>,
+	res: AltWitnessingConsensusResult<Vec<SolAddressLookupTableAccount>>,
 ) {
 	let mut vote = BoundedBTreeMap::new();
 	vote.try_insert(
@@ -504,7 +504,7 @@ fn can_send_solana_ccm_v1() {
 			// Wait until swap is complete and ALT election started
 			vote_for_alt_election(
 				29,
-				AltConsensusResult::ValidConsensusAlts(vec![SolAddressLookupTableAccount {
+				AltWitnessingConsensusResult::ValidConsensus(vec![SolAddressLookupTableAccount {
 					key: user_alt().key,
 					addresses: vec![Default::default()],
 				}]),
@@ -586,7 +586,7 @@ fn ccms_can_contain_overlapping_alts() {
 			// Let election come into Consensus
 			vote_for_alt_election(
 				29,
-				AltConsensusResult::ValidConsensusAlts(vec![
+				AltWitnessingConsensusResult::ValidConsensus(vec![
 					SolAddressLookupTableAccount {
 						key: user_alts[0].into(),
 						addresses: vec![SolPubkey([0xE0; 32]), SolPubkey([0xE1; 32])],
@@ -599,7 +599,7 @@ fn ccms_can_contain_overlapping_alts() {
 			);
 			vote_for_alt_election(
 				30,
-				AltConsensusResult::ValidConsensusAlts(vec![
+				AltWitnessingConsensusResult::ValidConsensus(vec![
 					SolAddressLookupTableAccount {
 						key: user_alts[1].into(),
 						addresses: vec![SolPubkey([0xE2; 32]), SolPubkey([0xE3; 32])],
@@ -611,8 +611,6 @@ fn ccms_can_contain_overlapping_alts() {
 				]),
 			);
 
-			// TODO Fix this test once this problem is fixed.
-			// Currently only 1 ccm can be egressed, the other will hang forever.
 			testnet.move_forward_blocks(1);
 
 			System::assert_has_event(RuntimeEvent::SolanaIngressEgress(
@@ -1100,7 +1098,7 @@ fn invalid_alt_triggers_refund_transfer() {
 
 			testnet.move_forward_blocks(1);
 
-			vote_for_alt_election(13, AltConsensusResult::AltsInvalidNoConsensus);
+			vote_for_alt_election(13, AltWitnessingConsensusResult::InvalidNoConsensus);
 
 			// Let the election come to consensus.
 			testnet.move_forward_blocks(1);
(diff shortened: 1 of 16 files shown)


state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -227,11 +234,22 @@ impl
...
 		alts: AltConsensusResult<Vec<SolAddressLookupTableAccount>>,
 	) {
-		Environment::add_sol_ccm_swap_alts(alt_identifier.swap_request_id, alts);
-	}
-
-	fn should_expire_election(alt_identifier: SolanaAltWitnessingIdentifier) -> bool {
-		crate::System::block_number() >= alt_identifier.election_expiry_block_number
+		// use set_unsynchronised_state_map to write result onto the state map
+		match alts {
+			AltConsensusResult::ValidConsensusAlts(res) => res.into_iter().for_each(|r| {
```
kylezs, line 239: res implies result, but it's actually just lookup tables here
```suggestion
			AltConsensusResult::ValidConsensusAlts(lookup_tables) => lookup_tables.into_iter().for_each(|lookup_table| {
```

state-chain/chains/src/sol/api.rs
```diff
@@ -767,3 +767,30 @@ pub enum AltConsensusResult<T> {
 	ValidConsensusAlts(T),
 	AltsInvalidNoConsensus,
```
kylezs, line 768: Naming is inconsistent, we end one with Alts, and start one with Alts.... we don't even need to include "Alts" in the variants, it's on the type.
```suggestion
	ValidConsensus(T),
	InvalidNoConsensus,
```

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -705,20 +719,14 @@ pub(crate) fn initiate_solana_alt_election(alts: Vec<SolAddress>) {
...
 				SolanaAltWitnessing,
 				RunnerStorageAccess<Runtime, SolanaInstance>,
 			>,
-		>(SolanaAltWitnessingIdentifier {
-			swap_request_id: Default::default(),
-			alt_addresses: alts,
-			election_expiry_block_number: crate::System::block_number() +
-				EXPIRY_TIME_FOR_ALT_ELECTIONS,
-		})
+		>(SolanaAltWitnessingIdentifier { alt_addresses: alts })
```
kylezs, line 729: nit: naming

are they alt_addresses or are they alts?

state-chain/cf-integration-tests/src/solana.rs
```diff
@@ -428,35 +429,133 @@ fn can_send_solana_ccm_v1() {
...
+				},
+				SolAddressLookupTableAccount {
+					key: user_alts[2].into(),
+					addresses: vec![SolPubkey([0xE4; 32]), SolPubkey([0xE5; 32])],
 				}],
 			);
 
-			// 2 calls should be built - one for each CCM.
+			// TODO Fix this test once this problem is fixed.
+			// Currently only 1 ccm can be egressed, the other will hang forever.
```
kylezs, line 558: Why is this?

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -227,11 +234,22 @@ impl
...
 	) {
-		Environment::add_sol_ccm_swap_alts(alt_identifier.swap_request_id, alts);
-	}
-
-	fn should_expire_election(alt_identifier: SolanaAltWitnessingIdentifier) -> bool {
-		crate::System::block_number() >= alt_identifier.election_expiry_block_number
+		// use set_unsynchronised_state_map to write result onto the state map
+		match alts {
+			AltConsensusResult::ValidConsensusAlts(res) => res.into_iter().for_each(|r| {
+				SolanaAltWitnessingElectoralAccess::set_unsynchronised_state_map(
```
kylezs, line 240: There's a bug here (in combination with a bunch of other pieces).

1. Start 2 CCMs using the same ALTs in the same block.
2. The ALTs for each of these is witnessed and stored *against the same key*. 
3. Both are ready in the same X blocks. So they are both in the `ccms_to_send` in `do_egress_scheduled_ccm()` in the ingress-egress pallet.
4. We call new_unsigned for the first one, it succeeds, the ALTs are there.... then it deletes them.
5. We try to construct the second one. It fails. The ALTs have already been deleted. 

... We need to index them by something else... the swap request id perhaps 🙈

state-chain/runtime/src/chainflip.rs
```diff
@@ -620,10 +621,30 @@ impl ChainEnvironment<Vec<SolAddress>, AltConsensusResult<Vec<SolAddressLookupTa
...
+		alts: Vec<SolAddress>,
 	) -> Option<AltConsensusResult<Vec<SolAddressLookupTableAccount>>> {
-		// TODO: To be implemented with ALT election refactor
-		Some(AltConsensusResult::ValidConsensusAlts(vec![]))
+		let mut res = vec![];
+
+		// Verify all the data is ready
+		for alt in &alts {
+			if let Ok(Some(consensus)) =
+				solana_elections::SolanaAltWitnessingElectoralAccess::unsynchronised_state_map(alt)
```
dandanlen, line 631: I would be nice to not have to deal with elections pallet internals here. 

For example the `ExactValue` election could expose an api like this:

```
// This method already exists:
fn witness_exact_value(id: ES::Identifier);
// It should be possible to implement this, where ES::Result is the AltConsensusResult
fn check_value(id: ES::Identifier) -> ES::Result;
```

state-chain/runtime/src/chainflip/solana_elections.rs
```diff
@@ -227,11 +234,22 @@ impl
...
+					r.key.into(),
+					Some(AltConsensusResult::ValidConsensusAlts(r)),
+				)
+			}),
+			AltConsensusResult::AltsInvalidNoConsensus =>
+				alt_identifier.alt_addresses.into_iter().for_each(|alt| {
+					SolanaAltWitnessingElectoralAccess::set_unsynchronised_state_map(
```
dandanlen, line 247: Similar comment here, ideally we don't want to have to deal with pallet internals here. We should replace the hook with a storage write, and then provide a lookup function to 'pull' the data.

state-chain/runtime/src/chainflip/solana_elections.rs
dandanlen: As discussed - Roy and I have already talked about this and agreed to tackle it once everything else was in place. I'm not in favour of using the swap request id 😅 

The way I see it, there are three options:
- Reference counting: we store the results as `(count, result)` and increment/decrement as needed, only deleting when the count reaches zero. This should work feels a little too fine-grained for our needs (ie. making N writes/reads seems like a waste when we always want to read/write atomically anyway). 
- Index by `election_id`: I don't like the idea of needing to carry around additional state, but if we do it makes more sense for it to be the election_id that the swap_request_id. 
- Index by the election's ES::Identifier (in this case, `Vec<SolAddress>`). I feel like this might be a nice trade-off and works nicely with the existing abstractions around the ExactValue electoral system. 

Currently I'm leaning towards the last option...

kylezs: > Currently I'm leaning towards the last option...

Doesn't this have the same issue though? Two CCMs can use the same set of ALTs

kylezs (approved):
I don't think the reference counting bloats things too much. Looks good to me

state-chain/cf-integration-tests/src/solana.rs
syan095: I fixed this test after Ref Counting was added.
Good catch! I forgot to remove the comments

dandanlen (commented):
I added a draft PR with a refactor that IMO simplifies this a bit and also moves the state access into the pallet as mentioned above.

LMK what you think. 

#5803

dandanlen (approved):

PR #5796, opened 2025-04-10
feat: vault swap minimum broker fee

j4m1ef0rd:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

- New extrinsic that brokers can use to set a minimum broker fee for vault swaps.
   - Added an RPC to the broker API to use this new extrinsic.
- Added a check when encoding vault swap params that makes sure the broker fee is >= the new minimum.
    - Also added validation of the broker fees there, because it seemed like it was missing.
- Added logic to the processing of the vault swap witness that will enforce the new minimum, increasing the broker fee to the minimum if needed.
- Added test and benchmark


state-chain/runtime/src/lib.rs
```diff
@@ -2243,6 +2243,17 @@ impl_runtime_apis! {
...
+			// Validate broker fee
+			if broker_commission < pallet_cf_swapping::Pallet::<Runtime>::get_minimum_fee_for_broker(&broker_id) {
+				return Err(DispatchErrorWithMessage::from("Broker commission is too low"));
+			}
+			let beneficiaries = pallet_cf_swapping::Pallet::<Runtime>::assemble_broker_fees(
+				broker_id.clone(),
+				broker_commission,
+				affiliate_fees.clone(),
+			);
+			pallet_cf_swapping::Pallet::<Runtime>::validate_broker_fees(&beneficiaries)?;
```
dandanlen, line 2255: In both cases where we use `assemble_broker_fees`, we immediately `validate_broker_fees` afterwards. We can just make `assemble_broker_fees` return a Result and always validate inside that fn?

state-chain/traits/src/lib.rs
```diff
@@ -1061,6 +1061,7 @@ pub trait SwapLimitsProvider {
 	fn validate_broker_fees(
 		broker_fees: &Beneficiaries<Self::AccountId>,
 	) -> Result<(), DispatchError>;
+	fn get_minimum_fee_for_broker(broker_id: &Self::AccountId) -> BasisPoints;
```
dandanlen, line 1064:
```suggestion
	fn get_minimum_vault_swap_fee_for_broker(broker_id: &Self::AccountId) -> BasisPoints;
```

Since it shouldn't be applied to deposit channels, we should make this explicit in the name.

dandanlen: Also, a nit, but `SwapLimitsProvider` is not really a good name for this trait any more since we've slowly been adding various other bits and pieces.

Ether we rename this to something more general, eg. `SwapValidationParameters`, or we break it up into more specific traits.

dandanlen (commented):
Some minor nits but LGTM, thanks.

state-chain/pallets/cf-swapping/src/benchmarking.rs
```diff
@@ -255,5 +255,18 @@ mod benchmarks {
...
+	fn set_vault_swap_minimum_broker_fee() {
+		let caller = <T as Chainflip>::AccountRoleRegistry::whitelisted_caller_with_role(
+			AccountRole::Broker,
+		)
+		.unwrap();
+
+		#[extrinsic_call]
+		set_vault_swap_minimum_broker_fee(RawOrigin::Signed(caller.clone()), 100);
+
+		assert_eq!(VaultSwapMinimumBrokerFee::<T>::get(caller.clone()), 100);
```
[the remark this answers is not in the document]
j4m1ef0rd: but it's 0 before the extrinsic is called?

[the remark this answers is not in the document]
msgmaxim: Sure, I just think it is good form to make this explicit. Feel free to keep unchanged.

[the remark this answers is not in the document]
j4m1ef0rd: oh sorry, i read that wrong. yeah i will add it.

j4m1ef0rd:
Sorry, I accidentally rebased main instead of merge.

dandanlen (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2601,13 +2617,14 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			..
 		} = vault_deposit_witness.clone();
 
-		let Some(broker_fees) =
-			Self::assemble_broker_fees(broker_fee.clone(), affiliate_fees.clone())
-		else {
+		let Some(broker_fees) = Self::assemble_broker_fees(
+			broker_fee.map(|broker_fee| Self::enforce_broker_fee_minimum(broker_fee)),
```
martin-chainflip, line 2621: nit:
`broker_fee.map(Self::enforcer_broker_fee_minimum),`

state-chain/runtime/src/lib.rs
```diff
@@ -2357,6 +2357,16 @@ impl_runtime_apis! {
 				.try_into()
 				.map_err(|_| pallet_cf_swapping::Error::<Runtime>::BoostFeeTooHigh)?;
 
+			// Validate broker fee
+			if broker_commission < pallet_cf_swapping::Pallet::<Runtime>::get_minimum_vault_swap_fee_for_broker(&broker_id) {
+				return Err(DispatchErrorWithMessage::from("Broker commission is too low"));
+			}
+			let _beneficiaries = pallet_cf_swapping::Pallet::<Runtime>::assemble_and_validate_broker_fees(
```
martin-chainflip, line 2364: why should we filter here (remove beneficiaries that have a 0bps fee), but then still encode the swap including these? In the interest of smaller transaction size, we should use the filtered beneficiaries from here.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Actually, we should enforce this for the affiliate fees too, no?

PR #5799, opened 2025-04-10
feat: save all already processed events

marcellorigotti:
# Pull Request

Closes: PRO-2162
## Checklist

## Summary

In order to handle some edge cases we changed the approach:
- we now save all processed events until the safety margin expires, in case of duplicated we always save the highest expiry_block (block_number + safety margin)

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
@@ -11,16 +11,16 @@ use frame_support::{pallet_prelude::TypeInfo, Deserialize, Serialize};
 use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, marker::PhantomData, vec, vec::Vec};
 
 ///
-/// DepositChannelWitnessingProcessor
+/// BlockProcessor
 /// ===================================
 ///
-/// This processor is responsible for handling block data from a blockchain deposit channel while
+/// This processor is responsible for handling block data from a blockchain while
 /// managing reorganization events (reorgs) within a safety margin. It maintains an internal state
-/// of block data and reorg events, applies chain-specific processing rules (such as pre-witness and
-/// witness event generation), deduplicates events to avoid processing the same deposit twice, and
-/// finally executes those events.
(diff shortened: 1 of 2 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -203,57 +181,62 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
...
+						if *stored_expiry < expiry {
+							*stored_expiry = expiry;
+						}
+						false
+					},
+					None => {
+						self.processed_events.insert(event.clone(), expiry);
+						true
+					},
+				}
```
MxmUrw, line 231: It's a bit surprising that this function mutates state, because it feels like one might to want to call it without the mutation effect. If that's not true then I'm fine with keeping it this way, though I would try to update the name of the function to make clear that it is mutating / has side effects.

Thoughts @kylezs ?

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -203,57 +181,62 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
...
 		events
 			.into_iter()
-			.filter(|(_, last_event)| {
-				!self
-					.reorg_events
-					.iter()
-					.flat_map(|(_, (events, _))| events)
-					.any(|event| event == last_event)
+			.filter(|(block_number, event)| {
+				let expiry = block_number.saturating_forward(safety_margin as usize);
```
MxmUrw, line 219: Nice, storing the expiry directly makes things definitely more understandable!

Can we, instead of using `current_block_number + safety` as expiry, use `max{all blocks we know, "end of current range"}`. That would make it a bit more resistant to bad behaved reorgs. We could compute that number in `process_block_data` and pass it in to this function I think.

MxmUrw (commented):
Thanks, looks good, left some comments!

marcellorigotti:
Closing this one, will open it against the new branch `feat/block-witnesser-es-rebase` which is rebased against main

PR #5804, opened 2025-04-14
feat: ability to delay boost

msgmaxim:
# Pull Request

Closes: PRO-2212

## Summary

- Prewitnessed deposits are now delayed by `BoostDelayBlocks` blocks before being processed. If `BoostDelayBlocks` is 0 (the current default), they are processed immediately as before.
- If a deposit is finalised before it is "prewitness" is processed (e.g. the delay is too long), the deposit is processed as not boosted.
- `BoostDelayBlocks` is configurable by governance

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -2563,8 +2563,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		block_height: TargetChainBlockNumber<T, I>,
 		origin: DepositOrigin<T, I>,
 	) -> Result<FullWitnessDepositOutcome, DepositFailedReason> {
-		// Don't not reject deposit for any reason since boosters
-		// would be losing funds:
+		// Deposits can only fail or be rejected if we haven't already boosted.
 		if !matches!(boost_status, BoostStatus::Boosted { .. }) {
 			if deposit_amount < MinimumDeposit::<T, I>::get(asset) {
 				// If the deposit amount is below the minimum allowed, the deposit is ignored.
@@ -2649,7 +2648,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				used_pools: Vec<BoostPoolTier>,
 			},
 			PerformChannelAction {
-				should_cancel_pending_boost: bool,
+				deposit_outcome: FullWitnessDepositOutcome,
 			},
 		}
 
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1058,6 +1104,27 @@ pub mod pallet {
...
+					BoostStatusLookup::Channel { deposit_address } =>
+						DepositChannelLookup::<T, I>::get(deposit_address)
+							.map(|details| details.boost_status),
+				};
+
+				if boost_status == Some(BoostStatus::BoostPending) {
+					Self::process_prewitness_deposit_inner(deposit);
+				} else {
+					log_or_panic!("Invariant broken: attempt to process prewitness deposit with status: {:?} ", boost_status);
+				}
```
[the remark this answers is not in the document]
msgmaxim: I did consider this initially, but it seemed a bit sloppy not to clean up when we know it is no longer needed. If we keep it, we'd also need to consider rare edge cases like what if we prewitness again on the same channel after the first tx is finalised (which we do allow), and then this stale prewitnessed deposit will still be executed...

I know that `cancel_pending_prewitnessed_deposit` is slow, but keep in mind that it will almost never be called (we are talking about what I imagine will be a very rare edge case), and it is here just for correctness.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2098,7 +2102,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			let new_boost_status = if delay > Default::default() {
 				let process_at_block = frame_system::Pallet::<T>::block_number() + delay;
 				PendingPrewitnessedDeposits::<T, I>::append(process_at_block, deposit);
-				BoostStatus::BoostPending
+				BoostStatus::BoostPending { amount }
```
dandanlen, line 2119: Is the amount enough to distinguish this? 

Maybe it would be better to assign the `prewitness_deposit_id` and use that instead?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2539,6 +2545,8 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		block_height: TargetChainBlockNumber<T, I>,
 		origin: DepositOrigin<T, I>,
 	) -> Result<FullWitnessDepositOutcome, DepositFailedReason> {
+		// Don't not reject deposit for any reason since boosters
+		// would be losing funds:
```
dandanlen, line 2567: nit: this comment is a bit unclear

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: I think i get it: as long as the amounts match, we can consider them 'fungible' in a sense. We should think about how this works alongside the assumptions in tx screening: imagine there are two txs with matching amounts, but one is flagged for rejection. Does this create any problems?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2497,109 +2643,133 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		// Add the deposit to the balance.
 		T::DepositHandler::on_deposit_made(deposit_details.clone());
 
+		enum ActionToPerform {
+			FinaliseBoost {
+				prewitnessed_deposit_id: PrewitnessedDepositId,
+				used_pools: Vec<BoostPoolTier>,
+			},
+			PerformChannelAction {
+				should_cancel_pending_boost: bool,
```
dandanlen, line 2652: nit: we could assign the outcome here, no need for the intermediate boolean.

dandanlen (commented):
LGTM - I will resolve the minor issue mentioned below. Would be good to write a test for the mentioned edge case (two identical deposits, one rejected, one boosted).

dandanlen:
I didn't have time to finish the additional test - if you can, would be great...

dandanlen (approved):

PR #5806, opened 2025-04-14
fix: correctly clean up TransactionOutIdToBroadcastId storage

kylezs:
# Pull Request

Closes: PRO-2214

## Checklist

## Summary

- Use correct threshold signer types in broadcast mocks
- Refactor broadcast tests to use a more accurate transaction_out_id.
- Don't clean up storage when re-signing. If we clean it up there's an edge case that means we can successfully witness at the point of resigning, but this will be rejected, because the appropriate storage is not there. When we resign we will overwrite the data associated with the broadcast id later down the line anyway.
- Add failing test to demonstrate no clean up bug
- Introduce `BroadcastIdToTransactionOutIds` to properly clean up `TransactionOutIdToBroadastId` (use a Vec for append semantics), fixing the failing test

### DIFF
--- a/state-chain/pallets/cf-broadcast/src/lib.rs
+++ b/state-chain/pallets/cf-broadcast/src/lib.rs
@@ -911,7 +911,6 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		}
 	}
 
-	/// Start a broadcast by try to select a nominee.
 	fn start_broadcast_attempt(mut broadcast_data: BroadcastData<T, I>) {
 		let broadcast_id = broadcast_data.broadcast_id;
 		T::TransactionBuilder::refresh_unsigned_data(&mut broadcast_data.transaction_payload);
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -1098,8 +1097,9 @@ impl<T: Config<I>, I: 'static> Broadcaster<T::TargetChain> for Pallet<T, I> {
 		AbortedBroadcasts::<T, I>::mutate(|aborted| {
 			aborted.remove(&broadcast_id);
 		});
-		let mut api_call = Self::clean_up_broadcast_storage(broadcast_id)
-			.ok_or(Error::<T, I>::ApiCallUnavailable)?;
+
+		let mut api_call =
```
kylezs, line 1101: We don't delete here because it means that if we start to resign and then we succeed in witnessing it, the witness will not go through. We can leave the data here, and it will be overwritten at the appropriate time (on_signature_ready) if necessary.

syan095 (approved):
LGTM!

dandanlen (approved):

PR #5807, opened 2025-04-14
fix: lru cache fixes in the subscription gc

nmammeri:
# Pull Request

Closes: PRO-xxx

## Checklist

- [x] ~~I have written sufficient tests.~~ I have conducted manual tests

## Summary

* Minor fix that serializes Subscription ID and stores it as a String in the subscription gc
* Make sure to use the same chain head client for both the subscription and its GC.
* Removed debug messages to not pollute log file


dandanlen (approved):
Thanks for this 👌

PR #5815, opened 2025-04-22
refactor: revamp swap fees

j4m1ef0rd:
# Pull Request

part of: PRO-2203

## Checklist

## Summary

- Moved the 2 fee tracking amounts from the swap request to the chunk.
- Created a struct and functions to help track the network fee.
- The network fee (rate and minimum) is now "locked in" when the swap request is created, instead of getting the latest fee each chunk.
- Refactored the network fee tests, splitting them up a bit.
- Changed the network fee collected storage item to only write once per group swap.
- Removed duplication of broker fees.
- Removed the migrations for swapping pallet so i could add the new migration.

### TODO
- Change fee storage items to group rate and minimum (in separate PR)


state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -120,80 +111,220 @@ fn test_buy_back_flip() {
...
-					broker_fees: Default::default(),
-				},
-			},
-		);
+				);
+			}
+			// Normal swap with network fee
+			normal_swap(NORMAL_SWAP_AMOUNT);
+			// Normal swap that will be charged the minimum network fee
+			normal_swap(NORMAL_MINIMUM_NETWORK_FEE);
```
msgmaxim, line 167: It is a bit confusing that you use the fee itself as the amount. Might be useful to show that even swapping, say 5x the `NORMAL_SWAP_AMOUNT` would still result in NORMAL_SWAP_AMOUNT charged in fees.

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -120,80 +111,220 @@ fn test_buy_back_flip() {
...
-			MinimumNetworkFee::<Test>::set(0);
-			InternalSwapNetworkFee::<Test>::set(Permill::from_percent(network_fee_percent));
-			InternalSwapMinimumNetworkFee::<Test>::set(minimum_network_fee);
-		} else {
-			NetworkFee::set(Permill::from_percent(network_fee_percent));
-			MinimumNetworkFee::<Test>::set(minimum_network_fee);
-			InternalSwapNetworkFee::<Test>::set(Permill::from_percent(0));
-			InternalSwapMinimumNetworkFee::<Test>::set(0);
-		}
+fn swap_uses_correct_network_fee() {
```
msgmaxim, line 120: I think it might be worth splitting this into individual tests (grouping them into a module would help keeping them organised). There is not that much boilerplate (and you could create a `setup_fees` if needed). Right now it is hard to link initiated swaps to the corresponding events: e.g. it is not clear which event corresponds to `IngressEgressFee`; you can sort of guess that it is the one where min network fee is not applied, but right now the same field is used to effectively identify a swap and aslo check its behaviour which isn't ideal.

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -120,80 +111,220 @@ fn test_buy_back_flip() {
...
+		assert_eq!(
+			fee_tracker.take_fee(CHUNK_AMOUNT),
+			FeeTaken { remaining_amount: CHUNK_AMOUNT - MIN_NETWORK_FEE, fee: MIN_NETWORK_FEE }
+		);
+		// Second chunk gets partial network fee taken from it
+		assert_eq!(
+			fee_tracker.take_fee(CHUNK_AMOUNT),
+			FeeTaken {
+				remaining_amount: CHUNK_AMOUNT - (MIN_NETWORK_FEE - normal_fee_amount),
+				fee: MIN_NETWORK_FEE - normal_fee_amount
```
msgmaxim, line 282: This second fee should be the amount that would normally be charged from 2 chunks minus whatever you charged on the first chunk: `2 * normal_fee_amount - MIN_NETWORK_FEE`. The test currently only happens to work due to the parameters you chose, and would fail if you set `MIN_NETWORK_FEE` to 160, for example.

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -120,80 +111,220 @@ fn test_buy_back_flip() {
...
+#[test]
+fn test_network_fee_calculation() {
+	fn take_fees_from_swap(
+		network_fee_percent: u32,
+		minimum_network_fee: AssetAmount,
+		chunk_amount: AssetAmount,
+		network_fee_collected: AssetAmount,
+		accumulated_stable_amount: AssetAmount,
+	) -> (AssetAmount, AssetAmount) {
+		let FeeTaken { remaining_amount, fee } = NetworkFeeTracker::new_for_test(
```
msgmaxim, line 321: You might as well directly construct `NetworkFeeTracker` here and remove `new_for_test`. We have created test-only constructors in the past, but that was mainly to avoid making some private fields public (where it was important). Here you have access to those fields because you are in the child module.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2102,85 +2101,26 @@ pub mod pallet {
...
-							"Swap request {swap_request_id} not found, ignoring fee collection history for network fee calculation"
-						);
-						max(T::NetworkFee::get() * input, MinimumNetworkFee::<T>::get())
-					}
-				},
-				MinFeePolicy::NotEnforced => T::NetworkFee::get() * input,
 			};
 
-			if !fee.is_zero() {
+			if !swap_output.network_fee.is_zero() {
```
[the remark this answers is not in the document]
j4m1ef0rd: I did think about not adding it in there and just renaming the function to have "simulate" or something in it. But thought it would just be safer to just apply the network fee as usual in all places, just incase the function not used just for simulating in the future.

state-chain/pallets/cf-swapping/src/tests/fees.rs
j4m1ef0rd: I forgot that children can still use private fields 🤦‍♂️

state-chain/pallets/cf-swapping/src/lib.rs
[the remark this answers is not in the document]
msgmaxim: We should have tests (and we do) to check that the network fee is correctly collected where required. I know we sometimes like to write code defensively, but in this case the cost of making (a very unlikely) mistake is low (I imagine it would mean that the fee is not collected in some edge case), and it is better to keep the code clean.

msgmaxim (approved):

dandanlen (approved):

PR #5817, opened 2025-04-22
feat: decode vault swap rpc

syan095:
# Pull Request

Closes: PRO-2095

## Checklist

## Summary

Created a new RPC call + brokerAPI call that decodes Vaul Swap parameters encoded by us.

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to decode vault swap parameters for Bitcoin and Solana vault swaps via new RPC methods and APIs.
  - Added support for decoding Solana swap instructions into structured parameters, including affiliate and DCA details.
  - Exposed new data structures and conversions for vault swap inputs and extra parameters.
  - Added public constants and type aliases to improve cross-chain swap handling and instruction decoding.

- **Bug Fixes**
  - None.

- **Tests**
  - Added comprehensive unit tests for Solana swap decoding and instruction builder account validation.

- **Documentation**
  - Improved and clarified type aliases and conversion implementations for cross-chain assets and parameters.

- **Chores**
  - Minor code cleanup, trait derivations, and feature flag adjustments.

### DIFF
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -14,13 +14,12 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use cf_chains::{RefundParametersRpc, VaultSwapExtraParametersRpc, VaultSwapInputRpc};
 use cf_rpc_apis::{
 	broker::{
 		BrokerRpcApiServer, DcaParameters, GetOpenDepositChannelsQuery, SwapDepositAddress,
-		TransactionInId, WithdrawFeesDetail,
+		TransactionInId, VaultSwapExtraParametersRpc, VaultSwapInputRpc, WithdrawFeesDetail,
 	},
-	RpcApiError, RpcResult,
+	RefundParametersRpc, RpcApiError, RpcResult,
 };
 use cf_utilities::{
 	health::{self, HealthCheckOptions},
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -14,7 +14,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use cf_chains::{RefundParametersRpc, VaultSwapExtraParametersRpc};
+use cf_chains::{RefundParametersRpc, VaultSwapExtraParametersRpc, VaultSwapInputRpc};
 use cf_rpc_apis::{
 	broker::{
 		BrokerRpcApiServer, DcaParameters, GetOpenDepositChannelsQuery, SwapDepositAddress,
@@ -137,6 +137,21 @@ impl BrokerRpcApiServer for RpcServerImpl {
 				boost_fee,
 				affiliate_fees,
 				dca_parameters,
+				None,
+			)
+			.await?)
+	}
+
+	async fn decode_vault_swap_parameter(
+		&self,
+		vault_swap: VaultSwapDetails<AddressString>,
+	) -> RpcResult<VaultSwapInputRpc> {
+		Ok(self
+			.api
+			.raw_client()
+			.cf_decode_vault_swap_parameter(
+				self.api.state_chain_client.account_id(),
+				vault_swap,
 				None,
 			)
 			.await?)
--- a/state-chain/chains/src/cf_parameters.rs
+++ b/state-chain/chains/src/cf_parameters.rs
@@ -93,12 +93,33 @@ pub fn build_cf_parameters<C: Chain>(
 	}
 }
 
+pub fn decode_cf_parameters<C: Chain>(
+	data: Vec<u8>,
+) -> Result<VaultSwapParameters<C::ChainAccount>, &'static str> {
+	let VersionedCfParameters::V0(CfParameters { ccm_additional_data: _, vault_swap_parameters }) =
+		VersionedCfParameters::<C::ChainAccount>::decode(&mut &data[..])
+			.map_err(|_| "Failed to decode cf_parameter")?;
+
+	Ok(vault_swap_parameters)
+}
+
+pub fn decode_ccm_cf_parameters<C: Chain>(
+	data: Vec<u8>,
+) -> Result<(VaultSwapParameters<C::ChainAccount>, Vec<u8>), &'static str> {
+	let VersionedCcmCfParameters::V0(CfParameters { ccm_additional_data, vault_swap_parameters }) =
+		VersionedCcmCfParameters::<C::ChainAccount>::decode(&mut &data[..])
+			.map_err(|_| "Failed to decode cf_parameter")?;
+
+	Ok((vault_swap_parameters, ccm_additional_data.to_vec()))
+}
+
 #[cfg(test)]
 mod tests {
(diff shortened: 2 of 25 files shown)


state-chain/chains/src/cf_parameters.rs
```diff
@@ -93,19 +93,38 @@ pub fn build_cf_parameters<C: Chain>(
...
+pub fn decode_ccm_cf_parameters<C: Chain>(
+	data: Vec<u8>,
+) -> Result<(VaultSwapParameters<C::ChainAccount>, Vec<u8>), &'static str> {
+	let VersionedCcmCfParameters::V0(CfParameters { ccm_additional_data, vault_swap_parameters }) =
+		VersionedCcmCfParameters::<C::ChainAccount>::decode(&mut &data[..])
+			.map_err(|_| "Failed to decode cf_parameter")?;
+
+	Ok((vault_swap_parameters, ccm_additional_data.to_vec()))
+}
+
```
albert-llimos, line 115: There is a `decode_cf_parameters` that does exactly this in `engine/src/witness/evm/vault.rs`. We should use that instead.

state-chain/chains/src/lib.rs
```diff
@@ -1177,3 +1183,55 @@ impl VaultSwapExtraParametersRpc {
...
+			})
+			.expect("Address conversion is infallible")
+			.try_map_amounts(|n| Result::<NumberOrHex, ()>::Ok(n.into()))
+			.expect("Amount conversion is infallible")
+	}
+}
+
+#[derive(Clone, Debug, Encode, Decode, Serialize, Deserialize, TypeInfo)]
+pub struct VaultSwapInput<Address, Amount> {
+	pub broker_id: cf_primitives::AccountId,
```
albert-llimos, line 1202: This is called `broker` in `cf_request_swap_parameter_encoding`. Let's keep it consistent.

api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -142,6 +142,21 @@ impl BrokerRpcApiServer for RpcServerImpl {
 			.await?)
 	}
 
+	async fn decode_vault_swap_parameter(
+		&self,
+		vault_swap: VaultSwapDetails<AddressString>,
```
albert-llimos, line 147: The `VaultSwapDetails` are the same parameters that are to be passed when encoding a vault swap call in `cf_request_swap_parameter_encoding`, no? Could we not use the same approach in both?

state-chain/chains/src/cf_parameters.rs
syan095: The cf_chains won't have access to Engine, but Engine can use the cf_chain's fn instead.

state-chain/chains/src/sol.rs
```diff
@@ -495,10 +505,188 @@ pub struct SolApiEnvironment {
...
+	pub dst_address: crate::address::EncodedAddress,
+	pub dst_token: AnyChainAsset,
+	pub refund_parameters: ChannelRefundParameters<EncodedAddress>,
+	pub dca_parameters: Option<DcaParameters>,
+	pub boost_fee: u8,
+	pub broker_id: cf_primitives::AccountId,
+	pub broker_commission: BasisPoints,
+	pub affiliate_fees: Vec<AffiliateAndFee>,
+	pub ccm: Option<CcmChannelMetadata>,
+}
```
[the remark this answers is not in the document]
syan095: the input is slightly different for Swap Native and Swap Token - while we can use a generalized struct to describe the input, personally I would prefer the input of a function to be precise and provide exactly what is needed.

Also we cannot enfore variant of a enum the same way we enforce a Struct/Type. We will have to set the input type as the enum, then handle all variants of the enum:

```rust
pub enum Input {
A,
B,
C
}

fn test(a: Input) {
    if let Input::A = a {
        // do business logic here
    } else {
        // handle error for other variants
    }
}
```

it just doesn't feel very concise 

I don't think we need to worry so much about inconsistencies - if changes to one breaks another, it will be a compilation error and we would know immediately.

api/bin/chainflip-broker-api/src/main.rs
syan095: Are we ok with changing the interface for a existing RPC call?

state-chain/chains/src/cf_parameters.rs
syan095: Yep did just that

syan095: Yep did just that. I've moved the Engine's `decode_cf_parameters` logic to chains. But kept the original error logging

api/bin/chainflip-broker-api/src/main.rs
syan095: I think what we are thinking is to change the input of `cf_request_swap_parameter_encoding` to the new `VaultSwapInput` type

state-chain/chains/src/lib.rs
```diff
@@ -1177,3 +1183,53 @@ impl VaultSwapExtraParametersRpc {
...
+	pub extra_parameters: VaultSwapExtraParameters<Address, Amount>,
+	pub channel_metadata: Option<CcmChannelMetadata>,
+	pub boost_fee: BasisPoints,
+	pub affiliate_fees: Affiliates<cf_primitives::AccountId>,
+	pub dca_parameters: Option<DcaParameters>,
+}
+
+/// Type intended for RPC calls
+#[cfg(feature = "std")]
+pub type VaultSwapInputRpc = VaultSwapInput<AddressString, NumberOrHex>;
```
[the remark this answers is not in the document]
syan095: Got it. Moved the RPC specific types to the cf-rpc-types crate.
The conversion is a bit ugly though. I'll have to add a dedicated function instead of the From trait. (can't impl trait for types outside the crate the type is defined)

state-chain/runtime/src/chainflip/vault_swaps.rs
```diff
@@ -340,3 +366,81 @@ pub fn solana_vault_swap<A>(
...
+				chunk_interval,
+				boost_fee,
+				broker_fee,
+				affiliates,
+			},
+	} = UtxoEncodedData::decode(&mut &nulldata_payload[..])
+		.map_err(|_| "Failed to decode Bitcoin Null data Payload")?;
+
+	Ok(VaultSwapInputEncoded {
+		source_asset: None, // We lose source asset information in Vault swap
```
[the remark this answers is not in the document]
syan095: Oh yeah true!

api/bin/chainflip-broker-api/src/main.rs
albert-llimos: I've gotten to know now that some integrators are starting to use the broker api encoding directly (not SDK), for example for Solana Vault swaps, so I'm also a bit hesitant to introduce breaking changes now.

dandanlen: I added an issue for this https://linear.app/chainflip/issue/PRO-2235/add-bouncer-tests-for-vault-swap-decoding

dandanlen (approved):
Thanks @syan095 👌

PR #5818, opened 2025-04-22
fix: auto-sweep at the beginning of the next block

msgmaxim:
# Pull Request

Closes PRO-1660 (though not the main focus of the PR)

## Summary

- Order fills subscription relies on order not being swept (it derives executed amounts by "collecting" from existing orders), and recently added auto-sweeping broke this. This PR delays auto-sweeping until the beginning of the next block.
- Added an integration test to test order fills subscription (which caught this bug).
- Added `sold_amounts` to the range orders fills struct, which are simply derived from collcted fees (which we can do since the pool's fee in bps is known).


dandanlen (approved):

state-chain/custom-rpc/src/order_fills.rs
```diff
@@ -162,6 +165,9 @@ fn order_fills_for_pool<'a>(
 						base_asset: asset_pair.assets().base,
 						quote_asset: asset_pair.assets().quote,
 						id: id.into(),
+						sold_amounts: fees.map(|amount| {
```
niklasnatter, line 168: i think these are not the amounts the lp sold, but the ones that he bought, right?

niklasnatter: at least as far as i understand, the fees are earned in the asset that the lp buys. eg if somebody swaps 1 btc to 50000 usdc, the fees earned by the lp will be in btc

niklasnatter: for the product team it would be nice to expose the `sold_amounts`, because that is what we expose as volume filled by an lp. we could calculate that by using exact rate at which the assets were swapped for this fill (we estimate this rate in our code atm which makes the number slightly inaccurate). but im not sure if that rate is available anywhere

PR #5819, opened 2025-04-22
fix: use ci-benchmarks-release instead of ci-benchmarks

dandanlen:
This appears to have changed (?) since the last release.

Seems to work, based on [this run](https://github.com/chainflip-io/chainflip-backend/actions/runs/14596896538/job/40945101442).


dandanlen:
Question: what if we want to run benchmarks against the `ci-benchmarks-main` job?

vdoflip:
> Question: what if we want to run benchmarks against the `ci-benchmarks-main` job?

I'm missing quite a bit of context, but seems like the `profile`input parameter is not used, and maybe that should select between "main" and "release" benchmarks?

vdoflip (approved):

PR #5826, opened 2025-04-25
refactor: unify network fee storage

j4m1ef0rd:
# Pull Request

Closes: PRO-2203

## Checklist

## Summary

Changes both network fee's to be stored as a new struct that has fee rate and minimum inside it instead of storing them as separate items.
- Removed the constant `NetworkFee`
- Added 2 new storage items the store the regular and the internal swap network fee details.
- Added a migration
- Changed the config update to use the new structs. This also means that the regular network fee is now configurable.


state-chain/custom-rpc/src/lib.rs
```diff
@@ -1554,7 +1554,7 @@ where
 				maximum_swap_amounts: any::AssetMap::try_from_fn(|asset| {
 					api.cf_max_swap_amount(hash, asset).map(|option| option.map(Into::into))
 				})?,
-				network_fee_hundredth_pips: NetworkFee::get(),
+				network_fee_hundredth_pips: api.cf_network_fees(hash)?.regular_network_fee.rate,
```
j4m1ef0rd, line 1557: This `cf_swapping_environment` only has the regular network fee rate. Should we add the other network fee details? or maybe make another RPC just for network fee details? @dandanlen

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -423,14 +438,11 @@ pub enum PalletConfigUpdate<T: Config> {
...
-	/// Set the minimum fee in USDC paid per swap request
-	SetMinimumNetworkFee { min_fee: AssetAmount },
-	/// Set the network fee in USDC that will be used just for internal swaps (credit on-chain
-	/// swaps)
-	SetInternalSwapNetworkFee { rate: Permill },
-	/// Set the minimum network fee in USDC that will be used just for internal swaps (credit
-	/// on-chain swaps)
-	SetInternalSwapMinimumNetworkFee { min_fee: AssetAmount },
+	/// Set the network fee rate and minimum in USDC
+	SetNetworkFee { rate: Permill, minimum: AssetAmount },
```
msgmaxim, line 442: Might be annoying that you have to provide both values when you only want to change one (also comes with a risk of unintentionally changing the other). What if we made them optional and only apply the change for provided fields?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -629,20 +638,15 @@ pub mod pallet {
...
 	pub type BrokerBond<T: Config> = StorageValue<_, T::Amount, ValueQuery, DefaultBrokerBond<T>>;
 
-	/// Minimum network fee in USDC, charged per swap request. Only applies to regular swaps, i.e.
-	/// it excludes internal swaps like ingress/egress/network fees.
-	/// For DCA swaps, this fee will be taken from the first chunks until it is fulfilled.
-	#[pallet::storage]
-	pub type MinimumNetworkFee<T: Config> = StorageValue<_, AssetAmount, ValueQuery>;
-
-	/// The network fee in USDC used just for internal swaps (credit on-chain swaps).
+	/// Network fee rate and minimum in USDC, charged per swap request. Only applies to regular
```
msgmaxim, line 641: The commens needs to be updated. Only min fee does not apply to ingress/egress.

msgmaxim (approved):

state-chain/custom-rpc/src/lib.rs
dandanlen: Sorry for the late feedback here - I think doing both can't hurt. We can keep the new rpc and add a new field on the return value of this rpc, `network_fees`, containing all the output from `cf_network_fees` (and deprecate the old field). 

cc @acdibble

dandanlen (approved):

dandanlen:
I added a Linear issue to track documentation for this. 

https://linear.app/chainflip/issue/PRO-2252/document-changes-to-network-fees-in-cf-swapping-environment

PR #5827, opened 2025-04-25
feat: validator reform 1

Janislav:
# Pull Request

Closes: PRO-1817, PRO-2210

## Checklist

## Summary

In this PR, we add the possibility to move funds easily between accounts under the assumptions of the same restrictions. We do this to make it easier for validators the rebalance there nodes to improve a better auction dynamic.

Done so far:

- Added new extrinsic to trigger the internal transfer
- Added logic to move the account restrictions during the transfer
- Added logic to do the internal accounting directly without the overhead of the redemption accounting (withdrawal funds, hold it back in a reserve, etc)
- Added a new command to the chainflip-cli to trigger the transfer
- Added basic units tests
- Added benchmark

## Summary by CodeRabbit

- **New Features**
  - Introduced an internal transfer command in the CLI, enabling users to transfer FLIP tokens between accounts.
  - Added support for internal transfers on the state chain, with events and validation for restricted and unrestricted funds.
- **Bug Fixes**
  - Improved handling and validation of restricted balances during internal transfers.
- **Tests**
  - Added tests covering internal transfers, including scenarios with restricted funds and account binding restrictions.
- **Performance**
  - Defined execution weights for internal transfer operations to ensure predictable performance.
- **Refactor**
  - Modularized redemption logic for better maintainability.


Janislav:
@coderabbitai review

Janislav:
@coderabbitai review

Janislav:
@coderabbitai full review

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -769,6 +708,114 @@ impl<T: Config> Pallet<T> {
...
+		);
+		ensure!(
+			remaining_balance >=
+				restricted_balances.values().copied().sum::<FlipBalance<T>>() -
+					restricted_deficit,
+			Error::<T>::InsufficientUnrestrictedFunds
+		);
+
+		Ok((total_restricted_balance.saturating_sub(redemption_fee), redeem_amount))
+	}
```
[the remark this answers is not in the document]
Janislav: Maybe... But not in this PR...

state-chain/pallets/cf-funding/src/weights.rs
```diff
@@ -540,4 +549,12 @@ impl WeightInfo for () {
...
 			.saturating_add(ParityDbWeight::get().writes(1_u64))
 	}
+
+	fn internal_transfer() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `140`
+		//  Estimated: `3605`
+		// Minimum execution time: 13_287_000 picoseconds.
+		Weight::from_parts(13_695_000, 3605)
+	}
```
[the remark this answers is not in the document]
Janislav: Just a placeholder that gets overwritten by the real value when generate the benchmark.

state-chain/pallets/cf-funding/src/tests.rs
```diff
@@ -1835,3 +1835,139 @@ fn only_governance_can_update_settings() {
...
+			BOB,
+			RESTRICTED_ADDRESS,
+			AMOUNT_MINUS_FEE.into()
+		));
+
+		assert_eq!(RestrictedBalances::<Test>::get(ALICE).get(&RESTRICTED_ADDRESS), None);
+		assert_eq!(
+			RestrictedBalances::<Test>::get(BOB).get(&RESTRICTED_ADDRESS),
+			Some(&AMOUNT_MINUS_FEE)
+		);
```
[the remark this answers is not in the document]
Janislav: We test this where we define the transfer logic - in the Flip pallet

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +620,47 @@ pub mod pallet {
...
+			address: EthereumAddress,
+			amount: RedemptionAmount<FlipBalance<T>>,
+		) -> DispatchResult {
+			let source = ensure_signed(origin)?;
+
+			let address_is_restricted_for_account =
+				RestrictedBalances::<T>::get(&source).contains_key(&address);
+
+			let (_total_restricted_balance, redeem_amount) =
+				Self::try_redeem(amount, source.clone(), address, None)?;
```
dandanlen, line 644: In the normal redeem case, we also check if the source account should be removed (ie. if remaining balance is zero). We should check that in this case too.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +620,47 @@ pub mod pallet {
...
+
+			T::Flip::try_transfer_funds_internally(redeem_amount, &source, &account_id)?;
+
+			if address_is_restricted_for_account {
+				RestrictedBalances::<T>::mutate(account_id.clone(), |map| {
+					map.entry(address)
+						.and_modify(|balance| *balance += redeem_amount)
```
dandanlen, line 651: We are marking the full amount as restricted even though only some of it might be restricted. Should this not use the `total_restricted_balance`?

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -620,6 +622,20 @@ impl<T: Config> cf_traits::Funding for Pallet<T> {
...
+
+	fn try_transfer_funds_internally(
+		amount: Self::Balance,
+		from: &Self::AccountId,
+		to: &Self::AccountId,
+	) -> Result<(), DispatchError> {
+		ensure!(from != to, Error::<T>::CanNotTransferToSelf);
+		if let Some(from_imbalance) = Pallet::<T>::try_debit(from, amount) {
+			from_imbalance.offset(Pallet::<T>::credit(to, amount));
+			Ok(())
```
dandanlen, line 634: This should function (almost) exactly as if we called:
`try_initiate_redemption(from, amount)` -> `finalize_redemption(from)` -> `credit_funds(to, amount)`

So we need to call `T::OnAccountFunded::on_account_funded` for each account to trigger the callbacks for account balance changes.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -769,6 +713,114 @@ impl<T: Config> Pallet<T> {
...
+			Error::<T>::BelowMinimumFunding
+		);
+		ensure!(
+			remaining_balance >=
+				restricted_balances.values().copied().sum::<FlipBalance<T>>() -
+					restricted_deficit,
+			Error::<T>::InsufficientUnrestrictedFunds
+		);
+
+		Ok((total_restricted_balance.saturating_sub(redemption_fee), redeem_amount))
```
dandanlen, line 822: What is the purpose of returning `total_restricted_balance.saturating_sub(redemption_fee)` here? 

Couldn't we return the actual restricted amount:

```
min(
    total_restricted_balance.saturating_sub(redemption_fee),
    redeem_amount,
)
```

Then we can use this in the redeem extrinsic and also in the transfer extrinsic.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +620,47 @@ pub mod pallet {
...
+
+		/// Transfers funds from one account to another without the need of a full redemption.
+		/// The funds are getting transferred under the same conditions as a full redemption.
+		///
+		/// ## Events
+		///
+		/// - [InternalTransfer](Event::InternalTransfer)
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::internal_transfer())]
+		pub fn internal_transfer(
```
dandanlen, line 632: I think I remember discussing that we need to check that the source and destination have the same restrictions, including `BoundExecutorAddress` and `BoundRedeemAddress`. I can't see these checks anywhere.

api/bin/chainflip-cli/src/settings.rs
```diff
@@ -235,6 +235,15 @@ pub enum CliCommand {
...
 		epoch_index: Option<EpochIndex>,
 	},
+	#[clap(about = "Request an internal transfer of FLIP to another account")]
+	RequestInternalTransfer {
+		#[clap(help = "The amount to transfer")]
+		amount: Option<f64>,
+		#[clap(help = "The Ethereum address under which condition we transfer the FLIP")]
+		destination: String,
+		#[clap(help = "The account ID to transfer the FLIP to")]
+		account_id: String,
```
dandanlen, line 245: Similar comment to above: naming could be better / more descriptive. For example, `destination` is a confusing name for a thing that is not the destination...

Would it make sense to extend the existing `redeem` method instead of adding an entirely new one? 

`chainflip-cli redeem --exact=1000 $ETH_ADDRESS --internal-destination=$ACCOUNT_ID`

api/bin/chainflip-cli/src/main.rs
```diff
@@ -266,6 +269,50 @@ async fn request_redemption(
 	Ok(())
 }
 
+async fn request_internal_transfer(
```
dandanlen, line 272: This has a lot in common with `request_redemption`, seems like there should be plenty of scope to share code.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +620,47 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			let source = ensure_signed(origin)?;
+
+			let address_is_restricted_for_account =
+				RestrictedBalances::<T>::get(&source).contains_key(&address);
+
+			let (_total_restricted_balance, redeem_amount) =
+				Self::try_redeem(amount, source.clone(), address, None)?;
+
+			T::Flip::try_transfer_funds_internally(redeem_amount, &source, &account_id)?;
```
dandanlen, line 646: In the `funded` extrinsic we also check if the account exists, and create it if not. But in this extrinsic we don't check that amount and we don't check the destination. 

We factored out the redemption logic from `redeem`, why not do the same with `funded`? This would ensure that the same checks are applied in both cases.

state-chain/pallets/cf-funding/src/tests.rs
```diff
@@ -1835,3 +1835,139 @@ fn only_governance_can_update_settings() {
...
+			Funding::internal_transfer(
+				OriginTrait::signed(ALICE),
+				BOB,
+				RESTRICTED_ADDRESS,
+				AMOUNT_MINUS_FEE.into()
+			),
+			Error::<Test>::AccountBindingRestrictionViolated
+		);
+	});
+}
```
dandanlen, line 1973: A nice test might be to assert that doing a redeem, then funding using the redeemed amount results in the same outcome as using the `internal_transfer` extrinsic.

dandanlen (commented):
Looks like coderabbit missed a few things... 

One other thing - I think it might make sense to charge no redemption fee if it's a rebalance operation... Will discuss internally.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: Mhhh... No, we restrict what we transfer to the other account to the same address under a different account. It's pretty much the same as we would:

1. Redeem 50 of 100 FLIP from account A to restricted address 0x1
2. Now we stake 50 FLIP to account B under the same restriction to address 0x1
3. We end up in a situation where we have 50 FLIP in account A and 50 FLIP in account B both restricted to 0x1

** For simplicity, I didn't take the fees into account.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: Should be all in `try_redeem` If I'am not mistaken.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: Good point

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +620,47 @@ pub mod pallet {
...
+		///
+		/// ## Events
+		///
+		/// - [InternalTransfer](Event::InternalTransfer)
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::internal_transfer())]
+		pub fn internal_transfer(
+			origin: OriginFor<T>,
+			account_id: AccountId<T>,
+			address: EthereumAddress,
```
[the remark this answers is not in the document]
Janislav: I think because we ones discussed it like that 😅. But I agree that for unrestricted funds, we don't need one.

state-chain/pallets/cf-funding/src/lib.rs
dandanlen: When redeeming to a single address on Ethereum, there is no other way to do it. But when transferring internally, we can keep the amount equal - why not do this?

state-chain/pallets/cf-funding/src/lib.rs
dandanlen: I can't see it. `try_redeem` doesn't know what the destination is, so how can it check this?

dandanlen: It should be simple enough to check this with a unit test.

Janislav: The checks are in try_redeem, and they're also getting performed against the BoundRedeemAddress for the passed address. For the executor address we pass None right now, because we said we can ignore that, but actually I am not so sure about that anymore 🤔. I think we have to ensure that if there is an executor address registered for the source account, it has to be the same in the destination account.

Janislav: *and there is a test for this: ensure_bonded_address_condition_holds_during_internal_transfer

dandanlen: > I think we have to ensure that if there is an executor address registered for the source account, it has to be the same in the destination account.

Yes this is what I mean. 

We need to check:

`BoundRedeemAddress(origin) == BoundRedeemAddress(destination)`

and 

`BoundExecutorAddress(origin) == BoundRedeemAddress(destination)`.


The test just checks the address that is passed as an argument. It doesn't check that source and destination match.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: We could do this, but then we should also use the existing redemption code from the Flip pallet and accounting the transfer directly.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: I am not sure if I understand correctly 🤔. I think If we move funds under the assumptions of address restrictions, we have to do it the same as we would over a redemption and a new stake to another account.

The other possibly is we only move liquid funds without an address restriction, but then we don't care about this anyway.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +624,89 @@ pub mod pallet {
...
+					T::AccountRoleRegistry::has_account_role(
+						&account_id,
+						AccountRole::Validator
+					),
+				Error::<T>::RestrictedToValidators
+			);
+
+			Self::ensure_address_restrictions(&source, &account_id)?;
+
+			let address_is_restricted_for_account = if let Some(address) = address {
```
dandanlen, line 662: Why do we need to define a bool variable here which is only conditionally used ~30 lines below?

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -719,6 +624,89 @@ pub mod pallet {
...
+		///
+		/// - [RestrictedToValidators](Error::RestrictedToValidators)
+		/// - [ExecutorBindingRestrictionViolated](Error::ExecutorBindingRestrictionViolated)
+		/// - [AccountBindingRestrictionViolated](Error::AccountBindingRestrictionViolated)
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::rebalance())]
+		pub fn rebalance(
+			origin: OriginFor<T>,
+			account_id: AccountId<T>,
+			address: Option<EthereumAddress>,
```
dandanlen, line 646: argument names are still very vague

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -621,44 +625,86 @@ pub mod pallet {
...
+			ensure!(
+				T::AccountRoleRegistry::has_account_role(&source, AccountRole::Validator) &&
+					T::AccountRoleRegistry::has_account_role(
+						&account_id,
+						AccountRole::Validator
+					),
+				Error::<T>::RestrictedToValidators
+			);
+
+			Self::ensure_address_restrictions(&source, &account_id)?;
```
dandanlen, line 660: Not sure why this needs to be factored out into a function. I can see that there's a unit test for it - but the function alone is trivial. We should test that rebalancing between two accounts with different restrictions isn't possible, not because we are concerned that it might not work (it's a simple check), but in case someone in the future makes a change that accidentally breaks or removes this restriction.

api/bin/chainflip-cli/src/settings.rs
```diff
@@ -235,13 +235,15 @@ pub enum CliCommand {
...
+	Rebalance {
 		#[clap(help = "The amount to transfer")]
 		amount: Option<f64>,
-		#[clap(help = "The Ethereum address under which condition we transfer the FLIP")]
-		destination: String,
-		#[clap(help = "The account ID to transfer the FLIP to")]
+		#[clap(
+			help = "An optional Ethereum address under which restriction conditions we transfer the FLIP"
+		)]
+		destination: Option<String>,
```
dandanlen, line 245: As mentioned before, `destination` is confusing here.

state-chain/pallets/cf-funding/src/lib.rs
Janislav: But then we end up with a bigger chunk of logic that is harder to test and reason about. We would need to write a whole unit test with a way more complicated setup than just having this separate.

In general, I don't understand what would be bad about having a few more abstractions and breaking the code into smaller chunks, even if this would end in single-use functions.

It would make the code:

- Easier to understand and reason about when the abstractions are good and there is maybe even a doc comment for every function 
- Easier and better to unit test - since the smaller the unit of code you test, the easier to find edge cases and come up with good tests
- Maybe even easier to refactor in some cases, since the code is already a bit split up

Of course, it always depends on the situation and having a different function for every 5 lines of code doesn't make sense either, but we are definitely went the other extreme, and that's also not ideal...

By the way, I did exactly what you proposed in the first attempt and then end up writing super complex unit tests to test an actual simple thing...

state-chain/pallets/cf-funding/src/lib.rs
Janislav: Because there is a case where the address is getting removed from the source account when it was restricted. We have to remember this to be able to enforce the restrictions to the new account after the execution of the `try_reddem` function.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -769,6 +759,156 @@ impl<T: Config> Pallet<T> {
...
+		use cf_chains::Get;
+		use cf_traits::RedemptionCheck;
+
+		ensure!(T::SafeMode::get().redeem_enabled, Error::<T>::RedeemDisabled);
+
+		// Not allowed to redeem if we are an active bidder in the auction phase
+		T::RedemptionChecker::ensure_can_redeem(&account_id)?;
+
+		// The redemption must be executed before a new one can be requested.
+		ensure!(!PendingRedemptions::<T>::contains_key(&account_id), Error::<T>::PendingRedemption);
```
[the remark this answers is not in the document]
Janislav: I don't know... I don't think it hurts, and it is the safer option :D. Feels like there could be a hidden edge cases

state-chain/pallets/cf-funding/src/lib.rs
dandanlen: We return the restricted amount from `try_redeem` - instead of ignoring it, we could just use this (`if restricted_balance > 0 { .. }`) instead of duplicating that logic here.

state-chain/pallets/cf-funding/src/lib.rs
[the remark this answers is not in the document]
dandanlen: It's pretty annoying and unintuitive: if there's a pending redemption to Ethereum, you can't rebalance? 

The Ethereum contract can only handle one redemption at a time - this is why we need to check this for redemptions. But for rebalances, I don't think it's necessary.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -769,6 +759,156 @@ impl<T: Config> Pallet<T> {
...
+			);
+			ensure!(
+				remaining_balance >=
+					restricted_balances.values().copied().sum::<FlipBalance<T>>() -
+						restricted_deficit,
+				Error::<T>::InsufficientUnrestrictedFunds
+			);
+
+			Ok((total_restricted_balance.saturating_sub(redemption_fee), redeem_amount))
+		} else {
```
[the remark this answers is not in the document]
Janislav: I don't think it's that many:

1) No, we don't. It's the total amount of all restrictions of the account, that dosent change dosent matter if we're operating on a restricted address or not
2) The tax is getting ignored because we don't have network cost in this case what is the main reason for the tax
3) I think the only check is missing is the one for the MinimumFunding, all the others are address related

I think what we should do is restoring this function to what it was before we handle that special case with an Option<Address> and handle the case we're only redeem liquid funds separately. It's easier to understand, and we don't have to touch too much of the existing logic with the possibility of breaking something.

PR #5828, opened 2025-04-25
feat: log the received swap_deposit_address in the broker API

marcellorigotti:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

log the received swap_deposit_address in the broker API

### DIFF
--- a/api/bin/chainflip-broker-api/src/main.rs
+++ b/api/bin/chainflip-broker-api/src/main.rs
@@ -43,12 +43,16 @@ use futures::{stream, FutureExt, StreamExt};
 use jsonrpsee::{core::async_trait, server::ServerBuilder, PendingSubscriptionSink};
 use std::{
 	path::PathBuf,
-	sync::{atomic::AtomicBool, Arc},
+	sync::{
+		atomic::{AtomicBool, AtomicU32, Ordering},
+		Arc,
+	},
 };
 use tracing::log;
 
 pub struct RpcServerImpl {
 	api: StateChainApi,
+	request_counter: AtomicU32,
 }
 
 impl RpcServerImpl {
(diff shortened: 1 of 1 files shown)


api/bin/chainflip-broker-api/src/main.rs
```diff
@@ -86,7 +86,7 @@ impl BrokerRpcApiServer for RpcServerImpl {
 		refund_parameters: RefundParametersRpc,
 		dca_parameters: Option<DcaParameters>,
 	) -> RpcResult<SwapDepositAddress> {
-		Ok(self
+		let swap_deposit_address = self
```
acdibble, line 89: would be nice to log some of the input parameters too, this will give us a starting timestamp and make it maybe easier to correlate the logs.

ideally we would have some unique request id here to make it very simple

dandanlen:
Can I ask what this is for? 

I think it's already possible to log this (albeit not in the same format of course) through jsonrpsee.

dandanlen:
> Can I ask what this is for?
> 
> I think it's already possible to log this (albeit not in the same format of course) through jsonrpsee.

Otherwise, for example, I think it's possible to add middleware, and we could eg. optionally log out every request/response.

marcellorigotti:
> Can I ask what this is for?
> 
> I think it's already possible to log this (albeit not in the same format of course) through jsonrpsee.

Well debugging this took quite a while, mostly because the logs are quite misleading IMO (we always log `Request found in finalized block with hash .....`, even if we wait only for the extrinsic to be included in a block.
Logging this would have save us a lot of time. 

To add a bit of context I kind of remembered that there was a possibility of a bug with reorgs but that specific log line made me disregard that option (that's until your input on the matter and me checking the submission watcher code again to realize what was happening)

dandanlen:
Sorry - getting back a bit late. 

I guess my point is that this is a very specific solution to a general problem (ie. we're adding logging for a case that we had in the past, but we might want to log other things in the future). Adding this log output doesn't really solve any immediate issue - we already figured out what the problem was. So IMO better to concentrate on the larger issue (how to get the logs that we might need in future). 

Or, for example, we could log a warning if the output changed between the initial response and finality.

PR #5836, opened 2025-04-29
feat: limit order auto expire

j4m1ef0rd:
# Pull Request

Closes: PRO-1734

## Checklist

## Summary

- Added `dispatch_at` field to the `set_limit_order` and `update_limit_order` extrinsics.
- Added an `expire_at` field to `set_limit_order`, Allowing an LP to automatically schedule a cancel order at the desired block.
- Changed the storage item for scheduling limit order updates to a struct.
- Added a migration to the new struct.
- Refactored how the limit order updates are scheduled and dispatched.
- Changed the `schedule_limit_order_update` Call to use the new struct instead of a call.

TODO: Update docs


CumpsD:
This is a nice addition, simplifies LP code :)

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -986,6 +997,52 @@ pub mod pallet {
...
+					None => Ok::<(), DispatchError>(()),
+					Some(previous_tick) => {
+						Self::inner_update_limit_order_at_tick(
+							pool,
+							lp,
+							asset_pair,
+							side,
+							id,
+							previous_tick,
+							crate::pallet::IncreaseOrDecrease::Decrease(Amount::MAX),
```
msgmaxim, line 1036: Nit, but `IncreaseOrDecrease` and `NoOpStatus` should already be in scope, so you don't need the full path.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1402,3 +1420,156 @@ fn test_sweeping_when_updating_range_order() {
...
+			assert_eq!(
+				LiquidityPools::pool_orders(ASSET, STABLE_ASSET, Some(ALICE), false)
+					.unwrap()
+					.limit_orders
+					.asks
+					.len(),
+				1
+			);
+		})
+		.then_execute_at_block(EXPIRE_AT, |_| {
```
msgmaxim, line 1490: Probably better to use `then_process_blocks_until_block` since it is closer to what happens in practice (i.e. we don't skip blocks).

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -757,7 +759,15 @@ pub mod pallet {
 				id,
 				option_tick,
 				sell_amount,
-			)
+			)?;
+
+			if let Some(expire_at) = expire_at {
+				let call = Call::cancel_limit_order { base_asset, quote_asset, side, id };
+				Self::schedule_limit_order_update(origin, Box::new(call), expire_at)
```
dandanlen, line 766: This function checks if the expiry_at block is greater or equal than the current block. But for this particular case, it doesn't make much sense: If expire_at is equal to the current block, we would create the order and then immediately remove it. 

We should add a check to ensure that expire_at is greater than the current block.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -986,6 +997,52 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Cancels a limit order.
+		/// Main difference with `update_limit_order` is that this will not error if the order does
+		/// not exist. This is used for auto-expiring orders that may of already been swept and
+		/// removed, without spamming failure events.
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::cancel_limit_order())]
+		pub fn cancel_limit_order(
```
dandanlen, line 1007: If this is the same as update_limit_order, why not just implement it via update_limit_order, ignoring the result?

api/lib/src/lp.rs
```diff
@@ -362,8 +358,14 @@ pub trait LpApi: SignedExtrinsicApi + Sized + Send + Sync + 'static {
 		option_tick: Option<Tick>,
 		sell_amount: AssetAmount,
 		dispatch_at: Option<BlockNumber>,
+		expire_at: Option<BlockNumber>,
 		wait_for: WaitFor,
 	) -> Result<ApiWaitForResult<Vec<LimitOrder>>> {
+		if let (Some(dispatch_at), Some(expire_at)) = (dispatch_at, expire_at) {
+			if dispatch_at >= expire_at {
```
dandanlen, line 365: We could enforce this on the protocol to avoid any weird edge cases.

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Btw. I might make sense to break `schedule_limit_order_update` out into a pallet method instead of calling the extrinsic directly, otherwise we're duplicating things like the origin check.

state-chain/custom-rpc/src/lp.rs
```diff
@@ -413,8 +413,14 @@ where
...
 		sell_amount: NumberOrHex,
 		dispatch_at: Option<BlockNumber>,
+		expire_at: Option<BlockNumber>,
 		wait_for: Option<WaitFor>,
 	) -> RpcResult<ApiWaitForResult<Vec<LimitOrder>>> {
+		if let (Some(dispatch_at), Some(expire_at)) = (dispatch_at, expire_at) {
+			if dispatch_at >= expire_at {
+				Err(anyhow!("Invalid expire_at, must be larger than dispatch_at"))?;
+			}
+		};
```
dandanlen, line 423: Putting this check in the runtime would also avoid having to duplicated it here..

dandanlen (commented):
Looking good, some minor comments.

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -241,6 +241,7 @@ impl LpRpcApiServer for RpcServerImpl {
 		tick: Option<Tick>,
```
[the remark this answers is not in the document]
j4m1ef0rd: I didnt want to add it to the `update` because the user might expect it to update the expire block. This would involve looking up the existing cancel order and removing it before adding the new one. Not simple or cheap to do, so I avoided it for now.

state-chain/pallets/cf-pools/src/lib.rs
j4m1ef0rd: Because the `on_initialize` does not see the difference between a auto-cancel call and a scheduled update, so it will deposit the `ScheduledLimitOrderUpdateDispatchFailure`event. Having a seperate call was the easiest solution.

dandanlen: Yeah I get that. I don't see having a few extra instances of this event as a big downside really - might as well have the visibility of this. Not a strong opinion though, it's fine as it is.

j4m1ef0rd:
Realised that adding a field to the `set_limit_order` call would probably need a migration since we store the call on chain for schedule queue. Dropped it back into draft while work a refactor to decouple the call from the storage item and write a migration.
@dandanlen

j4m1ef0rd:
I have done a large refactor to address a few issues:
- (main issue) Decoupled the storage item from the Call.
- No need to check that the type of call is correct when storing/using it.
- Deduplicate the dispatching and scheduling of order updates.
- No need for the `cancel_limit_order` call and we still keep the feature of not depositing an event when an automatic cancel does not work because the order was already closed.
- Add a migration for the storage item. A lot easier now that it's a struct.
- Deduplicate the schedule or dispatch logic from the api's.

> Would it be easier to just add a new call and deprecate the old one?

That would still leave the main issue. It would work at not needing a migration, but I think it's much nicer to just use a struct. It's more standard and less prone to mistakes, like I almost made in this PR.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -837,12 +931,17 @@ pub mod pallet {
...
+			side: Side,
+			id: OrderId,
+			details: LimitOrderUpdateDetails<BlockNumberFor<T>>,
+			dispatch_at: Option<BlockNumberFor<T>>,
+		) -> DispatchResult {
 			let lp = T::AccountRoleRegistry::ensure_liquidity_provider(origin)?;
 
-			Self::schedule_or_dispatch_limit_order_update(lp, call, dispatch_at)
+			LimitOrderUpdate::<T> { lp: lp.clone(), base_asset, quote_asset, side, id, details }
+				.schedule_or_dispatch(dispatch_at)
```
msgmaxim, line 944: If this is the only place where `schedule_or_dispatch` is called I think it might make sense to inline it? (If we had more code in `schedule_limit_order_update`, it would probably be justified to split it up, but now it looks like there isn't much benefit to having a separate function).

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -837,12 +931,17 @@ pub mod pallet {
...
 		pub fn schedule_limit_order_update(
 			origin: OriginFor<T>,
-			call: Box<Call<T>>,
-			dispatch_at: BlockNumberFor<T>,
-		) -> DispatchResultWithPostInfo {
+			base_asset: any::Asset,
+			quote_asset: any::Asset,
+			side: Side,
+			id: OrderId,
+			details: LimitOrderUpdateDetails<BlockNumberFor<T>>,
```
msgmaxim, line 938: Is the assumption here that `schedule_limit_order_update` isn't documented, and thus breaking compatibility isn'ta problem since it is only used internally in lp-api (which you also updated)?

state-chain/pallets/cf-pools/src/migrations/scheduled_limit_order_updates.rs
```diff
@@ -0,0 +1,195 @@
...
+	#[derive(Clone, DebugNoBound, PartialEq, Eq, Encode, Decode, TypeInfo)]
+	struct OldStorage<T: Config> {
+		pub lp: T::AccountId,
+		pub id: OrderId,
+		pub call: crate::Call<T>,
+	}
+
+	new_test_ext().execute_with(|| {
+		// Create a call (The update_limit_order has not changed, so we can use it for the test)
+		let call = crate::Call::<Test>::update_limit_order {
```
msgmaxim, line 163: might be worth testing `set_limit_order` order too? (Just to show that we can decode it correctly too)

msgmaxim (approved):

state-chain/pallets/cf-pools/src/lib.rs
msgmaxim: Also (as discussed), if `schedule_limit_order_update` isn't used externally, maybe we could just remove it and instead add distpatch_at field to `set_limit_order` and `update_limit_order` directly (as long is it is done in a backwards compatible way)? Wdyt @dandanlen?

state-chain/pallets/cf-pools/src/migrations/scheduled_limit_order_updates.rs
j4m1ef0rd: As discussed, Not easy to do because I added a field to `set_limit_order`.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -179,32 +180,144 @@ impl<T> AskBidMap<T> {
...
+			order_id: self.id,
+			dispatch_at,
+		});
+		ScheduledLimitOrderUpdates::<T>::append(dispatch_at, self);
+	}
+
+	pub fn schedule_or_dispatch(self, dispatch_at: Option<BlockNumberFor<T>>) -> DispatchResult {
+		if let Some(dispatch_at) = dispatch_at {
+			let current_block_number = frame_system::Pallet::<T>::block_number();
+			ensure!(dispatch_at >= current_block_number, Error::<T>::LimitOrderUpdateExpired);
```
j4m1ef0rd, line 282: I do wonder if we should put a cap on how far in the future someone can schedule an update for. Because they could potentially schedule a lot of junk updates for a million years in the future just to fill up space on the chain. @dandanlen @msgmaxim

dandanlen: There are currently no scheduled updated on mainnet, but I agree it makes sense to limit the duration. I don't think it needs to be configurable, I doubt it's something that needs major fine-tuning, so a constant should be enough. What order of magnitude seems reasonable? 6 hours?

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Yes I think I agree.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -837,12 +931,17 @@ pub mod pallet {
...
 			origin: OriginFor<T>,
-			call: Box<Call<T>>,
-			dispatch_at: BlockNumberFor<T>,
-		) -> DispatchResultWithPostInfo {
+			base_asset: any::Asset,
+			quote_asset: any::Asset,
+			side: Side,
+			id: OrderId,
+			details: LimitOrderUpdateDetails<BlockNumberFor<T>>,
+			dispatch_at: Option<BlockNumberFor<T>>,
```
dandanlen: Why is this optional?

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Any addition to the call parameters will break the SCALE encoding. The only way to make it fully backwards-compatible would be to add a new extrinsic and deprecate the old one. If we assume that everyone uses the LP API then we can assume it's a non-breaking change - probably a fair assumption? 

I do agree that adding a single param to the existing calls is probably better that a separate call (if nothing else, it makes the feature more visible).

dandanlen (commented):
I agree using a struct here is much nicer 👌

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -847,32 +864,47 @@ pub mod pallet {
...
+					side,
+					id,
+					option_tick,
+					sell_amount,
+				)?;
+
+				if let Some(close_order_at) = close_order_at {
+					if close_order_at <= frame_system::Pallet::<T>::block_number() {
+						return Err(Error::<T>::InvalidCloseOrderAt.into());
+					}
```
dandanlen, line 895: nit: `ensure` is a bit more idiomatic.

dandanlen (approved):

dandanlen:
LGTM but can we please add a limit as discussed [here](https://github.com/chainflip-io/chainflip-backend/pull/5836#discussion_r2084428951). Can be in this PR or in a follow-up. 

We also need to update documentation.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -464,9 +465,25 @@ fn cant_schedule_in_the_past() {
...
 				IncreaseOrDecrease::Decrease(55),
-				Some(9) // 1 block in the past
+				// 1 block in the past
+				Some(CURRENT_BLOCK - 1)
 			),
-			Error::<Test>::LimitOrderUpdateExpired
+			Error::<Test>::InvalidDispatchAt
+		);
+
+		assert_noop!(
```
msgmaxim, line 474: Would be nice to test that for `CURRENT_BLOCK + SCHEDULE_UPDATE_LIMIT_BLOCKS` this does work.

dandanlen (approved):

PR #5840, opened 2025-04-29
chore: rename sold_amounts -> bought_amounts

msgmaxim:
# Pull Request

## Summary

As @niklasnatter correctly pointed out, the amount we are exposing here is the *bought* amount, thus renaming the field.


dandanlen (approved):

dandanlen:
FYI I removed the non-breaking label and cherry-picked to release manually.

PR #5842, opened 2025-04-30
feat: protect against priority fee spikes

kylezs:
# Pull Request

Closes: PRO-2237

## Checklist

## Summary

We saw a large ETH priority fee spike. We can protect against these single block spikes by taking the last two blocks of fees. Taking the highest base fee, and the lowest priority fee of the two.

### DIFF
--- a/engine/src/witness/eth/chain_tracking.rs
+++ b/engine/src/witness/eth/chain_tracking.rs
@@ -30,8 +30,8 @@ impl<T: EvmRetryRpcApi + Send + Sync + Clone> GetTrackedData<cf_chains::Ethereum
 		&self,
 		header: &Header<<cf_chains::Ethereum as cf_chains::Chain>::ChainBlockNumber, H256, Bloom>,
 	) -> Result<<cf_chains::Ethereum as cf_chains::Chain>::TrackedData, anyhow::Error> {
-		// We take the highest base fee (which is limited to a 12.5% increase per block) to ensure
-		// we don't estimate too low. Then we take the lowest priority fee, which is not limited,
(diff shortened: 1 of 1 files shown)


engine/src/witness/eth/chain_tracking.rs
```diff
@@ -31,16 +30,20 @@ impl<T: EvmRetryRpcApi + Send + Sync + Clone> GetTrackedData<cf_chains::Ethereum
 		&self,
 		header: &Header<<cf_chains::Ethereum as cf_chains::Chain>::ChainBlockNumber, H256, Bloom>,
 	) -> Result<<cf_chains::Ethereum as cf_chains::Chain>::TrackedData, anyhow::Error> {
-		const PRIORITY_FEE_PERCENTILE: f64 = 50.0;
+		// We take the highest base fee (which is limited to a 12.5% increase per block) to ensure
+		// we don't estimate too low. Then we take the lowest priority fee, which is not limited,
```
dandanlen, line 34: nit: Since base fee is deterministic we could just take the latest base fee otherwise there will be some (very rare) cases where we overpay when fees go down.

dandanlen (approved):

albert-llimos (approved):

engine/src/witness/eth/chain_tracking.rs
kylezs: Yeah, possible. Thinking was that at least if we take the lower priority fee, but actually the priority was higher, we'll at least have increased the base fee. I don't think it will matter too much either way, but using the latest might just be a little better overall if we go with the assumption that the latest is most likely to be close to the next (rather than the one before latest).

PR #5845, opened 2025-05-02
feat: upgrade Solana vault swaps

albert-llimos:
# Pull Request

Closes: PRO-2236

## Checklist

## Summary

This update is to match the new Solana Swap Endpoint program in https://github.com/chainflip-io/chainflip-sol-contracts/pull/93 .

Besides the Vault swap logic itself, I have also added the capability to upgrade on chain programs since we will need that to upgrade the currently deployed Swap Endpoint program.

There is a breaking change in the `request_swap_parameter_encoding` rpc replacing the  `event_data_account: Address` parameter  for a  `seed: Vec<u8>`  parameter.
We could make the seed optional and generate a random seed ourselves if the seed is not passed. However, it adds quite a bit of verbosity on the types and all the conversions (`VaultSwapExtraParametersRpc`, `VaultSwapExtraParameters`, `VaultSwapExtraParametersEncoded`) that might not be worth it, integrators/SDK can generate a seed on the fly very easily.


state-chain/chains/src/lib.rs
```diff
@@ -1086,7 +1086,7 @@ pub enum VaultSwapExtraParameters<Address, Amount> {
 	Arbitrum(EvmVaultSwapExtraParameters<Address, Amount>),
 	Solana {
 		from: Address,
-		event_data_account: Address,
+		seed: Vec<u8>,
```
dandanlen, line 1089: Actually - since this exposed in the rpc, can we make sure it serializes/deserializes as hex?

dandanlen: And another question - is it ok that this Vec is unbounded?

state-chain/chains/src/sol/api.rs
```diff
@@ -583,6 +591,40 @@ impl<Environment: SolanaEnvironment> SolanaApi<Environment> {
...
+			sol_api_environment.vault_program,
+			sol_api_environment.vault_program_data_account,
+			// Assumed that the agg_key is the gov_key in the on-chain programs.
+			// This assumption is valid until we change the key to some independent Governance key.
+			agg_key,
+			spill_address,
+			durable_nonce,
+			compute_price,
+			vec![sol_api_environment.address_lookup_table_account],
+		)?;
```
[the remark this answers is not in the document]
albert-llimos: Technically it could if we failed to derive a pda account which can't happen because we  pass accounts that have a derived pda, or if the built transaction is too long which can't happen for this transaction.

state-chain/runtime/src/chainflip/vault_swaps.rs
```diff
@@ -413,7 +422,7 @@ pub fn decode_solana_vault_swap(
 		amount,
 		src_asset,
 		src_address,
-		event_data_account,
+		event_data_account: _,
```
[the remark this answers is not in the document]
albert-llimos: Yes, will do that now.

state-chain/chains/src/lib.rs
albert-llimos: Good point. I have now made it bounded with the seeds maximum length.
I also made it serialize/deserialize as a hex and updated the bouncer accordingly.

dandanlen (approved):
LGTM 🚀

PR #5854, opened 2025-05-08
feat: Make transaction building transactional

Janislav:
# Pull Request

Closes: PRO-2238

## Checklist

## Summary

- Modifies the pattern of building a transaction to introduce a default function that automatically wraps the actual call functionality into a transactional wrapper
- This pattern applies only if the `new_unsigned` returns a Result type, since otherwise the call can not fail
- Added a test to verify the general storage rollback

### DIFF
--- a/state-chain/chains/src/lib.rs
+++ b/state-chain/chains/src/lib.rs
@@ -33,6 +33,7 @@ use crate::{
 };
 use ccm_checker::{CcmValidityChecker, CcmValidityError, DecodedCcmAdditionalData};
 use core::{fmt::Display, iter::Step};
+use frame_support::storage::transactional;
 use sol::api::VaultSwapAccountAndSender;
 use sp_std::marker::PhantomData;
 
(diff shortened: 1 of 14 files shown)


state-chain/chains/src/lib.rs
```diff
@@ -634,9 +718,18 @@ pub enum TransferFallbackError {
 	Unsupported,
 	/// Failed to lookup the given token address, so the asset is invalid.
 	CannotLookupTokenAddress,
+	/// Some other DispatchError occurred.
+	DispatchError(DispatchError),
 }
+
 pub trait TransferFallback<C: Chain>: ApiCall<C::ChainCrypto> {
-	fn new_unsigned(transfer_param: TransferAssetParams<C>) -> Result<Self, TransferFallbackError>;
+	fn new_unsigned(transfer_param: TransferAssetParams<C>) -> Result<Self, TransferFallbackError> {
```
dandanlen, line 726: It would be good to have doc comments on these default methods mentioning that they should not be overridden, and the `_impl` method should be implemented instead.

state-chain/chains/src/lib.rs
```diff
@@ -485,13 +486,27 @@ pub trait ChainEnvironment<
 pub enum SetAggKeyWithAggKeyError {
 	Failed,
 	FinalTransactionExceededMaxLength,
+	DispatchError(DispatchError),
+}
+
+#[derive(RuntimeDebug, Clone, PartialEq, Eq)]
+pub enum SetGovKeyWithAggKeyError {
+	Failed,
```
dandanlen, line 494: `Failed` is not very descriptive (obviously when there's an error something failed). Can we do better?

dandanlen (approved):
Thanks 👌

PR #5858, opened 2025-05-12
feat: minor bouncer test improvements

dandanlen:
Some changes to the AllSwaps bouncer tests:

Runs AllSwaps as a vitest test suite where each swap is an individual test with its own timeout. This means we get feedback on all tests that fail, not just `AllSwaps timed out`. 

We run vitest with maxConcurrency=100 to ensure we run lots of concurrent tests (default it 10).


dandanlen:
The above test run failed, which provides a nice example of the output after this change: 

https://github.com/chainflip-io/chainflip-backend/actions/runs/14971382448/job/42053283238

(I think it's just a flaky test, will re-run)

kylezs (approved):

bouncer/tests/all_swaps.ts
```diff
@@ -126,9 +126,10 @@ export function testAllSwaps(timeoutPerSwap: number) {
         context.logger.info(`🧪 Private broker channel opened`);
       },
       120,
+      'AllSwaps',
     );
     for (const swap of allSwaps) {
-      concurrentTest(swap.name, swap.test, timeoutPerSwap);
+      concurrentTest(swap.name, swap.test, timeoutPerSwap, 'AllSwaps');
```
dandanlen, line 132: Will this not cause the info to be overwritten? 

(btw. What is the purpose of `writeTestInfoFile`?)

j4m1ef0rd: It is just a tmp file that is used by the `run_test` command to link the function name with the `vitest` name. So the run test command can look at the function names in a given file and find the `vitest` name to run the test in `vitest`. Saves us having to look up the test name each time. Works great and I use it a lot.

Yeah it skips if it exists. Looks like my fix is broker. my bad, let me try again.

PR #5861, opened 2025-05-15
feat: Feature/pro 2246

marcellorigotti:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

### DIFF
--- a/engine/src/witness/btc.rs
+++ b/engine/src/witness/btc.rs
@@ -33,7 +33,7 @@ use pallet_cf_elections::{
 		block_height_tracking::{
 			primitives::Header, state_machine::InputHeaders, HWTypes, HeightWitnesserProperties,
 		},
-		block_witnesser::state_machine::BWElectionProperties,
+		block_witnesser::state_machine::{BWElectionProperties, BWElectionType},
 	},
 	VoteOf,
 };
@@ -60,6 +60,7 @@ use crate::{
 };
 use anyhow::Result;
 
+use pallet_cf_elections::electoral_systems::block_witnesser::state_machine::BWTypes;
 use sp_core::H256;
 use state_chain_runtime::chainflip::bitcoin_elections::{
 	BitcoinEgressWitnessingES, BitcoinFeeTracking, BitcoinVaultDepositWitnessingES,
@@ -90,7 +91,7 @@ impl VoterApi<BitcoinBlockHeightTrackingES> for BitcoinBlockHeightTrackingVoter 
 		let mut headers = VecDeque::new();
 
 		let header_from_btc_header =
-			|header: BlockHeader| -> anyhow::Result<Header<btc::Hash, btc::BlockNumber>> {
+			|header: BlockHeader| -> anyhow::Result<Header<TypesFor<BitcoinBlockHeightTracking>>> {
 				Ok(Header {
 					block_height: header.height,
 					hash: header.hash.to_byte_array().into(),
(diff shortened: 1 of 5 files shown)


engine/src/witness/btc.rs
```diff
@@ -274,7 +318,7 @@ impl VoterApi<BitcoinEgressWitnessingES> for BitcoinEgressWitnessingVoter {
 			tracing::info!("Witnesses from BTCE: {:?}", witnesses);
 		}
 
-		Ok(Some(witnesses))
+		Ok(Some((witnesses, response_block_hash.map(|hash| H256::from_slice(hash.as_ref())))))
```
MxmUrw, line 321: Previously we didn't transfer the hash between statechain <-> engine at all! Still a bit inconvenient that we have to do this now...

state-chain/pallets/cf-elections/src/electoral_systems/block_height_tracking.rs
```diff
@@ -72,6 +72,26 @@ pub enum ChainProgress<T: ChainTypes> {
...
+		range: RangeInclusive<T::ChainBlockNumber>,
+	) -> Self {
+		ChainProgress::Reorg(map, range)
+	}
+
+	pub fn none() -> Self {
+		ChainProgress::None
+	}
+}
+
```
MxmUrw, line 94: What's the reason for these constructor functions, it looks like they do the same as the enum variants?

state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -319,6 +335,10 @@ impls! {
 		}
 	}
 
+	Hook<HookTypeFor<Self, ElectionTrackerEventHook>> {
+		fn run(&mut self, _: <HookTypeFor<TypesFor<BitcoinVaultDepositWitnessing>, ElectionTrackerEventHook> as HookType>::Input) -> () {}
+	}
```
MxmUrw, line 340: Instead of implementing an empty `run()` function for the "debug hooks", we can use the `EmptyHook` struct. That is, in the `BWProcessorTypes` we can do
```rust
		type ElectionTrackerEventHook = EmptyHook;
```

engine/src/witness/btc.rs
```diff
@@ -249,23 +279,37 @@ impl VoterApi<BitcoinEgressWitnessingES> for BitcoinEgressWitnessingVoter {
...
+			match election_type {
+				BWElectionType::Optimistic => {
+					let block_hash = self.client.block_hash(block).await?;
+					let block = self.client.block(block_hash).await?;
+					response_block_hash = Some(block.header.hash);
+					txs.extend(block.txdata);
+				},
+				BWElectionType::ByHash(hash) => {
+					let block = self.client.block(bitcoin::BlockHash::hash(hash.as_ref())).await?;
+					response_block_hash = Some(block.header.hash);
```
MxmUrw, line 303: In case of a `ByHash` election we actually don't want to return a hash, right?

engine/src/witness/btc.rs
```diff
@@ -249,23 +279,37 @@ impl VoterApi<BitcoinEgressWitnessingES> for BitcoinEgressWitnessingVoter {
...
-			let block = self.client.block(block_hash).await?;
-
-			txs.extend(block.txdata);
+			match election_type {
+				BWElectionType::Optimistic => {
+					let block_hash = self.client.block_hash(block).await?;
+					let block = self.client.block(block_hash).await?;
+					response_block_hash = Some(block.header.hash);
+					txs.extend(block.txdata);
+				},
```
MxmUrw, line 300: It feels like we can improve the logical flow here a bit, we're currently matching on the `election_type` in every iteration of the loop, even though the election type cannot change between iterations. Also related: it would be nicer if we didn't have to have a mutable `respone_block_hash` variable... This currently only work correctly for the assumption that we only ever iterate once, I think.

MxmUrw:
Perfect, let's merge this!

MxmUrw (approved):

PR #5865, opened 2025-05-19
feat: Custom network for any asset

j4m1ef0rd:
# Pull Request

Closes: PRO-2262 & PRO-2255

## Checklist

## Summary

You can set a custom network fee rate (not minimum) for any asset. The swap with look at both the input and output assets network fee rate (custom or standard) and choose the highest.
So to get the desired behaviour of lower fees for internal stable coin swaps, we can send a pallet update gov extrinsic with 
``` ts
[SetInternalSwapNetworkFeeForAsset { asset:"Usdc", rate: 500 } ,
SetInternalSwapNetworkFeeForAsset { asset:"Usdt", rate: 500 } ,
// etc, for all stables
...]
```

- Added 2 new storage items that map a network fee rate to an asset. One for regular and one for internal.
- Added to the update pallet config to be able to set and remove them.
   - Changed the limit on how many updates can be put in a single extrinsic from 10 -> 20. Makes it easier for the unit test.
- Added 2 tests to cover the new behaviour

### DIFF
--- a/state-chain/pallets/cf-swapping/src/tests/fees.rs
+++ b/state-chain/pallets/cf-swapping/src/tests/fees.rs
@@ -1045,3 +1045,152 @@ fn gas_calculation_can_handle_extreme_swap_rate() {
 		);
 	});
 }
+
+#[test]
+fn test_get_network_fee() {
+	const REGULAR_NETWORK_FEE: u32 = 5;
+	const INTERNAL_SWAP_NETWORK_FEE: u32 = 6;
+	const MINIMUM_NETWORK_FEE: AssetAmount = 123;
+
+	fn test_get_fee(
+		input_asset_fee: (Asset, Option<u32>),
+		output_asset_fee: (Asset, Option<u32>),
+		is_internal: bool,
+		expected_fee: u32,
+	) {
+		new_test_ext().execute_with(|| {
+			// Set the standard network fee
+			if is_internal {
+				InternalSwapNetworkFee::<Test>::set(FeeRateAndMinimum {
+					rate: Permill::from_percent(INTERNAL_SWAP_NETWORK_FEE),
+					minimum: MINIMUM_NETWORK_FEE,
+				});
+			} else {
+				NetworkFee::<Test>::set(FeeRateAndMinimum {
+					rate: Permill::from_percent(REGULAR_NETWORK_FEE),
+					minimum: MINIMUM_NETWORK_FEE,
+				});
+			}
+
+			// Set the custom network fees for the assets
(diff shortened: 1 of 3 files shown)


state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -1045,3 +1045,152 @@ fn gas_calculation_can_handle_extreme_swap_rate() {
...
+		test_get_fee(
+			(Asset::Flip, None),
+			(Asset::Eth, Some(network_fee - 2)),
+			is_internal,
+			network_fee,
+		);
+
+		// Test that the highest fee is always used
+		test_get_fee((Asset::Flip, Some(20)), (Asset::Eth, Some(25)), is_internal, 25);
+		test_get_fee((Asset::Flip, None), (Asset::Eth, Some(25)), is_internal, 25);
```
msgmaxim, line 1132: As discussed, it is confusing that in this case we expect 25, while in the case above (where one of the fees is also missing) we expect the default fee. Migth we worth adding a comment to explain why that is.

msgmaxim (approved):

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -378,10 +378,16 @@ pub struct TradingStrategyLimits {
...
+pub struct NetworkFeeDetails {
+	pub standard_rate_and_minimum: FeeRateAndMinimum,
+	pub rates: AssetMap<Permill>,
+}
+
 #[derive(Encode, Decode, TypeInfo, Serialize, Deserialize, Clone)]
 pub struct NetworkFees {
-	pub regular_network_fee: FeeRateAndMinimum,
-	pub internal_swap_network_fee: FeeRateAndMinimum,
+	pub regular_network_fee: NetworkFeeDetails,
```
msgmaxim, line 389: This looks like a breaking change. Might be OK, but we could also consider making this backwards-compatible by adding two extra fields here with `AssetMap<Permill>` instead.

j4m1ef0rd: I think this RPC is not in use yet because it was only just merged. So breaking change is fine.

dandanlen (approved):

acdibble (commented):
In `cf_swap_rate_v3`, we need some sort of switch to indicate if it's an internal swap so the correct fees are used and accurate quotes can be provided

dandanlen:
> In `cf_swap_rate_v3`, we need some sort of switch to indicate if it's an internal swap so the correct fees are used and accurate quotes can be provided

Issue opened: PRO-2281

PR #5866, opened 2025-05-19
refactor: solana alt auction uses a BTreeSet instead of Vec as key

syan095:
# Pull Request

Closes: PRO-2261

## Checklist

## Summary

Refactored the Solana ALT auction's key to a BTreeSet.
This orders and dedups the alt addresses


engine/src/witness/sol.rs
```diff
@@ -228,12 +228,15 @@ impl VoterApi<SolanaAltWitnessing> for SolanaAltWitnessingVoter {
...
 	) -> Result<Option<VoteOf<SolanaAltWitnessing>>, anyhow::Error> {
-		lookup_table_witnessing::get_lookup_table_state(&self.client, alt_witnessing_identifier.0)
-			.await
-			// We wrap the vote in a Some here since the vote is always valid if there was no error
-			// in rpc while querying. This is so we come to consensus on "None" if the lookup
-			// table is not found.
-			.map(Some)
+		lookup_table_witnessing::get_lookup_table_state(
+			&self.client,
+			alt_witnessing_identifier.0.into_iter().collect(),
```
kylezs, line 233: better to pass in the BTreeSet imo, that way it's clear when looking at `fn get_lookup_table_state` that it receives a set without dupes

dandanlen (approved):

PR #5867, opened 2025-05-19
fix: return correct pool fee (zero)

dandanlen:
# Pull Request

Closes: PRO-2248

TODO: check if this is the correct long-term solution.


dandanlen:
@msgmaxim i seem to remember this wasn't actually required? I can't seem to find the message any more...

PR #5878, opened 2025-05-22
chore: add insta test for all rpc call return types

syan095:
# Pull Request

Closes: PRO-2024

## Checklist

## Summary
Added insta test for all rpc call return types


syan095:
The return type of `all_open_deposit_channels` has been changed from a tuple to a proper struct. This will impact Product. Who should I get permission/notify from the product team? @dandanlen

syan095:
> @syan095 Is there a particular reason for the change (apart from the fact that explicit structs are better)?

The main reason for this change is that we want to have regression tests for return types; having a `struct` means we can test against this return type when it changes. 
With tuples, the return type can change without making the regression test fail, therefore defeating the purpose.
Also having structs with named fields is nice :)

msgmaxim:
> > @syan095 Is there a particular reason for the change (apart from the fact that explicit structs are better)?
> 
> The main reason for this change is that we want to have regression tests for return types; having a `struct` means we can test against this return type when it changes. With tuples, the return type can change without making the regression test fail, therefore defeating the purpose. Also having structs with named fields is nice :)

I believe a tuple and a struct with the same fields (in the same order) should have idential scale encoding, no? I don't think this would be a breaking change then.

api/cf-rpc-apis/Cargo.toml
```diff
@@ -18,5 +18,6 @@ jsonrpsee = { workspace = true, features = ["full"] }
 cf-chains = { workspace = true, default-features = true }
 cf-primitives = { workspace = true, default-features = true }
 cf-rpc-types = { workspace = true }
+state-chain-runtime = { workspace = true }
```
dandanlen, line 21: I think one of the goals of the rpc-apis crate is to avoid dependencies on the runtime... can we avoid it?

cc @nmammeri

state-chain/custom-rpc/src/snapshots/custom_rpc__tests__chain_accounts_serialization.snap
```diff
@@ -0,0 +1,172 @@
...
+        2
+      ]
+    },
+    {
+      "Btc": [
+        1,
+        2,
+        3,
+        4
+      ]
```
dandanlen, line 73: Is this really how we serialize a bitcoin address?

dandanlen (commented):
Thanks Roy - only one comment re. the location of the new type definition. 

Having said that - I don't think changing the type to a struct is really needed. I agree it's objectively better, however IMO outside the scope of this PR. @msgmaxim is right in that a type alias or newtype wrapper would allow us to write tests without breaking the API. I think we should prefer that.

dandanlen (approved):

PR #5885, opened 2025-05-26
chore: ci to upload logs on failure

j4m1ef0rd:
## Summary

Fixes a mistake that stopped the ci from uploading logs when the bouncer fails 🤦‍♂️


msgmaxim:
We'll merge this since it seems reasonable and @j4m1ef0rd wants to start using it for debugging, but maybe either @ahasna  or @anton-chain could double check post merge?

msgmaxim (approved):

PR #5896, opened 2025-05-28
fix: deposit channels finality bug fix

nmammeri:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

### DIFF
--- a/state-chain/custom-rpc/src/broker.rs
+++ b/state-chain/custom-rpc/src/broker.rs
@@ -16,7 +16,7 @@
 
 use crate::{
 	backend::{CustomRpcBackend, NotificationBehaviour},
-	pool_client::SignedPoolClient,
+	pool_client::{is_transaction_status_error, PoolClientError, SignedPoolClient},
 	CfApiError,
 };
 pub use cf_chains::eth::Address as EthereumAddress;
@@ -24,7 +24,7 @@ use cf_chains::{address::AddressString, CcmChannelMetadataUnchecked, ChannelRefu
 use cf_node_client::{
 	extract_from_first_matching_event, subxt_state_chain_config::cf_static_runtime, ExtrinsicData,
 };
-use cf_primitives::{Affiliates, Asset, BasisPoints, ChannelId};
+use cf_primitives::{Affiliates, Asset, BasisPoints, ChannelId, ForeignChain};
 use cf_rpc_apis::{
 	broker::{
 		try_into_refund_parameters_encoded, try_into_swap_extra_params_encoded,
@@ -34,6 +34,7 @@ use cf_rpc_apis::{
 	},
 	RefundParametersRpc, RpcResult, H256,
 };
+use futures::StreamExt;
 use jsonrpsee::{core::async_trait, PendingSubscriptionSink};
 use pallet_cf_swapping::AffiliateDetails;
 use sc_client_api::{
@@ -41,6 +42,7 @@ use sc_client_api::{
 	HeaderBackend, StorageProvider,
 };
 use sc_transaction_pool::FullPool;
+use sc_transaction_pool_api::{TransactionStatus, TxIndex};
 use sp_api::CallApiAt;
 use sp_core::crypto::AccountId32;
 use sp_runtime::traits::Block as BlockT;
@@ -48,7 +50,7 @@ use state_chain_runtime::{
 	runtime_apis::{
 		ChainAccounts, ChannelActionType, CustomRuntimeApi, VaultAddresses, VaultSwapDetails,
 	},
-	AccountId, Nonce, RuntimeCall,
+	AccountId, Hash, Nonce, RuntimeCall,
 };
 use std::sync::Arc;
 
@@ -119,6 +121,42 @@ where
 			signed_pool_client: SignedPoolClient::new(client, pool, pair),
 		}
 	}
+
+	async fn extract_swap_deposit_address(
+		&self,
+		block_hash: Hash,
+		tx_index: TxIndex,
+	) -> RpcResult<SwapDepositAddress> {
+		let ExtrinsicData { events, header, .. } = self
+			.signed_pool_client
+			.get_extrinsic_data_dynamic(block_hash, tx_index)
+			.await
+			.map_err(CfApiError::from)?;
+
+		Ok(extract_from_first_matching_event!(
+			events,
+			cf_static_runtime::swapping::events::SwapDepositAddressReady,
+			{
+				deposit_address,
+				channel_id,
+				source_chain_expiry_block,
+				channel_opening_fee,
+				refund_parameters
+			},
+			SwapDepositAddress {
+				address: AddressString::from_encoded_address(deposit_address.0),
+				issued_block: header.number,
+				channel_id,
+				source_chain_expiry_block: source_chain_expiry_block.into(),
+				channel_opening_fee: channel_opening_fee.into(),
+				refund_parameters: ChannelRefundParameters::from(refund_parameters)
+				.map_address(|refund_address| {
+					AddressString::from_encoded_address(&refund_address.0)
+				}),
+			}
+		)
+		.map_err(CfApiError::from)?)
+	}
 }
 
 #[async_trait]
@@ -170,11 +208,11 @@ where
 		affiliate_fees: Option<Affiliates<AccountId32>>,
 		refund_parameters: RefundParametersRpc,
 		dca_parameters: Option<DcaParameters>,
-		wait_for_finality: Option<bool>,
+		_wait_for_finality: Option<bool>,
 	) -> RpcResult<SwapDepositAddress> {
-		let ExtrinsicData { events, header, .. } = self
+		let mut status_stream = self
 			.signed_pool_client
-			.submit_watch_dynamic(
+			.submit_watch(
 				RuntimeCall::from(
 					pallet_cf_swapping::Call::request_swap_deposit_address_with_affiliates {
 						source_asset,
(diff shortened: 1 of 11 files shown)


state-chain/custom-rpc/src/pool_client.rs
```diff
@@ -605,10 +577,65 @@ where
 		until_finalized: bool,
 		dry_run: bool,
 	) -> Result<ExtrinsicData<DynamicEvents>, PoolClientError> {
-		self.submit_watch(call, until_finalized, dry_run, |block_hash, extrinsic_index| {
-			self.get_extrinsic_data_dynamic(block_hash, extrinsic_index)
-		})
-		.await
+		let mut status_stream = self.submit_watch(call, dry_run).await?;
+
+		while let Some(status) = status_stream.next().await {
```
kylezs, line 582: would be nice to factor this out as it's very similar to above

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -3038,32 +3108,33 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-							Error::<T, I>::SolanaAddressDerivationError,
-						AddressDerivationError::MissingSolanaApiEnvironment =>
-							Error::<T, I>::MissingSolanaApiEnvironment,
-						AddressDerivationError::MissingAssethubVault =>
-							Error::<T, I>::MissingAssethubVault,
-					})?,
-				next_channel_id,
-			)
-		};
+			let new_chan = Self::next_channel(source_asset)?;
```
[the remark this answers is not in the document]
nmammeri: The reason for this is that Self::next_channel returns an error and is used in the mutate closure. Declaring new_chan makes it easy to propagate the error and be used in the mutate closure.

state-chain/custom-rpc/src/broker.rs
```diff
@@ -191,35 +228,49 @@ where
...
+						.iter()
+						.any(|channel_id| channel_id == &swap_deposit_address.channel_id)
+					{
+						return Ok(swap_deposit_address);
+					}
+				},
+				TransactionStatus::Finalized((block_hash, tx_index)) =>
+					return self.extract_swap_deposit_address(block_hash, tx_index).await,
+				_ =>
+					if let Some(e) = is_transaction_status_error(&status) {
```
kylezs, line 268: This is a bit of a strange construction to me. We return an Error inside a Some, meaning we have to unpack the option to throw the Error. 
If `is_transaction_status_error` returns a `Result<(),PoolClientError>` instead, it simplifies a lot of this. Here we can do `.map_err()` and in the other places we can simply use `?` directly.

state-chain/custom-rpc/src/broker.rs
```diff
@@ -191,35 +228,49 @@ where
...
-			cf_static_runtime::swapping::events::SwapDepositAddressReady,
-			{
-				deposit_address,
-				channel_id,
-				source_chain_expiry_block,
-				channel_opening_fee,
-				refund_parameters
+		// Get the pre-allocated channels from the previous finalized block
+		let source_chain: ForeignChain = source_asset.into();
+		let pre_allocated_channels = self.rpc_backend.with_runtime_api(
```
kylezs, line 238: nit: consistency
```suggestion
		let preallocated_channels = self.rpc_backend.with_runtime_api(
```

state-chain/custom-rpc/src/broker.rs
```diff
@@ -170,11 +208,10 @@ where
 		affiliate_fees: Option<Affiliates<AccountId32>>,
 		refund_parameters: RefundParametersRpc,
 		dca_parameters: Option<DcaParameters>,
-		wait_for_finality: Option<bool>,
 	) -> RpcResult<SwapDepositAddress> {
-		let ExtrinsicData { events, header, .. } = self
+		let mut status_stream = self
```
kylezs, line 212: I would leave a comment on top of each of these implementations to ensure that any change is reflected on the other to maintain consistency before we deprecate these old RPCs

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -3038,32 +3093,49 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				Some(channel) => channel,
+				// if there are no pre-allocated channels, take the next channel
+				// from the pool of recycled channels, or generate a new channel entirely.
+				None => match DepositChannelPool::<T, I>::drain().next() {
+					Some((_, deposit_channel)) => deposit_channel,
+					None => Self::generate_new_channel(source_asset)?,
+				},
+			};
+		// Make sure to set the asset on the channel, in case it was recycled
+		deposit_channel.asset = source_asset;
```
kylezs, line 3109: We should move this above into the branch above where we get the recycled channel - given it's only relevant for that case now... and makes it clear, removing the need for the comment.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -3038,32 +3093,49 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-						AddressDerivationError::MissingAssethubVault =>
-							Error::<T, I>::MissingAssethubVault,
-					})?,
-				next_channel_id,
-			)
-		};
+			// For chains that preallocation is not enabled such as Ethereum and Arbitrum,
+			// only pre-allocate channels from the global pool of addresses, for other chains,
+			// preallocate a newly generated channel.
+			if !Into::<ForeignChain>::into(source_asset).can_preallocate_addresses() {
```
kylezs, line 3124: also would invert the condition to remove this negation

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -3038,32 +3093,49 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+				// It doesn't matter what source asset we pass here, because source_asset
+				// is ignored by chains that support preallocation.
+				let new_channel = Self::generate_new_channel(source_asset)?;
+				PreallocatedChannels::<T, I>::mutate(requester, |queue| {
+					queue.push_back(new_channel)
+				});
+			}
+		}
 
 		let deposit_address = deposit_channel.address.clone();
```
kylezs, line 3140: don't actually need this separate clone it seems (and would just remove the variable in that case)

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -3359,6 +3431,7 @@ impl<T: Config<I>, I: 'static> DepositApi<T::TargetChain> for Pallet<T, I> {
 		(ChannelId, ForeignChainAddress, <T::TargetChain as Chain>::ChainBlockNumber, Self::Amount),
 		DispatchError,
 	> {
+		// Try to get a pre-allocated channel first, if not request new one
```
kylezs, line 3434: I don't think these comments are necessary, and if anything. If you do want to keep the comment, then I would put it on the `open_channel` function itself to avoid duplicating it

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1548,6 +1554,212 @@ fn all_batch_errors_are_logged_as_event() {
...
+		// which is actually the last element from the global pool
+		init_pool_ids.retain(|i| !preallocated_channels_1.contains(i));
+		assert_eq!(preallocated_channels_2.len(), 2);
+		assert_eq!(init_pool_ids.len(), 1);
+		assert!(preallocated_channels_2.contains(&init_pool_ids[0]));
+
+		// STEP 3: If we allocate another channel, it should be one from the preallocated list.
+		// however the preallocation list should fill to MAX 2 because we don't have any more pool
+		// channels.
+		let (deposit_channel_3, _, _, _) =
```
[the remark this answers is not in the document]
nmammeri: good idea, I've updated the tests to check for this. This needed changing the open_channel function to return the whole DepositChannel instead of just an ID and Address.

state-chain/pallets/cf-ingress-egress/src/tests.rs
```diff
@@ -1744,10 +1757,11 @@ fn preallocated_channels_no_global_pool() {
 			.map(|chan| chan.channel_id)
 			.collect::<Vec<_>>();
 		assert!(preallocated_channels_1.contains(&deposit_channel_3));
+		assert!(preallocated_channels_2.contains(&deposit_channel_3));
 		assert_eq!(preallocated_channels_3, vec![4, 5, 6, 7]);
 
-		// Since we have max 2 preallocated channels, the next allocation should be not from
-		// initial preallocated list.
+		// STEP 4: Since we initially had max of 2 preallocated channels that were consumed in steps
+		// 2 and 3, the next allocation should not from the initial preallocated_channels_1 list.
```
kylezs, line 1768: typo?

state-chain/pallets/cf-ingress-egress/src/mocks.rs
```diff
@@ -148,6 +149,7 @@ impl Config<Instance2> for Test {
 	type RuntimeEvent = RuntimeEvent;
 	type RuntimeCall = RuntimeCall;
 	const MANAGE_CHANNEL_LIFETIME: bool = true;
+	const CAN_PREALLOCATE_ADDRESSES: bool = true;
```
kylezs, line 152: This name doesn't really reflect what's happening, since we end up preallocating for all chains. 
Something like `OnlyPreallocateFromPool` perhaps? Where Ethereum and Arbitrum would then be true (and it clearly means we don't preallocate from elsewhere)
- open to other names, but the current one is inaccurate

state-chain/custom-rpc/src/broker.rs
nmammeri: I agree but this applies to all RPCs in the old binary not just this one.. So I feel no need to for an extra comment because we need to maintain consistency for all of them until the old RPC binary is deprecated.

dandanlen (approved):

PR #5899, opened 2025-05-30
fix: keep reorged events if replacement election doesnt resolve

MxmUrw:
# Pull Request

Closes: PRO-2299

## Checklist

## Summary

This PR adds another argument called `lowest_in_progress_height` to the `process_chain_progress()` method of the block explorer. This height is now used for expiry of block data and events, instead of the "highest seen block height" that was used previously.

This fixes the bug found in #5898 by only expiring block & event data when all replacement elections for "height + SAFETY_BUFFER" have finished (i.e. are not mentioned anywhere in the election tracker anymore).

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/state_machine.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/state_machine.rs
@@ -220,7 +220,7 @@ impl<T: BWTypes> Statemachine for BWStatemachine<T> {
 					// NOTE: we use the lowest "in progress" height for expiring block and event
 					// data, this way if one BW election is stuck (e.g. due to failing rpc
 					// call), no event data is going to be deleted. That's why we can't use
-					// the `highest_seen` block height, because that one progress always
+					// the `highest_seen` block height, because that one progresses always
 					// following data from the BHW, ignoring ongoing elections.
 					s.elections.lowest_in_progress_height(),
 				);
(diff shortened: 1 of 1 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
```diff
@@ -383,6 +392,10 @@ impl<N: Step + Ord> CompactHeightTracker<N> {
 			range_difference(r, &(range.start().clone()..N::forward(range.end().clone(), 1)))
 		}
 	}
+
+	fn get_all_heights(&self) -> BTreeSet<N> {
```
kylezs, line 396: We could just return an iterator here and let the consumer decide what it wants to collect into - especially if the only use so far is turning it into an iterator

kylezs (approved):

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
MxmUrw: Unfortunately this doesn't really improve things since the way we use this is that we want an iterator over `&N` references (we let this interact with the `&N`s that are the keys of the other ongoing and queued BTreeMaps). Here we can't return an `&N` because the ranges over that we iterate don't actually store all the `N`s, and so can only give us an iterator over `N`.

The alternative would be to change the other iterators to be over `N` as well, but that would mean more copying, in particular since we expect the safe block heights to be at most very sparsely populated.

PR #5901, opened 2025-06-02
fix: can close orders after sweeping

msgmaxim:
# Pull Request

If a limit order if fully executed, explicitly closing it by reducing the amount by u128::MAX may result in an error due to the order already having been removed by sweeping. This PR fixes the problem by effectively ignoring the error of not being able to find the order if our intention is to remove it.

The integration test cover this case when ASS is used (the test would previously fail).

### DIFF
--- a/state-chain/amm/src/limit_orders.rs
+++ b/state-chain/amm/src/limit_orders.rs
@@ -787,10 +787,22 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
 			let positions = &mut self.positions[!SD::INPUT_SIDE];
 			let fixed_pools = &mut self.fixed_pools[!SD::INPUT_SIDE];
 
-			let position = positions
-				.get(&(sqrt_price, lp.clone()))
-				.ok_or(PositionError::NonExistent)?
-				.clone();
+			let position = match positions.get(&(sqrt_price, lp.clone())) {
+				Some(position) => position.clone(),
+				None =>
+					if amount == u128::MAX.into() {
+						// We are removing the order anyway so not being able to
+						// find it is not a problem (this happens if implicit sweeping
+						// pre-removes the order if it was fully executed).
+						return Ok((
+							Default::default(),
+							Default::default(),
+							PositionInfo::default(),
+						));
+					} else {
+						return Err(PositionError::NonExistent);
+					},
+			};
 			let option_fixed_pool = fixed_pools.get(&sqrt_price);
 
 			let (collected_amounts, option_position) =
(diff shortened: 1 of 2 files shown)


state-chain/amm/src/limit_orders.rs
```diff
@@ -787,10 +787,22 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
+				None =>
+					if amount == u128::MAX.into() {
+						// We are removing the order anyway so not being able to
+						// find it is not a problem (this happens if implicit sweeping
+						// pre-removes the order if it was fully executed).
+						return Ok((
+							Default::default(),
+							Default::default(),
+							PositionInfo::default(),
+						));
```
dandanlen, line 801: Isn't the expectation that this should return some non-zero collected amount? 

ie. whatever was swept during the auto-sweep...

dandanlen (commented):
Do we not need some sort of migration to rescue the orders that are currently stuck?

dandanlen:
I had another look at this - I think maybe a better solution would be to change the implementation of `PoolApi::cancel_limit_order`: currently it calls into `update_limit_order`, which eventually calls into `inner_update_limit_order_at_tick` with `NoopStatus::Error`. 

What if we instead implemented `cancel_all_limit_orders` directly, instead of via `cancel_limit_order` (which is single-use). And then we call `inner_update_limit_order_at_tick` with `NoopStatus::Allow`.

Did you consider this?

dandanlen:
@msgmaxim I pushed a quick-and-dirty version of what I suggested - it seems to pass the new test. WDYT?

state-chain/amm/src/limit_orders.rs
msgmaxim: No, whatever was collected during `inner_sweep` is already in the strategy's balance at this point (however this gets rolled back due to the error later which this change was addressing).

msgmaxim:
> @msgmaxim I pushed a quick-and-dirty version of what I suggested - it seems to pass the new test. WDYT?

I think your solution is cleaner. However, I was also considering users triggering this by closing their orders manually by updating with `Decrease(u128::MAX)`, which my solution would also cover, but I suppose this doesn't happen very often or we would hear about this earlier. Happy to merge as is with your changes.

msgmaxim:
> Do we not need some sort of migration to rescue the orders that are currently stuck?

As discussed, no migration should be necessary. The orders were "stuck" because the way to cancel them was broken, not because there was something wrong with the state.

dandanlen:
> I was also considering users triggering this by closing their orders manually by updating with `Decrease(u128::MAX)`, which my solution would also cover, but I suppose this doesn't happen very often or we would hear about this earlier. Happy to merge as is with your changes.

Hmm, yes I agree this is weird behaviour either way. Will have another look before merging.

dandanlen (approved):

PR #5925, opened 2025-06-06
feat: add standalone aggregator engine witnessing logic

albert-llimos:
# Pull Request

Closes: PRO-2287

## Checklist

## Summary

Add witnessing for aggregator in EVM and Solana.
Update EVM and Solana images with changes from:
https://github.com/chainflip-io/chainflip-sol-contracts/pull/94
https://github.com/chainflip-io/chainflip-eth-contracts/pull/507
https://github.com/chainflip-io/chainflip-eth-contracts/pull/508

Various workarounds to make the upgrade test work, including a storage write because the protocol ALT address is not deterministic on each image. Similar to the nonces, which they have to be forced on each image update.
Remove old workarounds for the Swap Endpoint program.
Update the OLD_VERSION to ~~1.9.2~~ 1.9.6 because using 1.9.0 as the OLD_VERSION won't work in the upgrade test without the 1.9.2 updates in the Swap Endpoint Solana program. This will anyway be updated later on to the latest 1.9.X when preparing the 1.10 release.
For now this logic is not executed and will need the rest of work in the elections pallets to be completed. It might need further tweaks after that work is done but there are already are tests to ensure that it's working in Devnet.

I have manually decoded the Solana RPC return data because the Oracle Query Helper program interface doesn't declare the return type (long story about Anchor returning data). We might want to consider using Borsh deserialize to do that but this should already ensure correctness of the return data.

Right now a Solana single query gets all the data of multiple feeds in one single RPC call. This reduces the rpc load,otherwise it was multiple queries per asset. ~~In EVM we can query all the data of one asset in a single call. Also, we will use it as a fallback so for now it might not be worth it to write a small smart contract for that.~~
In EVM we will also use a function from the Address Checker to query all data in one single rpc call, especially with the block and block timestamp that are otherwise not possible to get in one call.

In PRO-2320 I will do the  deployment of the new contracts (both EVM and Solana) and the corresponding environment pallet migration to set the new values of the EVM Address checker. Inserting the Solana new contract values can be done together with the election system.


engine-upgrade-utils/src/lib.rs
```diff
@@ -26,7 +26,7 @@ pub mod build_helpers;
 // rest of the places the version needs changing on build using the build scripts in each of the
 // relevant crates.
 // Should also check that the compatibility function below `args_compatible_with_old` is correct.
-pub const OLD_VERSION: &str = "1.9.0";
+pub const OLD_VERSION: &str = "1.9.2";
```
kylezs, line 29: can use 1.9.6 now

engine/src/witness/sol/oracle_witnessing.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// Expected returned data is Vec<PriceFeedData>
+fn decode_query_return_data(
+	return_data: &UiTransactionReturnData,
+	expected_program_id: SolAddress,
+) -> Result<Vec<PriceFeedData>, anyhow::Error> {
+	let decoded_return_data = BASE64_STANDARD.decode(return_data.data.0.clone())?;
+	if return_data.data.1 != UiReturnDataEncoding::Base64 {
+		return Err(anyhow::anyhow!(
```
kylezs, line 113:
```suggestion
		bail!(
```
same below - for early returns of an anyhow error, we prefer bail

albert-llimos:
After thinking more about it I think there is a pretty nice way to add the auxiliary query logic in an EVM contract without too much work. I can add it in the already existing `AddressChecker` as a new function and then we can start using that. Deploying it should also be pretty simple as there it's stateless.
Putting it back to draft even if the majority of the logic is still valid but just to avoid rereviews.

albert-llimos:
Now this is ready for review @dandanlen

dandanlen (approved):
👌

PR #5931, opened 2025-06-16
Chore/deployment query contracts

albert-llimos:
# Pull Request

Closes: PRO-2320

## Checklist

## Summary

Migration will most likely not work as is because of the declaration of the declaration of the safe_mode migration but I'll fix it after 1.10 is out and that migration removed.

### DIFF
--- a/state-chain/pallets/cf-environment/src/migrations/oracle_query_environment.rs
+++ b/state-chain/pallets/cf-environment/src/migrations/oracle_query_environment.rs
@@ -1,0 +1,40 @@
+use crate::*;
+
+use frame_support::{pallet_prelude::Weight, traits::UncheckedOnRuntimeUpgrade};
+
+use cf_chains::evm::H256;
+pub struct OracleQueryEnvironmentMigration<T>(PhantomData<T>);
+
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-environment/src/migrations/oracle_query_environment.rs
```diff
@@ -0,0 +1,40 @@
+use crate::*;
+
+use frame_support::{pallet_prelude::Weight, traits::UncheckedOnRuntimeUpgrade};
+
+use cf_chains::evm::H256;
+pub struct OracleQueryEnvironmentMigration<T>(PhantomData<T>);
+
+impl<T: Config<Hash = H256>> UncheckedOnRuntimeUpgrade for OracleQueryEnvironmentMigration<T> {
+	fn on_runtime_upgrade() -> Weight {
+		log::info!("🌮 Running migration for Environment pallet: Updating AddressCheckerAddress.");
```
syan095, line 10: I always put a log::info at the end of the runtime upgrade as well.

Also it would be nice to add a pre- and post migration check to ensure the hash has changed (on non-local net of course) just in case

syan095 (changes requested):
Just some minor nitpicks

PR #5932, opened 2025-06-17
feat: PRO-2309 - Subaccounts

Janislav:
# Pull Request

Closes: PRO-2309

## Checklist

## Summary

Adds the concept of sub-accounts to the account roles pallet. An account can have up to 256 sub accounts and can execute extrinsics on behalf of a subaccount. Once registered, there is no way to unregister a subaccount.

## Summary by CodeRabbit

- **New Features**
  - Introduced sub-account support, allowing users to derive sub-accounts linked to their main account and manage multiple sub-accounts by index.
  - Added ability to execute runtime calls on behalf of a sub-account.
  - New events notify users when sub-accounts are created or when calls are executed as sub-accounts.

- **Bug Fixes**
  - Prevented duplicate sub-account creation for the same index.

- **Tests**
  - Added tests to verify sub-account creation, uniqueness enforcement, and call execution as sub-accounts.

### DIFF
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -546,9 +546,6 @@ pub mod pallet {
 		/// The sub-account already exists.
 		SubAccountAlreadyExists,
 
-		/// The sub-account ID derivation failed.
-		SubAccountIdDerivationFailed,
-
 		/// Can not derive sub-account ID if the parent account is bidding in an auction.
 		CanNotDeriveSubAccountIdIfParentAccountIsBidding,
 


Janislav:
@coderabbitai review

state-chain/pallets/cf-account-roles/src/lib.rs
```diff
@@ -152,6 +186,84 @@ pub mod pallet {
...
+			call.clone()
+				.dispatch(origin)
+				.map_err(|_| Error::<T>::FailedToExecuteCallOnBehalfOfSubAccount)?;
+
+			Self::deposit_event(Event::SubAccountCallExecuted {
+				account_id: account_id.clone(),
+				sub_account_id: sub_account_id.clone(),
+				sub_account_index,
+			});
+
```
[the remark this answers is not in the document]
Janislav: Just a dummy weight for now... Will be fixed

state-chain/pallets/cf-account-roles/src/lib.rs
```diff
@@ -152,6 +186,84 @@ pub mod pallet {
...
+			let mut origin = origin;
+			let account_id = ensure_signed(origin.clone())?;
+			let sub_account_id = SubAccounts::<T>::get(&account_id, sub_account_index);
+			ensure!(sub_account_id.is_some(), Error::<T>::UnknownAccount);
+			let sub_account_id = sub_account_id.unwrap();
+			origin.set_caller_from(frame_system::RawOrigin::Signed(sub_account_id.clone()));
+
+			call.clone()
+				.dispatch(origin)
+				.map_err(|_| Error::<T>::FailedToExecuteCallOnBehalfOfSubAccount)?;
```
dandanlen, line 257:
```suggestion
			call.clone()
				.dispatch(origin)?;
```

I feel like it would be useful to know the real error instead of swallowing it.

state-chain/pallets/cf-account-roles/src/lib.rs
```diff
@@ -152,6 +186,84 @@ pub mod pallet {
...
+				!SubAccounts::<T>::contains_key(&account_id, sub_account_index),
+				Error::<T>::SubAccountAlreadyExists
+			);
+			let sub_account_id: T::AccountId = Decode::decode(&mut TrailingZeroInput::new(
+				(*b"chainflip/subaccount", account_id.clone(), sub_account_index)
+					.blake2_256()
+					.as_ref(),
+			))
+			.map_err(|_| Error::<T>::SubAccountIdDerivationFailed)?;
+			SubAccounts::<T>::insert(&account_id, sub_account_index, &sub_account_id);
```
dandanlen, line 221: A subaccount needs to be a 'real' account, and should be created in the same way that accounts are created in the funding pallet (maybe we expose an internal 'fund_account' API. Then we also don't need the `SubAccounts` storage (we just derive the account and check if it exists, like any other account). It still might be useful to store *some* mapping to allow us to aggregate eg. balances from the subaccount to the parent in the rpcs. 

Using the current implementation you can't, for example, use rebalance to add funds to the account because the account doesn't exist. 

I think this implies that the subaccount needs to be created with some funds from the parent (basically like the `funded` extrinsic in the funding pallet).

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +987,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+		))
+		.map_err(|_| Error::<T>::SubAccountIdDerivationFailed)?;
+
+		Ok(sub_account_id)
+	}
+
+	fn derive_and_fund_sub_account(
+		parent_account_id: T::AccountId,
+		sub_account_index: u8,
+	) -> Result<T::AccountId, DispatchError> {
```
dandanlen, line 1009: I think it would be useful to allow specifying the starting balance of the subaccount instead of assuming `MinimumFunding`.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +987,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+			Error::<T>::SubAccountAlreadyExists
+		);
+
+		frame_system::Provider::<T>::created(&sub_account_id)?;
+
+		let restricted_balances = RestrictedBalances::<T>::get(&parent_account_id);
+
+		if !restricted_balances.is_empty() {
+			RestrictedBalances::<T>::insert(&sub_account_id, restricted_balances);
+		}
```
dandanlen, line 1028: This is taken care of by the rebalance operation. We don't need to do it here.

state-chain/traits/src/lib.rs
```diff
@@ -1261,3 +1261,23 @@ pub trait PoolPriceProvider {
...
+	fn derive_sub_account(
+		parent_account_id: AccountId,
+		sub_account_index: SubAccountIndex,
+	) -> Result<AccountId, DispatchError>;
+
+	fn sub_account_exists(
+		parent_account_id: AccountId,
+		sub_account_index: SubAccountIndex,
+	) -> Result<bool, DispatchError>;
+}
```
dandanlen, line 1283: I don't think this is right abstraction. 

We end up splitting sub account functionality between account-roles and funding pallets. 

Account creation logic should be in the funding pallet:

```
pub trait SpawnAccount {
	type AccountId;

	fn spawn_account(
		account_id: Self::AccountId,
		initial_balance: Balance,
	) -> Result<(), DispatchError>;
}
```

Everything else (subaccount deriviation, extrinsics) can be in the account roles pallet.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +987,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+			OriginFor::<T>::signed(parent_account_id.clone()),
+			sub_account_id.clone(),
+			None,
+			RedemptionAmount::Exact(MinimumFunding::<T>::get()),
+		)?;
+
+		Ok(sub_account_id)
+	}
+
+	fn sub_account_exists(
```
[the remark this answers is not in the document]
Janislav: true, this is avoidable.

state-chain/traits/src/lib.rs
dandanlen: Yes this is what I said: "Account creation logic should be in the funding pallet". 

Deriving and checking the sub-account is currently *also* defined in the funding pallet, but it doesn't need to be. It should be in the account roles pallet, which is the only place we need to use it.

Janislav: But we need all the data (parent account id fox ex) also in the funding pallet to do other necessary checks and write proper tests, and I don't really see why splitting it up makes it so much better.

dandanlen: I updated the code snippet above, should be clearer.

PR #5934, opened 2025-06-17
test: reenable BW proptests

MxmUrw:
# Pull Request

Closes: PRO-2321

## Checklist

## Summary

This re-enables the (statemachine) (unit-) proptests for the BW. They've been disabled a long time due to refactorings.

This surfaced a few bugs, so in particular this commit includes the following changes:
 - fix: expire optimistic blocks.
 - fix: forward Governance election block data to BP.
 - fix: only mutate ongoing ByHash elections to BySafeHeight, don't do anything with Optimistic.
 - test: remove the requirement for safeheight elections to be lower than ongoing. This is because in certain cases (impossible with the current BHW) we can't guarantee it.
 - test: add an alternative requirement for safeheight and ongoing elections to be disjoint.
 - fix: rewrite the CompactHeightTracker to use a BTreeMap internally. This also fixes a few bugs in its behaviour.

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
@@ -11,7 +11,6 @@ use serde::{Deserialize, Serialize};
 use sp_std::{
 	cmp::max,
 	collections::{btree_map::BTreeMap, btree_set::BTreeSet},
-	iter,
 	vec::Vec,
 };
 
@@ -619,7 +618,7 @@ pub struct CompactHeightTrackerExtract<'a, N: Ord> {
 	tracker: &'a mut CompactHeightTracker<N>,
 }
 
-impl<N: Step> Iterator for CompactHeightTrackerExtract<'_, N> {
+impl<N: SaturatingStep + Ord + Clone> Iterator for CompactHeightTrackerExtract<'_, N> {
 	type Item = N;
 
 	fn next(&mut self) -> Option<Self::Item> {
@@ -636,7 +635,6 @@ impl<N: Step> Iterator for CompactHeightTrackerExtract<'_, N> {
 		}
 	}
 }
-
 
 #[test]
 pub fn test_compact_height() {
@@ -668,7 +666,6 @@ pub fn test_compact_height() {
 	tracker.remove(1..6);
 	assert_eq!(tracker.elections, [].into_iter().collect());
 }
-
 
 #[cfg_attr(test, derive(Arbitrary))]
 #[derive(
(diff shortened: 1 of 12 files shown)


state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
```diff
@@ -169,6 +183,11 @@ impl<T: BWTypes> ElectionTracker<T> {
 		self.ongoing.last_key_value().inspect(|(height, _)| {
 			self.highest_ever_ongoing_election = max(self.highest_ever_ongoing_election, **height);
 		});
+
+		// clean up the optimistic block cache
+		self.optimistic_block_cache.retain(|height, _| {
+			height.saturating_forward(T::Chain::SAFETY_BUFFER) > self.seen_heights_below
+		});
```
marcellorigotti, line 190: I was thinking, maybe instead of removing the optimistic blocks here we should do it in the `schedule_election` and `mark_election_done` since it doesn't really make sense to keep these blocks around for more than necessary. 
I.E. once block data is received and it is valid we always removed it from the `optimistic_cache`

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
```diff
@@ -392,7 +410,9 @@ impl<T: BWTypes> ElectionTracker<T> {
 	pub fn lowest_in_progress_height(&self) -> ChainBlockNumberOf<T::Chain> {
 		*self
 			.ongoing
-			.keys()
+			.iter()
+			.filter(|(_, t)| **t != BWElectionType::Optimistic)
+			.map(|(h, _)| h)
```
marcellorigotti, line 415: nit:
```suggestion
			.filter_map(|(h, t)| if **t != BWElectionType::Optimistic { Some(h) } else { None })
```

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
MxmUrw: True, we don't want to keep the blocks around for long, I'll have a look how to optimize.

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
MxmUrw: Fair, but I have to say I find the first one a bit more readable, and it's shorter as well.

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/primitives.rs
MxmUrw: Ok, so I looked, we already do remove optimistic blocks from the cache when we receive a hash for the given height in `schedule_range`. So this code is more of a safeguard if for whatever reason we didn't that block, to eventually remove it.

PR #5936, opened 2025-06-17
feat: core chp

msgmaxim:
# Pull Request

Closes: PRO-2324

## Summary

(see details in the linked linear issue)

### DIFF
--- a/state-chain/pallets/cf-lending-pools/src/chp_lending.rs
+++ b/state-chain/pallets/cf-lending-pools/src/chp_lending.rs
@@ -1,0 +1,436 @@
+use super::*;
+
+#[cfg(test)]
+mod chp_lending_tests;
+
+#[derive(Clone, DebugNoBound, PartialEq, Eq, Encode, Decode, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct ChpLoan<T: Config> {
+	loan_id: ChpLoanId,
+	asset: Asset,
+	created_at_block: BlockNumberFor<T>,
+	expiry_block: BlockNumberFor<T>,
+	borrower_id: T::AccountId,
+	usdc_collateral: AssetAmount,
+	fees_collected_usdc: AssetAmount,
+	pool_contributions: Vec<ChpPoolContribution>,
+	// Interest charged on the principal amount every INTEREST_PAYMENT_INTERVAL blocks
+	interest_rate: Permill,
+	status: LoanStatus,
+}
+
+impl<T: Config> ChpLoan<T> {
+	fn total_principal_amount(&self) -> AssetAmount {
+		self.pool_contributions.iter().map(|c| c.principal).sum()
+	}
+
+	fn overcollateralisation_ratio(&self) -> Permill {
+		let principal_in_usdc =
+			usdc_equivalent_amount::<T>(self.asset, self.total_principal_amount());
+
+		let collateral_usdc = match self.status {
+			LoanStatus::Active => self.usdc_collateral,
+			LoanStatus::Finalising => {
+				// Should be unreachable
+				Default::default()
+			},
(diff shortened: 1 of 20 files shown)


state-chain/pallets/cf-lending-pools/src/lib.rs
```diff
@@ -394,6 +521,96 @@ pub mod pallet {
...
+
+			let core_pool_id = chp_pool.core_pool_id;
+
+			let (unlocked_amount, pending_core_loans) =
+				CorePools::<T>::mutate(asset, core_pool_id, |pool| {
+					let pool = pool.as_mut().ok_or(Error::<T>::PoolDoesNotExist)?;
+
+					pool.stop_lending(lender_id.clone()).map_err(|e| match e {
+						core_lending_pool::Error::AccountNotFoundInPool =>
+							Error::<T>::AccountNotFoundInBoostPool,
```
j4m1ef0rd, line 588: Wrong error? This is a boost error.
Why map the error?

msgmaxim: We need to map because it is a different enum. I changed the name of the error to just `AccountNotFoundInPool` for both pools (already discussed consolidation of errors with the product).

state-chain/pallets/cf-lending-pools/src/chp_lending.rs
```diff
@@ -0,0 +1,438 @@
...
+		amount: AssetAmount,
+	) -> Result<(), DispatchError> {
+		ChpLoans::<T>::try_mutate(asset, loan_id, |loan| {
+			let loan = loan.as_mut().ok_or(Error::<T>::ChpLoanDoesNotExist)?;
+
+			let principal_total = loan.total_principal_amount();
+
+			let amount = core::cmp::min(amount, principal_total);
+
+			T::Balance::try_debit_account(&loan.borrower_id, asset, amount)?;
```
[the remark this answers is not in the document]
msgmaxim: Maybe not, but I think we can update this once we start working on a use case that requires a different api?

state-chain/pallets/cf-lending-pools/src/chp_lending.rs
```diff
@@ -0,0 +1,438 @@
...
+	loan_id: ChpLoanId,
+	asset: Asset,
+	created_at_block: BlockNumberFor<T>,
+	expiry_block: BlockNumberFor<T>,
+	borrower_id: T::AccountId,
+	usdc_collateral: AssetAmount,
+	fees_collected_usdc: AssetAmount,
+	pool_contributions: Vec<ChpPoolContribution>,
+	// Interest charged on the principal amount every INTEREST_PAYMENT_INTERVAL blocks
+	interest_rate: Permill,
```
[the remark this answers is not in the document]
msgmaxim: Updated to be per block (and multiplying by INTEREST_PAYMENT_INTERVAL when charging it). As discussed, I will make a note about saving config/thresholds per loan so we can come back to this after some more discussion.

state-chain/pallets/cf-lending-pools/src/chp_lending.rs
```diff
@@ -0,0 +1,459 @@
...
+	})
+}
+
+pub fn process_collateral_topup<T: Config>(
+	loan: &mut ChpLoan<T>,
+	config: &ChpConfiguration,
+) -> Weight {
+	if loan.overcollateralisation_ratio() < config.overcollateralisation_topup_threshold {
+		let collateral_required =
+			usdc_collateral_required::<T>(loan.asset, loan.total_principal_amount());
```
syan095, line 193: We should be consistent here regarding what config we use. 
I think we should be passing the config through here

msgmaxim:
Closing in favour of #6061

PR #5942, opened 2025-06-20
Add benchmarking for the chp pallet

syan095:
WIP: add benchmarking for hook


state-chain/pallets/cf-lending-pools/src/chp_lending.rs
```diff
@@ -301,7 +316,8 @@ pub fn chp_upkeep<T: Config>(current_block: BlockNumberFor<T>) -> Weight {
 	}
 
 	// Calculate final weight
-	T::WeightInfo::upkeep_active(count_active) +
+	weight_consumed +
+		T::WeightInfo::upkeep_active(count_active) +
```
msgmaxim, line 320: There doesn't seem to be a significant difference in weight between different types of loan states (active/liquidating/finalising), so I think we could simplify and just use the "no_action" one. Then IMO the logic for how we compute the weight would be more clear: we use "no_action" as the baseline and add weight of the actions that we do perform.

state-chain/pallets/cf-lending-pools/src/chp_lending.rs
```diff
@@ -262,14 +262,13 @@ pub fn chp_upkeep<T: Config>(current_block: BlockNumberFor<T>) -> Weight {
 	let config = ChpConfig::<T>::get();
 
 	// For weight calculation
-	let mut count_active = 0u32;
-	let mut count_soft_liquidation = 0u32;
-	let mut count_no_action = 0u32;
+	let mut count_baseline = 0u32;
```
msgmaxim, line 265: This is just going to be equal to the number of elements in `ChpLoans`, right?

msgmaxim (approved):

PR #5955, opened 2025-06-25
feat: remove unwrap from NonemptyContinuousHeaders

marcellorigotti:
# Pull Request

Closes: PRO-2349

## Checklist


state-chain/pallets/cf-elections/src/electoral_systems/block_height_witnesser/primitives.rs
```diff
@@ -72,29 +83,63 @@ impl<T: ChainTypes> NonemptyContinuousHeaders<T> {
...
+
 	pub fn trim_to_length(&mut self, target_length: usize) {
-		while self.headers.len() > target_length && target_length > 0 {
-			self.headers.pop_front();
+		while self.len() > target_length && target_length > 0 {
+			self.safe_pop_front();
+		}
+	}
+
+	// Return None if there is 1 element only
```
MxmUrw, line 121:
```suggestion
	/// Extracts the first element. Returns None if there is 1 element only, as it cannot be removed.
```

PR #5956, opened 2025-06-27
refactor: ccm refund parameters

dandanlen:
See comments in #5816


state-chain/chains/src/refund_parameters.rs
```diff
@@ -0,0 +1,255 @@
...
+			refund_address: f(self.refund_address)?,
+			min_price: self.min_price,
+			refund_ccm_metadata: self.refund_ccm_metadata,
+		})
+	}
+}
+
+impl<A> ChannelRefundParametersUnchecked<A> {
+	/// Converts the refund address into the ForeignChainAddress type.
+	pub fn with_foreign_chain_refund_address<C>(
```
msgmaxim, line 164: How about:
```suggestion
	pub fn decode_chain_refund_address<C>(
```

Otherwise "with_" makes it sound like we adding something to the object rather than converting an existing field.

state-chain/chains/src/refund_parameters.rs
```diff
@@ -0,0 +1,255 @@
...
+			C::try_from_encoded_address(addr.clone()).map_err(|_| "Invalid refund address")
+		})?)
+	}
+}
+
+// NOTE: Currently CCM checking requires a ForeignChainAddress - hence the <ForeignChainAddress>
+// constraint on this impl. We should be able to remove this requirement somehow.
+impl ChannelRefundParametersUnchecked<ForeignChainAddress> {
+	/// Checks the CCM Refund metadata payload and converts the address to the internal
+	pub fn with_checked_metadata(
```
msgmaxim, line 209: Same reasoning as the comment above:

```suggestion
	pub fn check_metadata(
```

msgmaxim (approved):

state-chain/chains/src/refund_parameters.rs
```diff
@@ -0,0 +1,255 @@
...
+pub type ChannelRefundParametersChecked<A> =
+	ChannelRefundParameters<A, Option<CcmDepositMetadataChecked<ForeignChainAddress>>>;
+
+/// Refund parameters with CCM metadata that has not yet been checked for validity, but where the
+/// refund address has been parsed into the internal `ForeignChainAddress` type.
+pub type ChannelRefundParametersUncheckedInternalAddress =
+	ChannelRefundParametersUnchecked<ForeignChainAddress>;
+
+/// The legacy refund parameters type that does not include any ccm refund metadata.
+pub type ChannelRefundParametersLegacy<RefundAddress> = ChannelRefundParameters<RefundAddress, ()>;
```
syan095, line 97: This legacy type is not actually used anywhere (except for tests). This WAS originally used only in migrations, but not anymore. 
If we want, we do not need to create this type alias

msgmaxim: We need to keep legacy types of API compatibility, no? I'm pretty sure I'm using this in my PR. Not a problem for me to create it again, but I wouldn't worry about cleaning this PR up tbh, I can do it in my follow-up PR.

state-chain/chains/src/refund_parameters.rs
dandanlen: I wanted to find a somewhat common convention that communicates that something inside the struct is being converted. I agree `with_` is not ideal but imo check_metadata sounds like it just checks it without converting. Maybe something like `convert_into` or `map_into`.

PR #5965, opened 2025-07-02
feat: bouncer retry send liquidity transaction

j4m1ef0rd:
# Pull Request

Closes: PRO-2209

## Checklist

## Summary

Bouncer problems that i could find so far:
1) asset hub transaction subscription returns "invalid"
2) asset hub transaction subscription stops returning even if the transaction completes
3) btc transaction never gets confirmed
4) btc transaction is confirmed but never witnessed

This PR helps work around (1) and (3), should reduce bouncer flakiness by a noticeable amount. Not sure what the root problems are. Going to wait for new btc election to hopefully solve (4). Not sure how to solve (2) but maybe it's rare and we can ignore it.

- Added a rety to the assetHub and btc transfer transactions.
- (1) Added a check to the assethub deposit code to detect an "Invalid" transaction response. It will now error and retry.
- (3) Reduced the timeout of the btc transaction confirmation so it can timeout and use the retry.
- Cleaned up some bouncer things and added some logging.
- Added the ability to output substrate chain events to a file.

We will want these fixes to be in the upgrade test. So we might want to cherry pick it into the previous release @dandanlen .
Same with #5943


bouncer/shared/send_btc.ts
```diff
@@ -82,15 +86,31 @@ export async function sendBtc(
...
+      logger.warn(`Error sending BTC transaction (attempt ${attempts}): ${error}`);
+      if (attempts >= maxAttempts) {
+        throw error;
+      }
+      await sleep(1000);
+    }
   }
 
-  return txid;
+  return '';
```
dandanlen, line 113: Would we not want to throw an error in this case instead of returning the empty string?

dandanlen (approved):

PR #5971, opened 2025-07-03
feat: price oracle elections

MxmUrw:
# Pull Request

Closes: PRO-2289

## Checklist

## Summary

This PR adds another instantiation of the `elections` pallet to the runtime, with `I = ()`, called `GenericElections`. It contains a single electoral system, `ChainlinkOraclePriceES` that implements oracle price elections as described in the linear issue. There's a new safemode that stops the creation of oracle elections situated in the `GenericElections` instantiation of the elections pallet.

Features that are not part of this PR:
 - [ ] Price spike circuit breaker. We're going to defer this to a later date, when it is clearer how this feature would be used.


engine/src/witness/generic_elections.rs
```diff
@@ -0,0 +1,233 @@
...
+			updated_at,
+			answered_in_round: _,
+			decimals,
+			description,
+		} = value;
+		Self {
+			description,
+			answer: answer.try_into().unwrap(),
+			decimals,
+			timestamp: UnixTime { seconds: updated_at.try_into().unwrap() },
```
dandanlen, line 98: I can't find the PriceFeedData type... If these unwraps are safe, would be good to explain how/why in an `expect`.

engine/src/witness/generic_elections.rs
```diff
@@ -0,0 +1,233 @@
...
+		settings: <ChainlinkOraclePriceES as ElectoralSystemTypes>::ElectoralSettings,
+		properties: <ChainlinkOraclePriceES as ElectoralSystemTypes>::ElectionProperties,
+	) -> Result<Option<VoteOf<ChainlinkOraclePriceES>>, anyhow::Error> {
+		let price_feeds = match properties.chain {
+			ExternalPriceChain::Solana => {
+				let (price_feeds, _, _) = get_price_feeds(
+					&self.sol_client,
+					settings.sol_oracle_query_helper,
+					settings.sol_oracle_program_id,
+					settings.sol_oracle_feeds.clone(),
```
dandanlen, line 116: How / when do we populate these settings?

state-chain/pallets/cf-elections/src/electoral_systems/oracle_price/state_machine.rs
```diff
@@ -0,0 +1,541 @@
...
+}
+
+// This is defined here for the context, but it's used by the engines
+// to decide whether they should vote with a given oracle result.
+pub fn should_vote_for_asset<T: OPTypes>(
+	(time, price): &(UnixTime, T::Price),
+	conditions: &[VotingCondition<T>],
+) -> bool {
+	use VotingCondition::*;
+	conditions.iter().all(|condition| match condition {
```
dandanlen, line 153: Shouldn't the engines vote if either condition is true?

engine/src/witness/generic_elections.rs
MxmUrw: Good catch, appearently the `answer` is an `I256`.

state-chain/pallets/cf-elections/src/electoral_systems/oracle_price/state_machine.rs
MxmUrw: As discussed, in case of `UpToDate` they'll vote if the timestamp is newer AND the price moved.

state-chain/pallets/cf-elections/src/electoral_systems/oracle_price/primitives.rs
```diff
@@ -0,0 +1,137 @@
...
+
+pub trait AggregationValue = Ord + CommonTraits + MaybeArbitrary + 'static;
+
+def_derive! {
+	#[cfg_attr(test, derive(Arbitrary))]
+	#[derive(TypeInfo)]
+	pub struct Aggregated<A: AggregationValue> {
+		pub median: A,
+		pub iq_range: RangeInclusive<A>,
+	}
```
marcellorigotti, line 84: What is `iq_range` used for exactly? Is it only for testing?

marcellorigotti: I see! I like the idea 👌

PR #5979, opened 2025-07-04
SC calls via Ethereum

ramizhasan111:
# Pull Request

Closes: PRO-2379

## Checklist

## Summary

### DIFF
--- a/engine/src/witness/eth/sc_utils.rs
+++ b/engine/src/witness/eth/sc_utils.rs
@@ -15,7 +15,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use cf_chains::{Chain, Ethereum};
-use cf_primitives::AccountId;
 use ethers::{prelude::abigen, types::Bloom};
 use sp_core::{H160, H256};
 use tracing::{info, trace};
@@ -30,10 +29,8 @@ use super::super::{
 use crate::evm::retry_rpc::EvmRetryRpcApi;
 use cf_chains::evm::ToAccountId32;
 use cf_primitives::EpochIndex;
-use codec::{Decode, Encode, MaxEncodedLen};
 use futures_core::Future;
 use pallet_cf_funding::DepositAndSCCallViaEthereum;
-use scale_info::TypeInfo;
 
 abigen!(ScUtils, "$CF_ETH_CONTRACT_ABI_ROOT/$CF_ETH_CONTRACT_ABI_TAG/IScUtils.json");
 
@@ -123,15 +120,21 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 
 #[cfg(test)]
 mod tests {
-	use super::*;
+	use codec::Encode;
+	use frame_support::sp_runtime::AccountId32;
+	use pallet_cf_funding::AllowedCallsViaSCGateway;
+	use state_chain_runtime::Runtime;
 
 	#[test]
 	fn test_sc_call_encode() {
-		let sc_call_delegate =
-			ScCallViaGateway::DelegateTo { operator: AccountId::new([0xF4; 32]) }.encode();
+		let sc_call_delegate = AllowedCallsViaSCGateway::<Runtime>::Delegate {
+			delegator: [0xf5; 20].into(),
+			operator: AccountId32::new([0xF4; 32]),
+		}
+		.encode();
 		assert_eq!(
 			sc_call_delegate,
-			hex::decode("00f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4")
+			hex::decode("00f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4")
 				.unwrap()
 		);
 	}
--- a/engine/src/witness/eth/sc_utils.rs
+++ b/engine/src/witness/eth/sc_utils.rs
@@ -46,26 +46,6 @@ use anyhow::Result;
 // Undelegate and the UndelegateAndRedeem under the ScCallViaGateway enum.
 // TODO: To discuss if this is  the approach we want to take.
 
-#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
-pub enum ScCallViaGateway {
-	DelegateTo { operator: AccountId },
-	// Undelegate { current_operator: AccountId },
-	// UndelegateAndRedeem { current_operator: AccountId },
-	// TODO: We might just want to just use the Gateway for that, as
-	// funding from a different ETH account won't work here.
-	// NoOp {}, // basically just funding
-}
-
-// Can be used for `DepositToVaultAndScCall` event`. Could be used for both
-// lenders and borrowers if we wanted to.
-// #[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Debug)]
-// pub enum ScCallViaVault {
-// 	AddLoanCollateral { loan_id: H256 },
-//  AddFunds {},
-//  StopLending {amount: Amount}
-// 	...
-// }
-
 impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 	pub fn sc_utils_witnessing<
 		EvmRpcClient: EvmRetryRpcApi + ChainClient + Clone,
--- a/engine/src/witness/eth/sc_utils.rs
+++ b/engine/src/witness/eth/sc_utils.rs
@@ -28,6 +28,7 @@ use super::super::{
 	evm::contract_common::events_at_block,
 };
 use crate::evm::retry_rpc::EvmRetryRpcApi;
+use cf_chains::evm::ToAccountId32;
 use cf_primitives::EpochIndex;
 use codec::{Decode, Encode, MaxEncodedLen};
 use futures_core::Future;
@@ -99,6 +100,9 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 										             * the call from the sc_call bytes above */
 									},
 								caller: sender,
+								// use 0 padded ethereum address as account_id which the flip funds
+								// are associated with on SC
+								caller_account_id: sender.into_account_id_32(),
 								tx_hash: event.tx_hash.to_fixed_bytes(),
 							}
 							.into()
--- a/engine/src/witness/eth/sc_utils.rs
+++ b/engine/src/witness/eth/sc_utils.rs
@@ -30,19 +30,11 @@ use crate::evm::retry_rpc::EvmRetryRpcApi;
 use cf_chains::evm::ToAccountId32;
 use cf_primitives::EpochIndex;
 use futures_core::Future;
-use pallet_cf_funding::DepositAndSCCallViaEthereum;
+use pallet_cf_funding::{EthereumDeposit, EthereumDepositAndSCCall};
 
 abigen!(ScUtils, "$CF_ETH_CONTRACT_ABI_ROOT/$CF_ETH_CONTRACT_ABI_TAG/IScUtils.json");
 
 use anyhow::Result;
-
-// We could decode the ScCall bytes into a Vec<ScCall> to allow multiple actions into a single
-// transaction. The challenge with that is that we'd need to be very careful to make sure that none
-// of the possible combinations can be abused in any way.
-// Alternatively, the safer approach taken here is to limit explicitly the actions that can be taken
-// and allow for particular actions to be batched under the same enum. For example, having the
-// Undelegate and the UndelegateAndRedeem under the ScCallViaGateway enum.
-// TODO: To discuss if this is  the approach we want to take.
 
 impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 	pub fn sc_utils_witnessing<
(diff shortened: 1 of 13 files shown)


dandanlen (commented):
Take a look at the implementation in eth-contracts repo. There are four events:

```
    event DepositToScGatewayAndScCall(address sender, address signer, uint256 amount, bytes scCall);


    event DepositToVaultAndScCall(address sender, address signer, uint256 amount, address token, bytes scCall);


    event DepositAndScCall(address sender, address signer, uint256 amount, address token, address to, bytes scCall);


    event CallSc(address sender, address signer, bytes scCall);
```

Basically, a call can be 'pure' or it can paired with one of:
- deposit to the vault
- deposit to the sc gateway (native flip)
- transfer to another account (not sure what this is for - maybe @albert-llimos can shed some light). 

I think the SC apis could reflect this a bit better, for example:

```rust
enum Deposit {
	Vault { asset, amount },
	Native { flip },
	Transfer { asset, amount, destination },
}

enum EvmCall {
	Delegate { .. },
	Undelegate { .. },
	// etc..
}

fn execute(
	evm_origin: EvmAddress,
	deposit: Option<Deposit>,
	call: EvmCall,
) -> DispatchResult {
	match deposit {
		// .. do deposit
	}
	call.dispatch(..);
}
```

I also think we should consider doing the decoding on-chain. This way we can handle failures more gracefully, in particular in the cases where the call involves a transfer of funds (ie. we can refund on failure).

albert-llimos:
I have now pushed the other chunk of code:
https://github.com/chainflip-io/chainflip-backend/pull/5992
I think it could make sense to rebase this on that PR as there is the logic to decode the call and call the SC that needs to be put together:
https://github.com/chainflip-io/chainflip-backend/pull/5992/files#diff-60b178463337ac359460a617747e4981a88ce6f2cc51fc40ee8a7bb2db6b34cd

state-chain/chains/src/eth.rs
```diff
@@ -256,6 +256,32 @@ impl From<&DepositChannel<Ethereum>> for EvmFetchId {
...
+	PartialEq,
+	Eq,
+	Encode,
+	Decode,
+	MaxEncodedLen,
+	TypeInfo,
+	Serialize,
+	Deserialize,
+)]
+pub enum WhitelistedCallsViaEthereum<AccountId, FlipBalance> {
```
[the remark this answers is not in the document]
ramizhasan111: I put this now in the pallet itself as that was the best place I could find due to dependency issues.

engine/src/witness/eth/sc_utils.rs
```diff
@@ -99,32 +78,38 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
+						) => {
+							pallet_cf_funding::Call::execute_sc_call {
+								sc_call: sc_call.to_vec(),
+								deposit_and_call:
+									DepositAndSCCallViaEthereum::FlipToSCGatewayAndCall {
+										amount: amount.try_into().unwrap(),
+										call: None, /* This will be filled on the SC after SC
+										             * decodes
+										             * the call from the sc_call bytes above */
+									},
```
dandanlen, line 98: I don't understand why we need to construct this struct with a `None` value here... we could just pass the deposit amount and construct the call on-chain after decoding the data?

engine/src/witness/eth/sc_utils.rs
```diff
@@ -99,32 +78,38 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
+								caller: sender,
+								// use 0 padded ethereum address as account_id which the flip funds
+								// are associated with on SC
+								caller_account_id: sender.into_account_id_32(),
```
dandanlen, line 102: Similar thing here, we are basically duplicating information - why?

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +1063,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
 		BoundRedeemAddress::<T>::remove(account_id);
 	}
 }
+
+#[derive(Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+#[scale_info(skip_type_params(T))]
+pub enum AllowedCallsViaSCGateway<T: Config> {
```
dandanlen, line 1069: Somewhat pedantic point, but the calls are not via the sc gateway. They are via SC utils. 

Let's just call it the `EthereumInterface`, or something.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +1063,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+	},
+}
+
+// calls via vault and transfers for future use
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+pub enum AllowedCallsViaVault {}
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+pub enum AllowedCallsViaTransfer {}
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+pub enum AllowedCallsOnlyCalls {}
```
dandanlen, line 1086: I'm not sure these really need to be separate enums. Either way if they are unused we don't need to add them.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +1063,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+		amount: EthAmount,
+		destination: EthereumAddress,
+		call: Option<AllowedCallsViaTransfer>,
+	},
+	NoDepositOnlyCall {
+		call: Option<AllowedCallsOnlyCalls>,
+	},
+	_Marker(PhantomData<T>),
+}
+
```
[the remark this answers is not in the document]
ramizhasan111: ok, will remove. Added them to make explicit what I had in mind.

engine/src/witness/eth/sc_utils.rs
ramizhasan111: this is because we need both the original Eth address and the one converted to Account32. Since we are converting to Account32 in engine (for reasons discussed), we need to pass both.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +1063,67 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+pub enum AllowedCallsViaTransfer {}
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+pub enum AllowedCallsOnlyCalls {}
+
+#[derive(Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+#[scale_info(skip_type_params(T))]
+pub enum DepositAndSCCallViaEthereum<T: Config> {
+	FlipToSCGatewayAndCall {
+		amount: EthAmount,
```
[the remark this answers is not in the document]
albert-llimos: As I understand it,this whole pallet is using `T::Amount` to refer to FLIP amount e.g. `MinimumFunding` or the `RedemptionTax`.

engine/src/witness/eth/sc_utils.rs
```diff
@@ -99,32 +78,38 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
-                                    trace!("Successfully decoded ScCall::DelegateTo with operator: {operator}");
-                                },
-                                Err(_e) => {
-									println!("Failed to decode ScCall");
-                                }
-                            }
-                            continue
-                        },
+					let call: state_chain_runtime::RuntimeCall = match event.event_parameters {
+						ScUtilsEvents::DepositToScGatewayAndScCallFilter(
```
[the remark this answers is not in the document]
albert-llimos: I had a chat with Dan and he was arguing to implement all the events and then not perform any action in the SC if it's not implemented yet. You can align with @dandanlen on what he had in mind. At least we should implement the `CallSc` as there is an obvious use case for that one for undelegate and redeem actions, which don't require any FLIP.

state-chain/pallets/cf-funding/src/lib.rs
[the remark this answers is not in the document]
albert-llimos: As I understand it, in this pallet we use the `T::Amount` for FLIP amounts like `redeem_amount`, `redemption_fee` so I don't see why we shouldn't use that.
Fwiw, for `FlipToSCGatewayAndCall::amount` the amount will always be FLIP, while for other events such as `ScCall` the amount can be either ETH or ERC-20.

engine/src/witness/eth/sc_utils.rs
dandanlen: But is `caller` used anywhere? Why do we need it?

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -558,21 +590,7 @@ pub mod pallet {
...
-				});
-			}
-
-			Self::deposit_event(Event::Funded {
-				account_id,
-				tx_hash,
-				funds_added: amount,
-				total_balance,
-			});
+			Self::fund_sc_account(account_id, funder, amount, tx_hash);
```
Janislav, line 593: What about `fund_statechain_account`?

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -900,6 +918,56 @@ pub mod pallet {
...
+					match call
+						.clone()
+						.dispatch_bypass_filter(RuntimeOrigin::<T>::signed(caller_account_id))
+					{
+						Ok(_) => {
+							Self::deposit_event(Event::SCCallExecuted { sc_call: call, tx_hash });
+						},
+						Err(e) => {
+							log::warn!(
+								"SC call couldn't be executed. It returned an error: {:?}",
```
Janislav, line 958: We could maybe also return error here or fire an event with the error that indicates that the execution has failed instead of just swallow it and only log a warn.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -900,6 +918,56 @@ pub mod pallet {
...
+								e
+							);
+						},
+					}
+				},
+				Err(e) => {
+					log::warn!("SC call couldn't be decoded: {:?}", e);
```
Janislav, line 965: Same here.

state-chain/pallets/cf-funding/src/lib.rs
ramizhasan111: This was on purpose. We cant return an error because it will fail the extrinsic and the storages would revert. We dont want that. Even if the call fails, we want the deposit to be processed as stated in the comment above.

ramizhasan111: done.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -900,6 +918,56 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::call_index(12)]
+		#[allow(clippy::let_unit_value)]
+		#[pallet::weight(Weight::zero())]
```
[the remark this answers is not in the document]
ramizhasan111: will add

engine/src/witness/eth/sc_utils.rs
```diff
@@ -99,32 +74,108 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
+								DepositToScGatewayAndScCallFilter {
+									sender,    // eth_address to attribute the FLIP to
+									signer: _, // `tx.origin``. Not to be used for now
+									amount,    // FLIP amount deposited
+									sc_call,
+								},
+							) => pallet_cf_funding::Call::execute_sc_call {
+								deposit_and_call: EthereumDepositAndSCCall {
+									deposit: EthereumDeposit::FlipToSCGateway {
+										amount: amount.try_into().unwrap(),
```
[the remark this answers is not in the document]
dandanlen: I agree but we can make this assumption more explicit with `expect`

state-chain/pallets/cf-funding/src/lib.rs
dandanlen: What about simply `fund_account` 🙈 
The 'sc' part is kind of redundant.

dandanlen: (In fact this is true of many of the uses of 'SC')..

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +1088,54 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+		operator: T::AccountId,     // Operator the amount to delegate to
+	},
+	Undelegate {
+		delegator: EthereumAddress, // Ethereum Address of the delegator
+		operator: T::AccountId,     // Operator the amount was delegated to
+	},
+}
+
+#[derive(Clone, PartialEq, Eq, Encode, Decode, TypeInfo, DebugNoBound)]
+#[scale_info(skip_type_params(T))]
```
dandanlen, line 1106: nit: there's no T param here any more.


```suggestion
```

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -973,3 +1063,54 @@ impl<T: Config> OnKilledAccount<T::AccountId> for Pallet<T> {
...
+impl<T: Config> UnfilteredDispatchable for EthereumSCApi<T> {
+	type RuntimeOrigin = T::RuntimeOrigin;
+	fn dispatch_bypass_filter(
+		self,
+		_origin: Self::RuntimeOrigin,
+	) -> frame_support::dispatch::DispatchResultWithPostInfo {
+		match self {
+			EthereumSCApi::Delegation(delegation_api) => match delegation_api {
+				DelegationApi::Delegate { delegator: _, operator: _ } => todo!(),
+				DelegationApi::Undelegate { delegator: _, operator: _ } => todo!(),
```
dandanlen, line 1137: If you rebase against `main`, it should be possible to complete this now.

state-chain/pallets/cf-funding/src/lib.rs
ramizhasan111: yes, I have impled the emit events in the latest commit.
will do the transactional part.

dandanlen: I mean it looks like the log is still there...

state-chain/pallets/cf-funding/src/lib.rs
ramizhasan111: I will merge this into the base branch and will rebase against main. Then will impl this in a new PR against the base branch.

ramizhasan111:
> As discussed, let's finish the benchmark, then merger this into the base PR. Then, move the EthereumScApi type out of the pallet and inject it through the Config. Then, we can unit test the extrinsic using a simple mock call (like system::remark).

I ended up doing all of the above in the the current PR before merging.

PR #5983, opened 2025-07-07
fix: bouncer oracle price submission

albert-llimos:
# Pull Request

## Checklist

## Summary

Update the mock price feeds using submit instead of update_price to get around the fact that the timestamps are wrong (old) after starting the old Solana images.


MxmUrw (approved):
I'm not sure whether it works yet, but let's try to merge and debug then.

PR #5995, opened 2025-07-09
fix: avoid logging Error in case of Optimistic Election

marcellorigotti:
# Pull Request

Closes: PRO-2398

## Checklist

## Summary

When querying for optimistic blocks the RPC will always fail for non-existing block, since we are querying optimistically this is expected to happen very often. Changed the logic to query only if the block is actually new and return None otherwise.

### DIFF
--- a/engine/src/witness/btc.rs
+++ b/engine/src/witness/btc.rs
@@ -162,7 +162,7 @@ async fn query_election_block<C: ChainTypes>(
 	client: &BtcCachingClient,
 	block_height: btc::BlockNumber,
 	election_type: EngineElectionType<C>,
-) -> Result<(Vec<VerboseTransaction>, Option<btc::Hash>)>
+) -> Result<Option<(Vec<VerboseTransaction>, Option<btc::Hash>)>>
 where
 	ChainBlockHashOf<C>: AsRef<[u8]>,
 {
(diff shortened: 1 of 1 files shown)


engine/src/witness/btc.rs
```diff
@@ -199,8 +209,11 @@ impl VoterApi<BitcoinDepositChannelWitnessingES> for BitcoinDepositChannelWitnes
 			block_height, properties: deposit_addresses, election_type, ..
 		} = properties;
 
-		let (txs, response_block_hash) =
-			query_election_block(&self.client, block_height, election_type).await?;
+		let Some((txs, response_block_hash)) =
+			query_election_block(&self.client, block_height, election_type).await?
+		else {
+			return Ok(None)
+		};
```
MxmUrw, line 216: Ok cool, I never saw `let .. else {}` before!

engine/src/witness/btc.rs
```diff
@@ -162,23 +162,33 @@ async fn query_election_block<C: ChainTypes>(
...
+				let block_hash = client.best_block_hash().await?;
+				let best_block_header = client.block_header(block_hash).await?;
+				if best_block_header.height != block_height {
+					return Ok(None)
+				}
+				block_hash
+			} else {
+				client.block_hash(block_height).await?
+			};
+
```
MxmUrw, line 186: At the price of querying twice, I would have thought we can do it like this:
```rust
			// check whether a block exists with the given height
			if submit_hash {
				let block_hash = client.best_block_hash().await?;
				let best_block_header = client.block_header(block_hash).await?;
				if best_block_header.height < block_height {
					return Ok(None)
				}
			}

			let block_hash = client.block_hash(block_height).await?;
			let block = client.block(block_hash).await?;
			Ok(Some((
				block.txdata,
				if submit_hash { Some(block.header.hash.to_byte_array().into()) } else { None },
			)))
```
Then it'd potentially also work when there are 2 new blocks appearing immediately on chain.

The current version would not vote in such a case because best_height and requested height wouldn't match. Of course this case is unlikely and the consequence is just that we'll vote one SC block later.

PR #5997, opened 2025-07-11
fix: bouncer flakiness in observe-event

dandanlen:
Trying to fix the flakiness in observeEvent in the bouncer. 

Some of things that are changed:

- Make sure we always fill in any gaps between the subscription and historical blocks (explicit link between historical query and 'future' subscription).
- Cache by block hash rather than height (more re-org-resistant).
- Trace back through historical blocks by parent hash chain rather than block height lookup (similar reason).
- Use the AllHeads sub instead of NewHeads to ensure we see all blocks (not clear from the docs, but NewHeads might only return new best blocks and therefore might skip blocks).
 - Add a default value for `historicalCheckBlocks` to ensure we always check *some* already-seen blocks. (been experimenting with this a bit: too high causes some flaky tests to fail because we catch some old events; too low exposes a race condition where we subscribe too late and miss events). 
 - Moved the substrate header subscriptions into a background thread. Now there are two subs: one for finalized and another for all headers. This allows us to correctly update the finalized height without further queries. It also means the EventCache can provide its own subscription abstraction so that we can share the WS subscription, meaning less contention on this resource, and fewer opportunities for data races since fewer consumers are writing to the cache.

- Added a util to wait for extrinsic submission. Use it to simplify and reduce flakiness of some tests, for example when we submitted an extrinsic and then use observe-event to detect success: now we simply wait for the extrinsic to succeed 'inBlock'.
- Tried to avoid using the any type where possible.


dandanlen:
FWIW there is still some remaining flakiness, which seems to be related to submitting transactions to external chains, particularly Assethub and Solana: It seems the engine's rpc call can sometimes fail (even with the default 5 retries in the retrier-rpc module). And in this case, since it's a single-node network, there are no further validators to be nominated to retry the broadcast and the test times out. This isn't really something we can fix in the bouncer... we might have to investigate why the rpc requests / tx submissions are failing.

bouncer/commands/run_test.ts
```diff
@@ -77,5 +77,5 @@ if (!matchingTestName) {
   console.error('No matching test function found');
   process.exit(1);
 } else {
-  execSync(`pnpm vitest --maxConcurrency=100 run -t "${matchingTestName}"`, { stdio: 'inherit' });
+  execSync(`pnpm vitest --maxConcurrency=500 run -t "${matchingTestName}"`, { stdio: 'inherit' });
```
dandanlen: Setting the value this high causes Solana swaps to fail (I have discussed this with @MxmUrw , we think it's related to how engines process retries: if there is only one engine in the network, broadcasts are aborted much faster). 

I've reverted the concurrency back to 100. We can look at changing this selectively, maybe we can reduce it for Solana tests but keep it higher for other tests...

bouncer/tests/all_swaps.ts
```diff
@@ -111,8 +114,9 @@ export function testAllSwaps(timeoutPerSwap: number) {
   appendSwap('ArbEth', 'HubUsdc', testVaultSwap);
   appendSwap('ArbEth', 'HubUsdt', testVaultSwap);
 
-  manuallyAddTestToList('AllSwaps', 'testAllSwaps');
-  for (const swap of allSwaps) {
-    concurrentTest(swap.name, swap.test, timeoutPerSwap, true);
-  }
+  describe('AllSwaps', () => {
```
j4m1ef0rd, line 117: This stops the swap tests from running at the same time as the other tests. That's why I put it in fast_boucer.ts.

dandanlen: Ok I see... The vitest docs mention that nested describe is possible, but i didn't realise that it prevented the inner tests from running concurrently with tests from the parent suite. Will fix.

PR #5998, opened 2025-07-11
feat: add min bid to RPCs

dandanlen:
# Pull Request

Closes: PRO-2305 (see final comment)

## Summary

Exposes the min bid via RPC call.


dandanlen:
@acdibble this should be identical to the change added to the release branch, just needs an approval 🙏

PR #6001, opened 2025-07-15
feat: slashes and rewards are shared with delegators

syan095:
# Pull Request

Closes: PRO-2373

## Checklist

## Summary
* Block authoring rewards, and slashings are distributed among the delegators associated with the registered operator. Distribution% is proportional to the number of Flips delegated to that operator.
* Fixed integration tests for moving across multiple epochs
* Removed Backup validator rewards

### DIFF
--- a/state-chain/cf-integration-tests/src/lib.rs
+++ b/state-chain/cf-integration-tests/src/lib.rs
@@ -24,6 +24,7 @@ mod threshold_signing;
 
 mod account;
 mod authorities;
+mod delegation;
 mod fee_scaling;
 mod funding;
 mod genesis;
@@ -34,13 +35,10 @@ mod swapping;
 mod trading_strategy;
 mod witnessing;
 
-use cf_chains::{
-	eth::Address as EthereumAddress, evm::EvmTransactionMetadata, TransactionMetadata,
-};
+use cf_chains::eth::Address as EthereumAddress;
 use cf_primitives::{AuthorityCount, BlockNumber, FlipBalance};
 use cf_traits::EpochInfo;
 use frame_support::{assert_noop, assert_ok, sp_runtime::AccountId32, traits::OnInitialize};
-use pallet_cf_broadcast::AwaitingBroadcast;
 use pallet_cf_funding::EthTransactionHash;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
 use sp_consensus_grandpa::AuthorityId as GrandpaId;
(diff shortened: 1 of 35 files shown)


dandanlen (commented):
Some early comments.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -669,50 +690,3 @@ impl<T: Config> OnKilledAccount<T::AccountId> for BurnFlipAccount<T> {
...
-				amount: slash_amount,
-			});
-		}
-	}
-}
-
-impl<T: Config> Slashing for FlipSlasher<T>
-where
-	BlockNumberFor<T>: Into<T::Balance>,
-{
```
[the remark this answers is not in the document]
syan095: Will there be a situation where we want to slash a Node but not its associated delegators? because I was under the assumption that we would never want to do this. This is why the logic was added to FlipSlasher directly, instead of adding a new DelegateSlasher. 

Current logic supports slashing/rewarding someone with no delegator associated.

As to event, I can revert it so the slashing and reward events are deposited separately (instead of a single aggregated event)

I thought it was nice to have the reward and slashing in the same place. I can try to move both into the Validator pallet, but I don't really see an advantage of having them in the Validator pallet instead of in the Runtime.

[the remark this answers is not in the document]
syan095: I've moved the delegator slasher into the Validator pallet. But currently, it has to be closely coupled with the Flip pallet. Do you think this is OK? (see `DelegatorFlipSlasher`)
or would it be better to have the Delegator take in a Slasher generic type? i.e.
```pub struct DelegatorSlasher<T: Config, Slahser: Slasher>;

impl Config for Runtime {
     type Slasher = DelegatorSlasher<Runtime, FlipSlasher<Runtime>>;
}
```

I can try to decouple Validator's dependency on the Flip pallet.

state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -52,3 +60,101 @@ pub struct OperatorSettings {
...
+
+/// Distribute a settlement to a given validator for the current Epoch.
+/// The total amount is shared among all delegators associated with the operator controlling this
+/// validator.
+pub fn distribute_among_delegators<T: Config + pallet_cf_flip::Config>(
+	validator: &T::AccountId,
+	total: T::Balance,
+	settle: impl Fn(&T::AccountId, T::Balance),
+) {
+	if let Some((operator_fee, delegator_fees)) = crate::ManagedValidators::<T>::get(validator)
```
ramizhasan111, line 112: maybe a better name here would be ```delegator_cuts``` like in split_amount function.

state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -52,3 +60,101 @@ pub struct OperatorSettings {
...
+			split_amount(total, delegators, setting.fee_bps).ok()
+		}) {
+		settle(validator, operator_fee);
+		for (delegator, fees) in delegator_fees.iter() {
+			settle(delegator, *fees);
+		}
+	} else {
+		settle(validator, total);
+	}
+}
```
ramizhasan111, line 127: this will distribute all the rewards/slash among just the delegators of that operator. But the validators of this operator also have their own balance. The rewards/slash need to be distributed protional among the validator's balance also. 

Basically we need to find the total validator balance for all validators of this operator. Then divide the reward among total_staked by delegators and total validator balance propotionally. Then divide the delegator portion proportionally among delegators, substracting the fees, as we do here. Then credit the operator account with (validator prortion + fee from delegators)

state-chain/cf-integration-tests/src/lib.rs
```diff
@@ -101,106 +99,3 @@ pub fn witness_call(call: RuntimeCall) {
 		));
 	}
 }
-
-/// this function witnesses the rotation tx broadcasts for all chains. It takes in the broadcast ids
-/// of the rotation tx in the following chains order: Ethereum, Polkadot, Bitcoin, Arbitrum, Solana,
-/// Assethub (the order in which these chains were integrated in chainflip)
-pub fn witness_rotation_broadcasts(broadcast_ids: [cf_primitives::BroadcastId; 6]) {
```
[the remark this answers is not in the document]
ramizhasan111: discussed irl.

dandanlen:
FYI I'm going to rebase this against #6048 and untangle some of the changes.

dandanlen:
@ramizhasan111 this should be ready now, major changes are now split into individual commits.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -666,47 +672,22 @@ impl<T: Config> OnKilledAccount<T::AccountId> for BurnFlipAccount<T> {
...
 		account_id: &Self::AccountId,
-		blocks: Self::BlockNumber,
+		blocks_offline: Self::BlockNumber,
 	) -> Self::Balance {
 		let account = Account::<T>::get(account_id);
-		(SlashingRate::<T>::get() * account.bond / T::BlocksPerDay::get().into())
-			.saturating_mul(blocks.into())
+		(SlashingRate::<T>::get() * account.bond /
+			T::BlocksPerDay::get().into().saturated_into::<u128>().into())
+		.saturating_mul(blocks_offline.into().saturated_into::<u128>().into())
```
ramizhasan111, line 691: why this change?

Also, while we are here I think we should do all multiplications first and do division at the end to get more accurate value.

dandanlen: I moved it back into the trait. Otherwise we need to expose the Flip pallet itself in order to implement this (or implement it in the chainflip module).

I agree it's a bit ugly, i'll simplify.

state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -52,3 +63,203 @@ pub struct OperatorSettings {
...
+pub fn distribute<T: Config>(
+	validator: &T::AccountId,
+	total: T::Amount,
+	settle: impl Fn(&T::AccountId, T::Amount),
+) {
+	if let Some(snapshot) = DelegationResolver::<T>::resolve_for_account(validator) {
+		snapshot.distribute(total).for_each(|(account, amount)| settle(account, amount));
+	} else {
+		settle(validator, total);
+	}
```
ramizhasan111, line 204: given current logic for ```resolve_for_account```, this will return a ```Some(Snapshot)``` if the input is an operator account which shouldnt happen, it should return None. I know this case wont happen in reality but it should logically not do that. This function should basically check if the input validator is present in any of the operator's snaphots. If yes, return that snapshot, if not, return ```None```.

state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -198,12 +197,69 @@ pub fn distribute<T: Config>(
...
+				validator.into_ref(),
+				Self::Ref(snapshot.operator.clone()),
+			);
+		}
+		DelegationSnapshots::<T>::insert(
+			epoch_index,
+			snapshot.operator.clone(),
+			Self::Own(snapshot),
+		);
+	}
```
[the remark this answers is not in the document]
dandanlen: The thing is, when we trigger a slash or reward, we first need to check if the validator is associated with an operator. 

Using the suggested approach we need to decode the mapping of all validators, and the mapping of all delegations, just to check if we even need that data. 

I can see why the implementation is a little confusing because it 'squashes' both the epoch->validator->operator and epoch->operator->snapshot mappings into one. I do think it's provably correct :) but I'll split them up so that we have separate maps for each.

state-chain/pallets/cf-validator/src/mock.rs
```diff
@@ -225,28 +225,58 @@ macro_rules! assert_invariants {
...
 		// validator.
-		let current_epoch = ValidatorPallet::current_epoch();
-		for authority in ValidatorPallet::current_authorities() {
-			if let Some(snapshot) = DelegationSnapshots::<Test>::get(current_epoch, authority) {
-				assert!(
-					snapshot.validators.contains_key(&authority) && snapshot.delegators.is_empty(),
-					"Invalid Validator snapshot {:?} for authority {:?} at block {:?}",
-					snapshot,
-					authority,
+		let all_managed_validators = DelegationSnapshots::<Test>::iter_prefix(
```
ramizhasan111, line 228: the comment here is outdated.

ramizhasan111 (approved):

PR #6006, opened 2025-07-16
feat: delegate and undelegate

Janislav:
# Pull Request

Closes: PRO-2404

## Checklist

## Summary

Done in this PR:

- Added delegate and undelegate functionality
- Refactored operator signup
- Refactor block/allow of delegator to Exceptions

### DIFF
--- a/state-chain/pallets/cf-validator/src/benchmarking.rs
+++ b/state-chain/pallets/cf-validator/src/benchmarking.rs
@@ -474,9 +474,9 @@ mod benchmarks {
 		.unwrap();
 
 		#[extrinsic_call]
-		update_operator_settings(RawOrigin::Signed(caller.clone()), OperatorSettings::default());
+		update_operator_settings(RawOrigin::Signed(caller.clone()), OPERATOR_SETTINGS);
 
-		assert_eq!(OperatorSettingsLookup::<T>::get(caller), Some(OperatorSettings::default()));
+		assert_eq!(OperatorSettingsLookup::<T>::get(caller), Some(OPERATOR_SETTINGS));
 	}
 
 	#[benchmark]


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -801,6 +874,216 @@ pub mod pallet {
...
+
+		#[pallet::call_index(18)]
+		#[pallet::weight(T::ValidatorWeightInfo::delegate())]
+		pub fn delegate(origin: OriginFor<T>, operator_id: T::AccountId) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+
+			ensure!(
+				!ManagedDelegations::<T>::contains_key(&account_id),
+				Error::<T>::AlreadyDelegating
+			);
```
dandanlen, line 1041: We don't need this condition. It's ok to change delegation to another operator.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -801,6 +874,216 @@ pub mod pallet {
...
+
+			ensure!(
+				T::AccountRoleRegistry::has_account_role(&operator_id, AccountRole::Operator),
+				Error::<T>::NotOperator
+			);
+
+			// TODO: We should somehow ensure the operator has set his preferences...
+			match OperatorParameters::<T>::get(&operator_id)
+				.expect("an operator to have configured his parameters")
+				.delegation_acceptance
```
dandanlen, line 1056: Why not just error? 

Alternatively, we could:
- require operators to provide settings when they register
- default delegation acceptance to 'not allowed'
- 
I think i prefer the second option.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -345,6 +346,36 @@ pub mod pallet {
 	#[pallet::getter(fn active_bidder)]
 	pub type ActiveBidder<T: Config> = StorageValue<_, BTreeSet<T::AccountId>, ValueQuery>;
 
+	/// Maps an delegator to an associated operator account.
+	#[pallet::storage]
+	pub type ManagedDelegations<T: Config> =
```
dandanlen, line 351: `ManagedDelegations` is not the best name for this...

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -801,6 +874,216 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Executed by an operator to block a delegator from delegating to the operator.
+		#[pallet::call_index(14)]
+		#[pallet::weight(T::ValidatorWeightInfo::block_delegator())]
+		pub fn block_delegator(origin: OriginFor<T>, delegator_id: T::AccountId) -> DispatchResult {
+			let operator_id = T::AccountRoleRegistry::ensure_operator(origin)?;
+
+			// TODO: Check of delegator is currently delegation to this operator and error if so.
```
dandanlen, line 964: I don't think it would be an error - we would just remove the delegator->operator mapping.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -801,6 +874,216 @@ pub mod pallet {
...
+
+			AllowedDelegators::<T>::mutate(&operator_id, |delegators| {
+				delegators.remove(&delegator_id);
+			});
+
+			BlockedDelegators::<T>::mutate(&operator_id, |delegators| {
+				delegators.insert(delegator_id.clone());
+			});
```
dandanlen, line 972: We don't always need to update both of these. It depends on the default preference:

```pseudo
match default_preference {
    Allow => // add to block list,
    Block => // remove from allow list
}
```

state-chain/pallets/cf-validator/src/lib.rs
Janislav: I think the problem with 2 is that it's not only about the delegation acceptance, but also about the fee, and I don't think we really can provide a default for that. Probably the best way is to force it during the registration.

dandanlen: Basically, as an operator, you need to:
- register
- set fee
- claim validators
- either: change preference to 'allow' or: add whitelisted accounts

I think this is fine, we don't need to force any of these decisions during registration.

state-chain/pallets/cf-validator/src/lib.rs
Janislav: What about just `Delegations`?

state-chain/pallets/cf-validator/src/lib.rs
Janislav: But "deny" is not about not accepting any delegations, it's about only accepting actively allowed delegators, right? And in any case, we don't want to have an operator that is offering a 0 percent fee, so somehow we have to force them actively choosing one. I think what we should do here is either forcing them to set it during registration or error if there is no value present in the `delegate` extrinsic, so they have to set it.

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: I think DelegationChoice or DelegationPreference. 

`Delegations` can be reserved for the actual Delegation (ie. the struct with all the info about who is in the operator's delegation set for the current epoch).

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: Block/Allow means something different depending on what your default is. 

If your default is to allow, then blocking means adding to the block list. 
If your default is to block, then blocking means removing from the allow list. 

We don't always need to do both. If your default is block, your block list should be empty. If your default is allow, your allow list should be empty.

(In fact we don't need separate storage items for these. It could just be 'Exceptions' or something.)

Janislav: Possible but a bit strange because you would block everyone who was explicit allowed before and vice versa when you change the allowance mode...

Janislav: I refactored it to only have one StorageItem "Exceptions". Maybe I have to revisit some assumptions now, especially around if some events or test could be obsolete. Other than that I think we should pretty good to go.

Janislav:
@coderabbit review

state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -2,6 +2,14 @@ use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
 
+/// The minimum operator fee we charge.
```
dandanlen, line 5: nit: "we" don't charge it. Also could be useful to clarify that it's in bps.


```suggestion
/// The minimum delegation fee that can be charged, in basis points.
```

Janislav:
@dandanlen all your comments are addressed.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -999,18 +1034,89 @@ pub mod pallet {
...
+					!Exceptions::<T>::get(&operator_id).contains(&account_id),
+					Error::<T>::DelegatorRemovedFromExceptions
+				),
+				DelegationAcceptance::Deny => ensure!(
+					Exceptions::<T>::get(&operator_id).contains(&account_id),
+					Error::<T>::DelegatorRemovedFromExceptions
+				),
+			}
+
+			DelegationChoice::<T>::mutate(account_id.clone(), |maybe_operator| {
```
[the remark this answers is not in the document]
Janislav: No. I assume because we have to use it inside the Clojure

dandanlen (approved):

PR #6013, opened 2025-07-22
 chore: process all events in a block in parallel

ramizhasan111:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

### DIFF
--- a/engine/src/witness/btc/deposits.rs
+++ b/engine/src/witness/btc/deposits.rs
@@ -105,6 +105,7 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
 					})
 				};
 
+				let mut process_calls = vec![];
 				for (broker_id, channel_id, vault_address) in vault_addresses {
 					for tx in &txs {
 						if let Some(deposit) = super::vault_swaps::try_extract_vault_swap_witness(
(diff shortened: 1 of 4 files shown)


engine/src/witness/btc/deposits.rs
```diff
@@ -113,19 +114,20 @@ impl<Inner: ChunkedByVault> ChunkedByVaultBuilder<Inner> {
...
 								.into(),
 								epoch.index,
-							)
-							.await;
+							));
 						}
 					}
 				}
 
+				futures::future::join_all(process_calls).await;
```
msgmaxim, line 129: Just curious to know if there is any context for this, like was sequential processing causing some issues? I'm a little concerned about the code inside `process_call` which essentially simply submits an extrinsic, and if now all of them are submitted in parallel, I think you will end up with some nonce collisions, correct? (I know that the SC client should be able to handle them, but still, how do we know that this will be an improvement?)

msgmaxim: That's fair, though a short description would have still been nice ;) I'm listed as a reviewer (I know this was done automatically through code owners), but it is hard to review without any context.

msgmaxim (approved):

engine/src/witness/btc/deposits.rs
ramizhasan111: fait point. I meant to open it as a draft PR and then open for review once I put the context. I guess I accidently marked as ready for review.

PR #6014, opened 2025-07-24
feat: CFE check for CF chain node connection

syan095:
# Pull Request

Closes: PRO-1953

## Checklist

## Summary
When CFE is launched, it will try to check for CFE version - something unique to CF nodes. This ensures Engine is connecting to a valid CF node, not other Substrate-based chains by mistake

If an invalid CFE version is returned, the CFE exits early.

To reproduce the error and test the fix, replace `$SC_RPC_PORT` with `9947` in `localnet/init/scripts/start-engine.sh`

### DIFF
--- a/engine/src/state_chain_observer/client.rs
+++ b/engine/src/state_chain_observer/client.rs
@@ -736,10 +736,19 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 		)?);
 		let signer = signer::PairSigner::<sp_core::sr25519::Pair>::new(pair.clone());
 
+		// Pre-check to verify if the correct chain is connected.
+		if base_rpc_client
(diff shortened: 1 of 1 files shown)


engine/src/state_chain_observer/client.rs
```diff
@@ -736,10 +736,19 @@ impl SignedExtrinsicClientBuilderTrait for SignedExtrinsicClientBuilder {
 		)?);
 		let signer = signer::PairSigner::<sp_core::sr25519::Pair>::new(pair.clone());
 
+		// Pre-check to verify if the correct chain is connected.
+		if base_rpc_client
+			.storage_value::<pallet_cf_environment::CurrentReleaseVersion<state_chain_runtime::Runtime>>(
+				finalized_block_stream.cache().hash,
+			)
+			.await? == Default::default()
```
j4m1ef0rd, line 744: I don't understand what default is in this situation. is it the current version or zeros?

PR #6019, opened 2025-07-25
feat: don't error if tx is Confirmed

marcellorigotti:
# Pull Request

Closes: PRO-2424

## Checklist

## Summary

When witnessing solana egress we shouldn't throw an error in case the tx is confirmed and waiting for finalization (it is just confusing for NOs)


dandanlen:
> I logged the message as DEBUG, but IMO it can be removed completely. lmk what do you think

No strong opinion, I think debug is fine.

PR #6020, opened 2025-07-25
feat: deprecate polkadot

nmammeri:
# Pull Request

Closes: PRO-2388

## Checklist

## Summary

* This PR implements the required logic to transfer all state chain DOT balances to  HubDot balances:
* Implemented as a runtime migration that:
    * Cancels all current limit and range orders in the DOT/USDC pool
    * Then initiates the transfer of balances.
* Note it assumes that Dot deposit channel creation should be disabled using safe mode during the runtime upgrade
* This PR also adds a governance call to sign and broadcast an `xcm::limited_teleport_assets` tx to polkadot,
teleporting the polkadot vault contents to our assethub vault. It should be triggered shortly after the runtime upgrade goes through successfully.

### DIFF
--- a/bouncer/tests/minimum_deposit.ts
+++ b/bouncer/tests/minimum_deposit.ts
@@ -1,8 +1,8 @@
 import { requestNewSwap } from 'shared/perform_swap';
 import { setMinimumDeposit } from 'shared/set_minimum_deposit';
 import { observeEvent } from 'shared/utils/substrate';
-import { sendDot } from 'shared/send_dot';
 import { TestContext } from 'shared/utils/test_context';
+import { sendHubAsset } from 'shared/send_hubasset';
 
 export async function testMinimumDeposit(testContext: TestContext) {
   const logger = testContext.logger;
@@ -12,12 +12,12 @@ export async function testMinimumDeposit(testContext: TestContext) {
     await requestNewSwap(logger, 'HubDot', 'Eth', '0xd92bd8c144b8edba742b07909c04f8b93d875d93')
   ).depositAddress;
   const depositFailed = observeEvent(logger, ':DepositFailed');
-  await sendDot(depositAddress, '19');
+  await sendHubAsset(testContext.logger, 'HubDot', depositAddress, '19');
   logger.debug('Sent 19 DOT');
   await depositFailed.event;
   logger.debug('Deposit was ignored');
   const depositSuccess = observeEvent(logger, ':DepositFinalised');
-  await sendDot(depositAddress, '21');
+  await sendHubAsset(testContext.logger, 'HubDot', depositAddress, '21');
   logger.debug('Sent 21 DOT');
   await depositSuccess.event;
   logger.debug('Deposit was successful');


state-chain/runtime/src/migrations/polkadot_deprecation.rs
```diff
@@ -0,0 +1,136 @@
...
+///    1. Canceling all Polkadot orders in the Dot/USDC pool
+///    2. Converting all Polkadot balances to Assethub balances
+///
+/// Note that this call should be run after disabling Dot ingress using safemode and waiting
+/// for all Dot deposit channels to expire
+impl OnRuntimeUpgrade for PolkadotDeprecationMigration {
+	fn on_runtime_upgrade() -> Weight {
+		let mut weight: Weight = Weight::zero();
+
+		let dot_pool = pallet_cf_pools::pallet::Pools::<Runtime>::get(
```
martin-chainflip, line 26: why this extra step? If there is no dot pool, then we will get an error later anyways.

state-chain/runtime/src/migrations/polkadot_deprecation.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// Allows deprecating Polkadot. This will trigger:
+///    1. Canceling all Polkadot orders in the Dot/USDC pool
+///    2. Converting all Polkadot balances to Assethub balances
+///
+/// Note that this call should be run after disabling Dot ingress using safemode and waiting
+/// for all Dot deposit channels to expire
+impl OnRuntimeUpgrade for PolkadotDeprecationMigration {
+	fn on_runtime_upgrade() -> Weight {
+		let mut weight: Weight = Weight::zero();
```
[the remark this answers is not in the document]
nmammeri: I did it because this migration might consume more weights depending on how many orders and Dot accounts we have. I think is better to be complete and return estimated weight instead of zero weight.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -2430,3 +2430,51 @@ impl<T: Config> cf_traits::PoolPriceProvider for Pallet<T> {
 		})
 	}
 }
+
+impl<T: Config> cf_traits::PoolOrdersManager for Pallet<T> {
```
martin-chainflip, line 2434: i don't think we will need such a function again any time soon. As such it might be easier to put the code directly into the migration

state-chain/runtime/src/migrations/polkadot_deprecation.rs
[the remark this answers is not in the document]
martin-chainflip: weights are only used for transaction fees. this is a runtime update and pays no transaction fees.

state-chain/runtime/src/migrations/polkadot_deprecation.rs
martin-chainflip: but it errors in both cases?

state-chain/runtime/src/migrations/polkadot_deprecation.rs
```diff
@@ -0,0 +1,136 @@
...
+
+		log::info!("🍩 Transferring all Dot balances to HubDot balances");
+		for (account_id, asset, amount) in FreeBalances::<Runtime>::iter() {
+			if asset == Asset::Dot && amount > AssetAmount::zero() {
+				FreeBalances::<Runtime>::mutate(account_id.clone(), Asset::HubDot, |balance| {
+					*balance = balance.saturating_add(amount);
+				});
+				weight += <Runtime as frame_system::Config>::DbWeight::get().reads_writes(1, 1);
+
+				FreeBalances::<Runtime>::mutate_exists(account_id, asset, |balance| {
```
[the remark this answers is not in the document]
nmammeri: done

state-chain/runtime/src/migrations/polkadot_deprecation.rs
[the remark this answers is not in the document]
nmammeri: I thought the weight is also during block production to estimate how long the runtime upgrade is going to take. Anyway, I can remove the weight estimation and return zero weight.

state-chain/pallets/cf-pools/src/lib.rs
nmammeri: On the linear project there is comment stating that we might deprecate AssetHub itself if there is no liquidity. This is why I thought we might used it again

state-chain/runtime/src/migrations/polkadot_deprecation.rs
martin-chainflip: ah, i get what you mean.

MxmUrw:
Ready to review now.

dandanlen (approved):

PR #6021, opened 2025-07-28
…

syan095:
# Pull Request

Closes: PRO-2234

## Checklist

## Summary
Changed some LP rpc call's return type to BlockUpdate - which contains block info, such as block number, hash etc.

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -73,7 +73,7 @@ use sp_runtime::{
 };
 use sp_state_machine::InspectState;
 use state_chain_runtime::{
-	chainflip::{BlockUpdate, Offence},
+	chainflip::{get_header_timestamp, BlockUpdate, Offence},
 	constants::common::TX_FEE_MULTIPLIER,
 	runtime_apis::{
 		AuctionState, BoostPoolDepth, BoostPoolDetails, BrokerInfo, CcmData, ChainAccounts,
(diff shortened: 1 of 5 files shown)


j4m1ef0rd (approved):
What do we do about this breaking change? @dandanlen

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1086,6 +1086,35 @@ where
...
+		data: RpcResult<Data>,
+		at: Option<Hash>,
+	) -> RpcResult<BlockUpdate<Data>> {
+		let block_hash = self.rpc_backend.unwrap_or_best(at);
+		let block_number = self
+			.rpc_backend
+			.client
+			.number(block_hash)
+			.map_err(|e| call_error(e, CfErrorCode::OtherError))?
+			.ok_or(internal_error(format!("Could not fetch block number for block {:?}", at)))?;
```
dandanlen, line 1101: It would be good to understand the circumstances that can cause this to fail: what if the rpc is called on a block that is no longer in the block cache? (Like on a light rpc node).

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1086,6 +1086,35 @@ where
...
+		let header = self
+			.rpc_backend
+			.client
+			.header(block_hash)
+			.map_err(|e| call_error(e, CfErrorCode::OtherError))?
+			.ok_or_else(|| {
+				internal_error(format!("Could not fetch block header for block {:?}", at))
```
dandanlen, line 1108: Same here - not sure we always want to error. It might be better to optimistically return as much metadata as we can rather than erroring.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1086,6 +1086,35 @@ where
...
 		Self { rpc_backend: CustomRpcBackend::new(client, backend, executor) }
 	}
+
+	pub fn into_block_update<Data>(
+		&self,
+		data: RpcResult<Data>,
+		at: Option<Hash>,
+	) -> RpcResult<BlockUpdate<Data>> {
+		let block_hash = self.rpc_backend.unwrap_or_best(at);
+		let block_number = self
```
dandanlen, line 1096: This is repeating/duplicating some work:
- when we call `with_runtime_api` we already have the hash
- the header also contains the hash
- the header might also contain the height? 

I would like us to think a little harder about how to integrate the block wrapper nicely, without minimsl futher backend queries.

PR #6023, opened 2025-07-28
fix: remove race condition in bouncer observe event

nmammeri:
# Pull Request

Closes: PRO-xxx

## Checklist

## Summary

* Fixes a race condition in bouncer's `observeEvent` function between checking for the event in historic blocks and in the  new block from a subscription


MxmUrw (commented):
Looks good to me, the change moves the check for historical blocks into the loop that checks new blocks coming in from the subscription. This way we won't accidentally miss a block height. (Which could have happened previously)

j4m1ef0rd (approved):
This will conflict with a lot with #5997 which i'm still reviewing.
Find to merge this now and replicate the change on the other PR later.

PR #6026, opened 2025-07-30
chore: ensure all rpcs have "at" field

syan095:
# Pull Request

Closes: PRO-1368

## Summary
Added "at" to 2 rpc calls so all rpc calls (except Subscribe ones) contains the "at" field.

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -630,113 +630,92 @@ type BoostPoolFeesResponse = Vec<boost_pool_rpc::BoostPoolFeesRpc>;
 pub trait CustomApi {
 	/// Returns true if the current phase is the auction phase.
 	#[method(name = "is_auction_phase")]
-	fn cf_is_auction_phase(&self, at: Option<state_chain_runtime::Hash>) -> RpcResult<bool>;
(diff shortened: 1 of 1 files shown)


j4m1ef0rd (approved):

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1839,24 +1782,21 @@ where
...
-	) -> RpcResult<BlockUpdate<OrderFills>> {
+	fn cf_lp_get_order_fills(&self, at: Option<Hash>) -> RpcResult<BlockUpdate<OrderFills>> {
 		order_fills::order_fills_for_block(
 			self.rpc_backend.client.as_ref(),
 			at.unwrap_or_else(|| self.rpc_backend.client.info().finalized_hash),
 		)
 	}
 
-	fn cf_supported_assets(&self) -> RpcResult<Vec<Asset>> {
+	fn cf_supported_assets(&self, _at: Option<Hash>) -> RpcResult<Vec<Asset>> {
```
dandanlen, line 1792: We've added the param but we ignore it - in fact based on how this is implemented (ie. in the rpc layer), we can't support this for past runtime versions... We should push the implementation into a runtime rpc.

dandanlen (changes requested):

PR #6028, opened 2025-07-30
feat: configurable bitcoin fee polling interval [pick to 1.10]

MxmUrw:
# Pull Request

Related: PRO-2443

## Checklist

## Summary

Since 1.10 we use the `UnsafeMedian` electoral system for witnessing bitcoin fees. This creates new elections on every SC block. Unfortunately, the bitcoin rpc call that we use to get fees (`estimate_smart_fee`) only very rarely updates it's value (every 2 btc blocks or so), so it doesn't make sense to query the rpc and vote on every SC block.

This PR adds a new (unsynchronised) setting to the UnsafeMedian ES that lets us specify the number of SC blocks to wait until the next election is created.

### DIFF
--- a/state-chain/runtime/src/chainflip/bitcoin_elections.rs
+++ b/state-chain/runtime/src/chainflip/bitcoin_elections.rs
@@ -612,7 +612,7 @@ pub fn initial_state() -> InitialStateOf<Runtime, BitcoinInstance> {
 				safety_margin: 0,
 				safety_buffer: BITCOIN_MAINNET_SAFETY_BUFFER,
 			},
-			2, // wait 2 SC blocks until reopening fee election
+			10, // wait 10 SC blocks until reopening fee election
 			(),
 		),
 		settings: (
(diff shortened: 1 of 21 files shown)


state-chain/runtime/src/chainflip/bitcoin_elections.rs
```diff
@@ -611,7 +612,7 @@ pub fn initial_state() -> InitialStateOf<Runtime, BitcoinInstance> {
 				safety_margin: 0,
 				safety_buffer: BITCOIN_MAINNET_SAFETY_BUFFER,
 			},
-			Default::default(),
+			2, // wait 2 SC blocks until reopening fee election
```
marcellorigotti, line 615: Since the value changes every ~2 btc blocks maybe we can increase this a bit?

marcellorigotti: perfect! I think 10 is enough!

PR #6029, opened 2025-07-31
feat: spam protection for block/allow delegator

Janislav:
# Pull Request

Closes: PRO-2445

## Checklist

## Summary

Sets a limit of a storage item to preven spam.


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1007,9 +1009,14 @@ pub mod pallet {
...
+					Exceptions::<T>::try_mutate(&operator, |blocked| {
+						if blocked.len() <= EXCEPTIONS_LIMIT {
+							blocked.insert(delegator.clone());
+							Ok(())
+						} else {
+							Err(Error::<T>::ExceptionsLimitReached)
+						}
+					})?;
 				},
 			}
```
ramizhasan111, line 1022: I maybe think we should keep storages for allow lists and block lists separate since it would be easier to manage. For example, currently, if an operator changes his status from allow to deny his block list will get deleted and if he wants to allow again, he would have to add the whole block list again. Having separate storages would allow the lists to persist accross deny and allow changes.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1007,9 +1009,14 @@ pub mod pallet {
 				DelegationAcceptance::Allow => {
 					// If the operator is set to allow, exceptions are the delegators that are
 					// blocked.
-					Exceptions::<T>::mutate(&operator, |blocked| {
-						blocked.insert(delegator.clone());
-					});
+					Exceptions::<T>::try_mutate(&operator, |blocked| {
+						if blocked.len() <= EXCEPTIONS_LIMIT {
```
ramizhasan111, line 1013: this should be ```<``` instead of ```<=```

dandanlen:
> > I think it makes sense to set a limit for the allow-list but not for the block-list. For the block list we can simply require that the account exists.
> > @ramizhasan111 WDYT?
> 
> what would be the argument for treating block lists and allow lists differently? Also, I think one of the arguments in the issue was that we should be able to block accounts that not yet exist since we would already know of an eth address we want to blacklist (since its on some blacklist), even before the funds hit the sc.

It's just that blocking a non-existing address makes no sense, whereas *allowing* a non-existent address does seem useful (ie. you whitelist some addresses that you know are going to be staked). Either way, if we want to treat both the same, I think it makes more sense to require that the address exists than to impose some arbitrary limit.

ramizhasan111:
> It's just that blocking a non-existing address makes no sense, whereas _allowing_ a non-existent address does seem useful (ie. you whitelist some addresses that you know are going to be staked). Either way, if we want to treat both the same, I think it makes more sense to require that the address exists than to impose some arbitrary limit.

well, i mean in my mind blocking a non-existent address (non-existent on sc, to be clear what i mean. that eth address would exist on eth) would help in a situation where that eth address is blacklisted by a third party (for stolen funds, hack, etc).

The argument for having a limit was to stop an attack vector regarding unlimited state growth. A malicious could register as operator (with no stake) and just keep adding to its exception list, increasing the state size unbounded.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -482,6 +482,10 @@ pub mod pallet {
 		DelegatorBlocked,
 		/// The provided Operator fee is too low.
 		OperatorFeeTooLow,
+		/// Exceptions limit is reached.
+		ExceptionsLimitReached,
```
ramizhasan111, line 486: we can remove this event now.

ramizhasan111 (approved):

PR #6030, opened 2025-07-31
feat: add new vote storage

marcellorigotti:
# Pull Request

Closes: PRO-2400

## Checklist

## Summary

Add new `VoteStorage` to avoid using `SharedDataHash` in case the Value voted on is smaller than an Hash (32 bytes),
this makes less likely to delay consensus cause we don't have the full vote for some of the available hashes and vote extrinsics more readable since validators will always vote with the full vote.

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
@@ -273,7 +273,7 @@ macro_rules! generate_electoral_system_tuple_impls {
                         },)*
                         // If election ID and vote are not belonging to the same electoral system
                         #[allow(unreachable_patterns)]
-                        _ => Err(DispatchError::Other("VoteMisMatch")),
+                        _ => Err(DispatchError::Other("VoteMismatch")),
                     }
                 }
 


marcellorigotti:
Migration is there, what's missing is to make this works with the other migrations regarding the election pallet.
Waiting for others PR to get merged first such that this one can be based on those.

state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -201,7 +201,7 @@ mod benchmarks {
 
 		#[block]
 		{
-			let _ = Pallet::<T, I>::recheck_contributed_to_consensuses(epoch, &validator_id, epoch);
+			let _ = Pallet::<T, I>::recheck_contributed_to_consensuses(epoch, &validator_id, 0);
```
MxmUrw, line 204: What exactly was the problem here? I would think that epoch should be zero by default (at least judging by the looks of `impl_mock_epoch_info`)?

state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -253,43 +253,28 @@ mod benchmarks {
...
-				VoteOf<<T as Config<I>>::ElectoralSystemRunner>,
-			>(&BenchmarkValue::benchmark_value())),
-			Some(BenchmarkValue::benchmark_value())
-		);
+		ElectionConsensusHistoryUpToDate::<T, I>::insert(UniqueMonotonicIdentifier::from(0), 0);
+
+		#[extrinsic_call]
+		provide_shared_data(RawOrigin::Signed(validator_id), Box::new(shared_data_value.clone()));
+
+		assert!(SharedData::<T, I>::get(shared_data_hash).is_some());
```
MxmUrw, line 277: I think we could also check that the shared data we get is the one we provided, as we did previously, no?

state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -253,43 +253,28 @@ mod benchmarks {
...
-			Box::new(BenchmarkValue::benchmark_value()),
+		SharedDataReferenceCount::<T, I>::insert(
+			shared_data_hash,
+			UniqueMonotonicIdentifier::from(0),
+			ReferenceDetails::<BlockNumberFor<T>> {
+				count: 1u32,
+				created: BlockNumberFor::<T>::from(1u32),
+				expires: BlockNumberFor::<T>::from(10u32),
+			},
 		);
```
MxmUrw, line 270: Just for clarification, why did calling the `vote()` function break?

state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -374,7 +359,7 @@ mod benchmarks {
 	fn clear_election_votes() {
 		// Setup a validator set of 150 as in the case of Mainnet.
 		let election_identifier =
-			setup_validators_and_vote::<T, I>(150, BenchmarkValue::benchmark_value());
+			setup_validators_and_vote::<T, I>(3, BenchmarkValue::benchmark_value());
```
MxmUrw, line 362: The comment should also say 3. And I think it would be useful to explain to future readers why using 150 doesn't work.

state-chain/pallets/cf-elections/src/benchmarking.rs
marcellorigotti: I'll also add a comment here to explain

state-chain/pallets/cf-elections/src/benchmarking.rs
marcellorigotti: I'll update the comment. 
Same as the epoch index, I didn't investigate why 150 doesn't work, but I tested it and the authority set size get set to 3 max.

state-chain/pallets/cf-elections/src/electoral_systems/unsafe_median.rs
```diff
@@ -69,8 +69,7 @@ impl<
 	type ElectionIdentifierExtra = ();
 	type ElectionProperties = ();
 	type ElectionState = ();
-	type VoteStorage =
-		vote_storage::individual::Individual<(), vote_storage::individual::shared::Shared<Value>>;
+	type VoteStorage = vote_storage::bitmap_numerical::BitmapNoHash<Value>;
```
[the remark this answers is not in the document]
MxmUrw: > we should then use the `Identity` VoteStorage
Yes, I'm then fine with keeping it currently as you have it, and switch over to identity in my later fee PR.

state-chain/pallets/cf-elections/src/benchmarking.rs
MxmUrw: > I'll also add a comment here to explain

Perfect, a comment like what you wrote above would be already very helpful.

state-chain/pallets/cf-elections/src/benchmarking.rs
marcellorigotti: What happens is that the epoch authority size ends up being 3.
When a validator with authority index >= 3 votes the index is out of bound when accessing the bitmap causing CorruptedStorage, thinking about this the check `ensure_can_vote` in `vote()` should prevent this to happen though! I'll investigate this a bit further.

state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -561,6 +557,66 @@ mod benchmarks {
...
+		let mut electoral_data = Pallet::<T, I>::electoral_data(&caller.clone().into())
+			.unwrap()
+			.current_elections
+			.into_iter();
+
+		let (election_identifier0, ..) = electoral_data.next().unwrap();
+		let (election_identifier1, ..) = electoral_data.next().unwrap();
+
+		#[block]
+		{}
```
MxmUrw, line 573: This looks strange, what is this?

state-chain/pallets/cf-elections/src/benchmarking.rs
```diff
@@ -561,6 +557,66 @@ mod benchmarks {
 		assert_eq!(ElectionConsensusHistoryUpToDate::<T, I>::iter_keys().count(), 0);
 	}
 
+	#[benchmark]
+	fn test_mismatch<T: crate::pallet::Config<I>, I: 'static>() {
```
MxmUrw, line 561: Very nice!

state-chain/pallets/cf-elections/src/electoral_systems/composite.rs
```diff
@@ -262,6 +263,20 @@ macro_rules! generate_electoral_system_tuple_impls {
...
+                    election_identifier: ElectionIdentifierOf<Self>,
+                    partial_vote: &PartialVoteOf<Self>,
+                ) -> Result<(), DispatchError> {
+                    match (*election_identifier.extra(), partial_vote) {
+                        $((CompositeElectionIdentifierExtra::$electoral_system(extra), CompositePartialVote::$electoral_system(_)) => {
+                            Ok(())
+                        },)*
+                        // If election ID and vote are not belonging to the same electoral system
+                        #[allow(unreachable_patterns)]
+                        _ => Err(DispatchError::Other("VoteMisMatch")),
```
MxmUrw, line 276: nit:
```suggestion
                        _ => Err(DispatchError::Other("VoteMismatch")),
```

state-chain/pallets/cf-elections/src/electoral_systems/mock.rs
```diff
@@ -203,4 +203,11 @@ impl ElectoralSystemRunner for MockElectoralSystemRunner {
 	) -> bool {
 		Self::vote_needed()
 	}
+
+	fn check_vote_mismatch(
+		_election_identifier: ElectionIdentifierOf<Self>,
+		_partial_vote: &PartialVoteOf<Self>,
+	) -> Result<(), sp_runtime::DispatchError> {
+		Ok(())
+	}
```
MxmUrw, line 212: Does this need a default implementation?

state-chain/pallets/cf-elections/src/benchmarking.rs
marcellorigotti: we need either a `#[block]` or `#[extrinsic]` in the benchmarks.
Since we don't really care about measuring the weights here but it is more a test we just run an empty block just to make it compile. This could be moved into the tests but then we would have to mock another ES since the MockESRunner is a composite of 1 ES only.

state-chain/pallets/cf-elections/src/electoral_systems/mock.rs
marcellorigotti: 🤔 this is the mock impl

state-chain/pallets/cf-elections/src/benchmarking.rs
MxmUrw: I see, I wasn't aware of that 👍

PR #6032, opened 2025-08-01
feat: configurable poll intervall for unsafe median

MxmUrw:
# Pull Request

Closes: PRO-2443

## Checklist

## Summary

Since 1.10 we use the `UnsafeMedian` electoral system for witnessing bitcoin fees. This creates new elections on every SC block. Unfortunately, the bitcoin rpc call that we use to get fees (`estimate_smart_fee`) only very rarely updates it's value (every 2 btc blocks or so), so it doesn't make sense to query the rpc and vote on every SC block.

This PR adds a new (unsynchronised) setting to the UnsafeMedian ES that lets us specify the number of SC blocks to wait until the next election is created.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8299,6 +8299,7 @@ dependencies = [
  "serde",
  "serde_arrays",
  "sp-core 34.0.0",
+ "sp-runtime 39.0.0",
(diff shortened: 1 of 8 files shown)


marcellorigotti:
I think this PR won't pass the `upgrade_test` unless we run the migration for it.

PR #6039, opened 2025-08-07
feat: oracle swaps

j4m1ef0rd:
# Pull Request

Closes: PRO-2331

## Checklist

## Summary

- Replaced min_price param with a new SwapLimits struct that has the new oracle price field.
- Added logic to the swapping code to check the oracle price and give a price violation if the output of the swap is not enough.
- Moved some of the. oracle price feed api stuff.
- Added support to the btc vault swap to take a u8 for the new oracle price slippage bps.
   - Used u8::MAX as a special value to set None
   - Added a new version of the data struct to allow for backwards compatibility
- Added 2 tests for the new oracle price protection


state-chain/chains/src/cf_parameters.rs
```diff
@@ -59,7 +59,7 @@ pub struct VaultSwapParameters<R> {
 /// Original version of `VaultSwapParameters` that does not include
 /// and refund CCM metadata.
 pub type VaultSwapParametersV0<RefundAddress> =
-	VaultSwapParameters<ChannelRefundParameters<RefundAddress, ()>>;
+	VaultSwapParameters<ChannelRefundParameters<RefundAddress, (), ()>>;
```
syan095, line 62: the () design has caused some tricky problems. I would strongly advise using a field with Option<> instead of this () pattern.

j4m1ef0rd: I tried. Spent a day trying to change it and gave up in the end. Might re-visit it in another PR.

state-chain/chains/src/refund_parameters.rs
```diff
@@ -72,22 +72,26 @@ pub enum AccountOrAddress<AccountId, Address> {
 	PartialOrd,
 	Ord,
 )]
-pub struct ChannelRefundParameters<A, CcmRefundDetails> {
+pub struct ChannelRefundParameters<A, CcmRefundDetails, OracleSlippage> {
```
syan095, line 75: Same point as in the cf_parameter. If possible, avoid this and just use a field. Do we expect different types for the `OracleSlippage`?

state-chain/chains/src/cf_parameters.rs
syan095: yeah better to keep the previous one as is, and just add the new one as Option.
I considered changing this last time in the Solana refund parameter the refactor is too much

j4m1ef0rd: yeah I will give that a try

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd, line 1: Reminder to remove the `CancelSwap` event. Zoheb says they don't need it.

state-chain/chains/src/btc/vault_swap_encoding.rs
```diff
@@ -107,6 +124,43 @@ pub struct BtcCfParameters {
...
+		pub chunk_interval: u16,
+		// --- Boost fields ---
+		pub boost_fee: u8,
+		// --- Broker fields ---
+		// Primary's broker fee:
+		pub broker_fee: u8,
+		pub affiliates: BoundedVec<AffiliateAndFee, ConstU32<MAX_AFFILIATES>>,
+	}
+
+	impl From<BtcCfParametersV0> for BtcCfParameters {
```
albert-llimos, line 147: Nit: In order to keep it consistent with the other chain's `cf_parameters` I'd suggest renaming BtcCfParameters to BtcCfParametersV1.

state-chain/chains/src/cf_parameters.rs
```diff
@@ -59,7 +59,7 @@ pub struct VaultSwapParameters<R> {
 /// Original version of `VaultSwapParameters` that does not include
 /// and refund CCM metadata.
 pub type VaultSwapParametersV0<RefundAddress> =
-	VaultSwapParameters<ChannelRefundParameters<RefundAddress, ()>>;
+	VaultSwapParameters<ChannelRefundParametersV0<RefundAddress>>;
 
 /// New version of `VaultSwapParameters` that includes refund CCM metadata.
```
albert-llimos, line 64:
```suggestion
/// New version of `VaultSwapParameters` includes refund CCM and oracle price slippage.
```

state-chain/chains/src/cf_parameters.rs
```diff
@@ -135,6 +136,49 @@ pub fn build_and_encode_cf_parameters<RefundAddress: Encode>(
 	}
 }
 
+#[cfg(test)]
+pub fn build_and_encode_v0_cf_parameters<RefundAddress: Encode>(
+	refund_parameters: crate::ChannelRefundParametersV0<RefundAddress>, /* crate::ChannelRefundParametersUncheckedV0<RefundAddress>, */
```
albert-llimos, line 141: Remove commented code.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1668,10 +1678,53 @@ pub mod pallet {
...
+							});
+
+						// Use the larger of the two prices
+						let min_price = core::cmp::max(
+							min_price_from_oracle.unwrap_or_default(),
+							params.price_limits.min_price,
+						);
+
+						let min_output =
+							output_amount_ceil(swap.swap.input_amount.into(), min_price)
```
albert-llimos, line 1723: Any particular reason why we use `output_amount_ceil` instead of `output_amount_floor`? I don't think it matters here but normally the rounding tends to be done in "favor" of the protocol per say, just to cover any edge cases.
Since it's not AMM logic that's almost certainly not relevant though, so it's just curiosity.

bouncer/shared/sol_vault_swap.ts
```diff
@@ -118,6 +119,7 @@ export async function executeSolVaultSwap(
       gas_budget: fillOrKillParams?.refundCcmMetadata.gasBudget,
       ccm_additional_data: fillOrKillParams?.refundCcmMetadata.ccmAdditionalData,
     },
+    max_oracle_price_slippage: undefined, // TODO: update FillOrKillParamsX128 with max_oracle_price_slippage
```
albert-llimos, line 122: Right now the bouncer is using `"@chainflip/cli": "1.9.0-assethub.0-ccm-refunds"`, which is a little patch I made so `FillOrKillParamsX128` contains the `refund_ccm_metadata`. Then the web team will add that to the 1.11 SDK.
We can add `max_oracle_price_slippage` in that same branch, as it's a little patch anyway, so we can use this in the bouncer. @j4m1ef0rd lmk if you'd want me to help out with that.
Then I think we should have a oracle swap FoK test in `tests/fill_or_kill.ts` the same way that we have it for regular and CCM refunds.

j4m1ef0rd: I made the changes to your branch but I dont have permissions to push it. So I will this bouncer stuff in another PR so we can merge this now.

state-chain/pallets/cf-swapping/src/lib.rs
msgmaxim: No particular reason. Like you said, this shouldn't matter, but I'm not against changing this to `output_amount_floor`.

msgmaxim (approved):

PR #6052, opened 2025-08-19
fix: use fewer proptest iterations to decrease CI times.

MxmUrw:
# Pull Request

## Checklist

## Summary

We had long runs of the coverage-check in CI (which of course runs all tests to compute the coverage), due to a high number of iterations set for the proptests. This PR lowers the number of proptest iterations. When changing existing proptest-tested statemachine code, **the iteration numbers should be temporarily increased**, to ensure full coverage.


albert-llimos (approved):
13 min instead of one hour is a lot more dev-friendly :raised_hands:

PR #6056, opened 2025-08-20
feat: rpc for building ethereum tx for sc calls

ramizhasan111:
# Pull Request

Closes: PRO-2439

## Checklist

## Summary

### DIFF
--- a/state-chain/runtime/src/lib.rs
+++ b/state-chain/runtime/src/lib.rs
@@ -2859,7 +2859,7 @@ impl_runtime_apis! {
 			use chainflip::ethereum_sc_calls::DelegationApi;
 			let caller_id = EthereumAccount(caller).into_account_id();
 			let required_deposit = match call {
-				EthereumSCApi::Delegation(DelegationApi::Delegate { operator, increase: DelegationAmount::Some(ref increase) }) => {
+				EthereumSCApi::Delegation { call: DelegationApi::Delegate { increase: DelegationAmount::Some(ref increase), .. } } => {
 					pallet_cf_validator::MaxDelegationBid::<Runtime>::get(&caller_id).unwrap_or_default()
 						.saturating_add(*increase)
 						.saturating_sub(pallet_cf_flip::Pallet::<Runtime>::balance(&caller_id))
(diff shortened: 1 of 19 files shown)


dandanlen:
I added @acdibble to the reviewers in case he has relevant feedback.

state-chain/runtime/src/lib.rs
```diff
@@ -2825,6 +2832,37 @@ impl_runtime_apis! {
...
+				} else {
+					SCCall::new(call.encode()).abi_encoded_payload()
+				},
+				value: U256::zero(),
+				to: Environment::eth_sc_utils_address(),
+				source_token_address: Some(
+					Environment::supported_eth_assets(cf_primitives::chains::assets::eth::Asset::Flip)
+						.ok_or(DispatchErrorWithMessage::from(
+							"flip token address not found on the state chain: {e}",
+						))?,
```
dandanlen, line 2862: @albert-llimos do we need the source_token_address even if there is no FLIP deposit?

dandanlen:
I rebased and updated the api based on recent discussions.

It now uses delegate(increase) and undelegate(decrease). 

I also noticed that we need the caller address to be able to check the balance and return the correct deposit amount. 

The new API checks the current balance and max bid and only returns a payload with a deposit if the current balance isn't enough to cover the increase. (This support the use case @acdibble mentioned above where you just want to delegate your earnings). 

> i'd still prefer to have a simpler RPC that returns the SCALE encoded calldata that the state chain will parse. from a security point of view, i don't like the idea of requesting a blob of call data and a contract address in the browser for a user to execute

re. this - we can add this as an option (or return both) I guess but the idea was to use the same return format as for vault swaps. What is the added value this would give over simply encoding the data yourself using chain metadata?

dandanlen:
(This might not compile yet, I made the change pretty quickly)

state-chain/runtime/src/lib.rs
dandanlen: Ok i'll remove it then .

dandanlen (approved):

PR #6060, opened 2025-08-22
test: improve bouncer speed

albert-llimos:
# Pull Request

## Checklist

## Summary

I noticed that for some reason the `SwapLessThanED` is blocking the other tests from starting when running the bouncer. You can see it in any of the runs from the timestamps. e.g.
https://github.com/chainflip-io/chainflip-backend/actions/runs/17128788352/job/48587651646#step:16:74

This simple change reduces bouncer runs by 2-3 min.


dandanlen:
I'm not sure why this would block the other tests... but I had a quick look and there seem to be quite a few loops that repeated query the chain for various things like price - maybe adding some `sleep` commands inside the loops would prevent them from running too hot, and would improve things.

PR #6061, opened 2025-08-22
feat: Core Lending

msgmaxim:
# Pull Request

Closes: PRO-2324, PRO-2478, PRO-2479, PRO-2329, PRO-2472, PRO-2473, PRO-2474, PRO-2477, PRO-2476, PRO-2475, PRO-2503, PRO-2468, PRO-2471, PRO-2339, PRO-2492, PRO-2491, PRO-2481, PRO-2500, PRO-2469, PRO-2467, PRO-2490, PRO-2470, PRO-2548, PRO-2540

## Summary

Supported functionality:
- Lenders can add/remove to/from lending pools
- Borrowers can add/remove funds to be used as collateral
- Borrowers can borrow and repay loans
- Interest is dynamic and based on utilisation (for now it just grows linearly)
- Periodically collecting interest, checking each loan's CR triggering
  collateral topup or account's liquidation
- Liquidation swaps are DCA swaps which can be aborted at any point, notably
  if CR raises above a certain threshold
- Both hard and soft liquidations are supported (use different oracle swap parameters),
  and it is possible to transition between them

Left for future PRs:
- voluntary liquidation
- extra tests for some edge cases
- integration tests (exist in a separate branch but needs updating)
- bouncer tests
- benchmarking (exists in a separate branch but needs updating)


state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,978 @@
...
+	/// Computes account's total collateral value in USD, including what's in liquidation swaps.
+	pub fn total_collateral_usd_value(&self) -> Result<AssetAmount, Error<T>> {
+		match &self.liquidation_status {
+			LiquidationStatus::NoLiquidation => self
+				.collateral
+				.iter()
+				.map(|(asset, amount)| usd_value_of::<T>(*asset, *amount).ok())
+				.try_fold(0u128, |acc, x| acc.checked_add(x?))
+				.ok_or(Error::<T>::OraclePriceUnavailable),
+			LiquidationStatus::Liquidating { liquidation_swaps, .. } => {
```
j4m1ef0rd, line 55: as discussed, I think we need to add the collateral balance here when liquidating.

msgmaxim: updated

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1197 @@
...
+			let from_asset = collateral_asset;
+			let to_asset = loan_asset;
+
+			let max_slippage = if is_hard {
+				HARD_LIQUIDATION_MAX_ORACLE_SLIPPAGE
+			} else {
+				SOFT_LIQUIDATION_MAX_ORACLE_SLIPPAGE
+			};
+
+			let swap_request_id = initiate_swap::<T>(
```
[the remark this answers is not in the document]
msgmaxim: Yeah, as discused we should just test that swapping works, and I believe we have even seen such swaps on mainnet (likely performed by mistake). Disallowing loans means more work here and in the UI, and it is easier to just make sure it works.

>  maybe a bouncer test? 

Why would it need to be a bouncer test? A simple unit test in the swapping pallet should be sufficient.

state-chain/pallets/cf-lending-pools/src/general_lending_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+pub struct LendingPool<T: Config> {
+	// Total amount owed to active lenders (includes what's currently in loans)
+	pub total_amount: AssetAmount,
+	// Amount available to be borrowed
+	pub available_amount: AssetAmount,
+	// Maps lenders to their shares in the pool; each lender is effectively owed their `share` *
+	// `total_amount` of the pool's asset.
+	pub lender_shares: BTreeMap<T::AccountId, Perquintill>,
+	// Fees collected by the pool but not yet swapped into the pool's asset
+	pub collected_fees: BTreeMap<Asset, AssetAmount>,
```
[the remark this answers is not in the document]
msgmaxim: The fees are collected (updating this item) in `Pallet::accrue_fees`. The field is here because each pool collects fees seperately from other pools.

state-chain/pallets/cf-lending-pools/src/general_lending_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+
+	pub fn get_utilisation(&self) -> Permill {
+		let in_use = self.total_amount.saturating_sub(self.available_amount);
+
+		// Note: `from_rational` does not panic on invalid inputs and instead returns 100%.
+		Permill::from_rational(in_use, self.total_amount)
+	}
+
+	/// Either make a repayment of loan principal or pay fees into the pool.
+	pub fn accept_payment(&mut self, amount: AssetAmount, is_principal: bool) {
```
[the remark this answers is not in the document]
dandanlen: I suppose what i'm saying is that we could be clearer about the responsibilities of the `LendingPool`. 

It doesn't seem to concern itself with fees at all. It accepts deposits from accounts, deposits for *all* accounts (fees) and handles borrow  requests and repayments. And ensures everything adds up. But there is no notion of what fees are owed or should be paid, what the current interest rate is, etc.

I suppose that seeing the mention of fees here led me to believe that fee were somehow managed here, or that this struct somehow give guarantees about fees, which isn't the case. They are entirely external to this. 

Which is why I also suggested that `collected_fees` maybe doesn't belong here. This item does't interact in any way with the other items. Even the access patterns are different. It seems they were just added for convenience to avoid declaring another storage item.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1571 @@
...
+			)?;
+
+			let primary_collateral_asset = account.primary_collateral_asset;
+
+			ensure!(
+				extra_collateral.contains_key(&primary_collateral_asset),
+				Error::<T>::InvalidLoanParameters
+			);
+
+			account.primary_collateral_asset = primary_collateral_asset;
```
dandanlen, line 841:
```
let primary_collateral_asset = account.primary_collateral_asset;
//..
account.primary_collateral_asset = primary_collateral_asset;
```

Seems redundant. Also looks like it overrides the user's provided `primary_collateral_asset`. Is this intentional?

state-chain/pallets/cf-lending-pools/src/general_lending_pool.rs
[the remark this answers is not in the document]
msgmaxim: The fees are payed to the pool though, so it doesn't seem entirely inappropriate to store them there as well (and like you mentioned, it is more convenient).

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+					T::Balance::credit_account(
+						&borrower_id,
+						liquidation_swap.to_asset,
+						remaining_amount,
+					);
+
+					// If this swap is the last liquidation swap for the loan, we should
+					// "settle" it (even if it hasn't been repaid in full):
+					if is_last_liquidation_swap {
+						loan_account.settle_loan(loan_id, true /* via liquidation */);
```
dandanlen, line 1167: Do we not want to indicate / does it not matter if the loan was not fully repaid?

Also, I geuss we're relying on the fact that we always try to liquidate the full loan, ie we don't trigger partial liquidations?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+	pub fn total_collateral_usd_value(&self) -> Result<AssetAmount, Error<T>> {
+		self.get_total_collateral()
+			.iter()
+			.map(|(asset, amount)| usd_value_of::<T>(*asset, *amount).ok())
+			.try_fold(0u128, |acc, x| acc.checked_add(x?))
+			.ok_or(Error::<T>::OraclePriceUnavailable)
+	}
+
+	pub fn check_ltv(&mut self, borrower_id: &T::AccountId) {
+		let config = LendingConfig::<T>::get();
```
dandanlen, line 118: There are a few places where we call a function that needs the config (or, as in this case, a subset of the config) and just reads it from storage. 

Passing the config as an argument would make it easier to test (you can't forget to set the config in storage) and would avoid unnecessary duplicate storage reads. For example `new_loan` reads the config, and later calls `charge_origination_fee` which also read the config.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+
+	/// Computes account's LTV (Loan-to-Value) ratio. Takes liquidations into account.
+	/// Returns error if oracle prices are not available, or if collateral is zero.
+	pub fn derive_ltv(&self) -> Result<FixedU64, Error<T>> {
+		let collateral = self.total_collateral_usd_value()?;
+		let principal = self.total_owed_usd_value()?;
+
+		if collateral == 0 {
+			return Err(Error::<T>::InsufficientCollateral);
+		}
```
dandanlen, line 271: Can this ever happen? (would `log_or_panic` be more appropriate?)

And if it can happen, is an error the right return value? For example I can imagine we would want to top-up if the account has zero capital, but we would in fact return early (in `process_auto_topup`).

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+
+		let (provided_amount_after_fees, liquidation_fee) = if should_charge_liquidation_fee {
+			let liquidation_fee =
+				config.get_config_for_asset(loan.asset).liquidation_fee * provided_amount;
+			let after_fees = provided_amount.saturating_sub(liquidation_fee);
+
+			// NOTE: this may not be equal to the provided fee if the provided amount isn't
+			// sufficient
+			let actual_fee = provided_amount.saturating_sub(after_fees);
```
dandanlen, line 578: I don't understand this comment - since config.liquidation_fee is a Permill value, the calculated absolute liquidation fee is always <= provided_amount, so `actual_fee` is always the same as `liquidation_fee` here, or not?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+		SwapRequestType::Regular {
+			output_action: SwapOutputAction::CreditLendingPool { swap_type },
+		},
+		Default::default(), // broker fees
+		Some(PriceLimitsAndExpiry {
+			expiry_behaviour: ExpiryBehaviour::NoExpiry,
+			min_price: Default::default(),
+			max_oracle_price_slippage: Some(max_oracle_price_slippage),
+		}),
+		None, // no dca parameters
```
dandanlen, line 723: If we never use DCA, does this mean liquidations are always all-or-nothing? I would have expected we might want to use DCA for liquidations but not for fee swaps, for example.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+
+			ensure!(
+				account.derive_ltv()? <= config.ltv_thresholds.target,
+				Error::<T>::InsufficientCollateral
+			);
+
+			GeneralLendingPools::<T>::try_mutate(asset, |pool| {
+				let pool = pool.as_mut().ok_or(Error::<T>::PoolDoesNotExist)?;
+
+				pool.provide_funds_for_loan(amount_to_borrow)?;
```
dandanlen, line 869: From an API / correctness point of view, it might be nice to explicitly link loan creation with the pool to satisfy ourselves that taking a loan requires interaction with a lending pool. (ie. a method on GeneralLoan that takes a pool as argument). 

And if we structure it as eg. `GeneralLoan::new(..).expand(pool, loan_amount)` then we can share much more code between `new_loan` and `expand_loan`.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+					Error::<T>::LoanCreationDisabled
+				);
+
+				loan.owed_principal.saturating_accrue(extra_amount_to_borrow);
+			}
+
+			loan_account
+				.try_adding_collateral_from_free_balance(&borrower_id, &extra_collateral)?;
+
+			if loan_account.derive_ltv()? > config.ltv_thresholds.target {
```
dandanlen, line 934: In new_loan we check the LTV *before* taking the loan, here we take it *after* increasing the owed principal. Is this inentional? What am I missing?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: > Passing the config as an argument would make it easier to test (you can't forget to set the config in storage) 

Don't see this making a difference for testing tbh. All test just rely on the default values for the config without having to set them manually.

> would avoid unnecessary duplicate storage reads. 

I'm relying on caching behaviour here. We've made similar argument about safe mode in the past.

> For example new_loan reads the config, and later calls charge_origination_fee which also read the config.

Unless there is a noticable performace penalty (which I think there isn't?), I don't see a problem with this. On the other hand I think there is a slight benefit to readability as we don't need to explicitly pass config into almost every single method.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1635 @@
...
+			} else {
+				log_or_panic!("CHP Pool must exist for asset {}", self.asset);
+			}
+		});
+
+		self.owed_principal.saturating_reduce(amount);
+	}
+
+	/// Pays fee to the pool in pool's asset. Does NOT reduce the owed principal.
+	fn pay_fee(&mut self, amount: AssetAmount) {
```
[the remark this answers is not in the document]
msgmaxim: Yes, looks like `fees_paid` was being updated outside, so I now moved that line inside `pay_fee` instead.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Can't remember what this was about... removed `actual_fee` and the comment.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: This shouldn't be possible, but it is not locally obvious. 

> And if it can happen, is an error the right return value? 

I considered returning `FixedU64::max_value()`, but I think it is probably worse... If this returns error, then the topup won't work, but I think it is OK (if we got to a point where the user has loans but no collateral, something else must have gotten wrong, and we should investigate before taking more fund from the user). On the other hand, if the account has no loans and no collateral (but we failed to remove the account for some reason), it seems nicer to show LTV as `None` (what we would currently return in the RPC) instead of `max_value()`. 🤷 

I will add `log_or_panic`, but will leave the Error in for now.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Right, I think I was delaying enabling DCA until we decide what the parameters should be... I guess we can define a fixed chunk size controlled by governance? For simplicity could be the same value for each asset.

state-chain/pallets/cf-lending-pools/src/general_lending_pool.rs
```diff
@@ -0,0 +1,276 @@
+use super::*;
+
+/// This keeps track of each lender's share in the pool, the total amount of funds
+/// owed to lenders (and how much of it is available to be borrowed), and the collected
+/// fees in various assets that are yet to be swapped into the pool's asset.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Encode, Decode, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct LendingPool<T: Config> {
```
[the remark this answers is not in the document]
msgmaxim: Ok, I added a new error type `LendingPoolError`, and now LendingPool returns that instead of `Error<T>`, so I could change the generic parameter to be simply `AccountId`.

state-chain/pallets/cf-lending-pools/src/lib.rs
```diff
@@ -37,38 +47,78 @@ use cf_primitives::{
...
+			add_collateral_enabled: BTreeSet::new(),
+			remove_collateral_enabled: BTreeSet::new(),
+		}
+	}
+
+	fn code_green() -> Self {
+		Self {
+			add_boost_funds_enabled: true,
+			stop_boosting_enabled: true,
+			borrowing_enabled: BTreeSet::from_iter(Asset::all()),
```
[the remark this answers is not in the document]
msgmaxim: Created a ticked for this https://linear.app/chainflip/issue/PRO-2540/provide-a-convenient-way-to-enable-safe-mode-in-lending-for-all-assets.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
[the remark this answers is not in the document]
msgmaxim: Update: `pay_fee` was only used once (and incorrectly...) so I inlined the function and reworked `accrue_fees` to be able to take fees in the pools "native" asset too.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: > Do we not want to indicate / does it not matter if the loan was not fully repaid?

We already emit all repayments in `LoanRepaid` so consumers of events should be able to work out the total amount repaid and whether or not it is less than what was borrowed. I suppose we could explicitly emit the shortfall here. @GabrielBuragev would that be helpful?

> Also, I geuss we're relying on the fact that we always try to liquidate the full loan, ie we don't trigger partial liquidations?

Not sure what you mean here. We do support partial liquidations (ignoring the fact that DCA parameters were missing). `process_loan_swap_outcome` is only called on execution of liquidation swaps, however if liquidation swaps are aborted (see `abort_liquidation_swaps`) we won't necessarily be calling `settle_loan`.

state-chain/pallets/cf-lending-pools/src/general_lending/general_lending_tests.rs
```diff
@@ -0,0 +1,2212 @@
...
+			);
+
+			assert_eq!(
+				PendingPoolFees::<Test>::get(LOAN_ASSET),
+				BTreeMap::from([(COLLATERAL_ASSET, take_network_fee(origination_fee).1)])
+			);
+
+			assert_eq!(
+				LoanAccounts::<Test>::get(BORROWER),
+				Some(LoanAccount {
```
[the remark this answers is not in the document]
msgmaxim: That's fair. Though I'm now thinking instead of returning this to the account's free balance we should add it to the user's collateral instead? This will likely work better for users who don't use CF for anything other than lending/borrowing (e.g. a user coming via an intergration).

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -704,6 +694,36 @@ fn initiate_swap<T: Config>(
 	swap_type: LendingSwapType<T::AccountId>,
 	max_oracle_price_slippage: BasisPoints,
 ) -> SwapRequestId {
+	let dca_params = match swap_type {
+		LendingSwapType::Liquidation { .. } => {
```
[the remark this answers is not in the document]
msgmaxim: This particular code *is* tested because the mock only implements the `init_swap_request` part. Also the same general comment about additionall testing coming in a later PR applies here.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
dandanlen: `expand_loan_inner` is a little awkward in that expanding a loan requires removing it from the map so that it can be inserted again... it seems like it could take just an id, and then expand or create as necessary.

In fact, what's the reason for differentiating between expand and create? Aren't they essentially the same operations? The only difference seems to be that expand can fail if you accidentally provide the wrong loan id...

msgmaxim: I tried moving loan creation into`expand_loan` and ended up with a similar borrow checked problem: you end up having to hold a mut reference (in case the loan already existed), while also modifying the loan_account. There are ways to address this of course (e.g. inlining `try_adding_collateral_from_free_balance`, turning it into a free standing function, or reading the loan multiple times), but they are all felt like compromises, and not an improvement over what I had initially, so I think in the end I prefer to leave the code as is.

state-chain/pallets/cf-lending-pools/src/core_lending_pool.rs
```diff
@@ -106,6 +104,14 @@ mod scaled_amount {
 				.saturated_into()
 		}
 
+		pub fn take_whole_amount(&mut self) -> AssetAmount {
```
j4m1ef0rd, line 107: would be good to have a description on this. I misunderstood it at first, because `whole != all` in this case.

msgmaxim: update: I ended up renaming the method to `take_non_fractional_part` to make it more clear hopefully (also added a comment).

state-chain/traits/src/lending.rs
```diff
@@ -42,3 +48,59 @@ pub trait BoostApi {
...
+
+impl core::ops::Add<u64> for LoanId {
+	type Output = Self;
+
+	fn add(self, rhs: u64) -> Self::Output {
+		LoanId(self.0 + rhs)
+	}
+}
+
+pub trait LendingApi {
```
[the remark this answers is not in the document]
msgmaxim: It is left over from when were prioritising liquidity lending (which will likely need some trait like that, though maybe not exactly this), but I decided to keep it for now since it helped to organise code somewhat better IMO.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -0,0 +1,1846 @@
...
+	let mut i1 = i1.deconstruct() as u64;
+	let u0 = u0.deconstruct() as u64;
+	let u1 = u1.deconstruct() as u64;
+	let u = u.deconstruct() as u64;
+
+	let negative_slope = if i1 >= i0 {
+		false
+	} else {
+		// Ensure that we can subtract without underflow
+		core::mem::swap(&mut i0, &mut i1);
```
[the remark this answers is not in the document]
msgmaxim: > Also, not sure why we multiply/divide by ACCURACY... this is just so that we round to nearest?

`ACCURACY` is used... for better accuracy 😉. Because we are dealing with integers we can't just divide one by another and not expect huge errors. It is essentially doing what Permill does internally for multiplication and division.

Having said that, I tried using `FixedI64` (`Permill` itself is no good for this) instead and it works just as well, so I will keep the `FixedI64` version.

PR #6063, opened 2025-08-25
fix: allow multiple UTXOs spending to the same deposit channel

MxmUrw:
# Pull Request

Closes: PRO-2136

## Checklist

## Summary

Previously we always witnessed only a single UTXO, in case of multiple UTXOs spending to the same deposit channel. This PR takes instead the highest 10. This also adds a test that ensures that we properly witness all of the UTXOs.


dandanlen (approved):
👌 Thanks

PR #6065, opened 2025-08-26
feat: oracle delta

j4m1ef0rd:
# Pull Request

Closes: PRO-2454

## Checklist

## Summary

- Calculating the new `oracle_delta` when we check for price violations (regardless of refund params)
- Added the new `oracle_delta` to the `SwapExecuted` Event
- Using a `Option<i16>` for the `oracle_delta`. Will be `None` if the oracle price is not supported for one of the assets.
- The delta is in basis points but most likely will be a negative number.
- Added some tests to cover the new calculations.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -80,6 +80,8 @@ pub(crate) const DEFAULT_SWAP_RETRY_DELAY_BLOCKS: u32 = 5;
 const DEFAULT_MAX_SWAP_RETRY_DURATION_BLOCKS: u32 = 3600 / SECONDS_PER_BLOCK as u32; // 1 hour
 const DEFAULT_MAX_SWAP_REQUEST_DURATION_BLOCKS: u32 = 86_400 / SECONDS_PER_BLOCK as u32; // 24 hours
 
+type SignedBasisPoints = i16;
+
 pub struct DefaultSwapRetryDelay<T> {
 	_phantom: PhantomData<T>,
 }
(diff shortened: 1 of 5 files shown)


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1682,14 +1687,48 @@ pub mod pallet {
...
 
-		fn check_swap_price_violation(swap: &SwapState<T>) -> Result<(), SwapFailureReason> {
+		fn check_swap_price_violation(
+			swap: &SwapState<T>,
+		) -> Result<Option<SignedBasisPoints>, SwapFailureReason> {
+			let input_oracle_price = T::PriceFeedApi::get_price(swap.input_asset());
+			let output_oracle_price = T::PriceFeedApi::get_price(swap.output_asset());
+
+			// Calculate executed price delta from the oracle price.
+			let oracle_delta = if let (Some(input_oracle), Some(output_oracle)) =
```
msgmaxim, line 1697: As discussed, we could put this in a separate function considering that this delta is not used in the actual checks below (but it almost looks like it might).

j4m1ef0rd:
@MxmUrw im getting `Failed to get election state for the ChainlinkOraclePrice ES due to corrupted storage` error in the integration tests now that i'm trying to get the oracle price during a swap regardless of the refund params. 
Can you help please?

MxmUrw:
Found it. The reason was that the new elections pallet wasn't initialized in the mock runtime. Also pushed a fix.

msgmaxim (approved):

PR #6070, opened 2025-08-27
fix: use LP withdrawal to test ETH broadcasts during rotation

dandanlen:
Previously we used an Arbitrum vault swap which introduced further delays and therefore made it more
susceptible to timing issues.

### DIFF
--- a/.github/workflows/_40_post_check.yml
+++ b/.github/workflows/_40_post_check.yml
@@ -159,6 +159,12 @@ jobs:
           ./commands/explorer.ts 0 $BLOCK_HEIGHT > /tmp/chainflip/statechain_explorer.txt
           BLOCK_HEIGHT=$(curl -s -H "Content-Type: application/json" \
           -d '{"jsonrpc":"2.0","method":"chain_getHeader","params":[],"id":1}' \
+          http://localhost:9947 \
+          | jq -r '.result.number' \
+          | xargs printf "%d\n")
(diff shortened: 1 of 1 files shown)


bouncer/tests/rotation_barrier.ts
```diff
@@ -1,24 +1,48 @@
...
+  const broadcastAborted = observeBadEvent(logger, ':BroadcastAborted', {});
+
+  const withdrawalAddress = await newAssetAddress(InternalAssets.Eth);
+
+  const api = await getChainflipApi();
+  const { promise, waiter } = waitForExt(api, logger, 'InBlock', await lpMutex.acquire(lpUri));
+  const lp = createStateChainKeypair(lpUri);
+  const nonce = await api.rpc.system.accountNextIndex(lp.address);
+  const unsub = await api.tx.liquidityProvider
+    .withdrawAsset(1, InternalAssets.Eth, withdrawalAddress)
```
albert-llimos, line 39: There seems to be something flaky about this step, there  is some error popping up that I haven't seen before.
See:
https://github.com/chainflip-io/chainflip-backend/actions/runs/17288081219/job/49069502377#step:16:1733
It has happened multiple times when trying to merge. I think we should address it before merging.

dandanlen: I'm fixing - it's an polkaJs encoding issue.

PR #6077, opened 2025-08-28
fix: API changes (increase/decrease; remove set_max_bid) & delegator account cleanup

dandanlen:
Fixes: PRO-2461

- `set_max_bid` didn't check that the account was delegating.
- added a test
- added `OnKilledAccount` implementation to ensure cleanup

ALSO: 

Changed the API: delegate and undelegate now take increase/decrease to update the max bid. 
Setting the max bid directly is no longer supported.

### DIFF
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -1,10 +1,6 @@
 # This CI runs a lot of the jobs in parallel to speed up development time. We also run a simpler suite of bouncer tests.
 name: Release Chainflip Development
-on:
-  pull_request:
-    branches:
-      - main
-      - "release/*"
+on: pull_request
 concurrency:
   group: ${{ github.ref }}-release-development
   cancel-in-progress: true
--- a/bouncer/tests/delegate_flip.ts
+++ b/bouncer/tests/delegate_flip.ts
@@ -177,7 +177,7 @@ async function testDelegate(parentLogger: Logger) {
   scCallExecutedEvent = observeEvent(logger, 'funding:SCCallExecuted', {
     test: (event) => event.data.ethTxHash === receipt.transactionHash,
   }).event;
-  const undelegatedEvent = observeEvent(logger, 'validator:UnDelegated', {
+  const undelegatedEvent = observeEvent(logger, 'validator:Undelegated', {
     test: (event) => {
       const delegatorMatch = event.data.delegator === evmToScAddress(whalePubkey);
       const operatorMatch = event.data.operator === operator.address;
--- a/bouncer/tests/delegate_flip.ts
+++ b/bouncer/tests/delegate_flip.ts
@@ -38,7 +38,7 @@ async function encodeAndSendDelegationApiCall(
 
   logger.info(`Requesting EVM encoding for ${caller} ${JSON.stringify(call)}`);
 
-  const payload = await chainflip.raw_rpc('cf_evm_calldata', caller, {
+  const payload = await chainflip.rpc('cf_evm_calldata', caller, {
     API: 'Delegation',
     call,
   });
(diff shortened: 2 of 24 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1179,41 +1191,81 @@ pub mod pallet {
...
+				DelegationChoice::<T>::contains_key(&delegator),
+				Error::<T>::AccountIsNotDelegating
+			);
 
-			Ok(())
-		}
+			let unlink = MaxDelegationBid::<T>::mutate_exists(&delegator, |max_bid| {
+				use frame_support::sp_runtime::traits::Zero;
+				match (max_bid.clone(), decrease) {
+					(current, DelegationAmount::Some(decr)) => {
```
ramizhasan111, line 1234: we can match over just the delegation amount. We dont need to include current in the match.

state-chain/cf-integration-tests/src/delegation.rs
```diff
@@ -131,66 +111,76 @@ fn block_author_rewards_are_distributed_among_delegators() {
...
-					) <= 10u16
-				);
-			}
-			assert_eq!(
-				PerU16::from_rational(op_cut, total_auth_reward),
-				PerU16::from_rational(
-					delegators.iter().map(|d| d.pre_balance).sum::<u128>(),
-					total_pre_balance
-				) * PerU16::from_percent(5)
-			);
```
ramizhasan111, line 193: why remove these checks?

dandanlen:
> @ramizhasan111 fyi. i'm fixing the bouncer tests

Fixed in #6056 ([63f59c4](https://github.com/chainflip-io/chainflip-backend/pull/6056/commits/63f59c43f1f66b7fd9a7637cfdeb49339f319ed9))

.github/workflows/ci-development.yml
```diff
@@ -1,10 +1,6 @@
 # This CI runs a lot of the jobs in parallel to speed up development time. We also run a simpler suite of bouncer tests.
 name: Release Chainflip Development
-on:
-  pull_request:
-    branches:
-      - main
-      - "release/*"
+on: pull_request
```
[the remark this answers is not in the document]
dandanlen: I want to keep it :)

bouncer/tests/delegate_flip.ts
```diff
@@ -12,110 +12,66 @@ import {
...
-            },
-          },
-        }),
-      );
-    default:
-      throw new Error('Invalid payload type');
-  }
+  logger.info(`Requesting EVM encoding for ${caller} ${JSON.stringify(call)}`);
+
+  const payload = await chainflip.raw_rpc('cf_evm_calldata', caller, {
```
albert-llimos, line 41: Why don't we use `await chainflip.rpc(' cf_evm_calldata' ...` as we do for all other rpc calls? Then we don't need to add this `raw_rpc` nor the evmCallDetails's zod.
e.g. https://github.com/chainflip-io/chainflip-backend/blob/3d7568ab90a4a9744ddc70b31644ca274b842669/bouncer/shared/sol_vault_swap.ts#L135

dandanlen: Ah! That's how it works 😅

albert-llimos: Perfect, I see you changed it.
You can also use i.e. `as unknown as SolVaultSwapDetails;` and then define it as an interface instead of zod, which is a bit cleaner:
```
interface SolVaultSwapDetails {
  chain: string;
  program_id: string;
  accounts: RpcAccountMeta[];
  data: string;
}
```
But otherwise it's fine as is :+1:

albert-llimos: Pretty sure the interface approach does that, you're right. Then this seems better :+1:

PR #6086, opened 2025-09-02
refactor: concise fee estimation trait

MxmUrw:
# Pull Request

Part of: PRO-2441

## Checklist

## Summary

This PR was part of my work to get better bitcoin fees. Since these are a few things that are standalone refactors, and the direction of the bitcoin fee estimation problem changed a bit, it seems useful to get these merged independently.

In particular:
 - This PR changes the `FeeEstimationApi` and `AdjustedFeeEstimationApi` such as to be easier extendable in the future. 
Instead of 4 separate functions, the api now has a single `estimate_fee()` which takes an additional parameter of type `IngressOrEgress`. Its 4 cases represent exactly the same information that the 4 functions' input types carried previously.
 - By doing this, it also fixes an inconsistency, where one of the 4 functions previously didn't apply the fee multiplier (`estimate_ingress_fee_vault_swap()`).
 - Accordingly, the definition of `IngressOrEgress` is moved to the `cf_primitives` crate.
 - The macro `def_derive!{}` that was previously only used in the `cf_elections` pallet is renamed to `derive_common_traits!{}` and moved to `cf_utilities::macros`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6220,7 +6220,7 @@ dependencies = [
  "lalrpop-util",
  "petgraph",
  "regex",
- "regex-syntax 0.8.5",
+ "regex-syntax",
  "string_cache",
  "term",
  "tiny-keccak",
@@ -9511,7 +9511,7 @@ dependencies = [
  "rand",
  "rand_chacha",
  "rand_xorshift",
- "regex-syntax 0.8.5",
+ "regex-syntax",
  "rusty-fork",
  "tempfile",
  "unarray",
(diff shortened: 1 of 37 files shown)


marcellorigotti (approved):

MxmUrw:
> I now the distinction between the primitives and traits crates is a bit fuzzy, but...
> 
> If we only ever use the `IngressOrEgress` type in the context of the trait (either implementing it or calling into it), then there's no real need to put the type in cf_primitives, it can next to the trait where it's used.
> 
> Not a strong opinion though. Feel free to leave it (especially if it's used elsewhere outside of the trait context).

The `IngressOrEgress` type was previously defined in the ingress-egress pallet and used there, as well as in `runtime/lib`. This PR adds additional use-sites by also using it for the trait (&implementations).

PR #6099, opened 2025-09-05
feat: update account_info rpc with delegation and redemption estimate

dandanlen:
# Pull Request

Closes: PRO-2506 PRO-2375

one breaking change (at least):

For LPs, `balances` is now `asset_balances`.

I tried to keep everything else non-breaking and backwards compatible.


state-chain/runtime/src/lib.rs
```diff
@@ -1708,6 +1709,23 @@ impl_runtime_apis! {
...
+			RpcAccountInfoCommonItems {
+				flip_balance: flip_account.total(),
+				asset_balances: AssetBalances::free_balances(account_id),
+				bond: flip_account.bond(),
+				estimated_redeemable_balance: pallet_cf_funding::Redemption::<Runtime>::for_rpc(
+					account_id,
+				).map(|redemption| redemption.redeem_amount).unwrap_or_default(),
+				bound_redeem_address: pallet_cf_funding::BoundRedeemAddress::<Runtime>::get(account_id),
+				restricted_balances: pallet_cf_funding::RestrictedBalances::<Runtime>::get(account_id),
+				delegating_to: pallet_cf_validator::DelegationChoice::<Runtime>::get(account_id),
```
acdibble, line 1726: this is just their current choice for operator right? if they are switching in the middle of an epoch, this will be the new operator they've chosen and not the one they are "bonded" with

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -447,6 +508,41 @@ pub struct NetworkFees {
...
+	pub flip_balance: Balance,
+	pub asset_balances: cf_chains::assets::any::AssetMap<Balance>,
+	pub bond: Balance,
+	pub estimated_redeemable_balance: Balance,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub bound_redeem_address: Option<EthereumAddress>,
+	#[serde(skip_serializing_if = "BTreeMap::is_empty")]
+	pub restricted_balances: BTreeMap<EthereumAddress, Balance>,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub delegating_to: Option<AccountId32>,
```
acdibble, line 522: maybe this can be an optional struct and contain info about the :

```suggestion
	pub delegating_to: Option<DelegationInfo>,
	pub delegating_to_next: Option<DelegationInfo>,
```

```rs
struct DelegationInfo {
    delegating_to: AccountId32,
    bid: Balance,
}
```

`delegating_to` captures the info from the current epoch, `delegating_to_next` (or whatever better name there is) captures the current delegation info, i.e. if the auction resolves now, this will be their new delegation for the upcoming epoch

state-chain/runtime/src/lib.rs
dandanlen: Yes, this is the current choice.

state-chain/runtime/src/runtime_apis.rs
```diff
@@ -119,6 +121,50 @@ impl<BtcAddress> VaultSwapDetails<BtcAddress> {
...
+			apy_bp: old.apy_bp,
+			restricted_balances: old.restricted_balances,
+			estimated_redeemable_balance: old.estimated_redeemable_balance,
+			operator: None,
+		}
+	}
+}
+
 #[derive(Encode, Decode, Eq, PartialEq, TypeInfo, Serialize, Deserialize)]
 pub struct ValidatorInfo {
```
acdibble, line 169: would be nice to have the `total_stake` or whatever which is the stake + delegation for this node but i suppose we can read it from the operator if it's a pain

dandanlen: We can discuss this - I think another rpc is appropriate.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1658,6 +1659,15 @@ impl<T: Config> Pallet<T> {
...
 			Ok((auction_outcome, delegation_snapshots)) => {
+				let auction_outcome = AuctionOutcome {
+					winners: auction_outcome
+						.winners
+						.iter()
+						.map(ValidatorIdOf::<T>::from_ref)
+						.cloned()
+						.collect(),
+					bond: auction_outcome.bond,
+				};
```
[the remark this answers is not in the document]
dandanlen: Yes, but we only use it here, once. 
But I agree it would make things cleaner.

ramizhasan111 (approved):
some small comments but in general LGTM

PR #6100, opened 2025-09-06
fix: swap rate network fee calculation

j4m1ef0rd:
# Pull Request

Closes: PRO-2447

## Checklist

## Summary

I now ignore the network fee minimum when simulating the swap. This was the easiest solution (I tried others) and will only be inaccurate if the total network fee is smaller than the minimum, which is not a real concern.

- Network fee is now correct for small chunks that are affected the minimum.
- Moved the swap simulation RPC code to a file.
- Added integration tests to cover most of the swap simulation code.


state-chain/runtime/src/chainflip/simulate_swap.rs
```diff
@@ -0,0 +1,232 @@
...
+
+				(amount_after_fees, fees_withheld)
+			},
+		}
+	}
+
+	let include_fee = |fee_type: FeeTypes| !exclude_fees.contains(&fee_type);
+
+	// Default to using the DepositChannel fee unless specified.
+	let (amount_to_swap, ingress_fee) = if include_fee(FeeTypes::IngressDepositChannel) {
```
msgmaxim, line 147: Maybe subjective, but I find the `include_fee`/`remove_fees`/`exclude_fees` terminology very confusing. I'm assuming `include_fee` means the fee will be appied and `remove_fee` is you applying the fee? How about `charge_fee` instead of `remove_fee` (otherwise `remove` and `include` sounds like they have opposite meaning)?

state-chain/runtime/src/chainflip/simulate_swap.rs
```diff
@@ -0,0 +1,232 @@
...
+						index as OrderId,
+						tick,
+						sell_amount.into(),
+					)?;
+				},
+			}
+		}
+	}
+
+	fn remove_fees(
```
msgmaxim, line 85: Nit: there is no reason to nest this function inside `simulate_swap`. You already have a new scope since you created this new file, and so you won't be polluting external scope as long as the function is private.

state-chain/runtime/src/chainflip/simulate_swap.rs
```diff
@@ -0,0 +1,232 @@
...
+		CcmData, DispatchErrorWithMessage, FeeTypes, SimulateSwapAdditionalOrder,
+		SimulatedSwapInformation,
+	},
+	LiquidityPools, Runtime, Swapping,
+};
+
+use sp_runtime::traits::UniqueSaturatedInto;
+use sp_std::{collections::btree_set::BTreeSet, vec, vec::Vec};
+
+pub fn simulate_swap(
```
msgmaxim, line 42: Would be nice to document here what this function does and what its current limitations are.

state-chain/runtime/src/chainflip/simulate_swap.rs
```diff
@@ -0,0 +1,265 @@
...
+		(0, 0)
+	};
+
+	amount_to_swap.saturating_reduce(network_fee_amount_input_asset);
+
+	let amount_per_chunk: u128 = amount_to_swap / number_of_chunks;
+
+	if broker_commission > 0 {
+		let fee = FeeType::BrokerFee(
+			vec![Beneficiary { account: AccountId::new([0xbb; 32]), bps: broker_commission }]
```
msgmaxim, line 215: I think you can construct `bounded_vec!` directly here without having to convert?

PR #6101, opened 2025-09-08
feat: allow `lp_subscribe_order_fills` to return best blocks

marcellorigotti:
# Pull Request

Closes: PRO-2444

## Checklist

## Summary

Added optional parameter to the subscription to allow subscribing to best blocks instead of finalized only

### DIFF
--- a/api/bin/chainflip-lp-api/src/main.rs
+++ b/api/bin/chainflip-lp-api/src/main.rs
@@ -304,7 +304,11 @@ impl LpRpcApiServer for RpcServerImpl {
 			.tx_hash)
 	}
 
-	async fn subscribe_order_fills(&self, pending_sink: PendingSubscriptionSink, wait_finalized: Option<bool>) {
+	async fn subscribe_order_fills(
+		&self,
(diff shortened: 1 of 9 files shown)


dandanlen (approved):

api/bin/chainflip-lp-api/src/main.rs
```diff
@@ -304,9 +304,13 @@ impl LpRpcApiServer for RpcServerImpl {
 			.tx_hash)
 	}
 
-	async fn subscribe_order_fills(&self, pending_sink: PendingSubscriptionSink) {
+	async fn subscribe_order_fills(
+		&self,
+		pending_sink: PendingSubscriptionSink,
+		wait_finalized: Option<bool>,
```
dandanlen, line 310: Can we not expose the `NotificationBehaviour`?

I feel like it's important to be aware of the tradeoffs: 
`New` might duplicate events.
`Best` might miss events.

marcellorigotti: Yes makes sense, I wasn't aware of the difference!

PR #6131, opened 2025-09-18
feat: Use oracle price to estimate gas

marcellorigotti:
# Pull Request

Closes: PRO-2538

## Checklist

## Summary

Use oracle price as default choice to compute the amount of ingress/egress fee in case we don't have an internal price.
If oracle price is not available, default to hardcoded values.

### DIFF
--- a/state-chain/chains/src/arb.rs
+++ b/state-chain/chains/src/arb.rs
@@ -23,15 +23,13 @@ use crate::{
 	evm::{DeploymentStatus, EvmFetchId},
 	*,
 };
-use cf_amm_math::output_amount_ceil;
+use cf_primitives::chains::assets;
 pub use cf_primitives::chains::Arbitrum;
-use cf_primitives::{chains::assets, PriceFeedApi};
 use codec::{Decode, Encode, MaxEncodedLen};
 pub use ethabi::{ethereum_types::H256, Address, Hash as TxHash, Token, Uint, Word};
 use frame_support::sp_runtime::{traits::Zero, FixedPointNumber, FixedU64, RuntimeDebug};
 use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
-use sp_runtime::helpers_128bit::multiply_by_rational_with_rounding;
 use sp_std::{cmp::min, str};
 
 use self::evm::EvmCrypto;
@@ -44,6 +42,8 @@ impl Chain for Arbitrum {
 	const NAME: &'static str = "Arbitrum";
 	const GAS_ASSET: Self::ChainAsset = assets::arb::Asset::ArbEth;
 	const WITNESS_PERIOD: Self::ChainBlockNumber = 24;
+	const NATIVE_TOKEN_PRICE_IN_USD: Self::ChainAmount = eth::REFERENCE_ETH_PRICE_IN_USD;
+	const ONE_UNIT_IN_SMALLEST_UNITS: Self::ChainAmount = eth::ONE_ETH;
 
 	type ChainCrypto = EvmCrypto;
 	type ChainBlockNumber = u64;
@@ -63,31 +63,6 @@ impl Chain for Arbitrum {
 	type TransactionRef = H256;
 	type ReplayProtectionParams = Self::ChainAccount;
 	type ReplayProtection = evm::api::EvmReplayProtection;
-
-	fn input_asset_amount_using_reference_gas_asset_price<T: PriceFeedApi>(
-		input_asset: Self::ChainAsset,
-		required_gas: Self::ChainAmount,
-	) -> Self::ChainAmount {
-		match input_asset {
-			assets::arb::Asset::ArbEth => required_gas,
-			assets::arb::Asset::ArbUsdc =>
-				if let Some(relative_price) =
-					T::get_relative_price(Self::GAS_ASSET.into(), input_asset.into())
-				{
-					output_amount_ceil(U256::from(required_gas), relative_price.price)
-						.try_into()
-						.unwrap_or(0u128)
-				} else {
-					multiply_by_rational_with_rounding(
-						required_gas,
-						super::eth::REFERENCE_ETH_PRICE_IN_USD,
-						super::eth::ONE_ETH,
-						sp_runtime::Rounding::Up,
-					)
-					.unwrap_or(0u128)
-				},
-		}
-	}
 }
 
 #[derive(
--- a/state-chain/chains/src/arb.rs
+++ b/state-chain/chains/src/arb.rs
@@ -42,8 +42,9 @@ impl Chain for Arbitrum {
 	const NAME: &'static str = "Arbitrum";
 	const GAS_ASSET: Self::ChainAsset = assets::arb::Asset::ArbEth;
 	const WITNESS_PERIOD: Self::ChainBlockNumber = 24;
-	const NATIVE_TOKEN_PRICE_IN_FINE_USD: Self::ChainAmount = eth::REFERENCE_ETH_PRICE_IN_USD;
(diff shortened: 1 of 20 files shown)


state-chain/chains/src/arb.rs
```diff
@@ -66,16 +67,22 @@ impl Chain for Arbitrum {
...
-			.unwrap_or(0u128),
+			assets::arb::Asset::ArbUsdc =>
+				if let Some(price) = oracle_price {
+					output_amount_ceil(U256::from(required_gas), price).try_into().unwrap_or(0u128)
+				} else {
+					multiply_by_rational_with_rounding(
+						required_gas,
+						super::eth::REFERENCE_ETH_PRICE_IN_USD,
+						super::eth::ONE_ETH,
+						sp_runtime::Rounding::Up,
```
dandanlen, line 82: It's a bit odd that we need to use different methods depending on the source of the price... but I suppose we can address this another time.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2772,8 +2772,14 @@ pub mod pallet {
 			)
 			.and_then(|amount| C::ChainAmount::try_from(amount).ok())
 			.unwrap_or_else(|| {
-				log::warn!("Unable to calculate input amount required for gas of {required_gas:?} for input asset ${input_asset:?}. Estimating the input amount based on a reference price.");
-				C::input_asset_amount_using_reference_gas_asset_price(input_asset,required_gas)
+				log::warn!("Unable to calculate input amount required for gas of {required_gas:?} for input asset ${input_asset:?}. Estimating the input amount based on either oracle price or reference price.");
```
dandanlen, line 2775: I think we should remove this warning, especially now that we have a reliable fallback option. We don't need this polluting the node logs.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2772,8 +2772,14 @@ pub mod pallet {
...
+				let oracle_price = T::PriceFeedApi::get_price(input_asset.into());
+				let price = if oracle_price.as_ref().is_some_and(|price| !price.stale) {
+					Some(oracle_price.unwrap().price)
+				} else {
+					None
+				};
```
dandanlen, line 2781: We could just use the price even if it's stale? Even a stale price is likely to be better than the hard-coded estimate (which is by definition *more* stale)?

dandanlen (commented):
Implementation LGTM, just some suggestions.

state-chain/chains/src/arb.rs
```diff
@@ -66,16 +67,22 @@ impl Chain for Arbitrum {
 	fn input_asset_amount_using_reference_gas_asset_price(
 		input_asset: Self::ChainAsset,
 		required_gas: Self::ChainAmount,
+		oracle_price: Option<Price>,
```
MxmUrw, line 70: Could we try to move the querying of the price *into* this function? As it is now, to verify correctness, one has to look at all the call sites of this function, to make sure that the correct price for the correct asset is passed in, and there are no typing guardrails to prevent one from using the wrong asset for example.

state-chain/chains/src/arb.rs
```diff
@@ -66,16 +67,22 @@ impl Chain for Arbitrum {
...
-			assets::arb::Asset::ArbUsdc => multiply_by_rational_with_rounding(
-				super::eth::REFERENCE_ETH_PRICE_IN_USD,
-				required_gas,
-				1_000_000_000_000_000_000u128,
-				sp_runtime::Rounding::Up,
-			)
-			.unwrap_or(0u128),
+			assets::arb::Asset::ArbUsdc =>
+				if let Some(price) = oracle_price {
+					output_amount_ceil(U256::from(required_gas), price).try_into().unwrap_or(0u128)
```
MxmUrw, line 76: So, if I'm not mistaken, we have the following variables with the following units:
$$\textrm{oracle-price} \quad{in}\quad \frac{ETH_{fine}}{USD_{fine}}$$
$$\textrm{required-gas} \quad{in}\quad ETH_{fine}$$

This means the fees, if they would be taken in $USD_{fine}$, should be computed as:
$$\textrm{fee} = \frac{\textrm{required-gas}}{\textrm{oracle-price}} \quad{in}\quad \frac{USD_{fine}}{ETH_{fine}}$$.

But actually, we should also account for the USDC/USD price...

MxmUrw (commented):
Calculation looks correct.

state-chain/primitives/src/lib.rs
```diff
@@ -579,3 +579,16 @@ pub enum IngressOrEgress {
...
+	/// denominator is 2^128.
+	pub price: Price,
+
+	/// Whether the price is stale according to the oracle price ES settings.
+	pub stale: bool,
+}
+
+pub trait PriceFeedApi {
+	fn get_price(asset: Asset) -> Option<OraclePrice>;
+}
```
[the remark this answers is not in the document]
dandanlen: Hmm.. yeah - any reason why these moved out of `cf-traits`?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2824,6 +2811,55 @@ pub mod pallet {
 				}
 			}
 		}
+
+		fn input_asset_amount_using_reference_gas_asset_price<C: cf_chains::Chain>(
```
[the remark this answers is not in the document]
msgmaxim: But we can now remove `reference`, no? 🙂

state-chain/primitives/src/lib.rs
```diff
@@ -591,4 +591,5 @@ pub struct OraclePrice {
 
 pub trait PriceFeedApi {
 	fn get_price(asset: Asset) -> Option<OraclePrice>;
+	fn get_relative_price(asset1: Asset, asset2: Asset) -> Option<OraclePrice>;
```
msgmaxim, line 594: Should we move this back to cf_traits now that it is not necessary here?

state-chain/traits/src/mocks/price_feed_api.rs
```diff
@@ -29,4 +29,27 @@ impl PriceFeedApi for MockPriceFeedApi {
 		Self::get_storage::<_, Option<Price>>(ORACLE_PRICE, asset)
 			.and_then(|price| price.map(|price| OraclePrice { price, stale }))
 	}
+	fn get_relative_price(asset1: Asset, asset2: Asset) -> Option<OraclePrice> {
```
msgmaxim, line 32: You should be able to implement this directly on `PriceFeedApi` trait by means of calling `Self::get_price` (as the "default" implementation). Then implementors of `PriceFeedApi` (like this mock) won't need to duplicate this code.

state-chain/chains/src/lib.rs
```diff
@@ -307,6 +307,14 @@ pub trait Chain: Member + Parameter + ChainInstanceAlias {
 
 	const DEPRECATED: bool = false;
 
+	/// Reference price of 1 full native token (e.g. 1 ETH, 1 BTC, 1 SOL)
+	/// denominated in fineUSD (1e6 = $1.00).
+	const NATIVE_TOKEN_PRICE_IN_FINE_USD: Self::ChainAmount;
+
+	/// Number of smallest units that make up 1 full token.
+	/// For example: 1 ETH = 1_000_000_000_000_000_000 wei.
+	const SMALLEST_UNIT_PER_UNIT: Self::ChainAmount;
```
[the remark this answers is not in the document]
MxmUrw: I like `FINE_AMOUNT_PER_UNIT` as well.

msgmaxim (approved):

PR #6141, opened 2025-09-24
feat: support for non-native signed runtime calls and batching

albert-llimos:
## Pull Request

Closes: PRO-2509

## Checklist

## Summary

This adds support for off-chain signatures for arbitrary Ethereum and Solana payloads, pending support fotr the EIP-712 EVM offchain transactions.
https://eips.ethereum.org/EIPS/eip-712

While we could choose to just support EIP-712 for both EVM and Solana and ditch the domain approach, we support them as they are convenient for any party using libraries instead of wallets. Also, the EVM `personal_sign` (domain) is actually supported for wallets, just that's it's not human-readable.

So for both Phantom and Metamask we can sign using EIP-712, which is very user readable, with the Ethereum key. Signing with the Solana key is only supported using libraries for now, since Phantom's Solana key doesn't support this signing.

EIP-712 is being developed separately and is left as todo here, even if there was a first implementation done in #6109 . This is because we will most likely be leveraging ethers' code to implement that instead of implementing it from scratch.

More implementation details:

- This supports general runtime calls so this can substitute polkadot js calls one-to-one.
- Substrate native nonce accounts are used for the user account.
- They are submited as unsigned extrinsics and we have implemented the validation logic. This gets us as close as possible to native verification and avoids double signature verification.
- If the runtime call fails, the nonce is consumed but the rest of storage changes caused by the runtime call are rolled back.
- Batching is atomic.
- The Runtime Call are encoded in scale encoding and displayed as bytes in EIP-712.

To discuss/implement:
- Fee payment. Currently the tx fee is not paid by the signer. Since we use unsigned extrinsics we need to manually implement the fee taking. We will do in a separate PR.
- EIP-712 support (being developed separately)

### DIFF
--- a/state-chain/pallets/cf-environment/src/lib.rs
+++ b/state-chain/pallets/cf-environment/src/lib.rs
@@ -21,7 +21,7 @@
 
 use crate::submit_runtime_call::{batch_all, weight_and_dispatch_class, SignatureData};
 pub use crate::submit_runtime_call::{
-	BatchedCalls, TransactionMetadata, MAX_BATCHED_CALLS, UNSIGNED_CALL_VERSION,
+	BatchedCalls, Message, TransactionMetadata, MAX_BATCHED_CALLS, UNSIGNED_CALL_VERSION,
 };
 use cf_chains::{
 	btc::{
@@ -649,14 +649,13 @@ pub mod pallet {
 		#[allow(clippy::useless_conversion)]
 		#[pallet::call_index(10)]
 		#[pallet::weight({
-			let di = call.get_dispatch_info();
+			let di = message.call.get_dispatch_info();
 			let dispatch_weight = di.weight.saturating_add(T::WeightInfo::non_native_signed_call());
 			(dispatch_weight, di.class)
 		})]
 		pub fn non_native_signed_call(
 			origin: OriginFor<T>,
-			call: scale_info::prelude::boxed::Box<<T as Config>::RuntimeCall>,
-			_transaction_metadata: TransactionMetadata,
+			message: Message<<T as Config>::RuntimeCall>,
 			signature_data: SignatureData,
 		) -> DispatchResultWithPostInfo {
 			// unsigned extrinsic - validation happens in ValidateUnsigned
@@ -665,7 +664,7 @@ pub mod pallet {
 			let signer_account: T::AccountId =
 				signature_data.signer_account().map_err(|_| Error::<T>::FailedToDecodeSigner)?;
 
-			let _ = call.dispatch_bypass_filter(OriginTrait::signed(signer_account))?;
+			let _ = message.call.dispatch_bypass_filter(OriginTrait::signed(signer_account))?;
 
 			Self::deposit_event(Event::<T>::NonNativeSignedCall);
 			Ok(().into())
@@ -696,20 +695,19 @@ pub mod pallet {
 
 		fn validate_unsigned(_source: TransactionSource, call: &Self::Call) -> TransactionValidity {
 			if let Call::non_native_signed_call {
-				call: inner_call,
-				transaction_metadata,
+				message: Message { call: inner_call, metadata },
 				signature_data,
 			} = call
 			{
 				let Ok(signer_account) = signature_data.signer_account() else {
 					return Err(InvalidTransaction::BadSigner.into());
 				};
-				let valid_tx = validate_metadata::<T>(transaction_metadata, &signer_account)?;
+				let valid_tx = validate_metadata::<T>(metadata, &signer_account)?;
 				ensure!(
 					is_valid_signature(
 						inner_call,
-						ChainflipNetworkName::<T>::get().as_str(),
-						transaction_metadata,
+						ChainflipNetworkName::<T>::get(),
+						metadata,
 						signature_data
 					),
 					InvalidTransaction::BadProof
@@ -721,14 +719,18 @@ pub mod pallet {
 		}
 
 		fn pre_dispatch(call: &Self::Call) -> Result<(), TransactionValidityError> {
-			if let Call::non_native_signed_call { transaction_metadata, signature_data, .. } = call
+			if let Call::non_native_signed_call {
+				message: Message { metadata, .. },
+				signature_data,
+				..
+			} = call
 			{
 				let Ok(signer_account) = signature_data.signer_account() else {
 					return Err(InvalidTransaction::BadSigner.into());
 				};
 
 				// Signature validity already checked in `validate_unsigned`
-				let _ = validate_metadata::<T>(transaction_metadata, &signer_account)?;
+				let _ = validate_metadata::<T>(metadata, &signer_account)?;
 				frame_system::Pallet::<T>::inc_account_nonce(&signer_account);
 				Ok(())
 			} else {
--- a/state-chain/pallets/cf-environment/src/submit_runtime_call.rs
+++ b/state-chain/pallets/cf-environment/src/submit_runtime_call.rs
@@ -31,6 +31,12 @@ pub const UNSIGNED_CALL_VERSION: &str = "0";
 
 pub type BatchedCalls<T> = BoundedVec<<T as Config>::RuntimeCall, ConstU32<MAX_BATCHED_CALLS>>;
 
+#[derive(Clone, Debug, Encode, Decode, TypeInfo, PartialEq)]
+pub struct Message<C> {
+	pub call: scale_info::prelude::boxed::Box<C>,
+	pub metadata: TransactionMetadata,
+}
+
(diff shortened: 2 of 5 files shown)


state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -576,6 +619,73 @@ pub mod pallet {
...
+			let (dispatch_weight, dispatch_class) = weight_and_dispatch_class::<T>(calls);
+			let dispatch_weight = dispatch_weight.saturating_add(T::WeightInfo::submit_unsigned_batch_runtime_call(calls.len() as u32));
+			(dispatch_weight, dispatch_class)
+		})]
+		pub fn submit_unsigned_batch_runtime_call(
+			origin: OriginFor<T>,
+			calls: Vec<<T as Config>::RuntimeCall>,
+			transaction_metadata: TransactionMetadata,
+			user_signature_data: UserSignatureData,
+		) -> DispatchResult {
```
dandanlen, line 635: A couple of things:

- We don't need a Vec of calls here. We have a `batch` runtime call so it can just be a single call (which might be `batch`. 
- It follows from the above then that `atomic` doesn't need to be in the transaction metadata. 
- It's not really an unsigned call - it's signed with a non-native wallet, so calling it unsigned and then telling users to submit a signed call via submit_unsigned is a bit confusing. I would rename this to something like `non_native_signed_call` or even `submit_eip712_call`.

state-chain/pallets/cf-environment/src/submit_runtime_call.rs
```diff
@@ -0,0 +1,311 @@
...
+	let calls_len = calls.len();
+
+	for (index, call) in calls.into_iter().enumerate() {
+		let info = call.get_dispatch_info();
+
+		let origin = frame_system::RawOrigin::Signed(signer_account.clone()).into();
+
+		// Don't allow users to nest calls.
+		if let Some(Call::submit_unsigned_batch_runtime_call { .. }) |
+		Some(Call::submit_batch_runtime_call { .. }) = call.is_sub_type()
```
dandanlen, line 155: Why no nesting?

state-chain/pallets/cf-environment/src/submit_runtime_call.rs
```diff
@@ -0,0 +1,311 @@
...
+		let result = call.dispatch_bypass_filter(origin);
+
+		weight =
+			weight.saturating_add(frame_support::dispatch::extract_actual_weight(&result, &info));
+
+		if let Err(mut err) = result {
+			let base_weight = weight_fn(index.saturating_add(1) as u32);
+			err.post_info = Some(base_weight.saturating_add(weight)).into();
+			return Err((index, err));
+		};
```
dandanlen, line 174: We always exit early - is this likely to be surprising? Either way it should be made clear in the doc comment. 

ie. if the user submits `batch { calls: [A, B, C], atomic: false }` and it fails at B, then only A will be processed and C will not even be attempted.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -576,6 +619,73 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		#[pallet::call_index(11)]
+		#[pallet::weight({
+			let (dispatch_weight, dispatch_class) = weight_and_dispatch_class::<T>(calls);
+			let dispatch_weight = dispatch_weight.saturating_add(T::WeightInfo::submit_batch_runtime_call(calls.len() as u32));
+			(dispatch_weight, dispatch_class)
+		})]
+		pub fn submit_batch_runtime_call(
```
dandanlen, line 664: The extrinsics should be documented - basically the doc comments added to the inner functions should be here as well (or instead). This way they will appear in the metadata.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -576,6 +619,73 @@ pub mod pallet {
...
+			let account_id = ensure_signed(origin)?;
+
+			let _ = dispatch_user_calls::<T>(
+				calls.clone(),
+				account_id.clone(),
+				atomic,
+				T::WeightInfo::submit_unsigned_batch_runtime_call,
+			);
+
+			Ok(())
```
dandanlen, line 678: This always returns ok - I would imagine that might be unexpected if for example an atomic batched call fails.

dandanlen (commented):
tbh. I would leave batching out of this PR, it can be added separately. 

I think we can revisit if/when it's clear what the use cases are. For example, right now I can't think of a good use-case for non-atomic batches... Non-atomic batches are weird from the user's POV because they don't get a simple success/fail: they need to inspect the events to see what happened. So if we can decide for example that we only support atomic batches, the API and implementation is much simpler.

state-chain/pallets/cf-environment/src/submit_runtime_call.rs
albert-llimos: It could end up with an infinite loop of calls. I would expect this to be handled appropriately with transaction cost, but I saw that the utility pallet has some check:
https://paritytech.github.io/polkadot-sdk/master/src/pallet_utility/lib.rs.html#330
Happy to remove it if you deem it to be fine.

state-chain/pallets/cf-environment/src/submit_runtime_call.rs
albert-llimos: > is this likely to be surprising?
I don't know what you are asking. I will document the expected behavior though (the example you gave).

albert-llimos:
> Hence my suggestion to remove it until there is a concrete use-case that can help us to decide which of the many options we should support.

Seems reasonable. I will remove the support for non-atomic batches.

state-chain/primitives/src/chains/assets.rs
```diff
@@ -132,6 +132,11 @@ macro_rules! assets {
 						)+
 					}
 				}
+				pub fn as_bytes(&self) -> &'static [u8] {
+					match self {
+						$( $( Self::$asset_variant => concat!($chain_string, "-", $asset_string).as_bytes(), )+ )+
```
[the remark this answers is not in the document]
albert-llimos: Actually we don't use this now that we have settled on the RuntimeCall being hex encoded. I will remove this logic.

state-chain/pallets/cf-environment/src/submit_runtime_call.rs
albert-llimos: I removed the check for `non_native_call` as when calling it from the batch it will fail anyway with BadOrigin.

state-chain/custom-rpc/src/eip_712_types.rs
```diff
@@ -0,0 +1,138 @@
...
+			],
+			"RuntimeCall": [
+				{
+					"name": "value",
+					"type": "bytes"
+				}
+			],
+			"Transaction": [
+				{
+					"name": "Call",
```
dandanlen, line 115: nit: field names should probably be lower case. (call and metadata)

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -576,6 +621,82 @@ pub mod pallet {
...
+		#[pallet::weight({
+			let di = call.get_dispatch_info();
+			let dispatch_weight = di.weight.saturating_add(T::WeightInfo::non_native_signed_call());
+			(dispatch_weight, di.class)
+		})]
+		pub fn non_native_signed_call(
+			origin: OriginFor<T>,
+			call: scale_info::prelude::boxed::Box<<T as Config>::RuntimeCall>,
+			_transaction_metadata: TransactionMetadata,
+			user_signature_data: UserSignatureData,
```
dandanlen, line 642: I think that after Ramiz' is finished with his work it might make sense to define a wrapper that includes the TransactionMetadata and Domain. 

Since both of these need to be included in the signed payload, they will need to have the corresponding type info extracted etc. If we define a wrapper and derive TypeInfo etc. it should all 'just work'.

And we can remove 'signature_type' from the UserSignatureData. It doesn't really describe the signature, it describes how to derive the payload - ie. the domain.

Ie. something like this:

```rust
#[derive(Encode, Decode, TypeInfo)]
struct Message<C> {
    domain: Domain,
    metadata: Metadata,
    call: C,
}

pub fn non_native_signed_call(
    origin: OriginFor<T>,
			call: Box<Message<T::RuntimeCall>>,
			user_signature_data: UserSignatureData,
) -> DispatchResult { 
    // ..
}
```

cc @ramizhasan111

dandanlen: Ok I see what you mean - sometimes we don't want to construct an EIP-712 compatible payload, an in that case we don't need to encode any domain data at all.

But I don't really see a need to keep it separate since when it's required it *is* part of the payload. 

... I'll have a think about it.

PR #6142, opened 2025-09-25
fix: ensure bond for managed validators is distributed correctly

dandanlen:
Previously it was possible for the total managed validator bonds to be set too low if any of the validators had more stake than the current bond. 

I also noticed that validators were able to register at lower than the min stake *if* the number of authorities were low. I removed this, it doesn't make much sense IMO.

### DIFF
--- a/state-chain/pallets/cf-validator/src/delegation.rs
+++ b/state-chain/pallets/cf-validator/src/delegation.rs
@@ -421,6 +421,42 @@ mod tests {
 		}
 	}
 
+	proptest! {
+		#[test]
+		fn validator_bond_amounts_capped(
+			validator_amounts in prop::collection::vec(1u128..1_000_000u128, 1..10),
+			bond in 1u128..1_000_000u128,
+		) {
+			// Create a delegation snapshot
+			let snapshot = DelegationSnapshot::<ValidatorId, u128> {
+				validators: validator_amounts.iter().enumerate()
+					.map(|(i, bid)| (i as u64 + 100, *bid))
+					.collect(),
+				..Default::default()
+			};
+
+			// Distribute the total amount
+			new_test_ext().execute_with(|| {
+				let dist = snapshot.validator_bond_distribution(bond);
+
+				// Total bond cannot exceed
+				let max_expected_bond = bond * snapshot.validators.len() as u128;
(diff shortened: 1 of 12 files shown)


state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -168,6 +168,22 @@ where
...
+		validator_bids.sort_by_key(|(_, v)| *v);
+		validator_bids
+			.into_iter()
+			.map(|(id, bid)| {
+				let individual_bond = core::cmp::min(bid, total_bond);
+				total_bond.saturating_reduce(individual_bond);
+				(id, individual_bond)
+			})
+			.collect()
+	}
```
ramizhasan111, line 185: I think this works well when not considering the delegators and we just have to distribute bonds among managed validators. i am thinking maybe this might not be optimal when considering that the operator will also have delegators. Delegators will raise indirectly raise the bond at auction resolution and since the delegators, for a particular operator, are also bonded, we might not need to bond each validator the full amount (since some economic security is achieved by bonding the validators). Maybe we should divide the bond value proportionally based on total validator stake and total delegator stake and then bond the validators using this reduced bond.

dandanlen: I think I disagree with this - validators are rewarded proportionally to their share of MAB. The bond should reflect this.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1565,6 +1559,16 @@ impl<T: Config> Pallet<T> {
...
 		}
 
+		for (index, account_id) in new_authorities.iter().enumerate() {
+			AuthorityIndex::<T>::insert(new_epoch, account_id, index as AuthorityCount);
+			EpochHistory::<T>::activate_epoch(account_id, new_epoch);
+			if let Some(bond) = new_managed_validator_bonds.get(account_id.into_ref()) {
+				T::Bonder::update_bond(account_id, *bond);
+			} else {
+				T::Bonder::update_bond(account_id, EpochHistory::<T>::active_bond(account_id));
+			}
```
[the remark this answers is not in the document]
dandanlen: Update - this is done now.

PR #6143, opened 2025-09-26
chore: removed unwraps, log_or_panic instead.

j4m1ef0rd:
# Pull Request

Closes: PRO-2528

## Checklist

## Summary

The problem is that check_swap_price_violation and calculate_oracle_delta both must be called on a finished swap.
Refactoring with a new swap struct to avoid the unwrap would be quite difficult.
So instead I just ignore the check and delta calculation and log_or_panic. This is simple fix and avoids the unwraps.


dandanlen:
I changed it to use `let Some(_) else`, it makes it a bit more readable (the diff is more readable too).

PR #6144, opened 2025-09-26
refactor: better channel action handling in ingress-egress pallet

MxmUrw:
# Pull Request

Motivated as part of PRO-2451.

Closes: PRO-2553.

This extracts the ingress-egress pallet refactors of https://github.com/chainflip-io/chainflip-backend/pull/6038.

## Checklist

## Summary

While working on the "refund undecodable bitcoin vault swaps"-PR (PRO-2451), I tried to understand the ingress-egress pallet, and in that process, came upon various places where the code could be improved. This PR is the result. It would be useful to merge this before working on PRO-2551.

### Refactors:
1. When constructing ChannelAction::Refund, directly include the CcmDepositMetadata with source_chain and source_address. Then in BLS of "full witness" uses that instead of reconstructing it.
2. Make the `ChannelAction` struct parametric over it's CcmMetadata. At channel construction time we don't have deposit data, so we want to use `CcmChannelMetadata`. When we have an actual deposit, we want to use `CcmDepositMetadata`.
3. Move the validation of vault swap refund parameters into `try_validate_vault_swap`. Previously this logic was duplicated between witness and prewitness.
4. Change the type of the refund address inside of the `refund_params` field of `ChannelAction::Swap` from AccountOrAddress to ForeignChainAddress. We never used the `AccountOrAddress::InternalAccount` variant and it seems like this change was more or less accidental.
 
### Question:
1.  @albert-llimos: The current behaviour (on main) is: If we have a Ccm refund due to failure of `try_validate_vault_swap`, then the `CcmDepositMetadata` when passing
 the refund to `Self::schedule_egress()` has `source_address == None` due to how we propagate the error. Is this intentional?
2. @albert-llimos @dandanlen: I'm motioning to rename the struct `ChannelRefundParameters<A, CcmRefundDetails>` to simply `RefundParameters<A, CcmRefundDetails>` because we use it in two different ways, we either insert either deposit or channel metadata into the `CcmRefundDetails` type parameter. I haven't done this yet, and of course we can do that in a later PR.

 ### Impact:
 Types that are changed and require migration:
  - [x] ChannelAction. Changed the type of `Swap { refund_params }`.
      - from: `ChannelRefundParameters<AccountOrAddress, Option<CcmDepositMetadataChecked<ForeignChainAddress>>>`
      - to : `ChannelRefundParameters<ForeignChainAddress, Option<CcmChannelMetadataChecked>>`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1416,6 +1416,7 @@ dependencies = [
  "itertools 0.13.0",
  "libsecp256k1",
  "log",
+ "n-functor",
  "parity-scale-codec",
  "proptest",
  "rand",
@@ -1550,6 +1551,7 @@ version = "0.1.0"
 dependencies = [
  "frame-support",
  "hex",
+ "n-functor",
  "parity-scale-codec",
  "proptest",
  "proptest-derive",
@@ -7521,6 +7523,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "n-functor"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e279584350f406d0be657b1f66ed3450c2da43cfa154f8320470984e81b28b9d"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.96",
+]
+
+[[package]]
 name = "nalgebra"
 version = "0.33.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -8347,6 +8360,7 @@ dependencies = [
  "generic-typeinfo-derive",
  "hex-literal",
  "log",
+ "n-functor",
  "pallet-cf-governance",
  "parity-scale-codec",
  "scale-info",
--- a/state-chain/pallets/cf-ingress-egress/src/benchmarking.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/benchmarking.rs
@@ -111,22 +111,24 @@ mod benchmarks {
 			let source_asset: <<T as Config<I>>::TargetChain as Chain>::ChainAsset =
 				BenchmarkValue::benchmark_value();
 			let block_number = TargetChainBlockNumber::<T, I>::benchmark_value();
-			let mut channel =
-				DepositChannelDetails::<T, I> {
-					owner: account("doogle", 0, 0),
-					opened_at: block_number,
-					expires_at: block_number,
-					deposit_channel: DepositChannel::generate_new::<
-						<T as Config<I>>::AddressDerivation,
-					>(1, source_asset)
+			let mut channel = DepositChannelDetails::<T, I> {
+				owner: account("doogle", 0, 0),
+				opened_at: block_number,
+				expires_at: block_number,
+				deposit_channel:
+					DepositChannel::generate_new::<<T as Config<I>>::AddressDerivation>(
+						1,
+						source_asset,
+					)
 					.unwrap(),
-					action: ChannelAction::<T::AccountId, <T::TargetChain as Chain>::ChainAccount>::LiquidityProvision {
+				action:
+					ChannelAction::<T::AccountId, TargetChainAccount<T, I>>::LiquidityProvision {
 						lp_account: account("doogle", 0, 0),
 						refund_address: ForeignChainAddress::benchmark_value(),
 					},
-					boost_fee: 0,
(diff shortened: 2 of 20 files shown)


Cargo.lock
```diff
@@ -7520,6 +7522,17 @@ dependencies = [
  "unsigned-varint 0.7.2",
 ]
 
+[[package]]
+name = "n-functor"
```
msgmaxim, line 7526: What do we know about this new dependency? Doesn't look like it is widely known/used. Does the benefit from using it justify the extra cost/risk?

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -563,15 +536,21 @@ pub mod pallet {
...
+	pub enum ChannelAction<AccountId, ChainAccount, CcmMetadata = CcmChannelMetadataChecked> {
 		Swap {
 			destination_asset: Asset,
 			destination_address: ForeignChainAddress,
+			#[map_with(|arg, f| map_bounded_vec(arg, |x: Beneficiary<_>| x.map(f)))]
 			broker_fees: Beneficiaries<AccountId>,
-			channel_metadata: Option<CcmChannelMetadataChecked>,
-			refund_params: ChannelRefundParametersCheckedInternal<AccountId>,
+			channel_metadata: Option<CcmMetadata>,
+			#[map_with(|arg: ChannelRefundParameters<_,_>, f0, f1| arg.map(|x: AccountOrAddress<_,_>| x.map(f0, |a| a), |z: Option<_>| z.map(f1)))]
```
msgmaxim, line 549: Personally I find this less readable than the more explicit version he had before 🤷

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2743,60 +2692,98 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
+		let refund_action = |reason| ChannelAction::Refund {
+			reason,
+			refund_address,
+			refund_ccm_metadata: checked_refund_params.refund_ccm_metadata.clone().map(
+				|mut refund_ccm_metadata| {
+					// TODO: Check: @Albert is this intentional? setting the
+					// source_address to None is what implicitly happened in the
+					// refund flow until now
+					refund_ccm_metadata.source_address = None;
```
albert-llimos, line 2716: It was not intentional. It'd make no difference as noone is relying on it but we need to fix it of course. We should be using the  source address from `deposit_metadata`. We can map it directly here or after having checked it below, it doesn't really matter as the `to_checked` doesn't do any check on the source_address anyway.

state-chain/pallets/cf-ingress-egress/src/benchmarking.rs
```diff
@@ -76,7 +76,7 @@ mod benchmarks {
 						source_asset,
 					)
 					.unwrap(),
-				action: ChannelAction::<T::AccountId, T::TargetChain>::LiquidityProvision {
+				action: ChannelAction::<T::AccountId, <T::TargetChain as Chain>::ChainAccount>::LiquidityProvision {
```
dandanlen, line 79: FYI there's a type alias `TargetChainAccount<T>`

dandanlen (commented):
I can see why the use of `n-functor` might be a little controversial but IMO it's a net benefit if we can remove the boilerplate of adding these mapping functions everywhere.

dandanlen (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
MxmUrw: > We can map it directly here or after having checked it below.
It's already mapped slightly above, so I deleted this line (`... = None`) and we should be good to go.

state-chain/pallets/cf-ingress-egress/src/lib.rs
MxmUrw: Yeah you're totally right about this line, I'll try to make it more readable.

MxmUrw:
@dandanlen @albert-llimos I've also changed the refund address from `AccountOrAddress` to `ForeignChainAddress`, it was in the `ChannelAction` struct which is pretty easy to change since the only point it's referenced is the `DepositChannelLookup` storage.

Ran the upgrade-tests again, passed.

PR #6146, opened 2025-09-29
feat: Ethereum EIP712 Encoding

ramizhasan111:
# Pull Request

Closes: PRO-2535

## Checklist

## Summary

### DIFF
--- a/state-chain/pallets/cf-environment/src/tests.rs
+++ b/state-chain/pallets/cf-environment/src/tests.rs
@@ -814,8 +814,8 @@ fn can_non_native_signed_call() {
 }
 // Our encoding type encodes the path, encoding `*_mock_*` into the type names for these unit tests.
 // That is not the same for the bouncer. Therefore to generate this test vector, we must manually
-// create the EIP-712 payload in this test and and sign it externally. Then paste the signature here.
-// this can be removed once the testing is expanded.
+// create the EIP-712 payload in this test and and sign it externally. Then paste the signature
+// here. this can be removed once the testing is expanded.
 #[test]
 fn can_build_eip_712_payload_and_validate() {
 	new_test_ext().execute_with(|| {


state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+								.collect()
+						},
+						//todo
+						_ => Default::default(),
+					},
+				))
+			})
+			.collect(),
+		primary_type: T::type_info().path.segments.last().unwrap().to_string(),
+		message: match serde_json::to_value(&value).unwrap() {
```
dandanlen, line 55: The more I think of it, the more I'm convinced that the split JSON/TypeInfo approach is not a good one. JSON is used for other things (mainly rpcs). So imagine that someone decides to alter the JSON encoding of some value using a serde annotation (for example using a tagged enum instead of the default encoding). This will break our EIP712 compatibility. 

Either:
- we convert to JSON and use the JSON value to derive the EIP type information (but as discussed, JSON removes things like the type names, which will make things tricky).

Or:
- We use SCALE/TypeInfo and derive the value/type information this way.

I still think the second option is the better way :)

We can discuss later.

albert-llimos: (I don't have context on what you have discussed so please disregard if the following comment doesn't make sense)

I think the nice thing about JSON here for the message, besides that we can use ethers as is with it, is that it works well with the encoding for the product side of things.
We will need to encode the message in an rpc call for Product to then plug it into the wallets. That will need to be an object json-like, without the types like this:

```
  const message = {
    Batch: {
      Borrow: {
        amount: 1234,
        collateralAsset: 'Bitcoin-BTC',
        borrowAsset: 'Ethereum-USDC',
      },
      Withdraw: {
        target: address,
      },
    },
    metadata: {
      from: address,
      nonce: 1,
      expiryBlock: 10000,
    },
  };
  ```
 I suspect that this is why ethers does it with JSON. I imagine it will also work via SCALE/TypeInfo but I'm wondering if we are causing extra headaches to ourselves by not using this as is.

Comment (dandanlen) [reply state-chain/ethereum-eip712/src/lib.rs:55]: What I mean is that just calling `serde_json::to_value(&value)` does not give us the above format.

We need to construct some other value that conforms to this and simultaneous can be derived from the type info at runtime.

Comment (albert-llimos) [reply state-chain/ethereum-eip712/src/lib.rs:55]: I am testing it out and having the message in this format does give us the type exactly as product wants:
`pub message: BTreeMap<String, serde_json::Value>`

File: state-chain/ethereum-eip712/src/eip712.rs [hunk 2]
```diff
@@ -0,0 +1,1032 @@
...
+	primary_type: &str,
+	data: &scale_value::Value,
+	types: &Types,
+) -> Result<Vec<Token>, Eip712Error> {
+	let hash = hash_type(primary_type, types)?;
+	let mut tokens = vec![Token::Uint(U256::from(hash))];
+
+	if let Some(fields) = types.get(primary_type) {
+		for (field, value) in
+			fields.iter().zip(data.get_scale_value_fields().map_err(|e| {
```
dandanlen: Are we guaranteed the the fields in the Eip712DomainType are in the same order as the fields in the `scale_value::Value`?

state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+	fields: Vec<Field>,
+	comp_value: Composite<C>,
+	types: &mut BTreeMap<String, Vec<Eip712DomainType>>,
+) -> Result<Vec<Eip712DomainType>, &'static str> {
+	match comp_value {
+		Composite::Named(fs) => {
+			let fs_map = fs.into_iter().collect::<BTreeMap<_, _>>();
+			fields
+				.clone()
+				.into_iter()
```
dandanlen: Seems like a lot of unnecessary cloning... but also, why do we collect into a map? Do we not want to preserve the order of the fields?

*edit: I see it's so that we can look up the types by field name, but elsewhere we assume the ordering is preserved...*

state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+
+		#[derive(TypeInfo, Clone, Encode, Decode)]
+		pub struct Person {
+			pub name: String,
+		}
+
+		let payload = Mail {
+			from: Person { name: String::from("Ramiz") },
+			to: Person { name: String::from("Albert") },
+			message: String::from("hello Albert"),
```
dandanlen: I doubt this is in the official test vectors 😅 
How did we generate the reference output? Would be good to document this, or otherwise to use an official reference value from the EIP.

state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+						process_composite(type_def_composite.fields, comp_value, types)?,
+					),
+				},
+			(TypeDef::Variant(type_def_variant), ValueDef::Variant(value_variant)) => (
+				t.path
+					.ident()
+					// Should never error. Its a composite type so the name has to exist
+					.ok_or("Type doesnt have a name")?
+					.to_string() + "__" +
+					&value_variant.name.to_string(),
```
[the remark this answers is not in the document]
ramizhasan111: the value variant name will be the exact variant the value is instantiated with and so we want to use that variant name as the struct name.

state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+			(TypeDef::BitSequence(_), _) => return Err("Unreachable"),
+
+			_ => return Err("Type and Value do not match"),
+		};
+
+	// If there are generic parameters to this type, append uniqueness to the type name to avoid
+	// collisions
+	if t.type_params.len() > 0 {
+		type_name = type_name + "__" + &hex::encode(&keccak256(format!("{fields:?}"))[..4]);
+	}
```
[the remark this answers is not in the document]
ramizhasan111: The fields here only have name and type names, not the value.

state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+				.clone()
+				.into_iter()
+				.map(|field| -> Result<_, &'static str> {
+					// shouldn't be possible since we are in Named variant
+					let field_name = field.name.ok_or("field name doesn't exist")?.to_string();
+					let value =
+						fs_map.get(&field_name).ok_or("field with this name has to exist")?.clone();
+					Ok(Eip712DomainType {
+						name: field_name.clone(),
+						// find out in which cases type name would be empty.
```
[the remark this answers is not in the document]
ramizhasan111: this looks like a stale comment. will remove.

state-chain/ethereum-eip712/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+					.clone()
+					.into_iter()
+					.zip(fs.into_iter())
+					.enumerate()
+					.map(|(i, (ty, value))| -> Result<_, &'static str> {
+						let type_name = recursively_construct_types(value.clone(), ty, types)?;
+						Ok(Eip712DomainType {
+							// In case of unnamed fields, we decide to name it by its type name
+							// appended by its index in the tuple
+							name: type_name.clone() + "__" + &i.to_string(),
```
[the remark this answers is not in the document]
albert-llimos: I agree that the type doesn't add much info for the user but at the same time having `field` doesn't add anything useful either. What about just having the index?

dandanlen (approved):
As discussed/mentioned previously - happy to merge this to accelerate integration with product, and improve testing in the meantime.

albert-llimos (approved):
Approved! As discussed, Ramiz will work on adding exhaustive testing for this in a followup PR.

PR #6149, opened 2025-10-01
[RELEASE CANDIDATE] 1.11.6

dandanlen:
Contains some runtime-only fixes to delegation.

From these PRs:
- #6148 
- #6142 

One of the commits from the second PR was left out ([this](https://github.com/chainflip-io/chainflip-backend/pull/6142/commits/be71505da50c50faee83c57fca04e86c472882c6) one) - it was a non-functional refactor of the storage items. This would have broken engine compatibility, so I saved it for the next release.

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -994,8 +994,23 @@ pub mod pallet {
 			ensure!(settings.fee_bps <= MAX_OPERATOR_FEE, Error::<T>::OperatorFeeTooHigh);
 
 			if let Some(current_settings) = OperatorSettingsLookup::<T>::get(&operator) {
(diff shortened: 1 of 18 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -834,13 +834,10 @@ pub mod pallet {
 		#[pallet::weight(T::ValidatorWeightInfo::register_as_validator())]
 		pub fn register_as_validator(origin: OriginFor<T>) -> DispatchResult {
 			let account_id: T::AccountId = ensure_signed(origin)?;
-			if Self::current_authority_count() >= AuctionParameters::<T>::get().max_size {
```
nmammeri, line 837: Looks like a functional change, but looks fine to me that we always require `MinimumValidatorStake`. might need to update docs to reflect this.

PR #6159, opened 2025-10-08
fix: initialize generic elections in migration

MxmUrw:
## Summary

Since in CI we're running the upgrade test from 1.10.4 to 1.11.11, we're actually upgrading from a version of the network without the generic_elections pallet. This means that we still have to include the initialization code, even though all our live networks are already on 1.11.x.


dandanlen (approved):

dandanlen:
This can be closed now, right?

PR #6166, opened 2025-10-14
feat: Exclude nodes failing keygen from delegation snapshots

dandanlen:
# Pull Request

Closes: PRO-2565 , PRO-2803

This fixes an issue where validator nodes failing keygen would still be included in the delegation snapshot when the keygen is retried. 

It also included a refactor of the logic to make the side-effects more explicit (functions return results and errors are handled at the call-site rather than deep inside the call stack).

### DIFF
--- a/state-chain/pallets/cf-validator/src/benchmarking.rs
+++ b/state-chain/pallets/cf-validator/src/benchmarking.rs
@@ -121,10 +121,11 @@ pub fn try_start_keygen<T: RuntimeConfig>(
 	init_bidders::<T>(primary_candidates, epoch, 100_000u128);
 	init_bidders::<T>(secondary_candidates, epoch + LARGE_OFFSET, 90_000u128);
 
-	Pallet::<T>::try_start_keygen(RotationState::from_auction_outcome::<T>(AuctionOutcome {
+	Pallet::<T>::start_keygen_attempt(RotationState::from_auction_outcome::<T>(AuctionOutcome {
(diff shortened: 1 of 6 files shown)


ramizhasan111 (commented):
It looks like when the keygen fails, we only work with the existing delegation snapshot by removing banned validators from the snapshot but I think we should also consider any changes to the delegation and validator stakes. So basically, i think the more correct way is to re-run the auction from the start as if you are starting from the beginning again. We would build the delegation snapshot again. at this point we can exclude the banned ones but maybe integrating that functionality in qualified bidders.

This would not only be more correct but  would also mean we re-use the same auction code and we would not need to write a lot of custom code we have written here. we can see here that a lot of the auction code is basically duplicated (```re_resolve_auction_after_ban()```). In the suggested option, we would just need to integrate the banning functionality and the marking of banned validators as delegators before we re-run the auction using the same flow. The banning functionality can be done as part of get_qualified_bidders() nicely and the moving of validator to delegator can be done as an optional step right after we build the delegation snapshot by calling a method on delegation snapshot and passing in banned ids.

dandanlen:
@ramizhasan111 @marcellorigotti I think you are both suggesting basically the same thing and i agree. Previously we would resolve the auction once and 'freeze' the balances at that point. But I suppose this is not a really a critical property, we can relax this and the benefits are like Ramiz said: the code should end up much simpler.

One thing we would need to consider is if too many validators are banned, we need a way to reset the bans.

ramizhasan111 (approved):
just a small comment, otherwise looks good.

PR #6396, opened 2026-02-03
feat: bsc integration

ramizhasan111:
# Pull Request

Closes: PRO-2693, PRO-2691, PRO-2685, PRO-2669, PRO-2667, PRO-2666, PRO-2665

## Summary

* This PR adds support for BSC chain
* Also implements BSC broker level screening 
* Only BNB and BscUsdt assets are supported
* As part of the upgrade test, Bsc is initialized without needing to rotate.

---

* The PR is complete:
  * Tests:


albert-llimos:
I know it's obvious but something that is left to do in this PR is to write migrations and add the extra logic in the upgrade test to make it work.  We can get inspiration from the Tron PR, which has exactly the same setup to start a new node and setup a new chain and asset in the CI upgrade test:
https://github.com/chainflip-io/chainflip-backend/pull/6508/changes#diff-fb56dbf5e717892486fea44843511f8f08436ed0bda8df952befdd9e165a04d5
Same thing with the environment migration:
https://github.com/chainflip-io/chainflip-backend/pull/6508/changes#diff-6e71c79c842c150d3670eb028cc31af95ef672da18cd3d5781b8d3aaa9a1d8e3

bouncer/shared/initialize_new_chains.ts
```diff
@@ -50,6 +50,34 @@ export async function initializeAssethubChain(logger: Logger) {
...
+    keyManagerAddress,
+  );
+  const txData = keyManagerContract.methods
+    .setAggKeyWithGovKey({
+      pubKeyX: bscKey.pubKeyX,
+      pubKeyYParity: bscKey.pubKeyYParity === 'Odd' ? 1 : 0,
+    })
+    .encodeABI();
+  await signAndSendTxEvm(logger, 'Bsc', { to: keyManagerAddress, value: '0', data: txData });
+}
```
albert-llimos: We should be able to reuse the same logic as for Arbitrum (`initializeArbitrumContracts`). It only changes the client (endpoint) and the address of the keyManager.

bouncer/shared/perform_swap.ts
```diff
@@ -263,6 +267,52 @@ async function waitForCcmExecution<A = []>(
...
+            (event) => event.broadcastId === broadcastId,
+          ),
+        },
+      });
+
+      if (broadcastResult.key === 'ccmBroadcastFailed') {
+        throw new Error(`CCM broadcast failed for ${destAsset} broadcast ${broadcastId}`);
+      }
+      break;
+    }
```
albert-llimos: Can't we make this common logic with Ethereum and Arbitrum? There is no difference afaik

bouncer/shared/send_evm.ts
```diff
@@ -32,7 +33,7 @@ export async function getNextEvmNonce(
     privkey: string;
   },
 ): Promise<number> {
-  if (chain !== 'Ethereum' && chain !== 'Arbitrum') {
+  if (chain !== 'Ethereum' && chain !== 'Arbitrum' && chain !== 'Bsc') {
```
albert-llimos: We can use `isEvmChain` here.

bouncer/shared/send_evm.ts
```diff
@@ -78,7 +79,7 @@ function isNonceError(error: unknown): boolean {
 }
 
 export async function warnIfEvmAddressHasNoCode(logger: Logger, chain: Chain, address: string) {
-  if (chain !== 'Ethereum' && chain !== 'Arbitrum') return;
+  if (chain !== 'Ethereum' && chain !== 'Arbitrum' && chain !== 'Bsc') return;
```
albert-llimos: Same here, `isEvmChain`.

bouncer/shared/setup_elections.ts
```diff
@@ -7,7 +7,7 @@ export async function setupIngressEgressPallet<A>(cf: ChainflipIO<A>) {
     extrinsic: (api) =>
       api.tx.solanaIngressEgress.updatePalletConfig([
         {
-          SetIngressDelaySolana: { delayBlocks: 10 },
+          SetIngressDelaySolana: { delayBlocks: 15 },
```
albert-llimos, line 10: Why is this changed? Rebase issue?

bouncer/shared/swapping.ts
```diff
@@ -196,7 +198,7 @@ export async function newCcmMetadata(
   const destChain = chainFromAsset(destAsset);
 
   let userLogicGasBudget;
-  if (destChain === 'Arbitrum' || destChain === 'Ethereum') {
+  if (destChain === 'Arbitrum' || destChain === 'Ethereum' || destChain === 'Bsc') {
```
albert-llimos: `isEvmChain`

bouncer/tests/gaslimit_ccm.test.ts
```diff
@@ -239,8 +242,8 @@ async function testGasLimitSwapToEvm<A = []>(
   const destChain = chainFromAsset(destAsset);
   const web3 = getWeb3(chainFromAsset(destAsset));
 
-  if (destChain !== 'Arbitrum' && destChain !== 'Ethereum') {
-    throw new Error(`Destination chain ${destChain} is not Ethereum nor Arbitrum`);
+  if (destChain !== 'Arbitrum' && destChain !== 'Ethereum' && destChain !== 'Bsc') {
```
albert-llimos: `isEvmChain`

.github/workflows/ci-development.yml
```diff
@@ -100,7 +100,7 @@ jobs:
     uses: ./.github/workflows/upgrade-test.yml
     secrets: inherit
     with:
-      run-job: false
+      run-job: true
```
albert-llimos: Just leaving a comment to not forget to revert this flag.

bouncer/shared/vault_swap/evm_vault_swap.ts
```diff
@@ -18,17 +18,17 @@ import { ChainflipIO, WithBrokerAccount } from 'shared/utils/chainflip_io';
...
 interface EvmVaultSwapDetails {
-  chain: 'Ethereum' | 'Arbitrum';
+  chain: 'Ethereum' | 'Arbitrum' | 'Bsc';
   calldata: string;
   value: string;
   to: string;
 }
 
 export interface EvmVaultSwapExtraParameters {
-  chain: 'Ethereum' | 'Arbitrum' | 'Tron';
```
albert-llimos, line 31: Maybe we can refactor this into a constant. Actually in the `EVM_VAULT_CHAINS` in `check_transfer_fallbacks`  BSC needs to be added. Maybe we can reuse that?

engine/src/witness/bsc_elections.rs
```diff
@@ -0,0 +1,235 @@
...
+
+	let supported_asset_address_and_event_type =
+		[(BscAsset::BscUsdt, EvmEventSource::new::<UsdtEvents>(usdt_contract_address))]
+			.into_iter()
+			.collect();
+
+	let vault_event_source = EvmEventSource::new::<VaultEvents>(vault_address);
+	let key_manager_event_source = EvmEventSource::new::<KeyManagerEvents>(key_manager_address);
+
+	scope.spawn(async move {
```
albert-llimos: This should be `spawn_with_restart` as in other EVM chains.

localnet/docker-compose.yml
```diff
@@ -311,7 +290,7 @@ services:
       - CFDM_NETWORK=Localnet
       - CFDM_PERSISTENT_STATE_DIR=/persistent_state
       - CFDM_CRITICAL_ELLIPTIC_RISK_SCORE=8.0
-      - CFDM_BROKER_API_URL=${CFDM_BROKER_API_URL}
+      - CFDM_BROKER_API_URL=${CFDM_CHAINFLIP_RPC_URL}
```
albert-llimos, line 297: is this on purpose?

state-chain/node/src/chain_spec.rs
```diff
@@ -1154,6 +1217,11 @@ fn testnet_genesis(
 		tron_broadcaster: state_chain_runtime::TronBroadcasterConfig {
 			broadcast_timeout: 2 * BLOCKS_PER_MINUTE_TRON,
 		},
+		// instance8
+		bsc_broadcaster: state_chain_runtime::BscBroadcasterConfig {
+			// todo: revisit this
```
albert-llimos: I think this is fine and we can remove this todo

state-chain/chains/src/bsc.rs
```diff
@@ -0,0 +1,225 @@
...
+	type ChainBlockNumber = u64;
+	const WITNESS_PERIOD: Self::ChainBlockNumber = 24;
+}
+
+impl Chain for Bsc {
+	const NAME: &'static str = "Bsc";
+	const GAS_ASSET: Self::ChainAsset = assets::bsc::Asset::Bnb;
+	const WITNESS_PERIOD: Self::ChainBlockNumber = 24;
+	const FINE_AMOUNT_PER_UNIT: Self::ChainAmount = eth::ONE_ETH;
+	const BURN_ADDRESS: Self::ChainAccount = H160([0; 20]);
```
albert-llimos: We can use `evm::Address::zero()`

state-chain/chains/src/bsc.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl ChainWitnessConfig for Bsc {
+	type ChainBlockNumber = u64;
+	const WITNESS_PERIOD: Self::ChainBlockNumber = 24;
+}
+
+impl Chain for Bsc {
+	const NAME: &'static str = "Bsc";
+	const GAS_ASSET: Self::ChainAsset = assets::bsc::Asset::Bnb;
+	const WITNESS_PERIOD: Self::ChainBlockNumber = 24;
```
albert-llimos: In Arbitrum for 0.25 sec blocks we use 24 to get 6 sec intervals. In BSC as I understand it the block time is ~0.45sec. Therefore we should probably decrease this a bit. Maybe 14?

state-chain/chains/src/bsc.rs
```diff
@@ -0,0 +1,225 @@
...
+impl BscTrackedData {
+	pub fn max_fee_per_gas(
+		&self,
+		priority_fee_multiplier: FixedU64,
+	) -> <Bsc as Chain>::ChainAmount {
+		// the priority fee has to be non zero and ideally above a certain threshold otherwise the
+		// tx will likely be rejected by the validator
+		priority_fee_multiplier
+			.saturating_mul_int(self.priority_fee)
+			.max(20_000_000_000)
```
albert-llimos: This should be a declared constant. I think we should call it `MIN_GAS_FEE`

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
```diff
@@ -0,0 +1,517 @@
...
+
+define_empty_struct! { pub struct BscChain; }
+
+impl ChainTypes for BscChain {
+	type ChainBlockNumber = BlockWitnessRange<Bsc>;
+	type ChainBlockHash = evm::H256;
+	const NAME: &'static str = "Bsc";
+}
+
+pub const BSC_MAINNET_SAFETY_BUFFER: u32 = 8;
```
albert-llimos, line 93: We probably should discuss this, same with the safety margin.

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
```diff
@@ -0,0 +1,517 @@
...
+			},
+			Default::default(),
+			(),
+		),
+		settings: (
+			Default::default(),
+			Default::default(),
+			Default::default(),
+			Default::default(),
+			(FEE_HISTORY_WINDOW, PRIORITY_FEE_PERCENTILE),
```
albert-llimos, line 450: Setting these values here seem to be different than other chains where they are default. Is this so they are already initialized?

state-chain/runtime/src/configs.rs
```diff
@@ -533,6 +549,37 @@ impl pallet_cf_ingress_egress::Config<Instance7> for Runtime {
 	type LpRegistrationApi = LiquidityProvider;
 }
 
+impl pallet_cf_ingress_egress::Config<Instance8> for Runtime {
+	type RuntimeCall = RuntimeCall;
+	const MANAGE_CHANNEL_LIFETIME: bool = true;
+	const ONLY_PREALLOCATE_FROM_POOL: bool = true;
```
albert-llimos, line 555: 👍

albert-llimos (commented):
Generally looks good! Just some comments on small details.

bouncer/shared/perform_swap.ts
nmammeri: This was addressed by #6662, the logic is shared only events lookup tables are different.

bouncer/shared/setup_elections.ts
nmammeri: This was increased deliberately because of increased load the Solana broker level screening test was starting  to sometimes fail because the deposit monitor not flagging transactions on time.

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
nmammeri: This is needed because the  fee-tracking  settings are non-zero. It is similar to how ethereum_elections.rs declares it.

bouncer/tests/all_swaps.ts
```diff
@@ -78,12 +78,11 @@ export function testAllSwaps(timeoutPerSwap: number) {
...
 
-  // TODO: properly include TRON and BSC assets once they are fully integrated
-  // if we include Assethub swaps (HubDot, HubUsdc, HubUsdt) in the all-to-all swaps,
+  // If we include Assethub swaps (HubDot, HubUsdc, HubUsdt) in the all-to-all swaps,
   // the test starts to randomly fail because the assethub node is overloaded.
   const AssetsForTesting = Object.values(Assets).filter((id) => {
     const chain: string = chainFromAsset(id);
-    return chain !== 'Assethub' && chain !== 'Bsc';
+    return chain !== 'Assethub';
   });
```
albert-llimos: nit: we can inline all this in a nice way now:

```ts
const AssetsForTesting = Object.values(Assets).filter((id) => chainFromAsset(id) !== 'Assethub');
```

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
nmammeri: BSC claims to have 0.65s finality (around 2 BSC blocks) https://www.bnbchain.org/en/blog/how-bsc-achieved-sub-second-finality-a-70x-leap-from-45s-to-0-65s
I think we can set the safety margin to 4 Bsc blocks and the buffer to 6-8 blocks?

albert-llimos: Sounds good to me. Any input from you side @MxmUrw ?

bouncer/shared/vault_swap/evm_vault_swap.ts
nmammeri: I think this is the only place where 'Ethereum' | 'Arbitrum' | 'Tron' | 'Bsc' are defined, so no need to move them to a constant. This is a EvmVaultSwapExtraParameters type definition that captures what the SC rpc expects. In my opinion it should be independent of any other constants.  

The EVM_VAULT_CHAINS constant can be removed, it is just used to enumerate the fallback event names and can be replaced with the new lookup tables style instead.

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
nmammeri: I was assuming that these are in bsc blocks unit but actually the unit is in bactches of 14 bsc blocks = 1 state chain block. 
After discussing with @MxmUrw  we'll keep the BSC_MAINNET_SAFETY_BUFFER set 8 to be consistant with other chains as this only affects storage and set the safety margin to 2 for now we can reduce it later if we want to.

MxmUrw: As an addition for the reasons:
 - I think it's safer to start with a higher safety margin, it *is* easily modifiable through governance
 - The DM needs time for rejections, so there is a natural lower bound on how fast we can go

.github/workflows/upgrade-test.yml
```diff
@@ -215,11 +218,27 @@ jobs:
         run: |
           git checkout ${{ env.OLD_BIN_COMMIT_HASH }}
           git rev-parse HEAD
+          # Remove after 2.3 release.
+          # The old bouncer's rpc_tests filter out Tron (unreleased at the time); Bsc is the
+          # unreleased chain now, so retarget the filter to exclude Bsc assets instead.
+          sed -i "s/a.chain !== 'Tron'/a.chain !== 'Bsc'/" bouncer/tests/rpc_tests.ts
```
dandanlen, line 224: We don't want Tron in here any more do we?

state-chain/primitives/src/chains/assets.rs
```diff
@@ -936,6 +936,34 @@ assets!(
...
+			},
+			Asset {
+				variant: BscUsdt,
+				member: usdt,
+				string: "USDT" (aliases: ["Usdt", "usdt"]),
+				json: "USDT",
+				gas: false,
+				index: 20,
+				usd_stablecoin: true,
+				decimals: 18,
```
dandanlen, line 963: Not 6 decimals?

state-chain/custom-rpc/src/broker.rs
```diff
@@ -523,9 +528,24 @@ where
...
+					let api_version = api
+						.api_version::<dyn CustomRuntimeApi<state_chain_runtime::Block>>(hash)
+						.map_err(CfApiError::from)?
+						.unwrap_or_default();
+					Ok(if api_version < 17 {
+						#[expect(deprecated)]
+						api.cf_transaction_screening_events_before_version_17(hash)
+							.map_err(CfApiError::from)?
+							.into()
+					} else if api_version < 19 {
```
dandanlen, line 541: How come we introduced two new version here?

state-chain/custom-rpc/src/broker.rs
```diff
@@ -690,12 +725,21 @@ where
...
+		let at = Some(self.rpc_backend.client.info().best_hash);
+		self.rpc_backend.with_versioned_runtime_api(at, |api, hash, version| {
+			if version < 16 {
+				#[expect(deprecated)]
+				api.cf_vault_addresses_before_version_16(hash).map(Into::into)
+			} else if version < 17 {
+				#[expect(deprecated)]
+				api.cf_vault_addresses_before_version_17(hash).map(Into::into)
+			} else if version < 19 {
+				#[expect(deprecated)]
```
dandanlen, line 737: Similar question: did we not need this before?

state-chain/primitives/src/chains/assets.rs
albert-llimos: No, it is 18 decimals:
https://bscscan.com/address/0x55d398326f99059ff775485246999027b3197955

state-chain/custom-rpc/src/lib.rs
```diff
@@ -2767,9 +2844,24 @@ where
...
 				move |client, hash| {
-					Ok((*client.runtime_api())
-						.cf_transaction_screening_events(hash)
-						.map_err(CfApiError::from)?)
+					let api = client.runtime_api();
+					let api_version = api
+						.api_version::<dyn CustomRuntimeApi<state_chain_runtime::Block>>(hash)
+						.map_err(CfApiError::from)?
+						.unwrap_or_default();
+					Ok(if api_version < 18 {
```
dandanlen: Should this be 17?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -3480,6 +3614,16 @@ where
 			}]);
 		}
 
+		if chain.is_none_or(|chain| chain == ForeignChain::Bsc) {
+			result.insert(ForeignChain::Bsc, vec![AddressAndExplanation {
+				name: "bsc_vault_contract".into(),
+				address: AddressString::from_encoded_address(bsc),
+				explanation: none_if_compact("Holds BNB and all tokens on Tron. Directly receives user funds in case of smart contract-based vault swaps.".into()),
```
dandanlen:
```suggestion
				explanation: none_if_compact("Holds BNB and all tokens on BSC. Directly receives user funds in case of smart contract-based vault swaps.".into()),
```

.github/workflows/upgrade-test.yml
nmammeri: no we don't, since Tron is released we don't need to filter it out. All rpc calls that take Asset know about Tron but don't know about Bsc yet. All of this workaround is not needed once we get the Asset auto migrations in.

state-chain/custom-rpc/src/tests.rs
```diff
@@ -169,6 +170,7 @@ fn test_environment_serialization() {
 				btc: btc::AssetMap { btc: Some(0u32.into()) },
 				dot: dot::AssetMap { dot: None },
 				arb: arb::AssetMap { eth: None, usdc: Some(0u32.into()), usdt: Some(0u32.into()) },
+				bsc: bsc::AssetMap { bnb: None, usdt: None },
```
dandanlen: nit: it would be nice to stay consistent with the ordering of the bsc members of these structs. Sometimes we add in the middle, sometimes at the end (I would prefer the end)

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -809,6 +845,26 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::call_index(13)]
+		// This weight is not strictly correct but since it's a governance call, weight is
+		// irrelevant.
+		#[pallet::weight(T::WeightInfo::witness_initialize_bsc_vault())]
+		pub fn witness_initialize_bsc_vault(
```
dandanlen, line 853: Can we remove this extrinsic since the chain now intializes automatically?

state-chain/pallets/cf-environment/src/migrations.rs
```diff
@@ -18,12 +18,14 @@ use crate::{Config, Pallet, STORAGE_VERSION_U16};
...
 use crate::{CurrentReleaseVersion, Get};
 use cf_runtime_utilities::PlaceholderMigration;
-use frame_support::traits::OnRuntimeUpgrade;
+use frame_support::{migrations::VersionedMigration, traits::OnRuntimeUpgrade};
 #[cfg(feature = "try-runtime")]
 use frame_support::{pallet_prelude::DispatchError, sp_runtime};
 #[cfg(feature = "try-runtime")]
 use sp_std::vec::Vec;
 
+mod bsc_assets;
```
dandanlen: nit: this is more than just assets, more like `bsc_chain_environment`.

state-chain/primitives/src/chains/assets.rs
```diff
@@ -936,6 +936,34 @@ assets!(
 			},
 		],
 	},
+	Chain {
+		variant: Bsc,
```
dandanlen, line 940: In the interest of consistency: for other chains we use the full names - can we make this `BnbSmartChain` ?

state-chain/chains/src/bsc.rs
```diff
@@ -0,0 +1,228 @@
...
+	#[track_caller]
+	fn default() -> Self {
+		frame_support::print("You should not use the default chain tracking, as it's meaningless.");
+
+		BscTrackedData { priority_fee: Default::default() }
+	}
+}
+
+/// Minimum gas price (wei) BSC validators will accept; below this, txs are likely dropped.
+const MIN_FEE_PER_GAS: <Bsc as Chain>::ChainAmount = 20_000_000_000;
```
dandanlen: Where did this info come from, is it documented somewhere? Official/inofficial?

dandanlen: Also, since this is used when we calculate how much to refund, doesn't it mean we could accidentally drain the vaults if the fees are in fact below this? 

Also: we don't seem to apply this one the outgoing (in `calculate_transaction_fee`). Should we?

cc @albert-llimos

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
```diff
@@ -0,0 +1,517 @@
...
+			Vec<
+				ElectionIdentifier<
+					<BscFeeTracking as ElectoralSystemTypes>::ElectionIdentifierExtra,
+				>,
+			>,
+		),
+	) -> Result<(), CorruptStorageError> {
+		let current_sc_block_number = crate::System::block_number();
+
+		let chain_progress = BscBlockHeightWitnesserES::on_finalize::<
```
dandanlen, line 356: This was probably adapted from an old version of the arb_elections. As of [this PR](https://github.com/chainflip-io/chainflip-backend/pull/6576) we have a code block like this to allow dynamically adjusting the safety margin:

```

		if let Some(ingress_safety_margin) =
			pallet_cf_ingress_egress::WitnessSafetyMargin::<Runtime, ArbitrumInstance>::get()
		{
			pallet_cf_elections::ElectoralUnsynchronisedSettings::<Runtime, ArbitrumInstance>::mutate_extant(|settings|  {
				// We apply the safety margin to *all* BlockWitnessers. On EVM chains we have the same safety margin for
				// ingresses and egresses since egresses are entangled with other other stuff in the KeyManager BW.
				settings.1.safety_margin = ingress_safety_margin as u32;
				settings.2.safety_margin = ingress_safety_margin as u32;
				settings.3.safety_margin = ingress_safety_margin as u32;
			});
		}

```

state-chain/runtime/src/migrations/bsc_integration.rs
```diff
@@ -0,0 +1,135 @@
...
+		Ok(().encode())
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		let result = pallet_cf_elections::Pallet::<Runtime, BscInstance>::internally_initialize(
+			crate::chainflip::witnessing::bsc_elections::initial_state(),
+		);
+		if result.is_err() {
+			log::error!("Failed to initialize Bsc election pallet");
+		}
```
dandanlen, line 41:
```suggestion
		if result.is_err() {
			log::info!("BSC Elections already initialised.");
		}
```

`internally_initialize` only errors if the elections are already initialised.

dandanlen (commented):
Looking good 👌

state-chain/chains/src/bsc.rs
albert-llimos: That's actually true, we should apply it consistently.
When it comes to the exact value, I also wanted to ask that to Ramiz as he is the one that figured that one out. @ramizhasan111

state-chain/primitives/src/chains/assets.rs
albert-llimos: Everyone refers to it as BSC. Also, it would be BinanceSmartChain instead. I would leave it as Bsc but I'm not against renaming it with it's full name for consistency.

state-chain/custom-rpc/src/broker.rs
nmammeri: There was a minor release that introduced version 18 that added `cf_supported_assets` commit: 4dcd6beb2462af315f0cfed4a801c23c6c3dc005. This didn't change anything related to cf_transaction_screening_events, keeping version 17 and 18 types the same

state-chain/primitives/src/chains/assets.rs
nmammeri: I recall we discussed this sometime ago and we agreed to name it just BSC. I think product also names it BSC only.

.github/workflows/upgrade-test.yml
dandanlen: Ok i misread the sed command, now i get it. It seems quite brittle: I think I remove this hack at the 2.2.6 tag, so it will break at some point... can we not somehow use the PRE_UPGRADE_BOUNCER environment variable? 

Also why do we need to filter out BSC in the *new* rpc_tests? It seems like we want to filter it out in the old or the new, but not both?

.github/workflows/upgrade-test.yml
```diff
@@ -174,13 +174,16 @@ jobs:
           sudo cp ./upgrade-to-bins/libchainflip_engine_v*.so /usr/lib/
           echo "/usr/lib after copy of .so files"
           ls -l /usr/lib
+          
+          # Remove after 2.3 release
+          # This makes the deposit monitor uses the broker rpc 2.0 not the broker binary
+          sed -i 's/CFDM_BROKER_API_URL=${CFDM_BROKER_API_URL}/CFDM_BROKER_API_URL=${CFDM_CHAINFLIP_RPC_URL}/' localnet/docker-compose.yml
+          
+          # Remove after 2.3 release
```
dandanlen, line 182: How did this sneak in? What is it referring to? The next couple of lines aren't new.

state-chain/primitives/src/chains/assets.rs
dandanlen: The official website has it as "BNB Smart Chain"... I guess they've rebranded.

But ok, I don't mind that much, we can leave it as BSC. Happy to save us some tokens 😅

state-chain/chains/src/bsc.rs
dandanlen: According to [this](https://docs.bnbchain.org/bnb-opbnb/core-concepts/gas-and-fees/), the default minimum is now 0.05 gwei. 20 gwei was from early mainnet days and it has dropped significantly since then.

(And it's a convention rather than a consensus rule so we should consider making this configurable (although I doubt it changes very frequently).

state-chain/custom-rpc/src/broker.rs
nmammeri: version 18 an 17 had same types no types migrations needed, then the types changed in version 19.

.github/workflows/upgrade-test.yml
nmammeri: This is needed to get the upgrade test working without needing to restart the broker-api binary and then  the deposit monitor.

.github/workflows/upgrade-test.yml
nmammeri: In the rpc tests we get a lit of all supported assets by querying `cf_swapping_environment` then whenever we make a test call to an rpc that takes a single Asset as param, we pick a random one from that list. During the runtime upgrade boundary the node binary has the new Assets but we are still on the old runtime that doesn't know about new BSC assets yet. Any rpc that uses new assets will fail such as `cf_pool_orders` etc... This is just a workaround, the proper solution would be to also migrate all rpcs that takes `Asset` type as param.

state-chain/runtime/src/chainflip/witnessing/bsc_elections.rs
nmammeri: good catch

state-chain/chains/src/bsc.rs
nmammeri: Adjusted it to be 0.05 gwei. No need to make it configurable because it won't change frequently.

.github/workflows/upgrade-test.yml
dandanlen: We have a cf-supported-assets rpc, why don't we use that? 

We also set a `PRE_UPGRADE_BOUNCER` env var for the pre-upgrade bouncer which we could use. 

Both of the above options seem more robust than editing files using sed (for example it will break when update the old-version to > 2.2.6).

.github/workflows/upgrade-test.yml
dandanlen: Sorry I meant the lines below this, the snippet isn't very clear. 

There is `# Remove after 2.3 release` but the lines below it are not part of this PR.

dandanlen:
Apart from the remaining comments, LGTM, good work!

.github/workflows/upgrade-test.yml
nmammeri: cf-supported-assets does't work because it doesn't properly handle Asset enum migrations.

.github/workflows/upgrade-test.yml
nmammeri: You are right, I just added this comment as a reminder to remove it. We don't want to forget it similar to the line:
docker compose -f localnet/docker-compose.yml -p "chainflip-localnet" stop indexer 
which we forgot to remove after release.

.github/workflows/upgrade-test.yml
dandanlen: Ok - as discussed i'll raise an issue to improve this for the next release.

state-chain/pallets/cf-environment/src/lib.rs
dandanlen: Looks like it's needed for the integration tests - let's leave it here.

dandanlen (approved):

albert-llimos (approved):
Everything addressed, I went over the comments and resolved them. LGTM, good work @nmammeri (and @ramizhasan111 before that)!

PR #6476, opened 2026-03-11
feat: network and broker fee taken from input and output

j4m1ef0rd:
# Pull Request

Closes: PRO-2727

Note: This PR is merged into the feature branch [feat/support-for-non-usdc-swaps](https://github.com/chainflip-io/chainflip-backend/tree/feat/support-for-non-usdc-swaps)

## Summary

Refactors the swapping code to take the network fee from the input asset and the broker fee from the output asset. This is not full support for non-usdc pools. I tried to keep the swapping flow the same and only touch the fees. Any refactoring of swapping will come later.
Brokers and affiliates will have to wait for a fee swap to usdc before getting there fees (if the output is not usdc). 

- Changed the network fee collection to handle all assets and swap them to flip as usual.
   -  Added a migration for the network fee storage.
- Added a conversion when getting the network fee minimum on swap request, that converts the minimum to the input asset.
- Removed refund fee and related tests.
- Updated the cf_swap_rate_v3 rpc to support the new fee system and removed the complex fee logic, yay.
   - swap simulation now returns `AssetAndAmount`'s (non-breaking thanks to RpcFee type).
- Updated a lot of tests. Cleaned up a bit as well. 
- Updated the integration tests
- Updated the bouncer tests
- All broker fees that are not USDC are auto-swaped to USDC in a new swap type `BrokerFee`. Otherwise the USDC is just credited to the free balance like before.
- Refactored how the network and broker fees are stored in the swap request
- Added a new broker fee tracker that is used to record the amounts for each broker until the swap request is complete and then start the broker fee swaps.

### Non-breaking changes
- `cf_subscribe_scheduled_swaps` will now show the input amount after fees
- `cf_swap_rate_v3` will now return non-usdc assets in the existing assets field for network and broker fees. 
- `SwapRequestComplete` event now has a list of broker fee swaps so we can track the fees to a USDC amount.
- Added a new event `NetworkFeeSwapsInitiated` to allow for linking the network fee swaps to the exact usdc amounts.
- The network fee minimum is now subject to oracle/pool price because the minimum usdc amount is converted into the input asset. So network fee in `cf_swap_rate_v3` may change each block if the minimum was enforced. 

### Breaking changes @zoheb391 
- `SwapExecuted` event now uses `AssetAndAmount` type for all amounts (input, output, intermediate & fees)
   - Input amount is always **after** fees have been taken
- `RefundEgressScheduled` and `RefundedOnChain` events no longer has a `refund_fee` field

---

* The PR is complete:
  * Tests:


state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -246,40 +252,42 @@ pub struct FeeRateAndMinimum {
 
 #[derive(Debug, PartialEq, Eq, Encode, Decode, DecodeWithMemTracking, TypeInfo)]
 pub struct NetworkFeeTracker {
+	// Fee rate and minimum in input asset terms.
```
msgmaxim: nit: we should prefer doc comments

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1901,111 +1988,127 @@ pub mod pallet {
...
+				SwapRequests::<T>::mutate(swap.swap_request_id(), |swap_request| {
+					let fee_tracker = if let Some(swap_request) = swap_request {
+						match &mut swap_request.state {
+							SwapRequestState::UserSwap {
+								network_fee_tracker: network_fee, ..
+							} => Some(network_fee),
+							SwapRequestState::IngressEgressFee |
+							SwapRequestState::BrokerFee { .. } =>
+							// Disposable network fee tracker with no minimum
+								Some(&mut NetworkFeeTracker::new_without_minimum(
```
msgmaxim: Seems a bit wasteful that you call `get_network_fee_for_swap` which calculates among other things the minimum (which involves simulating a swap) only to wrap it into something that ignores the minimum.

Maybe you could extract part of the code in `get_network_fee_for_swap` that calculate the rate only and use it here (and have `get_network_fee_for_swap` take the rate only).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -132,20 +134,28 @@ pub struct SwapRefundParameters {
 
 #[derive(CloneNoBound, DebugNoBound)]
 pub struct SwapState<T: Config> {
-	pub swap: Swap<T>,
+	swap: Swap<T>,
+	// Always in terms of input asset
 	pub network_fee_taken: Option<AssetAmount>,
```
msgmaxim: What is the purpose of having this be an `Option` as opposed to a number that defaults to 0? At first I assumed we are using None as a sanity check that we have taken the fee in the end, but I looked and everywhere it is used we simply `unwrap_or_default`().

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1901,111 +1988,127 @@ pub mod pallet {
...
+							.and_modify(|total| total.saturating_accrue(fee))
+							.or_insert(fee);
+					}
+				});
 
-						T::BalanceApi::credit_account(account, STABLE_ASSET, fee);
+				// Now that the fee is taken, we can update the swap state output and intermediate
+				// if they are the input
+				if swap.output_asset() == swap.input_asset() {
+					swap.output_amount = Some(swap.input_amount());
```
msgmaxim: What happens to these swaps? Do they still get executed against pools in `do_group_and_swap`? If yes, how can we guarantee the output already? IMO would be nice to short circuit these somehow.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1901,111 +1988,127 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		fn start_broker_fee_swaps_or_credit(
+			fee_tracker: &BrokerFeesTracker<T::AccountId>,
+			fee_asset: Asset,
+		) -> BTreeMap<T::AccountId, SwapRequestId> {
+			if fee_asset == Asset::Usdc {
+				// No need to swap if the fee asset is already in usdc, just credit directly.
+				fee_tracker.fee_and_accumulated.iter().for_each(
```
msgmaxim: Maybe I'm wrong, but I thought we preferred `for (Beneficiary { account, .. }, amount) in fee_tracker.fee_and_accumulated { }` for simple loop like this.

state-chain/pallets/cf-swapping/src/tests/fees.rs
```diff
@@ -14,146 +14,183 @@
...
+				output_asset: Asset::Flip,
+				request_type: SwapRequestTypeEncoded::NetworkFee,
+				origin: SwapOrigin::Internal,
+				..
+			})
 		);
 	});
 }
 
+// This test covers:
```
msgmaxim: Again, we should prefer doc comments.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -226,35 +231,27 @@ impl<T: Config> SwapState<T> {
 	}
 
 	pub fn take_network_fee(&mut self, fee_tracker: &mut NetworkFeeTracker) -> AssetAmount {
-		let FeeTaken { fee, .. } = fee_tracker.take_fee(self.swap.input_amount);
-		self.swap.take_fee_from_input(fee);
-		self.network_fee_taken = Some(self.network_fee_taken.unwrap_or_default() + fee);
+		let FeeTaken { fee, .. } = fee_tracker.take_fee(self.input_amount_after_fees);
```
msgmaxim: I know that `input_amount_after_fees == input_amount` at this point in time, but would it be more corect to use `input_amount` here?

dandanlen (commented):
Looking good, just some minor comments.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -240,46 +266,97 @@ pub enum FeeType<T: Config> {
...
+				|(Beneficiary { bps, .. }, accumulated_fee)| {
+					let fee =
+						Permill::from_parts(*bps as u32 * BASIS_POINTS_PER_MILLION) * input_amount;
+					accumulated_fee.saturating_accrue(fee);
+					total_fee.saturating_accrue(fee)
+				},
+			);
+		}
+
+		assert!(total_fee <= input_amount, "Broker fee cannot be more than the amount");
```
MxmUrw, line 358: Could we replace this by something like `log_or_panic!()`? Feels like there's no point in introducing a possible panic even though it should never happen.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2324,54 +2438,39 @@ pub mod pallet {
 					let total_input_remaining = swap.input_amount +
 						dca_state.remaining_input_amount +
 						canceled_swaps_amount;
```
MxmUrw, line 2440: I might be wrong, but it looks like we're now refunding without taking any min network fee into account. The linear issue states:

> With input-leg fee collection, the network fee minimum is always taken upfront even if zero DCA chunks execute. 

I don't think that's true currently? I might be missing something though...

j4m1ef0rd: good catch. I'll fix in my other PR to avoid conflicts.

j4m1ef0rd: It was working until I updated the code to not mutate the swap, only the swap state, that's how it slipped through. I will add a test to my other PR.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2614,14 +2717,20 @@ pub mod pallet {
...
+
+					(true, Default::default())
 				},
 			};
 
 			if request_completed {
 				Self::deposit_event(Event::<T>::SwapRequestCompleted {
 					swap_request_id,
 					reason: SwapRequestCompletionReason::Executed,
+					broker_fee_swaps,
```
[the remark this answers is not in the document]
j4m1ef0rd: Im not sure what you mean by this. Are you saying we make a new event `BrokerFeeSwapInitiated` that has a `originating_swap_request_id` filed and a `broker_fee_swap_request_id` field to link the 2 swaps together?
Seems more straightforward to leave it as is.

PR #6498, opened 2026-03-23
feat: Lending pools can be used in boosting

msgmaxim:
# Pull Request

Closes: PRO-2721, PRO-2713, PRO-2820

## Summary

- Lending pools can be used in boosting deposits:
   - Lending pool gets priority; boost pool covers the shortfall.
   - Fee is split proportionally between the two sources.
- Removed boost pool tiers:
   - The existing 5bps pool is left unchanged
   - Pools other than 5bps will be removed; any LP with funds in those pools will have the funds returned into free balance (`StoppedBoosted` event will be emitted as if they requested to `stop_boosting`)
   - 5bps boost fee is now hardcoded (might consider making this a governance controlled parameter in the future) 
- Network fee behavior:
   - Network fee collected on the amount boosted by the lending pool is taken as origination fee: exposed via `OriginationFeeTaken` event and credited to the network immeidately on boost (given there is available liquidity in the pool).
  - Network fee collected on the amount boosted by the boost (legacy) pools, will be credited on deposit finalisation and exposed via `DepositFinalised` exactly as before.

## API Changes

### RPCS

- `cf_loan_accounts`: added new field `loan_type` to every loan
- `cf_all_loans`: new RPC to get loans of both user and boost types (the existing `cf_loan_accounts` exposes user loans only)

### Events

- `DepositBoosted`: `amounts` now contain amounts per "boost source" (lending/boost) rather than "boost tier".
- `LoanCreated`: replaced `borrower_id` with `loan_type` (with `LoanType::User` being the regular loans, and `LoanType::Boost` for boost loans).

### Extrinsics

- `add_boost_funds`, `stop_boosting` and `create_boost_pools` are unchanged except that providing anything other than 5bps for `pool_tier` will now fail.

### Other Notes
- As discussed with @dandanlen, we will assume that there won't be boosted deposits at the time of runtime upgrade (boost will be disabled prior); this lets us simplify the migration.

FYI @GabrielBuragev

---

## *Checklist

* The PR is complete:
  * Tests:


j4m1ef0rd (approved):

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1727,29 +1727,58 @@ where
...
+				api.cf_loan_accounts_before_version_17(hash, borrower_id.clone())
+					.map(|accounts| {
+						accounts
 							.into_iter()
-							.map(|loan| RpcLoan {
-								loan_id: loan.loan_id,
-								asset: loan.asset,
-								created_at: loan.created_at,
-								principal_amount: loan.principal_amount.into(),
+							.map(|acc| RpcLoanAccount::<_, U256> {
```
dandanlen: nit: it would be nice if we can keep the conversion/defaulting of these in the respective modules (ie. in `before_version_17`) in this case.

And similarly below we could just have a mapper function (derived or not) to make this more readable since all it's doing is passing through most of the values.

state-chain/runtime/src/runtime_apis/custom_api.rs
```diff
@@ -303,9 +303,16 @@ decl_runtime_apis!(
...
+		#[changed_in(17)]
+		fn cf_loan_accounts(
+			borrower_id: Option<AccountId32>,
+		) -> Vec<before_version_17::RpcLoanAccount<AccountId32, AssetAmount>>;
 		fn cf_loan_accounts(
 			borrower_id: Option<AccountId32>,
 		) -> Vec<RpcLoanAccount<AccountId32, AssetAmount>>;
+		#[changed_in(17)]
+		fn cf_all_loans();
+		fn cf_all_loans() -> Vec<RpcLoan<AccountId32, AssetAmount>>;
```
dandanlen, line 329: Is this exposed via the rpc layer? I can't see it.

state-chain/pallets/cf-lending-pools/src/boost.rs
```diff
@@ -318,36 +405,127 @@ fn fee_from_boosted_amount(amount_to_boost: AssetAmount, fee_bps: u16) -> AssetA
...
+	// `<= lending_available` (since `required_amount <= lending + boost`), so the
+	// final `.min(lending_available)` is what enforces the cap when the floor exceeds it.
+	let min_lending = min_lending_share * required_amount;
+
+	let lending_pool_principal = proportional_lending.max(min_lending).min(lending_available);
+	let boost_pool_principal =
+		required_amount.saturating_sub(lending_pool_principal).min(boost_available);
+	// Handle any rounding shortfall (or lending slack freed up by the boost-pool clamp)
+	// by pushing the remainder back to the lending pool. Guaranteed to be within
+	// `lending_available` because total liquidity is at least `required_amount`.
```
dandanlen, line 450: nit: this is a non-local guarantee. Maybe we can fold the check (total > required) into the function itself?

state-chain/pallets/cf-lending-pools/src/boost.rs
```diff
@@ -138,129 +164,196 @@ impl<T: Config> BoostApi for Pallet<T> {
...
+
+		let (lending_pool_principal, boost_pool_principal) = split_required_amount(
+			required_amount,
+			lending_available,
+			boost_available,
+			BoostConfig::<T>::get().min_lending_pool_share,
+		);
+
+		let lending_pool_fee =
+			Permill::from_rational(lending_pool_principal, required_amount) * total_fee;
```
dandanlen, line 189: I though boost loans from the lending pool would be treated identically to regular loans (ie. origination fee + interest) but it looks like they charge the boost fee and zero interest. Is that right? (I think this makes sense but maybe worth clarifying)

state-chain/pallets/cf-lending-pools/src/boost.rs
```diff
@@ -318,36 +405,127 @@ fn fee_from_boosted_amount(amount_to_boost: AssetAmount, fee_bps: u16) -> AssetA
...
+		assert_eq!(split(1_000, 1_000, 0), (1_000, 0));
+		assert_eq!(split(1_000, 0, 1_000), (0, 1_000));
+	}
+
+	#[test]
+	fn min_lending_share_is_configurable() {
+		// With a 50% minimum, lending must cover 500 even though proportional is ~230.
+		assert_eq!(split_required_amount(1000, 600, 2000, Percent::from_percent(50)), (500, 500));
+		// With a 0% minimum, only proportional applies.
+		assert_eq!(split_required_amount(1000, 200, 2000, Percent::from_percent(0)), (91, 909));
```
dandanlen: It's worth adding the 100% case here too.

state-chain/pallets/cf-lending-pools/src/boost.rs
```diff
@@ -318,36 +405,127 @@ fn fee_from_boosted_amount(amount_to_boost: AssetAmount, fee_bps: u16) -> AssetA
...
+	// Handle any rounding shortfall (or lending slack freed up by the boost-pool clamp)
+	// by pushing the remainder back to the lending pool. Guaranteed to be within
+	// `lending_available` because total liquidity is at least `required_amount`.
+	let lending_pool_principal = required_amount.saturating_sub(boost_pool_principal);
+
+	(lending_pool_principal, boost_pool_principal)
+}
+
+#[cfg(test)]
+mod split_tests {
```
dandanlen, line 457: nit: this would have been a good candidate for property testing.

state-chain/pallets/cf-lending-pools/src/boost.rs
msgmaxim: Correct, no interest on boost loans, only origination fee.

bouncer/tests/boost.ts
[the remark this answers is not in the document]
msgmaxim: Changing it to an assert. I don't think boost config is setup before this point though, so I think I will keep it.

state-chain/pallets/cf-lending-pools/src/boost.rs
msgmaxim: Don't think it is necessary: the logic is pretty simple and clear; just a couple of test cases should to be be sufficient.

bouncer/tests/boost.ts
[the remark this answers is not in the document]
j4m1ef0rd: `setupBoostPools` is called in the setup_concurrent.t script, just like the other pools. isn't that all we need?

state-chain/runtime/src/runtime_apis/custom_api.rs
msgmaxim: Indeed, it is not. Added now.

bouncer/tests/boost.ts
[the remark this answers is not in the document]
msgmaxim: It doesn't update boost config via updatePalletConfig is what I mean.

state-chain/pallets/cf-lending-pools/src/tests.rs
```diff
@@ -459,194 +451,307 @@ fn test_add_boost_funds() {
...
-		// Should be able to boost without the 10bps pool:
-		assert_boosted(Asset::Eth, DEPOSIT_ID, [TIER_5_BPS, TIER_30_BPS]);
+		assert_noop!(
+			LendingPools::add_boost_funds(
+				RuntimeOrigin::signed(BOOSTER_1),
+				Asset::Btc,
+				MINIMUM_ADD_FUNDS_AMOUNT - 1,
+				BOOST_FEE_BPS
+			),
+			crate::Error::<Test>::AmountBelowMinimum
```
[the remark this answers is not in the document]
msgmaxim: done

state-chain/pallets/cf-lending-pools/src/tests.rs
```diff
@@ -16,33 +16,53 @@
...
-const TIER_10_BPS: BoostPoolTier = 10;
-const TIER_30_BPS: BoostPoolTier = 30;
+const BOOST_FEE_BPS: BoostPoolTier = 5;
 
 const INIT_BOOSTER_ETH_BALANCE: AssetAmount = 1_000_000_000;
 const INIT_BOOSTER_FLIP_BALANCE: AssetAmount = 1_000_000_000;
 
-fn setup() {
+fn setup_lending_pool_for_boost() {
+	assert_ok!(LendingPools::create_lending_pool(RuntimeOrigin::root(), Asset::Eth));
```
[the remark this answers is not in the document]
msgmaxim: done

state-chain/pallets/cf-lending-pools/src/tests.rs
```diff
@@ -459,194 +451,307 @@ fn test_add_boost_funds() {
...
-			assert!(pool.pending_loans.is_empty());
-		}
+		assert_eq!(BoostLoans::<Test>::get(LOAN_ID), None);
+		assert_eq!(BoostedDeposits::<Test>::get(Asset::Eth, DEPOSIT_ID), None);
+
+		assert_eq!(get_supply_position(Asset::Eth, BOOSTER_1), Some(BOOST_FUNDS + POOL_FEE));
 	});
 }
 
+/// Uses BTC lending pool only
```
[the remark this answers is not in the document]
msgmaxim: done

state-chain/pallets/cf-lending-pools/src/tests.rs
```diff
@@ -966,6 +1079,586 @@ fn boost_pool_details() {
...
+
+	use cf_traits::lending::BoostFinalisationOutcome;
+
+	use super::*;
+
+	// 20% of each pool's fee goes to the network; 80% stays in the pool.
+	const NETWORK_FEE_PERCENT: u128 = 20;
+
+	fn setup_both_pools() {
+		MockPriceFeedApi::set_price_usd_fine(Asset::Eth, 1_000_000);
```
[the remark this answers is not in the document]
msgmaxim: Added a test for boosting with stale prices

state-chain/custom-rpc/src/lib.rs
msgmaxim: done

state-chain/pallets/cf-lending-pools/src/boost.rs
msgmaxim: done

state-chain/pallets/cf-lending-pools/src/boost.rs
msgmaxim: done

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1731,53 +1737,23 @@ where
...
+	}
+
+	fn cf_all_loans(
+		&self,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<Vec<RpcLoan<state_chain_runtime::AccountId, U256>>> {
+		self.rpc_backend.with_versioned_runtime_api(at, |api, hash, api_version| {
+			if api_version < 17 {
+				Ok(Vec::new())
+			} else {
```
dandanlen: FYI I think this case is already handled by the wrapper (ie. in the older api version this method is unavailable and therefore it will just return an api error).

dandanlen (approved):

state-chain/pallets/cf-lending-pools/src/lib.rs
```diff
@@ -97,6 +96,10 @@ pub struct BoostConfiguration {
 	pub network_fee_deduction_from_boost_percent: Percent,
 	/// The minimum amount that can be added to the boost pool.
 	pub minimum_add_funds_amount: BTreeMap<Asset, AssetAmount>,
+	/// Minimum share of the borrowed amount that must come from the lending pool
+	/// (when it has enough available liquidity to cover it). The remainder is split
+	/// proportionally to each pool's available liquidity.
+	pub min_lending_pool_share: Percent,
```
dandanlen, line 102: I think this is causing the merge queue issue: it's returned via the ingress_egress_environment rpc, which can't decode it and breaks.

PR #6499, opened 2026-03-23
test: bouncer decode vault swaps as sanity check

j4m1ef0rd:
# Pull Request

Closes: PRO-2235

## Summary

- Moved all of the vault swap files into a folder. 
- Refactored the use of the `cf_request_swap_parameter_encoding` rpc so it was common.
- Used the `cf_decode_vault_swap_parameter` to always check the encoded params.
- Fixed the solana vault swap decoding RPC. Would always fail with "unsupported solana token".

### DIFF
--- a/state-chain/chains/src/sol.rs
+++ b/state-chain/chains/src/sol.rs
@@ -637,9 +637,16 @@ pub fn decode_sol_instruction_data(
 			)
 			.map_err(|_| "Failed to deserialize SolInstruction")?;
 
-			let token_mint_pubkey: SolAddress = instruction
+			let from_token_account: SolAddress = instruction
 				.accounts
 				.get(sol_tx_core::consts::X_SWAP_TOKEN_FROM_TOKEN_ACC_IDX as usize)
+				.ok_or("Invalid accounts in SolInstruction")?
+				.pubkey
+				.into();
+
+			let token_mint_pubkey: SolAddress = instruction
+				.accounts
+				.get(sol_tx_core::consts::X_SWAP_TOKEN_MINT_ACC_IDX as usize)
 				.ok_or("Invalid accounts in SolInstruction")?
 				.pubkey
 				.into();
@@ -655,7 +662,7 @@ pub fn decode_sol_instruction_data(
 			Ok((
 				amount,
 				src_asset,
-				Some(token_mint_pubkey),
+				Some(from_token_account),
 				dst_chain,
 				dst_address,
 				dst_token,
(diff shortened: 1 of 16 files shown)


msgmaxim (approved):

state-chain/chains/src/sol.rs
```diff
@@ -639,7 +639,7 @@ pub fn decode_sol_instruction_data(
 
 			let token_mint_pubkey: SolAddress = instruction
 				.accounts
-				.get(sol_tx_core::consts::X_SWAP_TOKEN_FROM_TOKEN_ACC_IDX as usize)
```
j4m1ef0rd, line 642: @albert-llimos Can you take a look at this. Claude did this fix for the error I was getting.
```
RpcError: -32021: DispatchError: Unsupported input solana token
```

albert-llimos: This is wrong, Claude just allucinated a new account index that doesn't even exist. I'd say that if you were hitting that error there is almost certaintly something in the refactoring/changes that is wrong.

dandanlen:
Is this waiting for anything?

albert-llimos:
> Is this waiting for anything?

This seems wrong to me: https://github.com/chainflip-io/chainflip-backend/pull/6499#discussion_r3038310858
(also the bouncer needs linting)

state-chain/chains/src/sol.rs
j4m1ef0rd: I reverted the change claude made. But I can't see how I changed the code at all.

albert-llimos: Something does look fishy. I'll look into it.

j4m1ef0rd:
@albert-llimos I think I have fixed the bug with the help of AI. Can you please take a look.

In the `x_swap_token` account layout, index 3 is `from_token_account` — the user's USDC ATA. This is where the tokens get transferred from.
Index 8 is mint — the USDC mint address itself. This is the identifier for the token type.

`decode_sol_instruction_data` was reading index 3 (the user's ATA) and comparing it to the mint pubkey to determine whether the swap is USDC or USDT. Those are completely different addresses — it should be reading index 8 (the mint). The unit test at `sol.rs:952` masks this bug by setting `from_token_account` = `USDC_TOKEN_MINT_PUB_KEY` as a placeholder, so the comparison accidentally works in tests but would fail on a real transaction.

albert-llimos (approved):
That was indeed the bug, thanks @j4m1ef0rd ! 🙏

PR #6508, opened 2026-03-24
feat: Tron

albert-llimos:
# Pull Request

Closes: PRO-2767, PRO-2679, PRO-2673, PRO-2751, PRO-2708, PRO-2706, PRO-2688, PRO-2676, PRO-2674, PRO-2766

## Summary

- Full support for Tron end-to-end with CCM support.
- Modified smart contracts compared to our  regular EVM contracts: https://github.com/chainflip-io/chainflip-eth-contracts/pull/523
- Add Tron client to localnet.
- All bouncer logic included with the same tests as other EVMs.
- There is a chain number gap for BSC because in the new chainflip-cli package BSC was added first, under the assumption that BSC would be launched first given the extra complexity in Tron.
- Add migrations and upgrade-test quirks.

### State Chain

- Tron addresses are represented internally as any other EVM-compatible chain addresses (20 bytes). They are converted into strings in the rpcs as that is the normal representation for TRON.
- The State Chain part of things works very similarly as other EVMs with the exception of a new function selector string for the Tron encoded calls. That is needed for the engine to easily encode transactions via Tron rpc. Instead of the normal 4-byte function selector, the tron rpc expects the string before being hashed to get the selector. That has been added for all EVM calls.
- There is no fee tracking. Fees are pretty much deterministic in TRON. There is no base fee, and the transaction fee is, in short, proportional to the energy and bandwidth consumed.
- There is some ugly/annoying type conversions, especially when it comes to amounts and block numbers because Tron uses signed integers (`i64`) while for EVM chains we tend to use unsigned integers. We could choose to change that for TRON but it just seemed better to keep the unsigned ones in our State Chain for consistency (besides the fact that negative block numbers make no sense).
- There is no concept of gas limit, nor  maxFeePerGas nor similar. There is only the fee_limit in Tron. While it is different than the gas limit, we use it similarly in the State Chain, setting None for non-CCM transactions so the engine does it's own estimation before broadcasting. We only refund the transactions that are exactly what we expect so griefing is not possible. CCM transactions have a set fee_limit (same as for gas_limit in other EVMs) but with a little twist explained below.

### Engine

- Engine using two endpoints, one for the Ethereum compatible json-rpc and the other one for the Tron specific http api. Some rpc providers also split those into different endpoints, so I have set them different endpoints in the engine configuration for maximum flexibility. Under the hood we are still reusing the EVM compatible rpc calls for the json-rpc compatible one.
- Engine uses 21-byte representation of the addresses (`0x41` + EVM address) which is very convenient for both making the rpc calls but also deserializing them into the right length. We use `visible = false` for that.
- On startup we do a `getBlockBalance` rpc call to ensure that the endpoint supports querying historical balance. The majority of mainnet rpc providers have that enable (I think all of them) but some don't in testnet.
- We rely on the Ethereum compatible json-rpc that might not be supported by some rpc providers. Afaik only TronBox might not support that so it's deemed as an OK trade-off.
- The building of transactions luckily does not require a bunch of new encoding logic (Tron does not use the regular rlp encoding). Tron provides some rpcs that encode transactions adequately for us. We then reuse the EVM signing logic from other chains to sign the returned raw payload and then proceed to broadcast.

### Deposit Channels
- Address derivation is the same as other EVMs except the derivation algorithm has a different prefix. Also, the Deposit contract bytecode is different because Tron uses a different compiler.
- Differently than other EVMs, native asset deposit (TRX) on a reused deposit channel won't funnel the funds directly to the Vault but instead requires a fetch from the State Chain. This is due to Tron TRX transfers not guaranteeing the triggering of logic execution on the receiving smart contract.
- Channels are reusable as other EVM chains.
- The witnessing approach is  described in PRO-2688.

### Vault Swaps
-  Vault Swaps in TRON work completely different than other EVM chains. There is no smart contract call but instead it is a TRX or USDT transfer to the Vault with a note/memo attached to it. This is more gas-efficient for TRX and, most importantly, removes the need for the ERC-20 approval for USDT. This note is returned as part of the vault swap encoding rpc on top of the other parameters. The bouncer contains an example of a Vault swap.
- For that encoding that USDT Vault swap, a simple `transfer_token` encoding is added.
- The witnessing approach is very similar to deposit channels, except the note needs to be extracted and decoded. Transaction where the note fails to decode will just be logged.

### Fees, energy and repayment
- See the approach in Linear for how to charge fees: https://linear.app/chainflip/issue/PRO-2677/tron-fee-payment-and-validator-s-repayment-logic
- The way I think about applying that is to use our fee multiplier to decrease the cost of all transactions in TRON proportionally to the amount of energy we can subsidize. A multiplier of <1 can be used to reduce the fees charged to users. Energy gets restored linearly over 24h and we will monitor the energy and fees at all times, like for other EVMs. We shouldn't need to adjust that often to get close to breakeven.
- Repayments will be done as normal. We track the real fees paid by each transaction, which will depend on the energy available at the time, and will refund that to the validators.

### CCM

CCM works the same as any other EVM chains, with the difference that we can't limit the amount of gas that a transaction can consume in Tron, we can only limit the fee paid in TRX. The consequence is that if we subsidize the energy for transactions that go through our Vault, then they could technically spend a lot more energy than what they paid for and "draining" our Vault energy. Therefore, the chosen approach is to charge them fully for the energy they request and set that TRX value as fee limit. In other words, we bypass the multiplier we use to tweak/decrease the fees charged and then make sure the transaction will never pay more than that amount in fees regardless of the energy used.
The alternative there to make it cheaper for aggregators to use CCM is to whitelist receiving contracts but I think this would be a second step, only if that is really necessary.

### Limitations

TRX transfers to smart contracts are not supported. This is because the energy/gas passed when egressing TRX to an address can't be set. Otherwise we would need to stop batching actions together. See https://discord.com/channels/775961728608895008/1478724009070694460
In the Vault contract there is a new check added before an egress is made to make sure the receiving address is not a smart contract. Then it is configurable if we want the Vault to emit a `TransferFallback` event, so the State Chain crafts an independent isolated transfer or if we emit a `TransferNativeSkipped` to be able to log it on the engines and/or react as we deem fit. For now we will do the former.
This new event is the reason why Vault has a separate abi file from Ethereum.

## API & RPC Changes

Addition of TRON for all rpcs. In some it is not a breaking change, as it is just an extra option an en Enum, but others like the monitoring api it is breaking.

### Extrinsics & Events

New events  related to TRON. They are all events already supported for other chains but for TRON, but no Tron-specific events per say.

## Tasks missing
- Write more testing and migrations for the smart contracts in the eth-contracts repo.
- Potentially some more Tron-specific test in the bouncer for some specific behavior (e.g. transfer TRX to smart contract).
- Deploy the smart contracts on Tron and update the chainspecs and migrations.
- Support Tron in the deposit monitor. Now the BLS test is not working because the deposit monitor is not handling Tron events. We want to both fix that and extend support for Tron BLS.

---

## *Checklist*

* The PR is complete:
  * Tests:

### Most critical parts to review 
- Deposit channel and Vault swap witnessing logic, as it's always critical.
- Elections logic. Both because much of it was AI coded, but also because it was done before some of the recent ethereum logic was improved/cleaned up so there might be improvements that can be implemented here too.
- Election initialization after upgrade. It does seem to work on the upgrade test but I might be missing something to initialize it correctly.

#### Note
There is a just few "todos" that require some testing that I will address but otherwise it's ready for review.


state-chain/runtime/src/chainflip/witnessing/tron_elections.rs
```diff
@@ -0,0 +1,461 @@
...
+			Default::default(),
+			Default::default(),
+			Default::default(),
+		),
+		unsynchronised_settings: (
+			BlockHeightWitnesserSettings { safety_buffer: TRON_MAINNET_SAFETY_BUFFER },
+			BlockWitnesserSettings {
+				max_ongoing_elections: 15,
+				max_optimistic_elections: 1,
+				safety_margin: 1,
```
MxmUrw: Same here, we probably want a higher safety margin as well. E.g., for Ethereum we use 4 at the moment.

MxmUrw (commented):
Had a first look over the parts that I'm familiar with. Looks good, some comments:

marcellorigotti:
`RotationBroadcastsPending` is missing `TronBroadcaster`

engine/src/tron/rpc.rs
```diff
@@ -0,0 +1,936 @@
...
+pub trait TronSigningRpcApi: TronRpcApi {
+	fn address(&self) -> H160;
+
+	fn sign_raw_bytes(&self, bytes: Vec<u8>) -> anyhow::Result<Signature>;
+}
+
+// TronRpcSigningClient wraps a TronRpcClient, so it can also satisfy EvmRpcApi by delegation.
+// This allows TronCachingClient<TronRpcSigningClient> to be constructed.
+#[async_trait::async_trait]
+impl EvmRpcApi for TronRpcSigningClient {
```
ramizhasan111: Instead of doing this, we can make TronRpcSigningClient generic:
```rust
pub struct TronRpcSigningClient<RpcClient> {
	wallet: LocalWallet,
	rpc_client: RpcClient,
}
```
and then do a generic impl: ```impl <RpcClient: EvmRpcApi> EvmRpcApi for TronRpcSigningClient<RpcClient>```
Its nicer and more expressive since it ties signing client's EvmRpcApi directly to the client's impl in generic way

engine/src/tron/rpc.rs
```diff
@@ -0,0 +1,936 @@
...
+		sig_bytes[0..32].copy_from_slice(&signature.r.to_big_endian());
+		sig_bytes[32..64].copy_from_slice(&signature.s.to_big_endian());
+		sig_bytes[64] = signature.v as u8;
+
+		Ok(Signature::from(sig_bytes))
+	}
+}
+
+#[async_trait::async_trait]
+impl TronRpcApi for TronRpcSigningClient {
```
ramizhasan111: same for ```TronRpcApi```

state-chain/chains/src/tron/api.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl<E> AllBatch<Tron> for TronApi<E>
+where
+	E: EvmEnvironmentProvider<Tron> + ReplayProtectionProvider<Tron>,
+{
+	fn new_unsigned_impl(
+		fetch_params: Vec<FetchAssetParams<Tron>>,
+		transfer_params: Vec<(TransferAssetParams<Tron>, EgressId)>,
+	) -> Result<Vec<(Self, Vec<EgressId>)>, AllBatchError> {
+		let (transfer_params, egress_ids) = transfer_params.iter().cloned().unzip();
```
ramizhasan111:
```suggestion
		let (transfer_params, egress_ids) = transfer_params.into_iter().unzip();
```

state-chain/chains/src/address.rs
```diff
@@ -596,44 +630,118 @@ pub fn clean_foreign_chain_address(
...
+			),
+			Err(_) => {
+				// Try 21-byte hex (0x41 prefix + 20 bytes), then fall back to raw
+				// 20-byte EVM hex.
+				if let Ok(tron_addr) = clean_hex_address::<TronAddress>(address) {
+					EncodedAddress::Tron(
+						tron_addr.to_evm_address().map_err(|e| anyhow::anyhow!(e))?.into(),
+					)
+				} else {
+					EncodedAddress::Tron(clean_hex_address(address)?)
```
ramizhasan111, line 646: nit: we call ```clean_hex_address``` twice in case of raw 20 byte evm address case. can optimize to call it once and then see whether its 21 byte or 20.

state-chain/pallets/cf-chain-tracking/src/lib.rs
```diff
@@ -227,11 +227,24 @@ impl<T: Config<I>, I: 'static> AdjustedFeeEstimationApi<T::TargetChain> for Pall
...
+		// can cost because we can't control the amount of energy the transaction can spend, we
+		// can only control the TRX burnt. Therefore we charge full TRX burnt for CCM
+		// bypassing the multiplier. Alternatively a whitelist for CCM receivers would be needed.
+		if chain == cf_primitives::ForeignChain::Tron &&
+			matches!(ingress_or_egress, IngressOrEgress::EgressCcm { .. })
+		{
+			estimated_fee
+		} else {
+			FeeMultiplier::<T, I>::get().saturating_mul_int(estimated_fee)
+		}
```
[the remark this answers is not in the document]
ramizhasan111: A more proper way to do this (similar to how we do this elsewhere) would be to inject a trait into the pallet with a function called ```adjust_fee``` or something. This trait would have a default impl of multiplying the multiplier and a tron specific impl that would have this special logic. For tron chain tracking pallet we would inject this specific impl and for all other chains' chain tracking pallets we would inject the default impl.

state-chain/chains/src/address.rs
albert-llimos: The thing is that the `clean_hex_address` fails in that scenario. So to do this we need to do the conversion without using the function, effectively duplicating some logic and matching by length. I think I would rather have it a bit less optimized but seems simpler to read.

```rust
// Hex-decode once, then check if it's a 21-byte Tron address or a
// 20-byte EVM address.
let hex_str = address.strip_prefix("0x").unwrap_or(address);
let bytes = hex::decode(hex_str).context("Invalid hex")?;
EncodedAddress::Tron(match bytes.len() {
	21 => TronAddress::try_from(bytes)
		.map_err(|e| anyhow::anyhow!(e))?
		.to_evm_address()
		.map_err(|e| anyhow::anyhow!(e))?
		.into(),
	20 => bytes
		.try_into()
		.map_err(|_| anyhow::anyhow!("Invalid address length"))?,
	_ => bail!("Invalid Tron address length: expected 20 or 21 bytes"),
})
```

engine/src/tron/rpc.rs
albert-llimos: Done

state-chain/pallets/cf-chain-tracking/src/lib.rs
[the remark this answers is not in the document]
albert-llimos: I thought it was a bit unnecessary to add that but you're right that it's the more correct way. I have implemented it now.

engine/src/tron/retry_rpc.rs
```diff
@@ -0,0 +1,610 @@
...
+								}
+							},
+							_ => {
+								return Err(anyhow::anyhow!(
+									"Failed to estimate energy (code: {:?}, message: {:?})",
+									energy_estimate.result.code,
+									energy_estimate.result.message,
+								));
+							},
+						}.min(i64::MAX as u128) as i64;
```
ramizhasan111, line 422:
```suggestion
						}.try_into()?;
```

state-chain/pallets/cf-chain-tracking/src/lib.rs
```diff
@@ -222,29 +225,43 @@ impl<T: Config<I>, I: 'static> GetBlockHeight<T::TargetChain> for Pallet<T, I> {
...
+impl<C: Chain> FeeMultiplierProvider<C> for DefaultFeeMultiplier {
+	fn adjust_fee(
+		estimated_fee: C::ChainAmount,
+		fee_multiplier: FixedU128,
+		_ingress_or_egress: &IngressOrEgress,
+	) -> C::ChainAmount {
+		fee_multiplier.saturating_mul_int(estimated_fee)
+	}
+}
+
```
ramizhasan111, line 250: we should move this to traits crate since that is where we define (and default impl) these kind of crates

engine/src/witness/tron_elections.rs
```diff
@@ -0,0 +1,519 @@
...
+			tracing::warn!("Transaction result is not SUCCESS: {result}");
+		}
+	}
+
+	// Pass all the relevant information to the State Chain in case we want to use it to adjust
+	// energy estimates in the future.
+	TronTransactionFee {
+		fee: tx_info.fee.unwrap_or(0).try_into().unwrap_or(0),
+		energy_usage: Some(receipt.energy_usage.unwrap_or(0).try_into().unwrap_or(0)),
+		energy_fee: Some(receipt.energy_fee.unwrap_or(0).try_into().unwrap_or(0)),
```
ramizhasan111:
```suggestion
		energy_fee: receipt.energy_fee.map(|f| f.try_into().unwrap_or(0)),
```
i think we should convert all lines here to something like this. I dont think it makes sense to unwrap and then wrap it in `Some` again. We should keep the `Option` as it is and convert the inner value from i64 to u64.

engine/src/witness/tron_elections.rs
```diff
@@ -0,0 +1,519 @@
...
+		vault: vault_address,
+		vault_events: vault_event_source,
+		supported_assets: [(TronAsset::TrxUsdt, usdt_event_source)].into_iter().collect(),
+	};
+
+	let key_manager_config = TronKeyManagerWitnessingConfig {
+		key_manager: EvmEventSource::new::<KeyManagerEvents>(key_manager_address),
+	};
+
+	scope.spawn(async move {
```
marcellorigotti: we should use `spawn_with_restart` here, like we do for other witnessing tasks

engine/src/witness/tron_elections.rs
albert-llimos: Fair enough. Initially I wrote that to avoid having `None`  in the `TronTransactionFee`, which is a bit weird, but it seems better to just pass along the Option and then let the FeeRefundCalculator deal with it when computing the fee owed.

engine/src/witness/tron_elections.rs
albert-llimos: You're right. I think that was changed after I started working in Tron so I used the old `spawn`. Fixing that now .

.github/workflows/upgrade-test.yml
```diff
@@ -220,7 +218,20 @@ jobs:
           sed -i '/testGovernanceDepositWitnessing/d' bouncer/tests/fast_bouncer.test.ts
           cd bouncer
           pnpm install --frozen-lockfile
-          ./test_commands/all_cuncurrent_tests.sh
+          ./fast_bouncer.sh
+          git reset --hard
```
dandanlen: Is this deliberate or a merge artifact?

api/bin/chainflip-ingress-egress-tracker/src/witnessing/state_chain.rs
dandanlen, line 1: This file is deleted on main, we don't want to re-add it do we?

state-chain/chains/src/tron.rs
```diff
@@ -0,0 +1,636 @@
...
+
+use self::evm::EvmCrypto;
+
+// Reference constants for the chain spec
+pub const CHAIN_ID_MAINNET: u64 = 728126428; // 0x2b6653dc
+pub const CHAIN_ID_NILE_TESTNET: u64 = 3448148188; // 0xcd8690dc
+
+// Tron uses i64 so we could use u64. However, using u128 makes it
+// simpler to share code with the rest of the EVM-based chains.
+pub type TronAmount = u128;
```
dandanlen, line 67: How do they deal with 18-decimal assets? Do they just not support any?

`i64::MAX` is `9,223,372,036,854,775,807` which for eg. ETH is `9.22 ETH`.

.github/workflows/upgrade-test.yml
albert-llimos: Definitely some merge artifact, as the upgrade test used to pass and this looks wrong. Reverted it.

state-chain/traits/src/lib.rs
```diff
@@ -1069,6 +1070,29 @@ pub trait AdjustedFeeEstimationApi<C: Chain> {
 	fn estimate_fee(asset: C::ChainAsset, ingress_or_egress: IngressOrEgress) -> C::ChainAmount;
 }
 
+/// Trait for adjusting the fee estimate with the fee multiplier.
+/// Different chains can provide different implementations to customize fee adjustment behavior.
+pub trait FeeMultiplierProvider<C: Chain> {
+	fn adjust_fee(
```
dandanlen, line 1076: There's probably a better way to do this without adding a new single-use trait. I'll create a ticket to restructure fee estimation in general, we can tackle it then.

state-chain/runtime/src/safe_mode.rs
```diff
@@ -183,6 +196,11 @@ impl CallDispatchFilter<RuntimeCall> for WitnesserCallPermission {
 			RuntimeCall::AssethubIngressEgress(..) => self.assethub_ingress_egress,
 			RuntimeCall::AssethubVault(..) => self.assethub_vault,
 
+			RuntimeCall::TronBroadcaster(..) => self.tron_broadcast,
+			RuntimeCall::TronChainTracking(..) => self.tron_chain_tracking,
+			RuntimeCall::TronIngressEgress(..) => self.tron_ingress_egress,
+			RuntimeCall::TronVault(..) => self.tron_vault,
```
dandanlen, line 202: These are all unnecessary, we don't use any legacy witnessing for tron - we can remove these all together with the other chains in a separate PR.

engine/src/tron/cached_rpc.rs
```diff
@@ -0,0 +1,313 @@
...
+#[async_trait::async_trait]
+pub trait TronRetryRpcApiWithResult: Clone {
+	// Tron HTTP API
+	async fn get_transaction_info_by_id(&self, tx_id: &str) -> anyhow::Result<TransactionInfo>;
+	async fn get_transaction_by_id(&self, tx_id: &str) -> anyhow::Result<TronTransaction>;
+	async fn get_block_balances(
+		&self,
+		block_number: BlockNumber,
+		hash: &str,
+	) -> anyhow::Result<BlockBalance>;
```
dandanlen: Design nit: we use `&str` arguments in some places here. Besides being inconsistent, this requires the caller of the api method to know the string format that is expected (hex/base58/leading `0x`? etc.). Better to keep the interface more strongly typed.

engine/src/witness/tron/vault_swaps_witnessing.rs
```diff
@@ -0,0 +1,504 @@
...
+					block_number as u64,
+				) {
+					Ok(result) => result,
+					Err(e) => {
+						tracing::warn!(
+							"Failed to decode cf_parameters without CCM for tx {}: {:?}",
+							tx_id,
+							e
+						);
+						continue;
```
dandanlen: There are a few failure cases in here where just silently log and continue, meaning any deposit is ignored and will need to be recovered.

Yet we *do* have support for handling failed vault txs via transfer failed witnessing, and it looks like we handle this in the calling context. Would we not want to map some of these parsing failures into transfer failed events or similar?

engine/src/witness/tron/vault_swaps_witnessing.rs
```diff
@@ -0,0 +1,504 @@
...
+	pub destination_address: EncodedAddress,
+	pub ccm_data: Option<(AssetAmount, CcmMessage)>,
+	pub cf_parameters: Vec<u8>,
+}
+
+pub async fn fetch_and_decode_transactions<Client>(
+	client: &Client,
+	vault_ingress_transactions: Vec<(TronAsset, u64, H256)>,
+	block_number: i64,
+) -> Result<Vec<VaultDepositWitness<Runtime, TronInstance>>, anyhow::Error>
```
dandanlen: This returns a Result type but but I can't see any Err being returned from any branch. As discussed below, we are logging and then skipping the errors. IMO it would be cleaner to define a local Error type using `thiserror` and then handle it in the calling context. (return `Vec<Result<_, TronFetchAndDecodeError>>` for example).

engine/src/witness/tron/vault_swaps_witnessing.rs
```diff
@@ -0,0 +1,504 @@
...
+pub async fn witness_vault_swaps<Client: TronRetryRpcApiWithResult + Send + Sync + Clone>(
+	client: &Client,
+	config: &VaultDepositWitnessingConfig,
+	query: &EvmSingleBlockQuery,
+) -> Result<Vec<VaultDepositWitness<Runtime, TronInstance>>, anyhow::Error> {
+	let block_number_u64 = query.get_lowest_block_height_of_query();
+	let block_number = i64::try_from(block_number_u64).map_err(|_| {
+		anyhow::anyhow!("Block number conversion to i64 failed: value too large or negative")
+	})?;
+	let block_hash = format!("{:064x}", query.block_hash);
```
dandanlen: Similar comment to elsewhere: the knowledge of how exactly the rpc string argument needs to be formatted should be in the rpc - the caller shoudn't have to know about this.

engine/src/witness/tron/vault_swaps_witnessing.rs
```diff
@@ -0,0 +1,504 @@
...
+		vault_ingress_transactions.into_iter().zip(transactions_info_result)
+	{
+		let tx_id_str = format!("{:x}", tx_id);
+		let transaction = match transaction_info_result {
+			Ok(tx) => tx,
+			Err(e) => {
+				tracing::warn!("Failed to get transaction {:?}: {:?}", tx_id_str, e);
+				continue;
+			},
+		};
```
dandanlen: Errors like this would be better written as:


```suggestion
		let transaction = match transaction_info_result {
			Ok(tx) => tx,
			Err(e) => {
				tracing::warn!("Failed to get transaction {:#x}: {:#}", tx_id, e);
				continue;
			},
		};
```

No need for the intermediate string allocation, but also, with `{:?}` the error will not display as cleanly in the logs. `{:#x}` will use `0x..` hex representation for hashes, and `{:#}` will display anyhow errors as a human-readable error chain. 

However, as mentioned elsewhere, a better approach for this would IMO be an Error type with clear variants, and we handle the errors and logging consistently in one place, at the call-site of this function.

engine/src/tron/retry_rpc.rs
```diff
@@ -0,0 +1,610 @@
...
+							.await
+							.context("Failed to estimate energy rpc call failed")?;
+
+						let fee_limit: i64 = match (energy_estimate.result.result, energy_estimate.energy_required) {
+							(Some(true), Some(estimated_energy)) => {
+								if estimated_energy <= 0 {
+									return Err(anyhow::anyhow!("estimated_energy must be greater than 0, got {}", estimated_energy));
+								}
+
+								let estimated_fee_limit_sun = (estimated_energy as u128)
```
marcellorigotti: Nit. `estimated_fee_limit_sum`

dandanlen (commented):
First batch of comments, just some minor issues. Will continue reviewing.

I'll also push a small commit - just moves the tron migrations to the 2_2 migrations. It's not really housekeeping.

engine/src/tron/retry_rpc.rs
```diff
@@ -0,0 +1,610 @@
...
+									.saturating_div(cf_chains::tron::fees::ENERGY_PER_TX_TRX_BURN);
+
+								if let Some(fee_limit) = transaction.fee_limit {
+									let fee_limit_u128: u128 = fee_limit.into();
+									if estimated_fee_limit_sun > fee_limit_u128 {
+										return Err(anyhow::anyhow!(
+											"Estimated fee limit: {} SUN, is greater than the transaction fee limit {} SUN", estimated_fee_limit_sun, fee_limit_u128
+										));
+									}
+									estimated_fee_limit_sun
```
marcellorigotti: shouldn't we return `fee_limit` here? like we do for other evm broadcasts?

engine/src/tron/retry_rpc.rs
albert-llimos: It is actually `sun`. That's the smallest unit (same as `wei` in EVM).

state-chain/chains/src/tron.rs
albert-llimos: That applies only to Trx, which has only 6 decimals.  Therefore `i64` is enough. Anyone could deploy an ERC20 with more decimals but that's totally different i.e. ERC20 events can be `uint128`. It's just the native Trx values that use the `i64`.

state-chain/traits/src/lib.rs
dandanlen: Also btw. - what's the reason for bypassing the multiplier for tron ccms?

engine/src/witness/tron/vault_swaps_witnessing.rs
dandanlen: A specific example of where we might not want to skip is when we fail to decode SCALE data.

engine/src/tron/retry_rpc.rs
marcellorigotti: ah my bad I though you meant sum, which also doesn't make much sense since it is not a sum 😅

state-chain/traits/src/lib.rs
albert-llimos: I originally just had a simple piece of logic but Ramiz suggested to refactor it into a common trait:
https://github.com/chainflip-io/chainflip-backend/pull/6508#discussion_r3094241925
You can see the change here:
https://github.com/chainflip-io/chainflip-backend/pull/6508/commits/8f2d5e5dc5c2be3bafb43a5c519debdc57838cd0
We might want to restructure it further anyway though.

albert-llimos: > Also btw. - what's the reason for bypassing the multiplier for tron ccms?

I left a comment in the implementation:
https://github.com/chainflip-io/chainflip-backend/blob/feat/tron-witnessing/state-chain/runtime/src/chainflip.rs#L571-L573

marcellorigotti:
There are a bunch of casts and plain addition throughout the PR, these shouldn't be a problem but it would be good if we were explicitly converting with `try_into()` and saturating when summing/subtracting. This to be future proof and to handle scenarios where a bad RPC returns the wrong value.

Also why do we use i64 for many types which can't be negative? i.e. block number or timestamp? is this to match Tron implementation?

engine/src/tron/retry_rpc.rs
albert-llimos: You're right. I'll fix it.

albert-llimos:
> Also why do we use i64 for many types which can't be negative? i.e. block number or timestamp? is this to match Tron implementation?

I wrote it in the description
"There is some ugly/annoying type conversions, especially when it comes to amounts and block numbers because Tron uses signed integers (i64) while for EVM chains we tend to use unsigned integers. We could choose to change that for TRON but it just seemed better to keep the unsigned ones in our State Chain for consistency (besides the fact that negative block numbers make no sense)."
Having said that, I agree about handling better some of the conversions.

engine/src/witness/tron/vault_swaps_witnessing.rs
albert-llimos: Afaik the mechanism you describe has only been implemented for BTC (`VaultDepositWitness::unrefundable`) and not in EVM chains. I'm not against implementing it here but for consistency and to make sure we are not introducing any new issues it might make sense to separately implement this for all EVMs at once later on. Wdyt?

state-chain/runtime/src/safe_mode.rs
albert-llimos: Agreed — leaving these in for now to keep this PR focused on adding Tron, and will remove the legacy witnessing entries for all chains together in a follow-up.

state-chain/chains/src/tron.rs
```diff
@@ -0,0 +1,636 @@
...
+	let hash1 = Sha256::digest(data);
+	Sha256::digest(hash1).into()
+}
+
+/// A Tron address: 0x41 prefix + 20-byte EVM address.
+///
+/// Display/FromStr use base58check encoding (always starts with 'T').
+/// Serde uses hex encoding (42 hex chars: "41" + 40 hex digits) for Tron HTTP API compatibility.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TronAddress(pub [u8; 21]);
```
dandanlen: The leading 0x41 is a constant, it doesn't need to be part of the runtime data. So a cleaner way to implement this is probably as a wrapper around EvmAddress or raw `[u8; 20]`.

For example it's possible to construct in invalid address here (`eg. `let addr = TronAddress(Default::default());`.

state-chain/chains/src/tron.rs
```diff
@@ -0,0 +1,636 @@
...
+		write!(f, "{}", self.to_base58check())
+	}
+}
+
+impl Serialize for TronAddress {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+	where
+		S: serde::Serializer,
+	{
+		serializer.serialize_str(&hex::encode(self.0))
```
dandanlen: This is not very consistent: ToStr/FromStr uses base58 and serialize/deserialize use hex? Any particular reason?

albert-llimos: ToStr/FromStr is used to convert between a base58 encoded string to the TronAddress bytes, for example with String human-readable addresses passed externally. The de/serialization is used for the engine rpc were we convert to/from State Chain TronAddress (bytes) into the hex, because we use `visible=false` (aka hex bytes instead of the String).
I get how this doesn't look consistent, I implemented in steps what was actually needed.

state-chain/traits/src/lib.rs
dandanlen: I agree we should keep tron-specific logic out of the pallet, but i don't think we're achieving our aim: we created what is effectively a tron-specific trait (even if the name doesn't mention it) and wired that into the pallet instead. It's not a big deal, just something we should address as part of a later tech debt issue.

dandanlen: > > Also btw. - what's the reason for bypassing the multiplier for tron ccms?
> 
> I left a comment in the implementation: [`feat`/tron-witnessing/state-chain/runtime/src/chainflip.rs#L571-L573](https://github.com/chainflip-io/chainflip-backend/blob/feat/tron-witnessing/state-chain/runtime/src/chainflip.rs?rgh-link-date=2026-04-22T09%3A44%3A13Z#L571-L573)

I read the comment and I still don't understand 😅

engine/src/tron/cached_rpc.rs
albert-llimos: True. I have now replaced both the transaction hashes and the block hashes with`H256` and then formatted appropriately in the rpc.

state-chain/chains/src/tron.rs
dandanlen: Ok, so the convention in Tron-land is that base58 is for humans but machines use hex by default? Makes (sort of) sense but let's add some comments to explain this, otherwise it's not obvious. 

Also btw. I saw that we add this `visible: false` even in queries that don't send addresses... does it have any effect there? (Like does is affect return values?)

engine/src/tron/rpc_client_api.rs
```diff
@@ -0,0 +1,389 @@
...
+	pub contract_address: Option<String>,
+	pub receipt: ResourceReceipt,
+	pub log: Option<Vec<Value>>,
+	pub result: Option<String>,
+	#[serde(rename = "resMessage")]
+	pub res_message: Option<String>,
+	#[serde(rename = "assetIssueID")]
+	pub asset_issue_id: Option<String>,
+	pub withdraw_amount: Option<Amount>,
+	pub unfreeze_amount: Option<Amount>,
```
dandanlen, line 74: I noticed some are renamed to camelCase, some aren't... is this seriously how these rpcs work??

engine/src/tron/rpc_client_api.rs
```diff
@@ -0,0 +1,389 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlockBalance {
+	pub timestamp: i64,
+	pub block_identifier: BlockIdentifier,
+	pub transaction_balance_trace: Vec<TransactionBalanceTrace>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlockIdentifier {
+	pub hash: String,
```
dandanlen: Why  not `U256`?

state-chain/chains/src/tron.rs
albert-llimos: The convention is base58 for humans and then optional for machines. I chose to use 20-bytes (visible=false) for the engine queries because it makes it easier when getting the 20-byte addresses from the State Chain, so we don't have to convert them to strings, do the rpc calls, get strings back, reconvert them to bytes and submit them to the State Chain.

> Also btw. I saw that we add this visible: false even in queries that don't send addresses... does it have any effect there? (Like does is affect return values?)

Yes exactly, it affects the return values.

albert-llimos: I will add a comment explaining that.

engine/src/tron/rpc_client_api.rs
albert-llimos: That's what is documented and working. Check it out:
https://developers.tron.network/reference/gettransactioninfobyid

engine/src/tron/rpc_client_api.rs
```diff
@@ -0,0 +1,389 @@
...
+	pub code: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TriggerConstantContractRequest {
+	pub owner_address: TronAddress,
+	pub contract_address: TronAddress,
+	pub function_selector: String,
+	pub parameter: Vec<u8>,
+}
```
dandanlen: This and `TriggerSmartContractRequest` seem unnecessary. We construct them manually, then we pass them through an interface, and the implementation immediately deconstructs it again and builds a `json!` payload manually. 

We should choose one or the other:
- just pass in the args to the interface and construct the json payload manually, delete the structs
- OR define keep the structs, implement serialize/deserialize correctly and use those. 

(I think the former is simpler and sufficient for this use case)

engine/src/witness/tron/vault_swaps_witnessing.rs
dandanlen: Ok, I misremembered, I thought this was already supported / got confused with SwapNativeFailed etc.

We should come back to this but ok to leave it in this PR.

state-chain/traits/src/lib.rs
albert-llimos: It's quite hard without writing a massive comment.
In EVM we can limit the gas consumed in the transaction itself setting the gas limit. In Tron we can't do that, we can only limit the fee paid. That is TRX burnt to pay for energy that has not been subsidized. We use the multiplier to adjust the energy we have available on average to make all calls cheaper. 
For example, let's say some energy is subsidized so the multiplier is at 0.5. A user requests 100k worth of energy and we calculate that at max, it will require burning 10 TRX. We charge the user 10 * 0.5  = 5 TRX. However, the crafted transaction has a `fee_limit` of 10 TRX, so the user could spend more than they have been charged. We don't really want to put the `fee_limit` to 5 either, as the user wanted 10 TRX worth of energy and we can't really guarantee that the energy available at the time of execution will pay for up to 5 TRX. Therefore, the most reasonable approach that guarantees that the protocol won't lose money is to charge 10 TRX to the user (not apply the multiplier) and set the fee_limit of the transaction to 10 TRX.
They could technically then be spending our energy (we can set a cap) but there is not really any incentive, energy gets replenished every day and it costs fees to do this.
The alternative to this is to whitelist receiver contracts but that seems quite heavy handed.

engine/src/tron/rpc_client_api.rs
dandanlen: LOL

state-chain/chains/src/tron.rs
```diff
@@ -0,0 +1,636 @@
...
+	type ChainCrypto = EvmCrypto;
+	type ChainBlockNumber = u64;
+	type ChainAmount = TronAmount;
+	type TransactionFee = TronTransactionFee;
+	type TrackedData = TronTrackedData;
+	type ChainAsset = assets::tron::Asset;
+	type ChainAssetMap<
+		T: Member + Parameter + MaxEncodedLen + Copy + BenchmarkValue + FullCodec + Unpin,
+	> = assets::tron::AssetMap<T>;
+	type ChainAccount = evm::Address;
```
dandanlen: Not TronAddress?

engine/src/tron/rpc_client_api.rs
albert-llimos: It's wild...

engine/src/tron/rpc_client_api.rs
albert-llimos: I missed it when updated block and transaction hashes. Done.

state-chain/chains/src/tron.rs
albert-llimos: Before my latest change TronAddress was defined as 21 bytes, which then would have made that very annoying. Now that it's defined as a wrapper of the EVM address we could do that but then we'd need to add extra conversion logic in some places (specific to Tron instead of using the Evm Address), some benchmarking etc... I think it's cleaner and easier to reason about to just use the EVM address here and just use the TronAddress for conversions at the edges i.e. rpcs.
Let me know if you'd rather change it anyway.

engine/src/witness/tron/vault_swaps_witnessing.rs
albert-llimos: Yes that's a different mechanism.
I opened a triage ticket so we don't forget. https://linear.app/chainflip/issue/PRO-2832/handle-vault-swap-ingress-failure-cases-for-evm-chains

state-chain/chains/src/tron.rs
albert-llimos: Right. Nothing is really stored with the extra byt, it's only used in conversions, but it's true. I have implemented it as a wrapper now, it does seem cleaner.

engine/src/tron/rpc_client_api.rs
albert-llimos: I have now implemented that. Also for `TriggerConstantContractRequest`, which is pretty much the same thing.

state-chain/chains/src/tron.rs
dandanlen: Ok, let's leave it as-is for now, I see we've taken the same approach for Arb and Eth, so it's consistent with that.

I think we might be making the wrong trade-off: we should be adding friction between different chains and removing friction in the rpc conversions. At the moment it seems we're doing the opposite.

state-chain/chains/src/tron.rs
```diff
@@ -0,0 +1,663 @@
...
+	evm::{DeploymentStatus, EvmFetchId},
+	ChainWitnessConfig, FeeEstimationApi, *,
+};
+use cf_primitives::chains::assets;
+pub use cf_primitives::chains::Tron;
+use codec::{Decode, Encode, MaxEncodedLen};
+pub use ethabi::{
+	ethereum_types::{H160, H256},
+	Address, Hash as TxHash, Token, Uint, Word,
+};
```
dandanlen: nit: these are already exported from the evm module and having two sources for this confusing and misleading (these would be accessible as tron::Address for example, even though it's not tron-specific).

state-chain/chains/src/address.rs
```diff
@@ -467,14 +496,16 @@ pub enum ForeignChainAddressHumanreadable {
...
 }
 
 #[cfg(feature = "std")]
 impl std::fmt::Display for ForeignChainAddressHumanreadable {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
 		match self {
 			ForeignChainAddressHumanreadable::Eth(address) |
-			ForeignChainAddressHumanreadable::Arb(address) => write!(f, "{:#x}", address),
+			ForeignChainAddressHumanreadable::Arb(address) |
+			ForeignChainAddressHumanreadable::Tron(address) => write!(f, "{:#x}", address),
```
dandanlen: This will format a (human readable) tron address in the same way as eth/arb addresses, ie. 0x-prefixed hex. Is that what we want? We mentioned elsewhere that human-readable meant base58.

albert-llimos: Totally right, I missed that. I've fixed it now.

state-chain/traits/src/lib.rs
dandanlen: Thanks I think I get it now (also saw your comment in the PR summary).

engine/src/tron/rpc_client_api.rs
```diff
@@ -0,0 +1,372 @@
...
+/// result bool is not existent.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TransactionExtentionEnergyResult {
+	// #[serde(skip_serializing_if = "Option::is_none")]
+	pub result: Option<bool>,
+	// #[serde(skip_serializing_if = "Option::is_none")]
+	pub code: Option<Value>,
+	// #[serde(skip_serializing_if = "Option::is_none")]
+	pub message: Option<String>,
+}
```
MxmUrw, line 237: We might want to remove these commented attributes? I assume these fields are always populated then.

state-chain/chains/src/eth/deposit_address.rs
```diff
@@ -88,7 +126,7 @@ pub fn get_create_2_address(
 	);
```
MxmUrw, line 126: Instead of storing the full tron bytecode on the statechain in `DEPOSIT_CONTRACT_BYTECODE_TRON`, could we theoretically precompute and just store the `deploy_transaction_bytes_hash` value?

If I see it correctly, since we're feeding the bytecode to a non-const function, it must be available in the runtime code and so must be stored somewhere, right?

This is just ~1400 bytes, but still wondering.

bouncer/tests/gaslimit_ccm.test.ts
```diff
@@ -376,6 +377,62 @@ async function testEvmInsufficientGas<A = []>(
...
+    undefined,
+    () => stopObservingCcmReceived,
+  ).then((event) => {
+    if (event !== undefined) {
+      throw new Error(`$CCM event emitted. Transaction should not have been broadcasted!`);
+    }
+  });
+  await observeEvent(cf.logger, `${destChain.toLowerCase()}Broadcaster:BroadcastAborted`, {
+    test: (event) => event.data.broadcastId === broadcastId,
+  }).event;
```
MxmUrw: Since we're running this test just for swaps to the tron chain, I think it's nicer to use ChainflipIO:

```suggestion
  await cf.stepUntilEvent(
    'TronBroadcaster.BroadcastAborted',
    tronBroadcasterBroadcastAborted.refine(
      (data) => data.broadcastId === broadcastId
    )
  );
```

state-chain/node/src/chain_spec/berghain.rs
```diff
@@ -449,3 +455,4 @@ pub const BITCOIN_SAFETY_MARGIN: u64 = 2;
 pub const ETHEREUM_SAFETY_MARGIN: u64 = 2;
 pub const ARBITRUM_SAFETY_MARGIN: u64 = 1;
 pub const SOLANA_SAFETY_MARGIN: u64 = 1; // Unused - we use "finalized" instead
+pub const TRON_SAFETY_MARGIN: u64 = 10; // TODO: to set the right number
```
MxmUrw: Actually, this (and also for all the other chains) can be deleted after PRO-2795, as it's unused.

state-chain/runtime/src/chainflip/witnessing/tron_elections.rs
```diff
@@ -0,0 +1,461 @@
...
+
+define_empty_struct! { pub struct TronChain; }
+
+impl ChainTypes for TronChain {
+	type ChainBlockNumber = <Tron as Chain>::ChainBlockNumber;
+	type ChainBlockHash = evm::H256;
+	const NAME: &'static str = "Tron";
+}
+
+pub const TRON_MAINNET_SAFETY_BUFFER: u32 = 20; // TODO: to set the right number
```
MxmUrw: I would set this to 25 or 30.

state-chain/runtime/src/chainflip/witnessing/tron_elections.rs
```diff
@@ -0,0 +1,461 @@
...
+			Default::default(),
+			Default::default(),
+			Default::default(),
+		),
+		unsynchronised_settings: (
+			BlockHeightWitnesserSettings { safety_buffer: TRON_MAINNET_SAFETY_BUFFER },
+			BlockWitnesserSettings {
+				max_ongoing_elections: 15,
+				max_optimistic_elections: 1,
+				safety_margin: 10, // TODO: To set the right number
```
MxmUrw: As discussed this depends on whether we can gather data about tron reorgs before launch. If we don't have any data, let's set it to 20 and then adjust depending on how it runs for a day or two.

state-chain/runtime/src/migrations/tron_integration.rs
```diff
@@ -0,0 +1,132 @@
...
+		let unsynchronized_settings =
+			ElectoralUnsynchronisedSettings::<Runtime, crate::TronInstance>::get();
+		assert_eq!(
+			unsynchronized_settings,
+			Some((
+				BlockHeightWitnesserSettings { safety_buffer: TRON_MAINNET_SAFETY_BUFFER },
+				BlockWitnesserSettings {
+					max_ongoing_elections: 15,
+					max_optimistic_elections: 1,
+					safety_margin: 1,
```
MxmUrw: We've updated the initial safety margin in `witnessing/tron_elections`, so we'll also have to update it here. Maybe we can make a new `const TRON_MAINNET_SAFETY_MARGIN` that we reference here.

MxmUrw (commented):
Had a second look at everything. Looks good!

state-chain/node/src/chain_spec/berghain.rs
albert-llimos: I will delete it once that is complete. There is no activity in the ticket so I'd rather leave this here for now.

albert-llimos:
> Had a second look at everything. Looks good!

Thanks Maxim! Regarding the safety witness comments, I will update all those values when we decide on the final settings.

state-chain/chains/src/eth/deposit_address.rs
albert-llimos: According to my friend Claude:

> The bytecode is needed at runtime to:
> 
> 1- Construct the actual deploy transaction bytes (passed to Tron's API)
> 2- Derive the deposit address (the contract address depends on the bytecode hash, but you still need the bytecode to build the transaction)
> So the bytecode must be available in the runtime binary regardless. Storing just the hash would save ~1400 bytes of storage on-chain, but since it's a const, it's compiled into the WASM blob anyway — it doesn't actually occupy chain storage. The distinction is: const values live in the runtime WASM, not in the Substrate key-value store.
> 
> So the 1400 bytes are in the WASM runtime binary regardless, and there's no storage to save.

MxmUrw: If I search in the codebase for the constant `DEPOSIT_CONTRACT_BYTECODE_TRON`, it appears only a single time, in the function `get_create_2_address_tron`, where it's passed to `get_create_2_address_prefixed`. Inside of this function it is used a single time (the local variable name is `bytecode`), namely in the following computation:
```rust
	let deploy_transaction_bytes_hash = Keccak256::hash(
		itertools::chain!(
			bytecode,
			[0u8; 12], // padding
			token_address.unwrap_or(ETHEREUM_ETH_ADDRESS).to_fixed_bytes(),
		)
		.collect::<Vec<_>>()
		.as_slice(),
	);
```
The result type of the computation is H256 and all inputs to it are statically known values. Is there some other use for the bytecode I'm missing here?

So:
1. To me it looks like we *could* save these bytes.
2. The WASM runtime *is* stored inside the key-value store (under the key `:code`).

albert-llimos: The `token_address` is not known at compile time, it depends on the asset.
I guess we could have a different `deploy_transaction_bytes_hash` precomputed for each different asset asset. Or actually, it would need to be precomputed for each different asset address, as that is what the `tokenAddress` is. I guess that would maybe work but I think that would look quite hacky..

MxmUrw: > Or actually, it would need to be precomputed for each different asset address, as that is what the tokenAddress is.

Yes, this is what I had in mind.

@dandanlen @marcellorigotti @ramizhasan111 does anyone of you have an opinion on this one?

MxmUrw: Talked with Dan about this, looks like it would be a bit cumbersome to support (also because we would need precomputed hashes for all testnets x tokens). So I agree that it's best to leave it as it is.

state-chain/custom-rpc/src/lp.rs
```diff
@@ -539,16 +539,17 @@ where
...
-			.cf_free_balances(
-				self.rpc_backend.client.info().finalized_hash,
-				self.signed_pool_client.account_id(),
-			)
-			.map_err(CfApiError::from)?
-			.map(Into::into))
+		let account_id = self.signed_pool_client.account_id();
+		let at = Some(self.rpc_backend.client.info().finalized_hash);
+		self.rpc_backend.with_versioned_runtime_api(at, |api, hash, version| {
+			if version < 17 {
```
nmammeri: This should be consistent with the `cf_free_balances` in custom-rpc/lib.rs.  The clones of the accountId are extra. something like:

```
if version < 16 {
	#[expect(deprecated)]
	api.cf_free_balances_before_version_16(hash, account_id).map(Into::into)
} else if version < 17 {
	#[expect(deprecated)]
	api.cf_free_balances_before_version_17(hash, account_id).map(Into::into)
} else {
	api.cf_free_balances(hash, account_id)
}
.map(|balances| balances.map(Into::into))
```

nmammeri (approved):

dandanlen (approved):

PR #6509, opened 2026-03-24
chore: reduce number of swaps in AllSwaps test

MxmUrw:
# Pull Request

Part of PRO-2776

## Summary

We're forced to fight with the external node being overloaded both for the evm chains as well as for bitcoin. There are various options to go forward, such as introducing multiple btc wallets to reduce contention, but it all adds quite some complexity. Additionally, since we're adding new chains, the contention will become even higher in the future.

The easier alternative is to reduce the number of swaps that run in the AllSwaps test. This can be done in a smart way, since in fact, we're interested in testing 3 main components:
1. Ingress of an asset
2. Swapping mechanism
3. Egress of an asset

So to reduce the number of asset pairs that we test, instead of taking all possible input-output pairs, we can randomly select asset pairs with the condition that every input asset has to be tested by at least one swap and every output asset has to be tested by at least one swap.

That's the idea. The actual implementation has the following three additions:
 - On the input side, we cover every input asset one time as deposit channel swap, and additionally one time as a vault swap
 - Whenever an output asset of a given pair supports CCM, we add two tests for this pair, one normal and one with CCM
 - To do a bit more testing, we run the above procedure twice, effectively testing every input and output asset at least twice

This results in 105 AllSwaps tests, compared to ~490 previously.

### RPCS

None

### Extrinsics & Events

None

---

* The PR is complete:


j4m1ef0rd (approved):

dandanlen (approved):

j4m1ef0rd:
What is the status of this? I'm still getting flaky tests with AssetHub getting BroadcastAborted because the AssetHub node can't handle it. I think this PR might help.

MxmUrw:
> What is the status of this? I'm still getting flaky tests with AssetHub getting BroadcastAborted because the AssetHub node can't handle it. I think this PR might help.

Good question. And yeah you're right this PR should help with all of the nodes being overloaded. Not sure what's going on here though, there seems to be some DM problem which is quite orthogonal to what this PR does, but last time I tried looking into it it was quite deterministic IIRC.

j4m1ef0rd:
I got the bouncer passing, but it required some small changes to how the engine handles the assethub transaction errors and retries. @MxmUrw @msgmaxim can you please take a look.

engine/src/state_chain_observer/sc_observer.rs
```diff
@@ -641,6 +641,30 @@ where
...
+                                                                    Ok(tx_hash) => info!("Assethub PendingBroadcast {awaiting_broadcast_id:?} success: tx_hash: {tx_hash:#x}"),
+                                                                    Err(error) => {
+                                                                        // Just log: this backfill is best-effort (the tx is usually
+                                                                        // already in the pool) and not our broadcast to report on.
+                                                                        warn!("Error on Assethub PendingBroadcast {awaiting_broadcast_id:?}: {error:?}");
+                                                                    }
+                                                                }
+                                                            }
+                                                    }
+
```
MxmUrw: Hmmm, interesting. I see that this might help, but it also feels like we shouldn't do raw storage queries here... Maybe @marcellorigotti also has an opinion on this?

bouncer/tests/all_swaps.ts
```diff
@@ -53,6 +61,114 @@ export async function initiateSwap(
...
+type Destination = {
+  asset: Asset;
+};
+
+type SwapPair = {
+  source: Source;
+  destination: Destination;
+};
+
+function generateSwapPairs(allTestedAssets: Asset[], testWithAllPossiblePartners: Asset[]) {
```
msgmaxim: Nit: the parameters to this function aren't clear at all (I can see that we have some comment on the call side, but it probably belongs here).

engine/src/state_chain_observer/sc_observer.rs
j4m1ef0rd: The behaviour and storage read is copied from the above code for `BtcTxBroadcastRequest`.

j4m1ef0rd: I moved the commit to it's own [PR](https://github.com/chainflip-io/chainflip-backend/pull/6741). we can continue the review there.

j4m1ef0rd:
I made it seedable like before and removed the Bsc from the `testWithAllPossiblePartners` list. Made it a parameter so we can still test new assets against everything. But i'm not sure what assets that should be now. 
Its down to just 66 swaps now. All assets are covered. For each `testWithAllPossiblePartners` asset, it adds around another 125 swaps.

bouncer/tests/all_swaps.ts
```diff
@@ -11,9 +11,31 @@ import {
 } from 'shared/utils';
 import { TestContext } from 'shared/utils/test_context';
 import { manuallyAddTestToList, concurrentTest } from 'shared/utils/vitest';
-import { SwapContext } from 'shared/utils/swap_context';
 import { ChainflipIO, newChainflipIO } from 'shared/utils/chainflip_io';
 
+const GENERATE_SWAPS_SEED = 1;
```
msgmaxim: I think we can still randomly generate the seed and print it. It would still be an improvement over the previous solution (can replay the same seed on failure) and at the same time improve test coverage over multiple run.

bouncer/tests/all_swaps.ts
```diff
@@ -14,6 +14,29 @@ import { manuallyAddTestToList, concurrentTest } from 'shared/utils/vitest';
 import { SwapContext } from 'shared/utils/swap_context';
 import { ChainflipIO, newChainflipIO } from 'shared/utils/chainflip_io';
 
+const GENERATE_SWAPS_SEED = 1;
+
+// A tiny seedable PRNG (values in [0, 1)) so the sampled swap set is reproducible for a given seed.
+// Distribution quality doesn't matter here — it only decides which swaps to sample.
```
msgmaxim: Might as well use a proper (similar loc) rng algorithm, e.g. https://github.com/cprosche/mulberry32. I know it probably doesn't matter in this case, but this code might get reused for something else in the future.

MxmUrw:
Thanks for picking this up btw, @j4m1ef0rd. Everything looks good to me, don't have too much of an opinion re. proper rng algorithm, but also why not?

The `testWithAllPossiblePartners` is supposed to be the chain(s) that we haven't released yet (so BSC), but also yeah it has already been subjected to the old `all_swaps` test for quite some time now so should be tested enough.

From my side I think we can merge this!

bouncer/tests/all_swaps.ts
```diff
@@ -78,43 +226,35 @@ export function testAllSwaps(timeoutPerSwap: number) {
...
+  );
+  for (const { source, destination } of pairs) {
+    const testFunction = source.trigger === 'DepositChannel' ? testSwap : testVaultSwap;
+    appendSwap(source.asset, destination.asset, testFunction, false);
+
+    // also do ccm version of the same swap if destination supports it
+    if (ccmSupportedChains.includes(chainFromAsset(destination.asset))) {
+      // bitcoin vault swaps don't support ccm, so we use ArbEth instead
+      const sourceAsset =
+        source.asset === 'Btc' && source.trigger === 'VaultSwap' ? 'ArbEth' : source.asset;
```
msgmaxim: As discussed, this may result in duplicate ArbEth swaps (duplication is done prior to adding these).

MxmUrw: I *think* this is supposed to be solved by this commit in the other PR: https://github.com/chainflip-io/chainflip-backend/pull/6741/changes/7fafc7e0a22c1dee4f1dac854e30f0843a4ed469

j4m1ef0rd: oops, it ended up in the wrong PR, sorry. ill move it.

bouncer/tests/all_swaps.ts
```diff
@@ -205,7 +227,8 @@ function generateSwapPairs(
     );
   }
 
-  return uniquePairs;
+  // Shuffle
```
msgmaxim, line 231: Why do we need to shuffle again? If this is intentionall, let's add a comment to say why (the comment right now is not very useful).

msgmaxim (approved):

bouncer/tests/all_swaps.ts
j4m1ef0rd: because all the ccm swaps are at the end, so i want to spread them out.

PR #6517, opened 2026-04-07
refactor: type safe swap state and swapping pallet cleanup

j4m1ef0rd:
# Pull Request

Closes: PRO-2809

Note: This PR is merged into the feature branch [feat/support-for-non-usdc-swaps](https://github.com/chainflip-io/chainflip-backend/tree/feat/support-for-non-usdc-swaps)

## Summary

Adds type safety to the swapping execution logic (swap state). No storage changes or API changes. Just internal refactoring of the swapping pallet. 
Changed the data flow of the swapping execution to take the Vec of SwapState's and return the new swap state of the next stages type, instead of mutating state. This means SwapState is not clone-able, stopping many mistakes around duplicating swaps or using the wrong swap state.
Fees can't be taken twice or at the wrong time. Cleaned up the updating of the swap state after each step, It's all in one location now. removed more usdc specific code. 
Split up the swapping pallet into smaller files. I checked the moved files to make sure the AI didn't make any unexpected changes.

## Behaviour changes:
- `get_scheduled_swap_legs` RPC now runs the same logic as an actual batch swap instead of possible partial execution (bug).
- Ignoring price protection for `get_scheduled_swap_legs` swap simulation.
- Intermediate is None for all single leg swaps. Instead of being set to USDC.
- Added sort order to the swap groups. So sequential swap events may of changed order.
- Fixed rounding dust bug in swapping group. Now the last swap in the group will get the possible spare 1 atomic unit instead of it going missing.

---

## *Checklist

* The PR is complete:
  * Tests:


state-chain/pallets/cf-swapping/src/fees.rs
```diff
@@ -0,0 +1,134 @@
...
+		if input_amount.is_zero() {
+			return FeeTaken { remaining_amount: 0, fee: 0 };
+		}
+		if self.fee_and_accumulated.is_empty() {
+			return FeeTaken { remaining_amount: input_amount, fee: 0 };
+		}
+		// Sanity check: it should already not be possible to open a channel with broker fees
+		// this high, but if the total broker fee would exceed 100% we charge no broker fee
+		// instead (for simplicity):
+		let total_fee_bps = self
```
msgmaxim: Seem a bit wasteful checking this on every chunk? Could we move this into the BrokerFeesTracker's constructor instead (and in case of violation replace `fee_and_accumulated` with Default::default, for example)?

state-chain/pallets/cf-swapping/src/fees.rs
```diff
@@ -0,0 +1,134 @@
...
+
+		self.accumulated_fee.saturating_accrue(fee_taken);
+
+		FeeTaken { remaining_amount: input_amount.saturating_sub(fee_taken), fee: fee_taken }
+	}
+}
+
+#[derive(DebugNoBound, PartialEq, Eq, Encode, Decode, DecodeWithMemTracking, TypeInfo)]
+pub struct BrokerFeesTracker<AccountId: core::fmt::Debug + Ord> {
+	pub fee_and_accumulated: BTreeMap<Beneficiary<AccountId>, AssetAmount>,
```
msgmaxim: Worth adding a comment describing this (similar to NetworkFeeTracker)? Otherwise the name it not very clear.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -132,44 +133,173 @@ pub struct SwapRefundParameters {
...
-#[derive(CloneNoBound, DebugNoBound)]
-pub struct SwapState<T: Config> {
+#[derive(DebugNoBound)]
+pub struct SwapState<T: Config, Stage: Debug> {
 	swap: Swap<T>,
+	pub stage: Stage,
+}
+
+#[derive(DebugNoBound)]
+pub struct Stage1 {
```
msgmaxim: Should we give each stage a descriptive name instead?

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -0,0 +1,394 @@
...
+	pub from: Asset,
+	pub to: Asset,
+}
+
+/// The 2 stages that are handed to the group swap logic (to & from intermediate) implement this
+/// trait to allow them to be grouped and executed.
+pub(crate) trait GroupSwapState<T: Config> {
+	type OutputState;
+
+	/// The input amount that is being swapped for this single leg swap.
```
msgmaxim: I think this would be more accurate?
```suggestion
	/// The input amount that is being swapped for the next swap let.
```

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -0,0 +1,394 @@
...
+	fn swap_amount(&self) -> AssetAmount;
+	/// Returns the "to" and "from" assets as a swap pair to allow grouping of swaps. Returns None
+	/// if the asset does not need to be swapped for this leg (e.g. if the input or output asset is
+	/// already the intermediate asset).
+	fn swap_group(&self) -> Option<SwapGroupPair>;
+	/// Updates the swap state with the output amount of a swap for this single leg.
+	fn update_swap_result(self, amount: AssetAmount) -> Self::OutputState;
+	/// Used when the asset does not need to be swapped for this leg, just passes through the input
+	/// amount as the output amount.
+	fn update_no_swap(self) -> Self::OutputState;
```
msgmaxim: Nit, but maybe `advance_no_swap` would be more accurate? Ditto for update_swap_result. I imagine we used to call this "update" because in the past we needed to update the existing struct by setting output amount, but that's no longer the case.

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -0,0 +1,394 @@
...
+		self.swap.refund_params.as_ref()
+	}
+
+	/// Consume the swap state and return the inner swap.
+	pub(crate) fn into_swap(self) -> Swap<T> {
+		self.swap
+	}
+
+	#[cfg(test)]
+	pub fn new_test_state(swap: Swap<T>, stage: Stage) -> Self {
```
msgmaxim, line 268: IMO this could be moved closer to the tests.

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -0,0 +1,394 @@
...
+					.ok()
+					.flatten()
+					.map(|delta| delta.pessimistic_rounded_into())
+				} else {
+					None
+				};
+
+				Ok(SwapState {
+					swap: self.swap,
+					stage: Stage5 {
```
msgmaxim: Do we really need Stage5? IMO we could instead just use a new struct `SwapResult` or something like that. Otherwise it feels like we are leaking the implementation detail (the state machine) into the callers.

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -0,0 +1,394 @@
...
+
+impl<T: Config> SwapState<T, Stage4> {
+	/// The final step of the swapping process. Checks for price violations and then sets the oracle
+	/// delta values in the swap state.
+	#[expect(clippy::result_large_err)]
+	pub(crate) fn check_for_price_violation(
+		self,
+	) -> Result<SwapState<T, Stage5>, (Self, SwapFailureReason)> {
+		match Pallet::<T>::check_swap_price_violation(&self) {
+			Ok(oracle_delta_ex_fees) => {
```
msgmaxim, line 365: Could we split price violation check and the derivation of oracle delta? Then you could only compute oracle delta once you know the swap outcomes will be accepted. Otherwise if 1 out of 10 swaps fail, you still have to compute it for 9 swaps only to throw the result away.

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -0,0 +1,394 @@
...
+				})
+			},
+			Err(reason) => Err((self, reason)),
+		}
+	}
+}
+
+impl<T: Config> From<SwapState<T, Stage5>> for SwapState<T, Stage2> {
+	/// Downgrades a Stage5 swap state to a Stage2. Used in the 'get_scheduled_swaps' RPC's
+	fn from(state: SwapState<T, Stage5>) -> Self {
```
msgmaxim: Since this conversion is a bit odd, maybe we should give it a more distinctive name, e.g. `downgrade` (easy to misuse otherwise)?

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+			.into_iter()
+			// We only need the swaps at the stage after the first leg is complete. So we
+			// strip away the unused part of the state for the fully executed swaps.
+			.map(SwapState::from)
+			.chain(failed_swaps)
+			.filter_map(|state| {
+				let swap_request = SwapRequests::<T>::get(state.swap_request_id())
+					.expect("Swap request should exist");
+				let dca_state = match swap_request.state {
+					SwapRequestState::UserSwap { dca_state, .. } => Some(dca_state),
```
msgmaxim: Here we have `dca_state` without the need to unwrap it. Might as well read `remaining_chunks` and `chunk_interval` directly here insted of returning dca_state wrapped in Option.

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+						state.execute_at(),
+					))
+				} else {
+					None
+				}
+			})
+			.collect()
+	}
+
+	pub(crate) fn trigger_withdrawal(
```
msgmaxim: This seem to have nothing to do with "execution" (the name of this module).

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2737,95 +2963,96 @@ pub mod pallet {
...
+					.enumerate()
+					.map(|(index, swap)| {
+						if index == number_of_swaps - 1 {
+							// Give the dust to the last swap to ensure all output is assigned
+							let swap_output = bundle_output.saturating_sub(total_assigned_output);
+							swap.update_swap_result(swap_output)
+						} else {
+							let swap_output = if bundle_input > 0 {
+								multiply_by_rational_with_rounding(
+						swap.swap_amount(),
```
msgmaxim: indentation

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+			Some(oracle_price) if oracle_price.stale => Err(SwapFailureReason::OraclePriceStale),
+			Some(oracle_price) => Ok(Price::sell_price(input.amount.into(), output.amount.into())
+				.map(|execution_price| {
+					execution_price.hundredth_bps_difference_from(&oracle_price.price)
+				})),
+			None => Ok(None), // Price unavailable - skip the check
+		}
+	}
+
+	/// Enforce price protections. Must be called after the final output has been set.
```
msgmaxim: Would be nice to say what it returns in the comment.

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+					} else {
+						log_or_panic!("Swap request {} should exist", swap.swap_request_id());
+						swap.no_broker_fee()
+					}
+				})
+			})
+			.collect()
+	}
+
+	pub(crate) fn start_broker_fee_swaps_or_credit(
```
msgmaxim: This does not need to be public.

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+			swap_request_id,
+			reason: SwapRequestCompletionReason::Aborted,
+			broker_fee_swaps,
+		});
+
+		Some(swap_progress)
+	}
+}
+
+impl<T: Config> AssetConverter for Pallet<T> {
```
msgmaxim, line 1521: How do you decide whether to put something in `execution.rs` vs in `impls.rs`?

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+						}
+					}
+				})
+			} else {
+				log_or_panic!("Swap request {} not found for rescheduling", swap.swap_request_id);
+			}
+		});
+	}
+
+	pub fn estimate_usdc_price_using_simulated_swap_or_fallback(asset: Asset, side: Side) -> Price {
```
msgmaxim: This also doesn't need to be pub (I didn't check all of them there seem to be others too).

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+				// Return USDC / Asset
+				.and_then(|estimation_output| {
+					Price::from_amounts(estimation_output.into(), estimated_input.into())
+				})
+				.unwrap_or_else(|| utilities::hard_coded_price_for_asset(asset))
+			},
+		}
+	}
+
+	pub(crate) fn egress_for_swap(
```
msgmaxim: Does not need pub

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+					asset,
+					amount,
+					reason: err.into(),
+				}),
+			},
+		}
+	}
+
+	pub fn assemble_and_validate_broker_fees(
+		broker_id: T::AccountId,
```
msgmaxim: This is used in other places, so IMO this should be moved to the parent module instead of having it here as pub.

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+				NetworkFeeForAsset::<T>::get(input_asset).unwrap_or(default_fee.rate),
+				NetworkFeeForAsset::<T>::get(output_asset).unwrap_or(default_fee.rate),
+				default_fee.minimum,
+			)
+		};
+
+		FeeRateAndMinimum { rate: input_asset_fee.max(output_asset_fee), minimum: usdc_minimum }
+	}
+
+	pub fn get_network_fee_rate_for_swap(
```
msgmaxim:
```suggestion
fn get_network_fee_rate_for_swap(
```

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -0,0 +1,1696 @@
...
+		Some(DefaultOraclePriceSlippageProtection::<T>::get(AssetPair::new(asset, STABLE_ASSET)?))
+	}
+
+	/// Returns the default price protection to apply to a one or two-leg swap.
+	///
+	/// Returns `None` if no oracle price is available for any leg of the swap (no
+	/// meaningful protection can be applied). For two-leg swaps where only one leg
+	/// has an oracle, the other leg contributes zero to the total limit so that
+	/// the oracle-priced leg is still protected.
+	pub(crate) fn get_default_oracle_price_protection(
```
msgmaxim:
```suggestion
  fn get_default_oracle_price_protection(
```

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -132,44 +133,173 @@ pub struct SwapRefundParameters {
...
 
-#[derive(CloneNoBound, DebugNoBound)]
-pub struct SwapState<T: Config> {
+#[derive(DebugNoBound)]
+pub struct SwapState<T: Config, Stage: Debug> {
 	swap: Swap<T>,
+	pub stage: Stage,
+}
+
+#[derive(DebugNoBound)]
```
msgmaxim: State does not have a generic parameter so you should be able to use regular `Debug` here and in other stages.

state-chain/pallets/cf-swapping/src/dca_state.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2025 Chainflip Labs GmbH
```
msgmaxim, line 1: Nit: I would just call the module dca_state.rs to be more descriptive (don't think it is likely that we will put anything else here).

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: I gave it a try but i didn't like it. `Stage1AfterNetworkFee` doesn't add much info and doesn't let you know that this is one of the stages that impls the GroupSwapState for example. Makes it more confusing IMO.

state-chain/pallets/cf-swapping/src/swap_state.rs
j4m1ef0rd: I am not able to move it because `swap` is private by design.

state-chain/pallets/cf-swapping/src/execution.rs
j4m1ef0rd: Was not too sure about this one. But because it uses swap simulation, I put it in execution. 🤷‍♂️

state-chain/pallets/cf-swapping/src/execution.rs
j4m1ef0rd: I am trying to keep things out of the lib file as it is still quite big, but I do think it should be moved to the fees file.

state-chain/pallets/cf-swapping/src/execution.rs
j4m1ef0rd: I also moved this to the fees file and kept it pub.

state-chain/pallets/cf-swapping/src/execution.rs
```diff
@@ -429,7 +434,15 @@ impl<T: Config> Pallet<T> {
 				violating_swaps,
 				non_violating_swaps: non_violating_swaps
 					.into_iter()
-					.map(|state| state.into_swap())
+					.map(|successful_swap| {
+						(
+							successful_swap.swap_id,
+							swaps
```
msgmaxim, line 419: Maybe this should be done on the caller side instead? Otherwise it is strage that you had to modify error type only because the input into this function changed.

state-chain/pallets/cf-swapping/src/swap_state.rs
```diff
@@ -105,17 +122,17 @@ pub(crate) struct SwapGroupPair {
...
-	/// The input amount that is being swapped for this single leg swap.
+	/// The input amount that is being swapped for the next leg.
 	fn swap_amount(&self) -> AssetAmount;
 	/// Returns the "to" and "from" assets as a swap pair to allow grouping of swaps. Returns None
 	/// if the asset does not need to be swapped for this leg (e.g. if the input or output asset is
 	/// already the intermediate asset).
 	fn swap_group(&self) -> Option<SwapGroupPair>;
 	/// Updates the swap state with the output amount of a swap for this single leg.
-	fn update_swap_result(self, amount: AssetAmount) -> Self::OutputState;
+	fn advance_swap_result(self, amount: AssetAmount) -> Self::OutputState;
```
msgmaxim, line 132: How about `advance_with_swap_result`? We are not advancing the result here, but rather the swap state.

state-chain/pallets/cf-swapping/src/execution.rs
msgmaxim: And execution uses fees, but we don't put execution code into the fees module.

state-chain/pallets/cf-swapping/src/execution.rs
msgmaxim: This one is only used inside `execution.rs`, so IMO it is better to keep it close to that module and private. IMO we should only expose a function to a wider scope if it is actually needed there (and when it is needed you could move it to a different module).

msgmaxim (approved):

dandanlen (commented):
We can discuss tomorrow. While I appreciate the idea and the initiative, I don't feel like formulating the swap execution pipeline like like this really has many benefits. 

We're not guarding against any major/likely classes of bug, but rather we're creating a strongly-encoded version of the existing sub-optimal structure: Previously, we had a big struct with a bunch of `Option_<..>` members. We sort of still have this: we've basically just taken every `struct S { .., x: Option<X> }` and then created a `struct S { .. }; struct SWithX { .., x: X }`. We haven't added any composability or flexibility, quite the opposite. So any future changes will be more difficult. 

Some code smells:
- We're still accumulating a bunch of somewhat misleading placeholder values like `intermediate`. 
- We have a method `advance_no_swap` as a workaround for the fact that some swaps don't have all the stages and we want to skip some.
- We need an `into_stage_2` method to 'reverse' the state machine for one of the RPCs.
- `SwapGroupPair` and `swap_group`: used in only two places and in one of those we override one of the values...


---
Separately (could have been a separate PR btw):

Breaking things into modules is a good idea, although it could be more considered. Some of the naming/decisions seem a bit arbitrary (eg. `impls` is vague unless is contains *all* impls, utilities is also very vague - why would something go in utilities instead of the main lib.rs? execution contains AssetConverter as maxim mentioned.. etc). I think we can do this as a separate PR against main.

j4m1ef0rd:
> I don't feel like formulating the swap execution pipeline like like this really has many benefits

- Clear step by step flow makes it much easier to understand the execution flow for outsiders.
- Type safe, so we can't deduct fees at the wrong time or deduct twice. This will stop future bugs like the ones we seen earlier this year with the rescheduled/refunded swaps having no fees deducted.
- Laying the groundwork for separation of execution and swap state. My next PR expands this.
- No unwrapping or asserting that "should exist at this point" during execution.
- No cloning of swap state. This already found a partial execution bug in the scheduled swaps rpc.

> We haven't added any composability or flexibility, quite the opposite. So any future changes will be more difficult.

I argue that it is much easier to work with now and made the future changes easier. I know this because I have already made the changes in a branch. Support for non-usdc routes and unlimited number of swap legs. All while keeping it type safe and organised.

 > Some code smells:
 
 Most (if not all) of these are addressed in my next PR.
 
 > I think we can do this as a separate PR against main.
 
 I am fine with rearranging the modules. But because the base PR and this one have a lot of changes from before I moved the files, it will create a lot of conflicts when trying to rebase if we make a seperate PR to rearrange the files on main. It can be done, but I'm not sure its worth the time.
 
 I wanted to merge this PR into the base branch first and then I can put up my next PR as draft pointing at the base branch.

PR #6518, opened 2026-04-08
refactor: keep track of balance changes in vector during swap processing

MxmUrw:
# Pull Request

Closes: PRO-xxxx

## Summary

This PR sketches how keeping track of fee deductions and swap legs would look like if done in a more uniform (but not strictly typed way).

The main part are the changes to `SwapState`:
 - instead of recording the updates to the swap state in dedicated fields (such as `output_amount_after_fees`, `output_amount_before_fees`, ...), all changes to the swap state are treated uniformly as "balance changes". The swap has a `current_balance: AssetAndAmount`, which can be modified in two ways: either by processing a swap leg (which changes both the asset and the amount) or by deducting fees (which changes just the amount). All balance changes are recorded chronologically in a vector `balance_changes: Vec<BalanceChange>`.
 - all required accounting information is extractable from the balance_changes vector. For example, for the purposes of LPP, we simply have to look at all balance changes that have been swap legs, and look at their `input: AssetAmount` and `output: AssetAmount`.

This reduces the number of places where the assumption of every swap having two legs (to-stable and from-stable) is made. Apart from being a step towards non-USDC routes, it should also allow us remove special-case logic that was required to deal with swaps that don't have a to-stable leg (because they already start with USDC).

**NOTE**: this is a WIP, I haven't come around to remove all old fields from `SwapState`. The next step would be to adjust `simulate_swap()` which I think can also be simplified in the new approach. Then the remaining fields can be removed.

## API Changes

### RPCS

### Extrinsics & Events

---

* The PR is complete:
  * Tests:


j4m1ef0rd:
Your solution is better than what we had, its flexible and robust, but It still offers no type safety. For example, we could deduct fees twice, and your swap state would handle it fine. It would show the correct fees deducted and everything. But that's not what we want, we don't want to ever deduct the same fee twice. 
Having a Vec of balance changes is like making everything an option. It means you have to iterate, filter and handle the `None` case every time you want to get a value. Having a current_amount is good because it avoids doing this for the output amount, I like the idea of having a nice tracked asset and amount that the swap is currently at. It makes using the wrong amount a lot harder to do. But then it's like we are just avoiding using the Vec by having an explicit field. Might as well just have fields for all of the values, and we are back to what we had before.

PR #6525, opened 2026-04-10
feat: drop vote tasks for closed elections

marcellorigotti:
# Pull Request

Closes: PRO-2780

## Summary

The voter implementation uses the `RetrierClient` which retries any voting task up to 3 times on failure. However, retrying vote tasks is unnecessary and potentially harmful — slow RPC calls can cause timeouts, but since new elections are processed every block, a failed vote for an already-resolved election is wasted work.

This PR removes vote tasks from the `FutureMap` when their corresponding election no longer appears in the current elections set. When a task is removed from the `FutureMap`, its future — which holds the `oneshot::Receiver` returned by `request_with_limit` is dropped. This causes the `oneshot::Sender` on the retrier side to detect that the receiver is closed (`response_sender.is_closed()`) which makes the retrier discard the request instead of scheduling another retry attempt.

Note, this doesn't completely abort the task, in-flight task will still running, this will only prevent the task to be retried in case it fails/times out.

* The PR is complete:
  * Tests:

### DIFF
--- a/utilities/src/with_std/future_map.rs
+++ b/utilities/src/with_std/future_map.rs
@@ -16,6 +16,7 @@
 
 use futures::{Future, Stream};
 use std::{collections::BTreeSet, iter::IntoIterator};
+use tracing::debug;
 
 #[pin_project::pin_project]
 struct Wrapper<Key, Fut> {
@@ -86,7 +87,8 @@ impl<Key: Ord + Copy, Fut: Future + Unpin> FutureMap<Key, Fut> {
 
 		for future in futures {
 			if predicate(&future.key) {
-				assert!(self.keys.remove(&future.key));
+				let was_present = self.keys.remove(&future.key);
+				debug_assert!(was_present);
 				removed_keys.push(future.key);
 			} else {
 				self.futures.push(future);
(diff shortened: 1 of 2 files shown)


dandanlen (commented):
This looks sensible - did we forget this or is it waiting for something?

marcellorigotti:
> This looks sensible - did we forget this or is it waiting for something?

I think we just forgot about it, PR is ready, just missing a review

utilities/src/with_std/future_map.rs
```diff
@@ -77,6 +77,25 @@ impl<Key: Ord + Copy, Fut: Future + Unpin> FutureMap<Key, Fut> {
...
+	pub fn remove_where<Predicate>(&mut self, mut predicate: Predicate) -> Vec<Key>
+	where
+		Predicate: FnMut(&Key) -> bool,
+	{
+		let mut removed_keys = Vec::new();
+		let futures = std::mem::take(&mut self.futures).into_iter();
+
+		for future in futures {
+			if predicate(&future.key) {
+				assert!(self.keys.remove(&future.key));
```
MxmUrw: I think it's worth changing this into a `debug_assert!()`: even if this property doesn't hold, it isn't worth panicking on mainnet because of it.

MxmUrw (approved):

PR #6526, opened 2026-04-10
fix: add exceptions to cargo audit

marcellorigotti:
# Pull Request

Closes: PRO-xxxx

## Summary

Added exceptions to our cargo audit.
All the new vulnerabilities are from a dependency we don't control directly (`wasmtime`) which is used in substrate.


MxmUrw (approved):
These don't really affect us due to requiring a malicious guest runtime for being triggered.

Additionally the two critical ones have restricted fallout due to respectively:
 - https://rustsec.org/advisories/RUSTSEC-2026-0096.html is restricted to aarch64 (does not affect standard intel x86 arch, but *does* affect apple M* processors)
 - https://rustsec.org/advisories/RUSTSEC-2026-0095.html is restricted to the Winch compiler of wasmtime. AFAIK our dependencies use the standard cranelift compiler instead.

PR #6529, opened 2026-04-14
chore: add migration, deploy eth channels and refund

marcellorigotti:
# Pull Request

Closes: PRO-xxxx

## Summary

Deploy Eth channels to fetch stuck deposits, credit funds to LP.
The migration is designed such that the channel get then recycled and reused.

---

* The PR is complete:

### DIFF
--- a/state-chain/runtime/src/migrations/housekeeping/deploy_stuck_eth_channels.rs
+++ b/state-chain/runtime/src/migrations/housekeeping/deploy_stuck_eth_channels.rs
@@ -1,0 +1,205 @@
+use crate::*;
+use cf_chains::{
+	assets::eth::Asset as EthAsset, deposit_channel::DepositChannel, evm::DeploymentStatus,
+};
+use cf_primitives::Asset;
+use cf_traits::BalanceApi;
+use frame_support::{traits::OnRuntimeUpgrade, weights::Weight};
+use pallet_cf_ingress_egress::{
+	BoostStatus, ChannelAction, DepositChannelDetails, DepositChannelLookup,
+	DepositChannelRecycleBlocks, FetchOrTransfer, ProcessedUpTo, ScheduledEgressFetchOrTransfer,
+};
+use sp_core::{crypto::Ss58Codec, H160};
+use sp_runtime::AccountId32;
+#[cfg(feature = "try-runtime")]
+use sp_runtime::DispatchError;
+use sp_std::collections::btree_set::BTreeSet;
+
+// Channels that need contract deployment + balance credit.
+// (channel_id, address, amount in wei)
+const CHANNELS_TO_DEPLOY: [(u64, [u8; 20], u128); 2] = [
+	// 10.319838215944994885 ETH
+	(
+		7584,
+		hex_literal::hex!("9d6ca7cd47c9b69173691971545446625e605ab7"),
+		10_319_838_215_944_994_885,
+	),
+	// 9.029838215944994885 ETH
+	(
+		7585,
+		hex_literal::hex!("5958f112ab25c46ede2ef9811410b52ea1388d90"),
+		9_029_838_215_944_994_885,
+	),
+];
+
+// Channels already deployed and fetched — only need balance credit.
+// (address, amount in wei)
+const CHANNELS_TO_CREDIT: [([u8; 20], u128); 2] = [
+	// 10.689838215944994885 ETH
+	(hex_literal::hex!("96d29eb80309491f0b997cc11aafa5bafb3be66a"), 10_689_838_215_944_994_885),
+	// 9.199838215944994885 ETH
+	(hex_literal::hex!("8354ac55fab3d4782f621acb85e90705407af613"), 9_199_838_215_944_994_885),
+];
+
+pub struct Migration;
+
+impl OnRuntimeUpgrade for Migration {
+	fn on_runtime_upgrade() -> Weight {
+		let owner = match AccountId32::from_ss58check(
+			"cFLW4PhasdivcJKuA2BGw9Y9dz7EFwks82K8Z6U3MfCk8WcNW",
+		) {
+			Ok(account) => account,
+			Err(err) => {
+				log::info!("Failed to decode the AccountId {err:?}");
+				return Weight::zero();
+			},
+		};
+
+		let preallocated: BTreeSet<H160> = pallet_cf_ingress_egress::PreallocatedChannels::<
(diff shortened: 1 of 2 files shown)


dandanlen:
I want to make some small changes: 

- We should use the open_channel method to open an LP channel and rely on witnessing, instead of writing straight to storage. (this should work, right?) This way we trigger the expected events etc, which will minimize disruption/surprise in the product. 
- The pre-upgrade check is not being triggered. I'll add it to the housekeeping migration.

Will run the try-runtime checks locally to ensure it passes.

albert-llimos:
> We should use the open_channel method to open an LP channel and rely on witnessing

Maybe I am misunderstanding, but if you mean "rely on witnessing" as in expecting the witnessing to see the ingress once the deposit is open, that won't work. Witnessing won't see any ingress.

MxmUrw:
Yeah we discussed with @marcellorigotti: witnessing only registers balance changes that happen once the deposit channel is opened on-chain. We *could* open a regular lp deposit channel and then deposit a small amount of ETH into it to trigger deployment + fetch though. The fetch will fetch all the ETH but it won't credit. So that would have to be done separately anyways.

**EDIT**: I think we should do it this way - the more code runs through the usual paths the better.

albert-llimos:
In the past we have handled this by triggering a fetch and crediting separately. I assume what you are trying to do it this way so it runs through the usual paths. However, as you are describing, we anyway need to do the crediting separately as there is no way to witness the ingress unless the new election system can do that. If we anyway have to do the crediting separately, I'm not sure we gain much by creating the LP deposit channel and making a small deposit so it all gets fetched (compared to what we had done in the past). Not a strong opinion, both ways would work.

MxmUrw:
> In the past we have handled this by triggering a fetch and crediting separately.

If there's a nice way to trigger the fetching logic for an undeployed channel I'm all for it!

dandanlen:
I discussed this with Marcello earlier. I was under the impression that we could trigger the witnessing of the deposit 'organically' but it appears that's not possible. 

Given that we can't do that, i'm not sure we gain much by opening a 'real' channel. We need it open so that the fetch can be constructed, otherwise it's just unnecessary. ~~Another approach could be dispatch the fetch directly (using `broadcast(ApiCall::all_batch(..))`)~~ not very practical, I looked. 

(also can we not force-witness specific blocks now, with the elections?)

MxmUrw:
> (also can we not force-witness specific blocks now, with the elections?)

Yes we can, true! It should use the current deposit channels and tell the engines to witness that specific block with these channels. But it will (re-)witness *all* deposits in that block that go into one of our currently open channels. I think it's worth adding an optional tx_hash parameter to the governance elections, so that we can restrict the election to witness only a certain tx that we know was missed.

As things currently are, yeah this is an option, but we'd have to check that there are no other deposits in that block going into our channels.

marcellorigotti:
> > (also can we not force-witness specific blocks now, with the elections?)
> 
> Yes we can, true! It should use the current deposit channels and tell the engines to witness that specific block with these channels. But it will (re-)witness _all_ deposits in that block that go into one of our currently open channels. I think it's worth adding an optional tx_hash parameter to the governance elections, so that we can restrict the election to witness only a certain tx that we know was missed.
> 
> As things currently are, yeah this is an option, but we'd have to check that there are no other deposits in that block going into our channels.

As far as I remember we pass the deposit channel/s we want to re-witness so there shouldn't be such risk! 
The problem I am thinking of is that the deposit channel doesn't exists hence I am not really sure what happens in the downstream logic.

I think when we dispatch the witness call (`process_full_witness`) that simply fails since the channel doesn't exists -> it doesn't know what to do with the witnessed ingress, no channel means no action.

dandanlen:
Yes I think we would need a combination of this migration (open the channel as an LP deposit), then re-witness the failed deposits.

MxmUrw:
> As far as I remember we pass the deposit channel/s we want to re-witness so there shouldn't be such risk!

You are right!

> Yes I think we would need a combination of this migration (open the channel as an LP deposit), then re-witness the failed deposits.

I agree.

state-chain/runtime/src/migrations/housekeeping/deploy_stuck_eth_channels.rs
```diff
@@ -0,0 +1,205 @@
...
+				is_marked_for_rejection: false,
+			};
+
+			DepositChannelLookup::<Runtime, EthereumInstance>::insert(address, details);
+
+			ScheduledEgressFetchOrTransfer::<Runtime, EthereumInstance>::append(
+				FetchOrTransfer::Fetch {
+					asset: EthAsset::Eth,
+					deposit_address: address,
+					deposit_fetch_id: None,
```
MxmUrw, line 108: This one is correct because of the following code in `do_egress_scheduled_fetch_transfer()`:
```rust
if can_fetch {
	deposit_fetch_id.replace(
		details.deposit_channel.fetch_id(),
	);
	details
		.deposit_channel
		.state
		.on_fetch_scheduled();
}
```
which ignores the current fetch_id and replaces it with a new generated from the deposit channel state, correct?

**EDIT**: that's also how we inject fetches during the normal deposit processing in `process_full_witness_deposit_inner`.

MxmUrw (commented):
Confirming that all addresses, amounts and channel ids are correct. Original report is here: https://discord.com/channels/775961728608895008/1491985991236583556/1491985996055580682

This is the summary:
```
addresses: 
0x96D29EB80309491F0B997cc11AAfA5bAFB3BE66a     10.69 ETH fetched => 10.689838215944994885 ETH
0x9d6CA7Cd47C9B69173691971545446625E605Ab7   10.32 ETH NOT fetched => balance: 10.319838215944994885 ETH / channel id: 7584
0x8354aC55FAb3d4782F621aCB85E90705407Af613      9.2 ETH fetched => 9.199838215944994885 ETH
0x5958f112Ab25c46eDE2eF9811410B52EA1388D90       9.03 ETH NOT fetched => balance: 9.029838215944994885 ETH / channel id: 7585
```

state-chain/runtime/src/migrations/housekeeping/deploy_stuck_eth_channels.rs
dandanlen: Yeah it's a bit awkward: the implicit assumption is that the deposit_fetch_id is always `None` when newly scheduled.

dandanlen (approved):

PR #6531, opened 2026-04-15
fix: use full swap logic for get scheduled swaps rpc

j4m1ef0rd:
# Pull Request

## Summary

Fix for a bug in the `cf_scheduled_swaps` rpc. If a single swap would fail due to low liquidity, it would stop the execution loop and return error, leaving any other swaps not-executed. The non-executed swaps would use the pool price fallback.

- Fixed the `get_scheduled_swap_legs` partial execution bug
- Using full swap execution logic instead of just the `swap_into_stable_taking_fees` function. So now it will do the failed swap split off logic just like the real execution.
- Removed the refund params so that the price protection logic will be ignored. We fallback to pool price anyway, so might as well get the best swap estimation we can.

Note:
- The `get_scheduled_swap_legs` is refactored a lot in #6476. That PR has all of the above changes, but it will not get merged for a while, so I put up this fix in the meantime.
- The pool price fallback does not take the network and broker fees into account. This is existing behaviour and is out of scope for this PR.

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -330,6 +330,18 @@ impl<T: Config> Swap<T> {
 	) -> Self {
 		Self { swap_id, swap_request_id, from, to, input_amount, refund_params, execute_at }
 	}
+
+	pub fn without_refund_params(self) -> Self {
+		Self {
+			swap_id: self.swap_id,
+			swap_request_id: self.swap_request_id,
+			from: self.from,
+			to: self.to,
+			input_amount: self.input_amount,
+			refund_params: None,
+			execute_at: self.execute_at,
+		}
+	}
 }
 
 #[derive(Debug, Encode, Decode, DecodeWithMemTracking, TypeInfo, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


dandanlen:
Trying to understand what was fixed here. Is this roughly correct? 

If there was a batch of swaps and one of them failed (eg. due to price limits) then any remaining swaps would fail, and we would fall back to using the pool price to estimate the second swap leg? 

Is this a behavioural change if we now also show the second swap legs?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -330,6 +330,18 @@ impl<T: Config> Swap<T> {
 	) -> Self {
 		Self { swap_id, swap_request_id, from, to, input_amount, refund_params, execute_at }
 	}
+
+	pub fn without_refund_params(self) -> Self {
+		Self {
```
msgmaxim, line 335: I believe you can do something like:

```rust
Self {
  refund_params: None,
  ..self
}
```

msgmaxim (approved):

j4m1ef0rd:
> If there was a batch of swaps and one of them failed (eg. due to price limits) then any remaining swaps would fail, and we would fall back to using the pool price to estimate the second swap leg?

Pretty much. Not due to price violation (because that was not being checked in the previous design), due to `swap_single_leg` failing with `InsufficientLiquidity` error. Seems like a real edge case, but was worth fixing with the other improvements.

> Is this a behavioural change if we now also show the second swap legs?

No, we always showed the second leg swaps, it was just that they were using the pool price fallback when they didnt need to sometimes.

dandanlen (approved):

PR #6532, opened 2026-04-15
fix: pre_dispatch signature check

dandanlen:
# Pull Request

This is a cherry pick from release plus added unit and integration tests.


state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -913,3 +913,304 @@ fn can_batch() {
...
+			);
+		});
+	}
+
+	/// Explicit regression test for the impersonation vector.
+	///
+	/// `SignatureData::Solana::signer_account()` copies the 32-byte signer directly into an
+	/// AccountId32, so the attacker picks any account's bytes. With the bug, pre_dispatch
+	/// accepted this without verifying the signature — so the call would dispatch as the
+	/// impersonated account. With the fix, the missing valid signature must fail closed.
```
MxmUrw: The comment describes a 32 byte to AccountId32 conversion, but the code uses a u64 account id. Understandable, but we should at least note in the comment that we don't faithfully represent the actual attack vector on a live network. I do suppose the representation is sufficient though.

state-chain/pallets/cf-environment/src/tests.rs
```diff
@@ -913,3 +913,304 @@ fn can_batch() {
...
+				Box::new(|| {
+					let (c, _) = valid_call(0);
+					c
+				}),
+			),
+			(
+				"wrong call variant",
+				Box::new(|| Call::<Test>::batch { calls: BoundedVec::default() }),
+			),
+		];
```
MxmUrw, line 1092: This is a nice test, we could even merge the other smaller tests from above into this one, to keep things streamlined?

state-chain/cf-integration-tests/src/unsigned_validation.rs
```diff
@@ -0,0 +1,138 @@
...
+	PalletExecutionOrder::on_finalize(block_number);
+	result
+}
+
+/// The canonical CVE scenario: a malicious block author embeds a forged
+/// `non_native_signed_call` whose Solana signer aliases an existing on-chain account
+/// (here, the genesis validator ALICE) with a garbage signature. With the original bug,
+/// `Executive::apply_extrinsic` would have accepted it and dispatched the inner call as
+/// ALICE. With the fix, it must be rejected with `BadProof`, leaving ALICE's nonce
+/// untouched.
```
MxmUrw, line 94: Just to double-check, did you run this against the old implementation and it failed?

state-chain/runtime/src/test_runner.rs
```diff
@@ -16,3 +16,184 @@
...
+/// validation that Substrate's executive runs for unsigned extrinsics during *block import*,
+/// a malicious block author could include a forged `non_native_signed_call` whose
+/// `SignatureData::Solana { signer: <any 32 bytes> }` impersonates an arbitrary on-chain
+/// account (the 32 bytes are copied straight into an `AccountId32`). The fix routes both
+/// `validate_unsigned` and `pre_dispatch` through `validate_unsigned_call`, which performs
+/// the full signature check.
+///
+/// These tests exercise the *real* `Runtime` rather than a per-pallet mock, so they catch
+/// regressions in how `Runtime` wires the pallet's `Config` (e.g. `ChainflipNetworkName`,
+/// `GetTransactionPayments`, etc.).
```
MxmUrw: There's already the integration test which checks against (hopefully) the exact block import pipeline. Is there concrete value in having this test as well? Also, if we were to keep this, it feels like it shouldn't go into the top-level `test_runner` file.

dandanlen: No sorry, this one is pointless, I meant to delete it. 

In fact i'm not sure there's even much value in the unit tests, they are almost too specific... but it seems better to be safe.

state-chain/cf-integration-tests/src/unsigned_validation.rs
dandanlen: Yep

state-chain/pallets/cf-environment/src/tests.rs
dandanlen: Yeah there's a bit of overlap... will roll them up.

state-chain/pallets/cf-environment/src/tests.rs
dandanlen: I removed this test in the end, the case is more accurately covered by the integration test.

MxmUrw (approved):

PR #6536, opened 2026-04-17
feat: no passive collateral

msgmaxim:
# Pull Request

Closes: PRO-2792

## Summary

- Removes passive collateral from loan accounts
- Any existing collateral is auto-supplied to lending pools in a migration
- Safe mode: removed `add_collateral` and `remove_collateral`

(Opened this on top of feat/lending-pools-can-boost-deposits to minimise conflicts)

### Extrinsics & Events

(FYI @GabrielBuragev)

- `CollateralAdded` and `CollateralRemoved` events have been removed
- `action_type` field added to `LendingFundsAdded` event
- `add_collateral` and `remove_collateral` extrinsics have been removed
- 
- removed `extra_collateral` field from `request_loan` and `expand_loan` extrinsics

---

## *Checklist

* The PR is complete:
  * Tests:


state-chain/pallets/cf-lending-pools/src/migrations/collateral_to_supply.rs
```diff
@@ -0,0 +1,179 @@
...
+mod old {
+	use super::*;
+	use crate::general_lending::{GeneralLoan, LiquidationStatus};
+
+	#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TypeInfo)]
+	pub struct LoanAccount<T: Config> {
+		pub borrower_id: T::AccountId,
+		pub collateral_topup_asset: Option<Asset>,
+		pub collateral: BTreeMap<Asset, AssetAmount>,
+		// This field is being removed:
```
j4m1ef0rd, line 15: Comment is in wrong place. The field above is being removed.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -153,9 +175,8 @@ impl<T: Config> LoanAccount<T> {
 		}
 	}
 
-	pub fn get_collateral_not_in_liquidation(&self) -> BTreeMap<Asset, AssetAmount> {
-		// Start with any collateral that may be sitting in the account:
-		let mut collateral = self.collateral.clone();
+	pub fn get_collateral_in_supply_pools(&self) -> BTreeMap<Asset, AssetAmount> {
+		let mut collateral = BTreeMap::<Asset, AssetAmount>::new();
 
 		// Add any funds supplied in lending pools
```
j4m1ef0rd, line 181: Comment is a little out of date.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
j4m1ef0rd, line 188:
```suggestion
		GeneralLendingPools::<T>::iter()
			.filter_map(|(asset, pool)| {
				pool.get_supply_position_for_account(&self.borrower_id)
					.ok()
					.map(|amount| (asset, amount))
			})
			.collect()
```

j4m1ef0rd (approved):

state-chain/pallets/cf-lending-pools/src/migrations.rs
```diff
@@ -37,5 +38,12 @@ pub type PalletMigration<T> = (
 		Pallet<T>,
 		<T as frame_system::Config>::DbWeight,
 	>,
-	PlaceholderMigration<4, Pallet<T>>,
+	VersionedMigration<
+		4,
+		5,
```
dandanlen, line 43: We need to bump the pallet storage version.

state-chain/pallets/cf-lending-pools/src/migrations/collateral_to_supply.rs
```diff
@@ -0,0 +1,179 @@
...
+
+		for (account_id, old_account) in &old_accounts {
+			// Supply each collateral asset into the corresponding lending pool.
+			for (asset, amount) in &old_account.collateral {
+				if *amount == 0 {
+					continue;
+				}
+
+				// Ensure the lending pool exists, creating it if needed.
+				if GeneralLendingPools::<T>::get(asset).is_none() {
```
dandanlen: nit: not that important in a migration, but genearlly speaking, prefer to use `contains_key` to check for existence otherwise we read and decode the pool on every iteration of the loop.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -196,46 +217,16 @@ impl<T: Config> LoanAccount<T> {
...
-		});
-	}
-
 	/// Return user's existing collateral (what hasn't been swapped during liquidation).
-	/// Unlike [add_new_collateral], we don't emit an event when collateral is returned.
-	fn return_collateral(&mut self, asset: Asset, amount: AssetAmount) {
-		self.collateral.entry(asset).or_default().saturating_accrue(amount);
+	/// Unlike [supply_funds], we don't emit an event when collateral is returned.
+	fn return_supplied_funds(&mut self, asset: Asset, amount: AssetAmount) {
+		GeneralLendingPools::<T>::mutate(asset, |maybe_pool| {
```
dandanlen: Can this be written in terms of `mutate_existing_pool`?

state-chain/pallets/cf-lending-pools/src/migrations/collateral_to_supply.rs
```diff
@@ -0,0 +1,179 @@
...
+						);
+						continue;
+					}
+				}
+
+				if let Err(e) = general_lending::supply_funds::<T>(
+					account_id.clone(),
+					*asset,
+					*amount,
+					SupplyAddedActionType::Manual,
```
dandanlen, line 53: ~~FWIW we can't rely on the events being triggered here because they will be erased again after the runtime upgrade during block initialization (but maybe we aren't relying on this anyway).~~ never mind, I confused this with something else.

If we are relying on the events, should we consider adding a special case for the migration like `SupplyAddedActionType::Migrated`? Maybe this would help disambiguate on the product side.

state-chain/pallets/cf-lending-pools/src/lib.rs
```diff
@@ -123,10 +123,6 @@ pub struct PalletSafeMode {
 	pub add_lender_funds: SafeModeSet<Asset>,
 	// whether lenders can withdraw funds from lending pools (stale oracle also disables this)
 	pub withdraw_lender_funds: SafeModeSet<Asset>,
```
dandanlen, line 125: We could rename these to remove the suggestion that it only applies to lenders? Something like `add_to_supply`/`withdraw_from_supply`?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1490,119 +1477,14 @@ impl<T: Config> LendingApi for Pallet<T> {
 			}
 
 			// NOTE: even if we settle the last loan here, we don't remove
```
dandanlen: I'm not sure I understand this - if there are no loans outstanding, shouldn't we delete the loan account? If nothing else it's one less thing to iterate over.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -679,12 +677,13 @@ impl<T: Config> LoanAccount<T> {
...
 		price_cache: &OraclePriceCache<T>,
 	) -> Result<Vec<AssetCollateralForLoan>, Error<T>> {
-		let mut collateral_to_liquidate = core::mem::take(&mut self.collateral);
+		let mut collateral_to_liquidate = BTreeMap::<Asset, AssetAmount>::new();
 
 		// Gather collateral from supply side of liquidity pools if available:
 		for asset in GeneralLendingPools::<T>::iter_keys().collect::<Vec<_>>().iter() {
 			GeneralLendingPools::<T>::mutate(asset, |pool| {
 				if let Some(pool) = pool.as_mut() {
+					// TODO: emit an event here?
```
dandanlen: We should clarify this before merging.

dandanlen (commented):
In the config, we still have `minimum_update_collateral_amount_usd` which is now unused. 

Implicitly, the `minimum_supply_amount_usd` is what will be used I suppose. We should decide if that's ok, and if so we can remove the old config item.

We might also decide that we want to keep the distinction between a min for 'pure' supply vs 'collateralising' supply. However i suspect that would be hard to enforce.

state-chain/pallets/cf-lending-pools/src/migrations.rs
msgmaxim: What do you mean? It is updated to 5 (the one currently released is 2).

state-chain/pallets/cf-lending-pools/src/migrations/collateral_to_supply.rs
msgmaxim: My initial reaction is that I'd rather not add a variant (and ask the product to handle it separately) that will immediately get deprecated after release, but I will check with the product team.

state-chain/pallets/cf-lending-pools/src/lib.rs
msgmaxim: We use this terminology throughout (e.g. also in extrinsics and events), it won't be a small change to update these to use the new terminology. Renaming would also be a breaking change (e.g. this will affect fields returned by in the correponding rpc). Not sure if it is worth it at this stage.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: The main motivation for this was "user preferences" like `collateral_topup_asset` (this is the only one I believe, but maybe in the future there will be more?). My thinking was that it may be not be uncommon for a user to fully repay a loan and then immediately (or after a short time) open a new loan against the existing supply/collateral. It could be confusing then if their preferences would get reset each time.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Yeah, we have been discussing this for a while, only just agreed on the solution. Will update.

msgmaxim:
> In the config, we still have `minimum_update_collateral_amount_usd` which is now unused.
> 
> Implicitly, the `minimum_supply_amount_usd` is what will be used I suppose. We should decide if that's ok, and if so we can remove the old config item.

I think we should replace `minimum_update_collateral_amount_usd` with `minimum_update_supply_amount_usd` (`minimum_supply_amount_usd` is just to ensure that the total/remaining amount is above some thresholds, but does not apply to updates).

> We might also decide that we want to keep the distinction between a min for 'pure' supply vs 'collateralising' supply. However i suspect that would be hard to enforce.

Yeah, I don't think there is any real need to differentiate the two cases.

state-chain/pallets/cf-lending-pools/src/migrations.rs
dandanlen: Sorry I meant the version in the pallet is still 4, so this migration won't run.

state-chain/pallets/cf-lending-pools/src/migrations/collateral_to_supply.rs
dandanlen: Sounds like manual is ok.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
dandanlen: It sounds like their topup asset is explicitly set anyway when they open a loan. 

Am I wrong about the iteration thing? Will we end up iterating over a bunch of empty loans while taking interest? Seems like an undesirable side-effect.

state-chain/pallets/cf-lending-pools/src/migrations.rs
msgmaxim: Again, it is set to 5 here: https://github.com/chainflip-io/chainflip-backend/pull/6536/files#diff-8d2868f417ae78e9420f213a040be69b81653fc0811446248665c69fe9803574R89. Unless you mean something else?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Hm, I suppose you are right. I will change to remove the account then (esp since we are considering removing auto topup). Update: done.

state-chain/pallets/cf-lending-pools/src/migrations.rs
dandanlen: 🤔 I must have been looking at the wrong file/commit..

dandanlen (approved):

PR #6540, opened 2026-04-23
feat: compile-time check for migration sequences

dandanlen:
# Pull Request

This improves the utility of the `VersionedMigration` sequences by:
- Implementing a trait to chain the sequence of `FROM`/`TO` versions of the versioned migrations.
- Adding compile-time checks to ensure the sequence is valid.
- Decoupling PALLET_VERSION and PALLET_STORAGE_VERSION so that we can use the actual pallet version in the migration declaration.

This makes it much harder to forget to update the version(s) in the right place(s).

The test is simply a reference to the composed `FROM`: in order to compute this the compiler needs to evaulate the chain, and if at any point it isn't contiuguous, we can get a compile-time panic.


msgmaxim (approved):

state-chain/pallets/cf-account-roles/src/lib.rs
```diff
@@ -46,7 +46,8 @@ use frame_system::{ensure_signed, pallet_prelude::OriginFor, RawOrigin};
 pub use pallet::*;
 use sp_std::{collections::btree_map::BTreeMap, marker::PhantomData, vec::Vec};
 
-pub const PALLET_VERSION: StorageVersion = StorageVersion::new(2);
+pub const PALLET_VERSION: u16 = 2;
+pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(PALLET_VERSION);
```
MxmUrw: I think it's confusing to have two different names for the same thing (which I think this is, no?). Could we call the first, numeric constant, e.g. `NUMERIC_STORAGE_VERSION`? Or even better, remove it if possible?

state-chain/runtime-utilities/src/lib.rs
```diff
@@ -255,3 +255,147 @@ impl UncheckedOnRuntimeUpgrade for NoopRuntimeUpgrade {
...
+		if $prev::TO != $next::FROM {
+			panic!(concat!(
+				"Migration sequence not contiguous: ",
+				stringify!($prev),
+				"::TO != ",
+				stringify!($next),
+				"::FROM",
+			));
+		}
+	};
```
MxmUrw: I think this could be simplified to not taking any args + an empty RHS:
```suggestion
	(@checks ) => {};
```
The recursive case should handle the last item correctly already. You might have to move the comma inside the repeating pattern, i.e. `$($rest,)+` instead of `$($rest),+` though.

state-chain/runtime-utilities/src/lib.rs
```diff
@@ -255,3 +255,147 @@ impl UncheckedOnRuntimeUpgrade for NoopRuntimeUpgrade {
...
+/// Implementations for tuples check at compile time that each step's `TO` version equals
+/// the next step's `FROM` version. `check` accesses `FROM`, triggering this evaluation.
+///
+/// Each pallet's `migrations.rs` should include:
+/// ```ignore
+/// #[test]
+/// fn migration_chain_is_current() {
+///     <PalletMigration<mock::Test> as MigrationSequence>::check(PALLET_VERSION);
+/// }
+/// ```
```
MxmUrw, line 320: Is this being actually done somewhere?

state-chain/pallets/cf-account-roles/src/lib.rs
dandanlen: I couldn't think of a way to have just one.. 

None of these seem workable without touching polkadot-sdk: 
`pallet::storage_version(StorageVersion(PALLET_VERSION))`
`pallet::storage_version(PALLET_VERSION)`
`PlaceholderMigration<{ STORAGE_VERSION }, Pallet<T>>,`

There also no const accessor for STORAGE_VERSION... 

I agree naming can be better, will update.

state-chain/runtime-utilities/src/lib.rs
dandanlen: Oops this is left over from the first iteration - now it's sufficient to just evaluate the constant.

dandanlen:
FWIW this is an example of the error. A bit verbose but the information is there:

```
error[E0080]: evaluation panicked: Migration sequence not contiguous: B::TO != C::FROM
    ╭▸ /Users/dan/.rustup/toolchains/nightly-2026-01-06-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/panic.rs:62:9
    │
 62 │         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
    │         ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ evaluation of `<(frame_support::migrations::VersionedMigration<2, 3, migrations::boost_refactor_migration::Migration<mocks::Test>, pallet::Pallet<mocks::Test>, ()>, frame_support::migrations::VersionedMigration<3, 4, migrations::add_min_lending_pool_share::Migration<mocks::Test>, pallet::Pallet<mocks::Test>, ()>, cf_runtime_utilities::PlaceholderMigration<3, pallet::Pallet<mocks::Test>>) as cf_runtime_utilities::MigrationSequence>::FROM` failed here
    │
    ⸬  state-chain/runtime-utilities/src/lib.rs:386:1
    │
386 │ impl_migration_sequence_for_tuple!(A, B, C);
    │ ─────────────────────────────────────────── in this macro invocation
    │
    ╰ note: this error originates in the macro `$crate::panic::panic_2021` which comes from the expansion of the macro `impl_migration_sequence_for_tuple` (in Nightly builds, run with -Z macro-backtrace for more info)

note: erroneous constant encountered
   ╭▸ state-chain/pallets/cf-lending-pools/src/migrations.rs:45:2
   │
45 │     <PalletMigration<crate::mocks::Test> as cf_runtime_utilities::MigrationSequence>::FROM;
   ╰╴    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

dandanlen:
I've addressed the comments and rebased.

MxmUrw (approved):

PR #6542, opened 2026-04-23
chore: dont allow dust flip accounts

ramizhasan111:
# Pull Request

Closes: PRO-2642

## Summary

In case of funding a flip account via the ScUtils contract, if the funding amount is below the Min funding and the account doesn't already exist on the state chain, we ignore the deposit and don't fund the account.

## API Changes

### RPCS

### Extrinsics & Events

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -320,7 +320,7 @@ pub mod pallet {
 	use super::*;
 	use cf_chains::eth::Ethereum;
 	use cf_primitives::BroadcastId;
-	use cf_traits::RedemptionCheck;
+	use cf_traits::{Issuance, RedemptionCheck};
 	use frame_support::{pallet_prelude::*, storage::with_transaction, Parameter};
 	use frame_system::pallet_prelude::*;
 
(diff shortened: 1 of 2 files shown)


ramizhasan111:
An alternative here is that in case of dust funding for a new account, we credit the funds in the FlipAccount storage in the flip pallet but not create the account in the system pallet. This would at least keep the accounting of where the funds went while simultaneously avoiding creating an account. However, this would violate consistency between the system pallet's definition of account existence vs Flip pallet's so probably we shouldn't do that. @dandanlen

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -912,11 +912,19 @@ pub mod pallet {
...
-					caller_account_id.clone(),
-					amount.into(),
-					FundingSource::EthTransaction { tx_hash: eth_tx_hash, funder: caller },
-				),
+				EthereumDeposit::FlipToSCGateway { amount } =>
+					if !frame_system::Pallet::<T>::account_exists(&caller_account_id) &&
+						amount < MinimumFunding::<T>::get().into()
+					{
+						// we ignore dust deposits and dont create a state chain account if the
+						// account doesnt already exist
```
dandanlen, line 920: We should:
- burn the funds
- emit an event

dandanlen (commented):
I made some changes - we shouldn't just silently ignore.

dandanlen (approved):

PR #6543, opened 2026-04-23
chore: remove dust delegations

ramizhasan111:
# Pull Request

Closes: PRO-2643

## Summary

## API Changes

### RPCS

### Extrinsics & Events

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-validator/src/lib.rs
+++ b/state-chain/pallets/cf-validator/src/lib.rs
@@ -2449,20 +2449,20 @@ impl<T: Config> Pallet<T> {
 		}
 		Ok(())
 	}
+
+	/// True if the account has no redemption restriction at all — neither an
+	/// active-bidder lock during auction phase, nor a delegation reservation.
+	fn is_redemption_unrestricted(validator_id: &ValidatorIdOf<T>) -> bool {
+		Self::ensure_not_active_bidder_during_auction(validator_id).is_ok() &&
+			!DelegationChoice::<T>::contains_key(<ValidatorIdOf<T> as IsType<T::AccountId>>::into_ref(
+				validator_id,
+			))
+	}
 }
 
 impl<T: Config> RedemptionCheck for Pallet<T> {
 	type ValidatorId = ValidatorIdOf<T>;
 	type Amount = T::Amount;
-
-	fn ensure_can_redeem(validator_id: &Self::ValidatorId) -> DispatchResult {
-		Self::ensure_not_active_bidder_during_auction(validator_id)?;
-		ensure!(
-			!DelegationChoice::<T>::contains_key(validator_id.into_ref()),
-			Error::<T>::StillBidding
-		);
-		Ok(())
-	}
 
 	fn ensure_can_redeem_amount(
 		validator_id: &Self::ValidatorId,
@@ -2486,6 +2486,22 @@ impl<T: Config> RedemptionCheck for Pallet<T> {
 			);
 		}
 		Ok(())
+	}
+
+	fn ensure_can_transfer(
+		source: &Self::ValidatorId,
+		dest: &Self::ValidatorId,
+	) -> DispatchResult {
+		// If the source can move funds out freely, the transfer is fine.
+		// Otherwise the recipient must share the same restriction so funds
+		// can't escape it by hopping to an unrestricted account.
+		if Self::is_redemption_unrestricted(source) ||
+			!Self::is_redemption_unrestricted(dest)
+		{
+			Ok(())
+		} else {
+			Err(Error::<T>::StillBidding.into())
+		}
 	}
 }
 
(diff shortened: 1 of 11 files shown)


dandanlen:
This doesn't work - I ran the migration and I get: 
`remove_underfunded_delegations: removed 0 underfunded delegations`

I think the problem is not that the delegator's max_bid is too low, it's that the max_bid is high, but the actual *balance* is zero.

ramizhasan111:
> This doesn't work - I ran the migration and I get: `remove_underfunded_delegations: removed 0 underfunded delegations`
> 
> I think the problem is not that the delegator's max_bid is too low, it's that the max_bid is high, but the actual _balance_ is zero.

Currently, it should not be possible to be in a situation where max bid is higher than the balance but I see that a bug was fixed in https://github.com/chainflip-io/chainflip-backend/pull/6166 where it was possible to delegate and then redeem flip before the rotation when it would have been bonded. Because of this it was possible to have a max bid that is higher the balance and so what you are saying would be possible. I, therefore, updated the migration to also include the case where you balance was below min balance. Also, as part of the migration, I am correcting the max bid of cases where it is higher than the balance and setting them equal to the balance. I ran the migration against mainnet and it looks like there are 2 cases where there is a delegation where balance < min funding.

Regarding the fix in the above mentioned PR where we dont allow delegators to withdraw, I think it is still not quite right. If someone's max bid is intentionally lower than their balance, we would want for them to be able to withdraw the rest which the current fix would not allow. I think we should fix that.

state-chain/pallets/cf-validator/src/migrations/remove_underfunded_delegations.rs
```diff
@@ -11,19 +11,29 @@ impl<T: Config> UncheckedOnRuntimeUpgrade for Migration<T> {
...
+
+			let balance = T::FundingInfo::balance(&delegator);
+			if max_bid > balance {
+				log::info!(
+					target: "cf-validator",
+					"max_bid should be lower or equal to balance. violation found and so setting max_bid equal to balance"
+				);
+				max_bid = balance;
+				DelegationChoice::<T>::insert(&delegator, (validator, max_bid));
+			}
```
dandanlen: Sorry - a bit late re-reviewing this. I'm not sure this is a universal truth: the max_bid *can* be lower than the current balance for example due to slashing or transaction fees. So IMO we don't need to apply this change, it's not necessary for correctness.

state-chain/pallets/cf-validator/src/migrations/remove_underfunded_delegations.rs
```diff
@@ -0,0 +1,97 @@
...
+			total_read += 1;
+
+			let balance = T::FundingInfo::balance(&delegator);
+			if max_bid > balance {
+				log::info!(
+					target: "cf-validator",
+					"max_bid should be lower or equal to balance. violation found and so setting max_bid equal to balance"
+				);
+				max_bid = balance;
+				DelegationChoice::<T>::insert(&delegator, (validator, max_bid));
```
dandanlen: We're inserting into the map while while iterating - it's probably fine but in the interest of certainty, better to collect the values first, then iterate.

*edit: as noted in the other comment, I don't think we should update the max bid anyway. All we need to do is undelegate (and possibly also remove the account, i'll check).

dandanlen (commented):
I had another look at the redemption restriction and I agree it needs to be changed. I've mostly done this while investigating the issue on my machine, will post a fix in this PR, and will also address the comments i made.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -2436,23 +2436,63 @@ impl<T: Config> QualifyNode<<T as Chainflip>::ValidatorId> for QualifyByMinimumS
...
+
+	fn ensure_can_transfer(source: &Self::ValidatorId, dest: &Self::ValidatorId) -> DispatchResult {
+		// If the source can move funds out freely, the transfer is fine.
+		// Otherwise the recipient must share the same restriction so funds
+		// can't escape it by hopping to an unrestricted account.
+		if Self::is_redemption_unrestricted(source) || !Self::is_redemption_unrestricted(dest) {
+			Ok(())
+		} else {
+			Err(Error::<T>::StillBidding.into())
+		}
```
ramizhasan111: I am not sure we would want this. There are two separate conditions in ```is_redemption_unrestricted```. Here, in case where both are restricted, it could be possible that both are restricted for different reasons (one because of delegation, one for bidding). We should deal with both cases separately here. If the source has delegation, we should not transfer or transfer only the free amount for example.

dandanlen: Yes I agree this is not ideal but it preserves existing functionality. It effectively replaces this:

```rust
	if !T::RedemptionChecker::can_redeem(&source_account_id) {
		ensure!(
			!T::RedemptionChecker::can_redeem(&recipient_account_id),
			Error::<T>::CanNotRebalanceToNotBiddingValidator
		);
	}
```

In the longer run we need to harmonize how delegator and validator bidding works, and then we can maybe harmonize how transferring from validator to/from delegator. But for now i feel like it's simpler to stick to the current overly-restrictive condition.

The comment is inaccurate though, I'll change it.

dandanlen (approved):

dandanlen:
Rebased and signed the unverified commit.

PR #6546, opened 2026-04-24
feat: Utilisation cap

msgmaxim:
# Pull Request

Closes: PRO-2821

## Summary

Introduces utilisation cap which prevents further borrowing if reached. 

- `liquidation_coverage_factor` is a new config item that controls how much of all loans we should be able to liquidate with the collateral that isn't borrowed (currently set to 100%)
- `cf_lending_pools` now additionally exposes current utilisation cap for the pool (cc @GabrielBuragev). Added `before_v17` for backwards compatibility. I removed `before_v12` however, I think this isn't needed anymore?

(Note that this is based on feat/no-passive-collateral branch, so would need to rebase once that one is merged.)

## API Changes

- new `UtilisationCapExceeded` error may be returned when trying to borrow

### RPCS

- `cf_lending_pools` (see above)


state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1320,12 +1329,21 @@ pub fn fund_loan<T: Config>(
...
 	origination_fee_network: AssetAmount,
+	price_cache: &OraclePriceCache<T>,
 ) -> Result<(), DispatchError> {
+	let utilisation_cap = compute_utilisation_cap::<T>(
+		loan.asset,
+		LendingConfig::<T>::get().liquidation_coverage_factor,
+		price_cache,
+	)?;
+
 	GeneralLendingPools::<T>::try_mutate(loan.asset, |pool| {
```
dandanlen, line 1340: Should we also check the cap in the other pool (ie. the collateral pool)?

Example: 
1. User A supplies 100 USDC, takes no loan
2. User B takes 40 USDC loan with 50 USDT collateral
	-> USDC is now at 0.4 utilisation, no cap 
3. User A takes 80 USDT loan using existing 100 USDC
	-> this implies a USDC utilisation cap of 0.2 but utilisation is at 0.4

However we don't check the cap in the collateral pool, only the loan pool. 

I think as long as the LTV of a new loan is greater than the aggregate LTV of the collateral pool, taking a loan can decrease the cap in the collateral pool.

dandanlen: In fact, given that the borrower might be backing the loan with more than 1 asset... we would need to check all the pools that might be backing the loan.

state-chain/pallets/cf-lending-pools/src/general_lending/config.rs
```diff
@@ -180,6 +180,10 @@ pub struct LendingConfiguration {
 	pub minimum_update_loan_amount_usd: AssetAmount,
 	/// Minimum equivalent amount that can be added to or removed from a lending pool supply.
 	pub minimum_update_supply_amount_usd: AssetAmount,
+	/// Fraction of outstanding loans the pool must remain able to liquidate with available (i.e.
+	/// not borrowed liquidity) at current oracle prices. Used when computing the utilisation cap
+	/// enforced on new borrows.
+	pub liquidation_coverage_factor: Percent,
```
dandanlen, line 186:
```suggestion
	/// Fraction of outstanding loans the pool must remain able to liquidate with available (i.e.
	/// not borrowed) liquidity at current oracle prices. Used when computing the utilisation cap
	/// enforced on new borrows.
	pub liquidation_coverage_factor: Percent,
```

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -904,7 +907,37 @@ impl<T: Config> LoanAccount<T> {
...
+			loan_asset,
+			extra_principal.saturating_add(origination_fee_total),
+		)?;
+
+		for collateral_asset in self.get_total_collateral().into_keys() {
+			let cap = compute_utilisation_cap::<T>(
+				collateral_asset,
+				config.liquidation_coverage_factor,
+				price_cache,
+				Some((&self.borrower_id, extra_load_usd)),
```
j4m1ef0rd: shouldn't the extra amount be in proportion to each asset's USD share of the collateral?

state-chain/pallets/cf-lending-pools/src/general_lending/general_lending_tests.rs
```diff
@@ -5759,3 +5762,394 @@ mod supply_as_collateral {
...
+				let expected_required_usd =
+					price_cache.usd_value_of(LOAN_ASSET, PRINCIPAL + ORIGINATION_FEE).unwrap();
+				let total_collateral_usd =
+					price_cache.usd_value_of(COLLATERAL_ASSET, INIT_COLLATERAL).unwrap();
+				let expected_cap = Permill::one().saturating_sub(Permill::from_rational(
+					expected_required_usd,
+					total_collateral_usd,
+				));
+
+				let utilisation_cap = compute_utilisation_cap::<Test>(
```
j4m1ef0rd: can use `get_utilisation_cap` utility here.

j4m1ef0rd (approved):

dandanlen (approved):
I'm happy for this to be merged (pretty sure it's correct) but I think a couple of changes are worthwhile in a follow up:

First: The optional `extra_borrower_load` param is a little confusing / a bit of a smell. It seems like we ought to be able to compute this once in one place instead of twice, slightly differently depending on context. Afaict the optional 'extra_borrow_load' is needed because when we do the check we haven't yet applied the loan, and so we need this argument as a workaround to communicate what we intend to do. 

For example, did you consider doing the check *after* we make the updates? If we create the loan as before, and then check coverage in all relevant pools and revert if the coverage limit is breached, that should be enough (?), and (IMO) would make for the code easier to follow and reason about.

The other thing we should improve is efficiency: there's a lot of iteration and I feel we can probably make the computation more incremental if we cache things cleverly, or even just index the accounts based on their loan/collateral exposure to avoid iterating over all of them every time.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: No, where it is used we just want to compute the total value of the acount's loans. It is at a later poin (inside `required_for_account` closure) that we take take into account which asset the ratio of the collateral in the target asset against to total collateral.

msgmaxim:
> For example, did you consider doing the check after we make the updates? If we create the loan as before, and then check coverage in all relevant pools and revert if the coverage limit is breached, that should be enough (?), and (IMO) would make for the code easier to follow and reason about.

There are a couple of tradeoffs with this approach: the main one is not being able to check everything in `expand_loan_inner` leading to having to repeat the check in both `request_loan` and `expand_loan` (and I didn't want to add another `inner` function). However, I gave this a try and I think duplicating the `check_collateral_pool_caps` is a reasoanble compromise. Updated.

> The other thing we should improve is efficiency: there's a lot of iteration and I feel we can probably make the computation more incremental if we cache things cleverly, or even just index the accounts based on their loan/collateral exposure to avoid iterating over all of them every time.

I didn't worry too much about efficiency because these checks will only be called on loan requests, which will probably be rare. Boost borrows are much more common, but they only check utilisation in one pool (BTC) and not the collateral pools.

dandanlen:
> There are a couple of tradeoffs with this approach: the main one is not being able to check everything in `expand_loan_inner` leading to having to repeat the check in both `request_loan` and `expand_loan` (and I didn't want to add another `inner` function). However, I gave this a try and I think duplicating the `check_collateral_pool_caps` is a reasoanble compromise. Updated.
> 

I think there might be a clearer way where we only check the caps once, after taking the loan - we currently still check once in fund_loan and then again outside. There are potentially redundant checks (ie. if the user borrows an asset that is also collateral, unlikely but still). It also just seems clearer to do all the checks in one place, after any mutations to the loan account storage. I'll see if I can put up a PR.

PR #6548, opened 2026-04-28
feat: deduplicate scheduled set/close orders for the same order

marcellorigotti:
# Pull Request

Closes: PRO-2814

## Summary

Deduplicate scheduled `Set`/`Close` orders targeting the same exact order (same lp, pool, side and id). 

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-pools/src/lib.rs
+++ b/state-chain/pallets/cf-pools/src/lib.rs
@@ -115,7 +115,7 @@ enum LimitOrderUpdateDetails<BlockNumber: sp_std::fmt::Debug> {
 }
 
 impl<T: Config> LimitOrderUpdate<T> {
-	fn targets_same_order(&self, other: &Self) -> bool {
+	fn can_override_order(&self, other: &Self) -> bool {
 		self.lp == other.lp &&
 			self.base_asset == other.base_asset &&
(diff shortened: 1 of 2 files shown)


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -115,6 +115,19 @@ enum LimitOrderUpdateDetails<BlockNumber: sp_std::fmt::Debug> {
 }
 
 impl<T: Config> LimitOrderUpdate<T> {
+	fn targets_same_order(&self, other: &Self) -> bool {
```
msgmaxim: Nit: the name is a little misleading. Here we don't only check that it is the same order, but also check the update type. How about `can_be_overridden_by`?

MxmUrw (approved):
Maybe we should add a test that verifies that a second, different order does not override the first one, and that the counter increases?

Everything looks good though!

dandanlen (approved):

state-chain/pallets/cf-pools/src/lib.rs
marcellorigotti: renamed to `can_override_order` since it is self replacing other 👍

PR #6549, opened 2026-04-29
feat: Lending broker fees

msgmaxim:
# Pull Request

Closes: PRO-2769

## Summary

- Adds a broker parameter to request_loan extrinsic (includes broker id and interest in basis points). The parameter is stored on the loan (migration in `loan_account_migration`). Whenever we calculate/take regular interest, we now additionally calculate/take "broker interest". Brokers will earn interest periodically (no repayment of loans is necessary), taking funds from the pool's available funds (the borrower will pay them back).
- `MAX_BROKER_FEE_BPS = 1000` (hardcoded currently) prevents brokers from setting broker interest that's too high. `BrokerFeeTooHigh` will be emitted if broker sets it incorrectly.

fiy @GabrielBuragev

## API Changes

### RPCS

- `cf_loan_accounts`: new `broker` field is added to loans

### Extrinsics & Events

- `request_loan` now takes broker of type `Beneficiary` (same as used in swapping)
- `InterestTaken`'s `broker_interest` field is now correctly set (this field already existed but was hardcoded to 0) 

---

## *Checklist*

* The PR is complete:
  * Tests:


j4m1ef0rd (approved):

dandanlen (commented):
There are still a couple of mentions of `// TODO: add support for broker fees` (OriginationFeeTaken and LiquidationFeeTaken).

=> We should resolve these: either we remove the broker_fee entry from the event or we keep it for future flexibility. Maybe linking an issue ID would help clarify what needs to be done to resolve the TODO.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1380,10 +1425,15 @@ impl<T: Config> LendingApi for Pallet<T> {
 		asset: Asset,
 		amount_to_borrow: AssetAmount,
 		collateral_topup_asset: Option<Asset>,
+		broker: Option<Beneficiary<T::AccountId>>,
 	) -> Result<LoanId, DispatchError> {
 		T::LpRegistrationApi::ensure_has_refund_address_for_asset(&borrower_id, asset)
 			.map_err(|_| Error::<T>::NoRefundAddressSet)?;
 
+		if let Some(broker) = &broker {
+			ensure!(broker.bps <= MAX_BROKER_FEE_BPS, Error::<T>::BrokerFeeTooHigh);
```
dandanlen, line 1439: We should also check if the account is registered as a broker.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1916,18 +1970,21 @@ impl<T: Config> Pallet<T> {
 		Ok(account)
 	}
 
-	/// Mutates for pool for `asset` expecting it to exist.
-	fn mutate_existing_pool<F>(asset: Asset, f: F)
+	/// Mutates the pool for `asset` expecting it to exist. If the pool is missing the
+	/// closure is skipped and the default value is returned.
+	fn mutate_existing_pool<R, F>(asset: Asset, f: F) -> R
```
dandanlen: nit: Returning `Option<R>` would be more explicit here.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -943,6 +946,8 @@ pub struct GeneralLoan<T: Config> {
 	pub owed_principal: AssetAmount,
 	/// Interest owed on the loan but not yet taken (it is below the threshold)
 	pub pending_interest: InterestBreakdown,
+	/// Broker and their fee/interest (if any)
```
dandanlen, line 953: nit: this comment is now slightly out of date: `pending_interest` now potentially contains broker fees that couldn't be taken because the pool was empty (unlikely but still...).

Also - there's no broker fee collection threshold, is there? Any particular reason not to just re-use the same logic? Also does this effectively mean that broker fees have (very minor) priority?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
dandanlen: Also, we might want to ensure the fee is non-zero?

dandanlen:
Feel free to merge once the above issues are addressed.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: > nit: this comment is now slightly out of date: pending_interest now potentially contains broker fees that couldn't be taken because the pool was empty (unlikely but still...).

Updated the comment.

> Also - there's no broker fee collection threshold, is there? Any particular reason not to just re-use the same logic? Also does this effectively mean that broker fees have (very minor) priority?

Not sure what you mean. We already use the same logic and apply the same threshold to broker fees (it uses `charge_fee_if_exceeds_threshold`).

Brokers fees do have a minor priority because of the way they handle 100% utilisation: network fees will only be re-attempted when there is a new contribution to network fees, while broker fees are more eager and will be re-attempted every block. It might make sense to change network fee behaviour to use the same logic, but would be a separate PR and probably not for 2.2 (I will create a ticket).

msgmaxim:
> There are still a couple of mentions of `// TODO: add support for broker fees` (OriginationFeeTaken and LiquidationFeeTaken).
> 
> => We should resolve these: either we remove the broker_fee entry from the event or we keep it for future flexibility. Maybe linking an issue ID would help clarify what needs to be done to resolve the TODO.

I will create a ticket. Will keep the TODOs, but link them to the linear ticket.

PR #6553, opened 2026-04-29
chore: remove unused mandatory settings from engine config

marcellorigotti:
# Pull Request

Closes: PRO-2835

## Summary

Removed:
- eth ws endpoint
- arb ws endpoint
- dot endpoints

* The PR is complete:
  * Tests:


vdoflip (approved):

MxmUrw (approved):
Very nice!

albert-llimos (approved):
Great 🙌

engine-upgrade-utils/src/lib.rs
```diff
@@ -32,10 +32,85 @@ pub const NEW_VERSION: &str = "2.2.0";
...
-// Sometimes we need to remove arguments that are valid for the new version but not for the old
-// version.
+// Sometimes we need to adapt arguments between the new and old versions while both CFEs can be run
+// by the upgrade runner.
 pub fn args_compatible_with_old(args: Vec<String>) -> Vec<String> {
-	args
+	let mut compatible_args = args;
+
+	// Old CFE 2.1.9 still requires and validates EVM websocket endpoints in its settings model,
+	// and constructs websocket subscription clients from them. The active ETH/ARB witnessing path
```
dandanlen, line 41: Is it true that the old version constructs a websocket client, even though it's unused?

dandanlen (approved):

engine-upgrade-utils/src/lib.rs
marcellorigotti: Actually not really, we did construct `ReconnectSubscriptionClient` but the actual websocker client wasn't constructed because we didn't call `subscribe_blocks`
We also did construct the `RetrierClient` which also was unused since no request was ever sent to it

PR #6554, opened 2026-04-30
feat: scheduled swaps rpc uses oracle price for estimation

j4m1ef0rd:
# Pull Request

Closes: PRO-2838

## Summary

Because we use JIT for most of our liquidity for most assets, Using swap simulation for the  `cf_scheduled_swaps` subscription is not great. Now that we have the oracle's we should use them instead.

Changed the `cf_scheduled_swaps` RPC to use oracle. price for estimating the second leg usdc amounts instead of doing swap simulation.
The pool price is still used as a fallback when no oracle is available (Flip and Dot). The stale flag is ignored.
Please note that this code will be replaced almost entirely in the non-usdc swapping branch. So this PR is a somewhat temporary stopgap until that feature is merged.

### RPCS

`cf_scheduled_swaps` should have no behaviour change, but it should be a bit more accurate now.

---

* The PR is complete:
  * Tests:


dandanlen (commented):
It occurred to me while reviewing this that this subscription probably only works correctly for non-USDC swaps. For USDC swaps we ignore any swap that doesn't start or end with USDC, even though both legs are usdc denominated.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1791,27 +1791,27 @@ pub mod pallet {
...
-			// Map the swaps to SwapLegInfo
-			successful_swaps
-				.into_iter()
-				// Turn the failed swaps into fresh swap states. Pool price fallback will be used
-				// for these swaps.
-				.chain(failed_swaps.into_iter().map(|(swap, _)| SwapState::new(swap)))
-				.filter_map(|state| {
-					let swap_request = SwapRequests::<T>::get(state.swap_request_id())
-						.expect("Swap request should exist");
 					let dca_state = match swap_request.state {
```
msgmaxim: Thought I've mentioned this before: we can extract remaining_chunks and chunk_interval inside the match, then we won't need to unwrap them later.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -1821,66 +1821,76 @@ pub mod pallet {
...
-									.output_amount_ceil(state.input_amount())
-									.saturated_into(),
-							)
-						})?;
+						let fee = core::cmp::max(
+							network_fee.rate * stable_amount_pre_fee,
+							network_fee.minimum,
+						);
+						let stable_amount = stable_amount_pre_fee
+							.saturating_sub(core::cmp::min(fee, stable_amount_pre_fee));
```
msgmaxim: Nit: if you use `saturating_sub`, then you shouldn't need to also min the fee with `stable_amount_pre_fee`.

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1236,7 +1236,10 @@ fn test_get_scheduled_swap_legs() {
 			]);
 		});
 
-		SwapRate::set(2f64);
+		// Oracle price of 2 USD/Eth is used to estimate the intermediate USDC amount
+		// when the input is Eth (swap 5).
+		MockPriceFeedApi::set_price_usd_fine(Asset::Eth, 2);
+		MockPriceFeedApi::set_price_usd_fine(Asset::Usdc, 1);
 		// The amount of USDC in the middle of swap (5):
 		const INTERMEDIATE_AMOUNT: AssetAmount = 2000;
```
msgmaxim: `const INTERMEDIATE_AMOUNT: AssetAmount = INIT_AMOUNT * 2;`

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1409,7 +1404,10 @@ fn test_get_scheduled_swap_legs_for_dca() {
 		const NUMBER_OF_CHUNKS: u32 = 3;
 		const CHUNK_INTERVAL: u32 = 10;
 		const BLOCK: u64 = INIT_BLOCK + SWAP_DELAY_BLOCKS as u64;
-		SwapRate::set(1_f64);
+
+		// Oracle price of 1 USD for both assets so the intermediate amount equals input.
+		MockPriceFeedApi::set_price_usd_fine(Asset::Flip, 1);
```
msgmaxim: Maybe we should use different prices to reduce false positives?

msgmaxim (approved):

dandanlen (approved):

PR #6565, opened 2026-05-05
feat: new `cf_all_account_infos` rpc call to get account info for all accounts

MxmUrw:
# Pull Request

Closes: PRO-2808

## Summary

This PR introduces a new rpc call, `cf_all_account_infos`, which returns the account info for *all* accounts. This was possible before by doing a `cf_accounts` call to get all account ids, and then a batched call to `cf_account_info` (which returns info for a single account) but it took a long time (>6s when querying against a production rpc node). The biggest performance problem was due to repeatedly sweeping all pools in every iteration of the batch.

This new RPC does a single sweep of *all* accounts in all pools and then mostly uses the preexisting runtime calls to get (common account info + role specific account info) and compose them in the runtime layer. (the legacy `cf_account_info` rpc call does the composition in the rpc layer.)

The result is a speedup of ~3x.

(Further speedups are possible if we improve how delegation related data is extracted, since there is quite a lot of inefficiency there currently)

**QUESTION for reviewers**: Which of the following names do you prefer?
 - `cf_all_accounts_info`
 - `cf_all_account_infos`
 - other proposals?

### Legacy method
The legacy call has to be done in two steps (first get all accounts, then do a batched account_info call for all of them):
```
[16:21:11.466] INFO: Fetching accounts from http://127.0.0.1:9944
[16:21:11.497] INFO: Fetched 1861 accounts in 30.41 ms
[16:21:11.497] INFO: Requesting data for 1861 accounts
[16:21:11.497] INFO: Sending batched cf_account_info request for 1861 accounts (218492 bytes)
[16:21:13.486] INFO: Batch completed in 1988.47 ms
[16:21:13.486] INFO: Returned 1861/1861 responses
[16:21:13.487] INFO: Batch response size: 1620417 bytes
[16:21:13.487] INFO: Average batch time per account: 1.07 ms
[16:21:13.491] INFO: All cf_account_info responses succeeded
```
~1990ms

### New method
Now we need only a single call:
```
[16:23:04.148] INFO: Sending native cf_accounts_info request for all accounts
[16:23:04.757] INFO: Request completed in 608.88 ms
[16:23:04.757] INFO: Returned 1861 account_info records
[16:23:04.758] INFO: Batch response size: 1552428 bytes
[16:23:04.758] INFO: Average batch time per account: 0.33 ms
[16:23:04.758] INFO: Native cf_accounts_info request succeeded
[16:23:04.758] INFO: All cf_account_info responses succeeded
```
~610ms

### RPCS

 - adds a new rpc `cf_all_account_infos`. It takes an optional parameter `roles: Option<Vec<AccountRole>>` that filters and returns accounts only with the provided roles.
 - adds a new field `account_id: AccountId32` to the existing `cf_account_info` rpc.

## Implementation details

1. Some of the existing runtime calls (`cf_common_account_info` and `cf_liquidity_provider_info`) have a new argument `ShouldSweep` to control whether they sweep or not. This is required because they're called from both the "legacy" and the new rpc and respectively have to sweep or not sweep.
2. The pool pallet's `inner_sweep` previously took a single account id and swept only for this account. It has been refactored to be able to sweep for a single account OR for all accounts. There's a new (quite generic) `Accessor` trait which allows the caller of `inner_sweep` to inject any account iteration logic that should be used for sweeping. This way the actual implementation of sweeping can stay readable while allowing for different iteration patterns.
3. The conversion in the custom rpc layer of the role specific account info from runtime encoding to the rpc encoding is boiler-platey and copies what the legacy `cf_account_info` method does. In the latter it's interleaved with calls to historical runtime apis so it wasn't straight-forward to extract the logic out. Since I intend to improve how custom-rpc is implemented anyways I'm keeping things like this for the time being.    

---

* The PR is complete:
  * Tests:


state-chain/custom-rpc/src/lib.rs
```diff
@@ -895,6 +896,12 @@ pub trait CustomApi {
 		account_id: state_chain_runtime::AccountId,
 		at: Option<state_chain_runtime::Hash>,
 	) -> RpcResult<RpcAccountInfoV2>;
+	#[method(name = "accounts_info")]
+	fn cf_all_accounts_info(
```
msgmaxim: Would this be slightly better?
```suggestion
	fn cf_all_account_infos(
```

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1549,14 +1560,22 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
-	fn inner_sweep(lp: &T::AccountId) -> DispatchResult {
+	fn inner_sweep(lp_accounts: &impl Accessor<BTreeMap<T::AccountId, !>>) -> DispatchResult {
```
msgmaxim: Wondering if this `Accessor` trait is actually necessary. Could you not just have this take an enum (where `SingleEntry(who)` and `AllEntries` are variants) and match on that here?

state-chain/custom-rpc/src/lib.rs
MxmUrw: Not sure, I feel like `cf_all_account_infos` could mean all "infos" for a *single* account. But I get that `cf_all_accounts_info` doesn't roll of the tongue either. Not a strong opinion I guess I'll go with whatever the majority is happier with.

state-chain/pallets/cf-pools/src/lib.rs
MxmUrw: I think it's worth having the trait because doing the enum matching inline (twice) would complicate the existing logic and make it more difficult to focus on the relevant parts of the code (the actual sweeping), because it would now be interleaved with logistical code that's just there to access accounts.

With the trait the code can be read and understood is if we were always iterating over all accounts; but with the benefit of actually also working for single-account access.

**EDIT**: I think you're actually right. Even with an enum I can have exactly the same code for `inner_sweep()` as I have currently. I'll give it a go.

MxmUrw: Done. Also FYI I renamed the enum to `Select<>`.

marcellorigotti (approved):
LGTM! I don’t have a strong opinion on the name, but if I have to choose one, I’d choose cf_all_accounts_info.

utilities/src/without_std/selector.rs
```diff
@@ -0,0 +1,23 @@
...
+/// Useful when e.g. processing rpc methods, where passing a parameter is
+/// a request for this particular key but passing nothing is meant to return all
+/// all the information for all keys.
+pub enum Select<'k, Key: 'k> {
+	Single(&'k Key),
+	All(),
+}
+
+impl<'k, Key: Ord + 'k> Select<'k, Key> {
+	pub fn with_values_from_btree_map<'a: 'k, A: 'a>(
```
dandanlen: nit: the `with_` prefix usually implies that the caller should pass a callback fn. 

Maybe replace it with `iter_` or `select_`?

MxmUrw: Sure, `select_` is fine as well.

dandanlen (approved):
LGTM - no strong opinion on the name. 

To me, all_accounts_info sounds like "a single info for all the accounts', and `all_account_infos` is `the individual infos for each account`. 

But it seems this is subjective, so i'm happy either way.

MxmUrw:
Renamed the rpc to `all_account_infos`, it slowly grew on me during the weekend. Thanks for the inputs everyone!

PR #6566, opened 2026-05-05
feat: collect just enough collateral for liquidation

msgmaxim:
# Pull Request

Closes: PRO-2716

## Summary

When preparing collateral for liquidation we now only withdraw just enough to cover what's needed for liquidation (owed amount + either 50 or 500 bps to cover max oracle slippage). This is mostly useful for manually triggered liquidations where there is likely more than enough collateral to cover the loans. Note that it is not a problem if we end up preparing less collateral than needed (can happen if prices move quickly) as we now support multiple liquidation rounds, so we would simply start more liquidation swaps with extra collateral.


state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1160,6 +1190,214 @@ pub(super) struct AssetCollateralForLoan {
...
+) -> Vec<(Asset, AssetAmount)> {
+	let total_available_usd =
+		positions.iter().fold(0u128, |acc, (_, _, usd)| acc.saturating_add(*usd));
+
+	if total_available_usd == 0 {
+		return Vec::new();
+	}
+
+	let total_required_with_slippage_usd =
+		required_collateral_with_slippage(total_owed_usd, max_oracle_slippage_bps);
```
dandanlen: We ignore the slippage liquidation fee here (and swap network fee?) - is this significant relative to the max_oracle_slippage?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1160,6 +1190,214 @@ pub(super) struct AssetCollateralForLoan {
...
+	let take_all = total_required_with_slippage_usd >= total_available_usd;
+	let target_total_usd = core::cmp::min(total_required_with_slippage_usd, total_available_usd);
```
dandanlen: nit: `target_total_usd` is always equal to `total_required_with_slippage_usd` - otherwise it's unused.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -605,34 +607,62 @@ impl<T: Config> LoanAccount<T> {
...
 		let mut collateral_to_liquidate = BTreeMap::<Asset, AssetAmount>::new();
 
-		// Gather collateral from supply side of liquidity pools if available:
-		for asset in GeneralLendingPools::<T>::iter_keys().collect::<Vec<_>>().iter() {
+		for (asset, amount_to_request) in compute_per_asset_collateral_takes(
+			total_owed_usd,
+			max_oracle_price_slippage_bps,
+			&positions_with_usd,
+		) {
 			GeneralLendingPools::<T>::mutate(asset, |pool| {
```
dandanlen: nit: we can use mutate_existing_pool here again?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1160,6 +1190,214 @@ pub(super) struct AssetCollateralForLoan {
...
+}
+
+/// Decide how much of each available collateral asset to pull into liquidation swaps.
+/// `positions` is `(asset, available_amount, available_usd_value)` for each pool the
+/// borrower has a positive supply position in. When the available collateral exceeds the
+/// owed-plus-slippage target, each asset is drawn proportionally to its share of the
+/// total available USD value (rounded up so the collected oracle value never falls short
+/// of the per-asset target). Otherwise the function returns the full available amount of
+/// each position.
+fn compute_per_asset_collateral_takes(
```
dandanlen: This name is kind of hard to interpret... how about compute_liquidation_estimates?

dandanlen (approved):

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Not insignificant I suppose (5 and 10 bps versus 50bps slippage), but I wanted to keep things simple as the other parameters are from a different pallet...

msgmaxim: Update: network and liquidation fees are now included in the estimates.

dandanlen (approved):

PR #6572, opened 2026-05-08
feat: Lending LP API

j4m1ef0rd:
# Pull Request

Closes: PRO-2849

## API Changes

- Added lending functionality to both the LP API and the node API
- Added a bouncer test for the new LP API functions
- Fixed a bug in the run test command so it will not confuse the new test with the existing LP api test

---

* The PR is complete:
  * Tests:


api/lib/src/lp.rs
```diff
@@ -505,4 +507,175 @@ pub trait LpApi: SignedExtrinsicApi + Sized + Send + Sync + 'static {
...
+	}
+
+	async fn request_loan(
+		&self,
+		loan_asset: Asset,
+		loan_amount: AssetAmount,
+		broker: Option<Beneficiary<AccountId>>,
+		wait_for: WaitFor,
+	) -> Result<ApiWaitForResult<LoanId>> {
+		if loan_amount == 0 {
```
msgmaxim: Do we have to check this (here and in other calls)? The actual min is greater than 0, and we already have to expect that extrinsic may fail if the amount is below that min. Might just let the extrinsic handle the 0 case as well.

api/cf-rpc-types/src/lp.rs
```diff
@@ -151,3 +153,28 @@ pub struct LiquidityDepositChannelDetails {
 	pub deposit_address: AddressString,
 	pub deposit_chain_expiry_block: <AnyChain as Chain>::ChainBlockNumber,
 }
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub enum RepaymentAmountJson {
+	Full,
+	Exact(NumberOrHex),
+}
```
dandanlen: Please don't create a new copy-paste type. 

Either make the original type abstract over `Amount`, or use serde helpers to control serialization behaviour:

```rust
pub enum RepaymentAmount {
	/// Full repayment
	Full,
	/// Only repay with the specified amount
	Exact(#[serde(with = "cf_utilities::serde_helpers::number_or_hex")] AssetAmount),
}
```

api/cf-rpc-apis/src/lp.rs
```diff
@@ -168,4 +168,50 @@ pub trait LpRpcApi {
...
+	) -> RpcResult<ApiWaitForResult<NumberOrHex>>;
+
+	#[method(name = "request_loan")]
+	async fn request_loan(
+		&self,
+		loan_asset: Asset,
+		loan_amount: NumberOrHex,
+		broker: Option<Beneficiary<AccountId32>>,
+		wait_for: Option<WaitFor>,
+	) -> RpcResult<ApiWaitForResult<LoanId>>;
```
dandanlen: We can think a bit harder about the return types. For example, just returning a raw number isn't very friendly. We could at least wrap it like `{ loan_id: 432 }` for example. 

But also, I can imagine there are more things someone might want to know about their loan when they open it: What was the fee? What is my LTV? etc.

dandanlen (commented):
FWIW I'm not totally sure we want to / need to add all these methods to the LP api... it's significant additional maintenance burden and the benefit is not obvious (we are just piping the methods through to the individual calls), so it's worth discussing.

j4m1ef0rd:
> not totally sure we want to / need to add all these methods to the LP api.

Are you talking about the entire PR or just having the new lending stuff in the legacy LP API server?
We had a small discussion [here](https://discord.com/channels/775961728608895008/1044538771967840276/1501115158401912883)

dandanlen:
> > not totally sure we want to / need to add all these methods to the LP api.
> 
> Are you talking about the entire PR or just having the new lending stuff in the legacy LP API server? We had a small discussion [here](https://discord.com/channels/775961728608895008/1044538771967840276/1501115158401912883)

I'm talking about the entire PR. I don't think we should be making these commitments based on a couple of discord comments... Adding this means longer-term support for the APIs, documentation etc. It goes beyond just this PR.

PR #6574, opened 2026-05-08
fix: Clamp price

msgmaxim:
# Pull Request

Closes: PRO-2855

## Summary

I chose to clamp price when adjusting it by fee as it is the simplest approach and at extreme values we are not concerned about accuracy.


dandanlen (approved):
As a longer term fix, I'd prefer we remove the implicit `From` impls to ensure we don't have any silent assumptions. I'll open a follow-up PR.

PR #6575, opened 2026-05-08
docs: bouncer claude skill

j4m1ef0rd:
# Pull Request

## Summary

Adds a `/bouncer` skill so Claude can:
- check if a localnet is running and if it's setup
- check if the running localnet is at the correct commit
- start/restart the localnet
- run a bouncer test or all bouncer tests
- regenerate the bouncer schema
- look for events on the indexer
- run other bouncer commands

Other changes:
- new command to check the current bouncer state.
- run_test command now supports being given a vitest name
- the perform swap command will generate an address if none is provided


.claude/skills/bouncer/SKILL.md
```diff
@@ -0,0 +1,193 @@
+---
+name: bouncer
+description: Use for anything involving the chainflip bouncer or localnet — running end-to-end tests, starting/stopping a localnet, getting a localnet ready for testing, running setup scripts, or regenerating the bouncer's event schemas. Triggered by phrases like "run the bouncer test", "test this in the bouncer", "kick off LpApi", "run the lending test", "spin up a localnet", "start a localnet", "get a localnet ready", "tear down the localnet", "regenerate the schemas", "regenerate event types", or naming a specific bouncer test. Covers localnet lifecycle, test setup scripts, schema regeneration, running individual tests, finding test names, debugging logs, and pre-commit checks.
```
msgmaxim: > "run the lending test"

Is this too generic? We don't want this to apply to unit tests, for example.

bouncer/commands/check_localnet_commit.ts
```diff
@@ -0,0 +1,81 @@
...
+      `Running localnet: ${version} → ${localnetOk ? '✓ matches HEAD' : `✗ STALE (${runningHash} ≠ ${head.slice(0, runningHash.length)})`}`,
+    );
+  } catch {
+    logger.warn('Running localnet: not reachable (is it running?)');
+  }
+
+  // Check debug binary
+  let binaryOk = false;
+  try {
+    const binaryVersion = execSync('../target/debug/chainflip-node -V', {
```
msgmaxim: This check the binary file rather than the running process? Would this give incorrect result if you rebuild the binary (but forgot to restart localnet)?

dandanlen: It will also fail if you built a debug binary on another commit but run the localnet on a (newer) release binary, for example.

tbh. not sure this check is really necessary.

msgmaxim (approved):

.claude/skills/bouncer/SKILL.md
```diff
@@ -0,0 +1,238 @@
...
+```bash
+cd bouncer
+./commands/generate_event_schemas.ts   # requires a running localnet
+```
+
+The generator deletes `bouncer/generated/events/` and rewrites it. After regeneration, run `pnpm prettier:write` and commit the diff alongside your pallet changes. Tests importing schemas that no longer exist (e.g. event renamed) need to be updated.
+
+## 6. Debugging a failure
+
+The TRACE-level log file is the source of truth — `stdout` is filtered. The bouncer test wrapper tags every log record with `test` (the test name) and `level` (pino numeric level), so jq queries work out of the box:
```
dandanlen, line 208: It's worth mentioning the indexer DB here, it contains a record of all extrinsics and event, queryable using sql. See the `debug-ci-failure` workflow for general debugging advice (maybe it can be extracted into another skill and referenced from here).

msgmaxim (approved):

PR #6576, opened 2026-05-08
chore: automatically copy ingress-egress's WitnessSafetyMargin to elections' BlockWitnesser settings on every block

MxmUrw:
# Pull Request

Closes: PRO-2842

## Summary

Currently the value for the witness safety margin in our ingress egress pallets (which is reported by the safety_margins rpc) and the actual safety margins used by the election-based witnessing are not connected with each other.

The best way forward is to connect them by using the ingress-egress pallets' safety margin for the elections. This is nicer than vice-versa because it's easier to update the value in the ingress-egress pallet and there's only a single value there. (Compared with the elections which have a separate setting for every BW instance.)

This PR does the following:
 - Properly initializes `ingress_egress_pallet::WitnessSafetyMargin` during the Tron setup migration
 - Adds code to copy the current witness safety margin to all electoral BlockWitnesser settings on every `on_finalize()` call.

NOTE: it sets the safety_margin of every BW *except the Bitcoin EgressWitnessingBW* because for this one we want to keep a safety margin that's independent of the ingress safety margin.

### RPCS

There are no changes to the rpc interface.

### Witnessing

I've checked and the current WitnessSafetyMargin in the ingress-egress pallets for Bitcoin, Arbitrum and Ethereum are the same as the `safety_margin` setting in their respective BWs. This means there will be no sudden change of safety margin after the upgrade.

### Extrinsics & Events

No changes here.

---

* The PR is complete:

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -912,11 +912,16 @@ pub mod pallet {
 	pub type DepositChannelRecycleBlocks<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, ChannelRecycleQueue<T, I>, ValueQuery>;
 
-	// Determines the number of block confirmations is required for a block on
-	// an external chain before CFE can submit any witness extrinsics for it.
+	// Election based witnessing: Determines the number of block confirmations required for a block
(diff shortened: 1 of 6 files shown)


marcellorigotti (approved):
Nice! LGTM

albert-llimos (approved):

PR #6580, opened 2026-05-11
chore: fix rust-base install issues

dandanlen:
# Pull Request

This fixes some issues I encountered while debugging the rust-base image. 

The main 'fix' is that previously, individual apt-get commands could fail silently. Now the `set eux` command:
    - set -e stops on the first failing command
    - set -x shows the exact command that failed
    - set -u catches unset-variable mistakes

I also broke the script up into more caching layers to help with debugging.

### DIFF
--- a/ci/docker/development/rust-base.Dockerfile
+++ b/ci/docker/development/rust-base.Dockerfile
@@ -46,8 +46,8 @@ RUN set -eux; \
     apt-get -o Acquire::Retries=5 -o Acquire::http::Timeout=60 update; \
     curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -; \
     DEBIAN_CODENAME="$(lsb_release -sc)"; \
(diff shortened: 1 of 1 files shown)


ci/docker/development/rust-base.Dockerfile
```diff
@@ -24,28 +22,42 @@ RUN DEBIAN_FRONTEND=noninteractive && export DEBIAN_FRONTEND; \
...
+    npm --version; \
+    npm install -g pnpm; \
+    apt-get clean; \
+    rm -rf /var/lib/apt/lists/*
+
+RUN set -eux; \
+    export DEBIAN_FRONTEND=noninteractive; \
+    apt-get -o Acquire::Retries=5 -o Acquire::http::Timeout=60 update; \
+    curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -; \
+    DEBIAN_CODENAME="$(lsb_release -sc)"; \
```
MxmUrw, line 48: We don't export `DEBIAN_CODENAME` anymore, not sure if this makes a difference.

dandanlen: I don't think so - DEBIAN_CODENAME was introduced in a previous PR for use here in this file, I don't think leaking it to the environment was intentional.

MxmUrw (approved):
LGTM

PR #6582, opened 2026-05-12
fix: Correct accounting in some rare edge cases

msgmaxim:
# Pull Request

Closes: PRO-2859, PRO-2860

## Summary

1. `lending_upkeep` is now transactional per borrower (split the inner code into a separate function `lending_upkeep_for_borrower`). This addresses accounting inconsistency when fees are collected while liquidations are disabled due to safe mode.
2. A slight behaviour change in `write_off_unrecoverable_debt`: in case of bad debt a portion of it may now be absorbed by the pending network fees (likely a very small amount). This addresses accounting inconsistency in an extremely unlikely edge case where all loans are completely lost at 100% utilisation. A more long term solution is to implement PRO-2850 which would let us remove `owed_to_network` completely.

---

## *Checklist*

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-lending-pools/src/general_lending.rs
+++ b/state-chain/pallets/cf-lending-pools/src/general_lending.rs
@@ -1301,6 +1301,50 @@ fn compute_per_asset_liquidation_estimates(
 		.collect()
 }
 
+/// Run a single upkeep tick for one borrower. Wrapped in `#[transactional]` so any
+/// error rolls back the full set of storage writes for this account (interest
+/// collection touches pool and `PendingNetworkFees` storage outside of
+/// `LoanAccounts`, which `try_mutate_exists` does not roll back on its own).
+#[transactional]
+fn upkeep_for_borrower<T: Config>(
+	borrower_id: &T::AccountId,
+	price_cache: &OraclePriceCache<T>,
+	config: &LendingConfiguration,
+	weight_used: &mut Weight,
+) -> DispatchResult {
+	LoanAccounts::<T>::try_mutate_exists(borrower_id, |maybe_account| {
+		let loan_account =
+			maybe_account.as_mut().expect("The caller guarantees that the key exists");
+
+		loan_account.derive_and_charge_interest(weight_used);
+
+		// Some of these may fail due to oracle prices being unavailable, but that's
+		// OK and doesn't need any specific error handling (they will simply be re-tried
+		// at a later point).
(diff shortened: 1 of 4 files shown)


state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1165,6 +1166,50 @@ pub(super) struct AssetCollateralForLoan {
...
+#[transactional]
+fn upkeep_for_borrower<T: Config>(
+	borrower_id: &T::AccountId,
+	price_cache: &OraclePriceCache<T>,
+	config: &LendingConfiguration,
+	weight_used: &mut Weight,
+) -> DispatchResult {
+	LoanAccounts::<T>::try_mutate_exists(borrower_id, |maybe_account| {
+		let loan_account =
+			maybe_account.as_mut().expect("The caller guarantees that the key exists");
```
marcellorigotti:
```
let Some(loan_account) = maybe_account.as_mut() else {
	return Ok(());
};
```
I'd change this since now it's the caller responsibility to be sure the account exists. I know this function is probably gonna be used only here but better be safe than sorry.

marcellorigotti (approved):

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1181,42 +1226,7 @@ pub fn lending_upkeep<T: Config>(current_block: BlockNumberFor<T>) -> Weight {
 	{
 		// Not being able to process a loan account is acceptable (expected when oracle
 		// prices are down).
```
marcellorigotti: nit. or when liquidations are disabled

state-chain/pallets/cf-lending-pools/src/general_lending.rs
marcellorigotti: Otherwise we can use `frame_support::storage::with_storage_layer` and keep this logic inline inside `lending_upkeep`, instead of declaring a separate function, such that the check is gonna be local to the expect call. Don't have a strong opinion here

dandanlen (approved):

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: `with_storage_layer` is what I had before deciding to make it a separate funciton. I will update the code to use `let Some`.

PR #6583, opened 2026-05-12
chore: fix tron Vault swap encoding string addresses

albert-llimos:
# Pull Request

## Summary

We always want to return string Tron addresses on the rpcs. For Vault swaps the `EvmCallDetails` returned when encoding swaps has the "to" and "source_token_address" which were reused from EVM, declared as 20-bytes. Updated to return a String for Tron instead.

I considered using `TronAddress` instead of `String` but the serialize/deserialize are already implemented for the hex addresses, which are the ones used in the engines (21-byte addresses). It seems unnecessary to add more wrappers or types for just this, that's why it's implemented with a `String`.

### DIFF
--- a/bouncer/shared/vault_swap/tron_vault_swap.ts
+++ b/bouncer/shared/vault_swap/tron_vault_swap.ts
@@ -70,7 +70,7 @@ export async function executeTronVaultSwap<A extends WithBrokerAccount>(
   } else {
     // TRC20 vault swap: transfer tokens to vaultSwapDetails.to
     const tokenContractAddress = getEncodedTronAddress(
-      getContractAddress(chainFromAsset(sourceAsset), sourceAsset),
+      '0x' + getContractAddress(chainFromAsset(sourceAsset), sourceAsset),
     );
     if (tokenContractAddress !== vaultSwapDetails.source_token_address)
       throw new Error(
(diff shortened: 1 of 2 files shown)


dandanlen (approved):

bouncer/shared/vault_swap/tron_vault_swap.ts
```diff
@@ -63,16 +63,16 @@ export async function executeTronVaultSwap<A extends WithBrokerAccount>(
...
-      getEncodedTronAddress(vaultSwapDetails.to),
+      vaultSwapDetails.to,
       Number(vaultSwapDetails.value),
       getEncodedTronAddress(pubkey),
     );
   } else {
     // TRC20 vault swap: transfer tokens to vaultSwapDetails.to
     const tokenContractAddress = getEncodedTronAddress(
-      getContractAddress(chainFromAsset(sourceAsset), sourceAsset),
+      '0x' + getContractAddress(chainFromAsset(sourceAsset), sourceAsset).slice(2),
```
nmammeri, line 73: This conversion is hidden here. It might be useful to put this into a function. That can understand tron addresses and do proper translation. 

I've noticed that in utils.ts getContractAddress the Tron CFTESTER has to be specified with 0x. I think we need to have some sort of unified conversion function that is used for the external address conversions.

state-chain/runtime/src/runtime_apis/custom_api/types.rs
```diff
@@ -115,25 +116,27 @@ pub enum VaultSwapDetails<BtcAddress> {
...
 		#[serde(with = "sp_core::bytes")]
 		note: Vec<u8>,
 	},
 }
 
+pub type TronCallDetails = EvmCallDetails<String>;
+
 #[derive(PartialEq, Eq, Clone, Encode, Decode, TypeInfo, Serialize, Deserialize)]
-pub struct EvmCallDetails {
+pub struct EvmCallDetails<Address = sp_core::H160> {
```
nmammeri, line 128: This is great.

PR #6584, opened 2026-05-12
feat: tron deposit monitoring

nmammeri:
# Pull Request

Closes: PRO-2817

## Summary

* This PR adds support for deposit monitoring Tron assets
* Adds a new RPC subscription that allows deposit monitor to subscribe to ingress events
* Refactored the broker level screening tests into separate files for each chain
* Adds broker level screening bouncer tests for Trx and TrxUsdt

### RPCS

* A new RPC subscription is added `subscribe_ingress_events` is added to the broker_api and custom_rpc layer

---

* The PR is complete:
  * Tests:


bouncer/shared/send_trx.ts
```diff
@@ -1,8 +1,9 @@
 import { getEncodedTronAddress, getTronWebClient, getTronWhaleKeyPair } from 'shared/utils';
 import { Logger } from 'shared/utils/logger';
 import { TronWeb } from 'tronweb';
+import { HexString } from '@polkadot/util/types';
 
-export async function sendTrx(logger: Logger, toAddress: string, amount: string): Promise<void> {
+export async function sendTrx(logger: Logger, toAddress: string, amount: string): Promise<string> {
```
MxmUrw: To be consistent with `send_trc20.ts` we could return `Promise<HexString>` here, I think.

state-chain/chains/src/address.rs
```diff
@@ -561,7 +561,7 @@ impl ToHumanreadableAddress for ForeignChainAddress {
 }
 
 #[cfg(feature = "std")]
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
MxmUrw, line 564: Out of curiosity: what are these new bounds required for?

state-chain/chains/src/lib.rs
```diff
@@ -1088,7 +1088,18 @@ pub trait FeeRefundCalculator<C: Chain> {
...
+	TypeInfo,
+	Encode,
+	Decode,
+	DecodeWithMemTracking,
+	PartialEq,
+	Eq,
+	Serialize,
+	Deserialize,
+)]
 pub enum TransactionInIdForAnyChain {
```
MxmUrw: I'm not sure whether we should expose this type on the rpc layer. (See other comment).

state-chain/custom-rpc/src/broker.rs
```diff
@@ -526,6 +529,27 @@ where
...
 	}
 
+	async fn subscribe_ingress_events(
+		&self,
+		pending_sink: PendingSubscriptionSink,
+		chain: ForeignChain,
+	) {
+		self.rpc_backend
+			.new_subscription(
+				NotificationBehaviour::Finalized,
```
MxmUrw: Actually I'm not sure we want finalized here. Why not unfinalized? Then we'll have a 6 seconds head start.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -2728,6 +2731,27 @@ where
...
 	}
 
+	async fn cf_subscribe_ingress_events(
+		&self,
+		pending_sink: PendingSubscriptionSink,
+		chain: ForeignChain,
+	) {
+		self.rpc_backend
+			.new_subscription(
+				NotificationBehaviour::Finalized, /* only_finalized */
```
MxmUrw: Also here, would Unfinalized work?

state-chain/runtime/src/runtime_apis/custom_api/types.rs
```diff
@@ -653,6 +655,45 @@ pub enum RawWitnessedEvents {
...
+pub enum DepositDetails {
+	Bitcoin(<Bitcoin as Chain>::DepositDetails),
+	Ethereum(<Ethereum as Chain>::DepositDetails),
+	Arbitrum(<Arbitrum as Chain>::DepositDetails),
+	Tron(<Tron as Chain>::DepositDetails),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, TypeInfo, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct DepositWitnessInfo {
```
MxmUrw: nit: there seems to be some inconsistency how (de-)serialization is handled: we feature gate it for `DepositWitnessInfo` (and the two structs below it) but we don't for `DepositDetails`.

state-chain/runtime/src/runtime_apis/custom_api/types.rs
```diff
@@ -653,6 +655,45 @@ pub enum RawWitnessedEvents {
...
+	#[cfg_attr(feature = "std", serde(serialize_with = "serialize_as_hex"))]
+	pub amount: AssetAmount,
+	pub asset: Asset,
+	pub deposit_details: DepositDetails,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, TypeInfo, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct VaultDepositWitnessInfo {
+	pub tx_id: TransactionInIdForAnyChain,
```
MxmUrw: This `TransactionInIdForAnyChain` wasn't previously used on our rpc layer (it didn't have the ser/de bounds and it looks a bit like its purpose is for testing). We *do* already have a similar type `TransactionInId` in `broker.rs` that's used by the DM (for submitting the `mark_transaction_for_rejection` extrinsic).

What do you think about using that one here? Then we wouldn't have two slightly different types for the same purpose exposed on the rpc layer.

MxmUrw (commented):
Very nice, also thanks for splitting the deposit_monitor tests! A few comments:

state-chain/custom-rpc/src/broker.rs
albert-llimos: I agree, assuming that works. The downside of an unlikely reorg for this seems very low / negligible.

state-chain/chains/src/address.rs
nmammeri: This was needed when I represented tron addresses using AddressString. I agree it is not needed by this PR anymore but I feel since it wraps a String it makes sense to derive PartialEq, Eq for future use.

state-chain/custom-rpc/src/broker.rs
nmammeri: I agree we can change this to unfinalized to gain some screening time.

state-chain/runtime/src/runtime_apis/custom_api/types.rs
nmammeri: good catch, I feel we should always gate it through std feature flag. It looks like other types in runtime types need to be changed accordingly as well but I think it is out of the scope of this PR

state-chain/runtime/src/runtime_apis/custom_api/types.rs
nmammeri: Both serve the same kind of purpose but not quite. The `TransactionInId` have 2 variants for Solana.  I used TransactionInIdForAnyChain because it is defined as a primitive and not as an RPC type. To me it looked like TransactionInId is geared for `mark_transaction_for_rejection` rpc call

MxmUrw (approved):
LGTM!

state-chain/custom-rpc/src/lib.rs
```diff
@@ -2715,9 +2718,9 @@ where
 	) {
 		self.rpc_backend
 			.new_subscription(
-				NotificationBehaviour::Finalized, /* only_finalized */
-				false,                            /* only_on_changes */
-				true,                             /* end_on_error */
+				NotificationBehaviour::Finalized,
```
albert-llimos, line 2721: Is this "Finalized" on purpose? Since the other ones have now been changed to "Best".

albert-llimos (approved):

state-chain/custom-rpc/src/lib.rs
nmammeri: Yes for the ingress events we want to get some head start hence we went for Best blocks but for screening events we want this to be Finalized

PR #6585, opened 2026-05-12
chore: add separate retry policy to retry the next block 

marcellorigotti:
# Pull Request

Closes: PRO-2867

## Summary

Add a separate `RetryPolicy` to be used in case of broadcast failure.
Keep the old policy, which delays for 10 blocks, for the other scenarios:
- no available authority
- waiting for broadcast barrier 

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/chains/src/lib.rs
+++ b/state-chain/chains/src/lib.rs
@@ -1441,6 +1441,15 @@ impl RetryPolicy for DefaultRetryPolicy {
 		Some(10u32)
 	}
 }
+pub struct RetryNextBlockPolicy;
+impl RetryPolicy for RetryNextBlockPolicy {
+	type BlockNumber = u32;
+	type AttemptCount = u32;
+
+	fn next_attempt_delay(_retry_attempts: Self::AttemptCount) -> Option<Self::BlockNumber> {
+		Some(1u32)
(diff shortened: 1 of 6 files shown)


MxmUrw (approved):
LGTM

state-chain/chains/src/lib.rs
```diff
@@ -1441,6 +1441,15 @@ impl RetryPolicy for DefaultRetryPolicy {
 		Some(10u32)
 	}
 }
+pub struct RetryNextBlockPolicy;
+impl RetryPolicy for RetryNextBlockPolicy {
+	type BlockNumber = u32;
+	type AttemptCount = u32;
+
+	fn next_attempt_delay(_retry_attempts: Self::AttemptCount) -> Option<Self::BlockNumber> {
+		Some(1u32)
```
dandanlen: On reflection, can we add a simple back-off mechanism? 

Like `if retry_attempts > 10 { 10 } else { 1 }`

This way if slow down the retry loop if there's an issue affecting multiple validators.

marcellorigotti: makes sense!

dandanlen (approved):

PR #6588, opened 2026-05-13
chore: remove serde derivations for `types/before_version_*.rs`.

MxmUrw:
# Pull Request

Part of PRO-2868

## Summary

This removes serde derivations from `types/before_version_16.rs` and `types/before_version_17.rs` (the others don't have them). The reason is that we don't use the serde implementations at all for historical types. These types only exist to provide backwards compatibility at the runtime api interface.

## Effects on runtime size

Removing these serde impls results in the runtime (build with `cf-build-release`) to be 3.5KB smaller. Which is a bit surprising, I would have assumed that these (unused) impls could have been optimized away by the compiler.
```
mxu@Maxims-Macbook chainflip-backend-21 % ls -la target/release/wbuild/state-chain-runtime/state_chain_runtime.compact.compressed.wasm
-rw-r--r--  1 mxu  staff  4744434 May 13 11:45 target/release/wbuild/state-chain-runtime/state_chain_runtime.compact.compressed.wasm
mxu@Maxims-Macbook chainflip-backend-21 % ls -la target/release/wbuild/state-chain-runtime/state_chain_runtime.compact.compressed.wasm 
-rw-r--r--  1 mxu  staff  4750980 May 13 11:59 target/release/wbuild/state-chain-runtime/state_chain_runtime.compact.compressed.wasm
```

## API Changes

None

---

* The PR is complete:

### DIFF
--- a/state-chain/runtime/src/runtime_apis/custom_api/types/before_version_16.rs
+++ b/state-chain/runtime/src/runtime_apis/custom_api/types/before_version_16.rs
@@ -1,7 +1,7 @@
 use super::*;
 use codec::{DecodeWithMemTracking, MaxEncodedLen};
 
-#[derive(Encode, Decode, TypeInfo, Serialize, Deserialize, Clone)]
+#[derive(Encode, Decode, TypeInfo, Clone)]
 pub struct VaultAddresses {
 	pub ethereum: EncodedAddress,
 	pub arbitrum: EncodedAddress,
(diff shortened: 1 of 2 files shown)


dandanlen (approved):

dandanlen:
> Removing these serde impls results in the runtime (build with cf-build-release) to be 3.5KB smaller. Which is a bit surprising, I would have assumed that these (unused) impls could have been optimized away by the compiler.

Interesting!

We used to feature-gate serde derives to generate them for `std` only, but stopped doing this when serde added no-std support, under the assumption that the impls would be pruned by the compiler... might be worth digging into.

PR #6596, opened 2026-05-18
refactor: zero price defense

j4m1ef0rd:
# Pull Request

Closes: PRO-2848

## Summary

Added defensive programming around 0 price. Not an active threat because it would require consensus on a 0 oracle price.

---

* The PR is complete:
  * Tests:


state-chain/amm-math/src/lib.rs
```diff
@@ -316,17 +316,22 @@ impl TryFrom<U512> for SqrtPrice {
...
-	mul_div(a, b, c).1
+pub fn mul_div_ceil_checked<C: Into<U512>>(a: U256, b: U256, c: C) -> Option<U256> {
+	let c: U512 = c.into();
+	if c.is_zero() {
+		return None
+	}
+	let (d, m) = U512::div_mod(U256::full_mul(a, b), c);
+	let ceil = if m.is_zero() { d } else { d + U512::from(1) };
+	ceil.try_into().ok()
 }
```
MxmUrw, line 338: It seems like we're duplicating code from `mul_div()` here, why did that become necessary?

state-chain/amm-math/src/lib.rs
```diff
@@ -454,33 +462,37 @@ impl Price {
...
+	pub fn divide_by(self, price: Price) -> Option<Self> {
+		if self.is_zero() || price.is_zero() {
+			return None
+		}
+		Some(Price(mul_div_floor_checked(self.0, Self::one().0, price.0)?))
 	}
 
 	pub fn multiply_by(self, price: Price) -> Self {
-		Price(mul_div_floor(self.0, price.0, Self::one().0))
+		Price(mul_div_floor_checked(self.0, price.0, Self::one().0).unwrap())
```
dandanlen: Is the unwrap warranted here? This can still overflow, right?

state-chain/amm/src/limit_orders/tests.rs
```diff
@@ -28,6 +28,14 @@ use rand::{Rng, SeedableRng};
 type LiquidityProvider = cf_primitives::AccountId;
 type PoolState = super::PoolState<LiquidityProvider>;
 
+fn mul_div_ceil(a: U256, b: U256, c: U256) -> U256 {
+	mul_div_ceil_checked(a, b, c).unwrap()
```
dandanlen, line 32: Similar comment here: whether or not we can unwrap this is dependent on the inputs, so we should do one of:

- return an option or error and allow the call site reason about whether or not overflow/underflow is possible, and add `unwrap` or `expect` as appropriate,  based on assumptions in the *calling* context.
- keep the panic, but state clearly in a doc comment which assumptions must hold, and under which conditions the function *might* panic.

j4m1ef0rd: This one is only for test code, So i left the unwrap. All other places I changed to the checked version of the function and handled the error as needed.

msgmaxim (approved):

state-chain/amm-math/src/lib.rs
```diff
@@ -454,33 +455,37 @@ impl Price {
 
 	/// Compute the price of asset 1 (self) in terms of asset 2 (given).
 	/// Both prices must have the same quote asset (eg. USD).
-	pub fn divide_by(self, price: Price) -> Self {
-		Price(mul_div_floor(self.0, Self::one().0, price.0))
+	/// Returns None if either price is zero.
+	pub fn divide_by(self, price: Price) -> Option<Self> {
+		if self.is_zero() || price.is_zero() {
+			return None
+		}
```
MxmUrw: Is this check necessary here?
 - I think `mul_div_floor_checked` does the check for `price != 0` already
 - and if `self == 0 && price != 0` we don't even want to error out but return 0, no?

state-chain/amm-math/src/lib.rs
```diff
@@ -454,33 +455,37 @@ impl Price {
...
 
-	pub fn multiply_by(self, price: Price) -> Self {
-		Price(mul_div_floor(self.0, price.0, Self::one().0))
+	pub fn multiply_by(self, price: Price) -> Option<Self> {
+		Some(Price(mul_div_floor_checked(self.0, price.0, Self::one().0)?))
 	}
 
 	pub fn output_amount_floor<I: Into<U256>>(self, input: I) -> Amount {
-		mul_div_floor(input.into(), self.0, Self::one().0)
+		mul_div_floor_checked(input.into(), self.0, Self::one().0).unwrap_or_default()
```
MxmUrw: It feels like returning 0 when mathematical operations fail is dangerous, as the caller can't distinguish whether it was a failure or a real result. Shouldn't we return an option here, as it will be much more obvious at the call-site whether 0 is approprate or not?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -3395,24 +3398,25 @@ pub mod pallet {
 								Self::estimate_usdc_price_using_simulated_swap_or_fallback(
 									asset, side,
 								); // USDC / Asset
-							asset_price.multiply_by(usdc_price) // USD / Asset
+							asset_price.multiply_by(usdc_price).unwrap_or(asset_price) // USD / Asset
```
MxmUrw, line 3401: Does this line mean that if `usdc_price = 0` we multiple `asset_price * asset_price`?

MxmUrw (approved):
LGTM, but one last thing that we should discuss:

state-chain/pallets/cf-swapping/src/lib.rs
j4m1ef0rd: No, it will use asset_price as the value if the multiply overflows. As we expect the udsc price to be close to 1, it's fine to ignore the multiply (even though it should never overflow).

MxmUrw: Right, I see. Well then it looks good to go!

PR #6597, opened 2026-05-18
fix: prevent use of expired btc deposit channels

dandanlen:


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1766,10 +1766,17 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-				DepositChannelPool::<T, I>::insert(
-					deposit_channel.channel_id,
-					DepositChannel { state, ..deposit_channel },
-				);
+				if !T::AddressDerivation::is_channel_state_current(&state) {
+					log_or_panic!(
+						"Logic error: Attempting to recycle a channel that is no longer valid: {:?}",
+						state
+					);
+				} else {
```
MxmUrw: Could the `is_channel_state_current()` be part of the `maybe_recycle()` check? Then we wouldn't have two nested if statements here and it seems to be semantically fitting?

state-chain/chains/src/address.rs
```diff
@@ -62,6 +62,11 @@ pub trait AddressDerivationApi<C: Chain> {
 		source_asset: C::ChainAsset,
 		channel_id: ChannelId,
 	) -> Result<(C::ChainAccount, C::DepositChannelState), AddressDerivationError>;
+
+	/// Whether a (possibly pre-allocated or recycled) channel state is spendable by the network.
+	fn is_channel_state_current(_channel_state: &C::DepositChannelState) -> bool {
+		true
+	}
```
MxmUrw: This *could* be part of the `LifeCycleHooks` trait, no?

MxmUrw (commented):
Looks good; not sure if the alternative approach I mention below is worth it.

state-chain/pallets/cf-ingress-egress/src/lib.rs
dandanlen: Yes I agree, there should be some way to fit this into `LifecycleHooks`, that's a much better fit.

dandanlen: Actually, the way the traits are currently organised, it's not quite straightforward: the LifecycleHooks are part of the `Chain` trait (so they cannot be 'effectful': they can't easily access chain state). 

Nonetheless, `AddressDerivation` is clearly not the right place for this either. I'll find another way.

state-chain/pallets/cf-ingress-egress/src/mocks.rs
```diff
@@ -109,11 +111,18 @@ impl AddressDerivationApi<Bitcoin> for MockAddressDerivation {
...
 			<Self as AddressDerivationApi<Bitcoin>>::generate_address(source_asset, channel_id)?,
-			DepositAddress::new([1u8; 32], 123),
+			DepositAddress::new(MockBtcActiveKey::get(), 123),
 		))
 	}
 }
 
+pub struct MockBtcChannelFreshness;
+impl DepositChannelFreshness<Bitcoin> for MockBtcChannelFreshness {
+	fn is_current(state: &<Bitcoin as Chain>::DepositChannelState) -> bool {
```
MxmUrw: nit: it might be worth calling this function `is_fresh()`, then the trait name and the method are syntactically linked.

MxmUrw (approved):
Yeah I think this is nicer!

PR #6598, opened 2026-05-18
fix: prevent use of expired btc deposit channels

dandanlen:
Cherry-picked from release 2.1.20


dandanlen:
Converted back to draft - I need to update this.

dandanlen:
This was replaced and merged as part of #6624

PR #6599, opened 2026-05-19
[codex] fix stale bitcoin broadcast signing payloads

zackben1git:
## Summary

- Detect Bitcoin broadcasts whose signed payload selects `Current`/`Previous` keys that do not match the deposit UTXO keys being spent.
- Recompute Bitcoin signing key selection against the current on-chain aggregate key before retrying, so broadcasts can be re-signed instead of repeatedly rebroadcasting an invalid transaction.
- Add focused unit coverage for the key-rotation case that can otherwise produce a Schnorr signature for the wrong Taproot script key.

## Root cause

Bitcoin retry handling assumed signed BTC transactions never need signature refresh. That is true for stable inputs, but it misses a key-rotation edge case where a transaction can be constructed with one aggregate-key view and signed after rotation with another. In that case the signature can be valid for the threshold signer payload while still not matching the deposit address key embedded in the Bitcoin input script.

## Validation

- `cargo test -p cf-chains detects_and_refreshes_stale_bitcoin_signing_key_selection`
- `cargo test -p state-chain-runtime btc_retry_refreshes_stale_key_selection_after_rotation`
- `cargo test -p cf-chains btc::test::test_payloads`
- `cargo test -p cf-chains btc::test::test_finalize`
- `git diff --check`

### DIFF
--- a/state-chain/chains/src/btc.rs
+++ b/state-chain/chains/src/btc.rs
@@ -828,6 +828,25 @@ fn extend_with_inputs_outputs(bytes: &mut Vec<u8>, inputs: &[Utxo], outputs: &[B
 }
 
 impl BitcoinTransaction {
+	fn old_utxo_input_indices_for(agg_key: &AggKey, inputs: &[Utxo]) -> VecDeque<u32> {
+		(0..)
+			.zip(inputs)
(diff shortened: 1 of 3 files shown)


dandanlen:
Thanks for the contribution but unfortunately this doesn't fix the underlying issue.

PR #6601, opened 2026-05-19
feat: introduce migration system

MxmUrw:
# Pull Request

Part of: PRO-2868
Closes: PRO-2907
Closes: PRO-2771

## Summary

This PR introduces auto-generated migrations. The goal is that all boiler-plate that's involved in writing migrations should be auto-generated. Only the actually important information (e.g. which fields are added / removed / changed) has to be stated. This targets both migrations between runtime versions as well as "migrations" on the RPC layer, where responses of runtime API queries from older runtimes have to be "migrated" to the latest version of that respective type before they can be returned by the RPC method.

This is the new architecture:
 - There's a new trait `HasChangelog` which is the main trait that all the other functionality is centered around. It has "changelog entries" for every runtime version upgrade (`in_20100`, `in_20200`, ...). Every changelog entry is effectively the migration that happens to this type in this runtime version (can be an identity migration of course). An important feature is that all of these entries are optional. If nothing is specified, they are populated with a default value that always has to be provided separately (`if_unspecified`).
 - The trait `HasChangelog` (and a few helpers around it) are generated by a macro called `define_all_released_runtime_versions!()` living in `cf-utilities/without_std/migrations.rs`. Whenever the runtime version is bumped (i.e. we branch release and then increment the runtime version), the call to this macro has to be updated to include that newly released version. Due to the "default value" for all changelog entries described above, **no other** code changes are required.
 - There is a new proc macro `#[cf_proc_macros::generate_module]` which can be attached to a struct and generates a helper module (with the same name except prepended with an underscore) that contains helpers for writing the `HasChangelog` implementation for that struct, in particular a default migration that just inherits all migrations from all of its fields. The proc macro is a thin wrapper around the macro `generate_module!` in `cf_utilities/without-std/generate_module`.
 - There are implementations of `HasChangelog` for primitive types and containers in `utilities/src/without_std/migrations/primitives.rs`.
 - For a type `T` that implements `HasChangelog`, the following is available:
   - All historical versions of that type, accessible through `<T as HasVersion<{VERSION}>>::HistoricalType`
   - Migrations in both directions between any of the historical versions and the current version through `migrate_from_historical_type()` and `migrate_to_historical_type()`.

---

The following notable types have been given a `HasChangelog` implementation:
1. The generic `any::AssetMap` as well as all chain-specific asset maps.
2. `NetworkFeeDetails` and `NetworkFees` (both inheriting from `AssetMap`)
3. `DelegationInfo` (doesn't have migrations)
4. `RpcAccountInfoCommonItems` (includes multiple fields containing`AssetMap`)

These types now don't require a manual definition of their historical variants. Previously there was a redefinition of all the asset maps in `types/before_version_16.rs` and `types/before_version_17.rs` as well as all the rpc types that contained them. Now, for example for `NetworkFees`, the historical definition is no longer required and is replaced by:
```rs
pub type NetworkFees = <super::NetworkFees as HasVersion<v20100>>::HistoricalType;
``` 
The implementation of `before_version_17::NetworkFees::Into` (for converting to the current version within a historical RPC call) is no longer required. It's possible to call `migrate_from_historical_type()` instead.

---

Additionally this PR also:
 - Adds a script to download (historical) metadata in v15 format from live networks for any given runtime version. If the runtime at that point in time didn't have entries for runtime api calls this won't make them magically appear though. So only useful for runtime versions >= 20100.
 - Adds historical metadata for 20119 and 20203. Files have to be located in `state-chain/cf-integration-tests/historical_metadata/`.
 - Adds a new system for testing historical types (lives in cf-integration-tests) by en- and decoding against historical representations:
   - against TypeInfo extracted from runtime API call definitions in historical metadata (works for versions >= 20100)
   - by runtime API queries at historical block hashes against a live archive node (works for all versions)

## Effects on `state_chain_runtime` and `chainflip-node` binary size

Since auto-generating code might potentially affect binary sizes, I've checked the fully-rebased PR against main. The results are a bit surprising since in fact the binary sizes were slightly reduced (!). Comparisons (first entry is *main*, second entry is *this PR*:
```
-rwxr-xr-x     1 mxu  staff  139004368 Jun 17 14:50 chainflip-node
-rwxr-xr-x     1 mxu  staff  139001664 Jun 17 14:40 chainflip-node
```

```
-rw-r--r--  1 mxu  staff   4753238 Jun 17 14:46 state_chain_runtime.compact.compressed.wasm
-rw-r--r--  1 mxu  staff   4725164 Jun 17 14:36 state_chain_runtime.compact.compressed.wasm
```
Quoting GPT, take it with a grain of salt:
> Bottom line: the PR adds a migration framework, but for these RPC historical return values it removes duplicated concrete historical structs and conversion implementations. The generic aliases/migrations give LLVM and Wasm compaction more opportunity to share/eliminate code, producing about 126 KB less Wasm code and 50 KB less name metadata before compression.

## API Changes

There are no API changes.

### Runtime API

All migrated runtime calls have been tested against their historic counterparts, so there shouldn't be any visible changes.

### RPCS

RPC methods haven't been migrated. They still use the same types as they did previously.

### Extrinsics & Events

No changes.

---

* The PR is complete:
  * Tests:


utilities/src/without_std/generate_module.rs
```diff
@@ -0,0 +1,241 @@
...
+                $( #[$($Field_Attributes)*])*
+                $field_vis $field: $field_ty,
+            )*
+        }
+
+        pub mod $mod {
+            #![allow(nonstandard_style)]
+            #![allow(unused)]
+
+            use super::*;
```
dandanlen, line 43: In general, it's better to use absolute paths in macros, or at last restrict the scope to specific values.

We don't know for sure what will be available in `super::*`.

For example this code assumes the presence of some external crates like `derive_where`, `codec` that aren't guaranteed to be in the calling context at all (and might have unexpected versions etc.), which can lead to confusing errors.

The common pattern for this to have a `__reexports` module in the root of the macro-defining crate and import it in the module as `$crate::__reexports::*`.

utilities/src/without_std/migrations/basics.rs
```diff
@@ -0,0 +1,157 @@
...
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// ---------- definition of migrations ------------
+
+use crate::migrations::HasChangelog;
+
+pub trait Version: Copy {
+	const CANONICAL_RUNTIME_PATCH_VERSION_FOR_COMPATIBILITY_TEST: Option<CanonicalPatchVersion>;
```
dandanlen, line 22: Can this version be `None`? What would that imply?

MxmUrw: There's a struct `vCurrent` that isn't a real runtime version but is an implementation detail required as target version of the migration from the "generic version" of a struct to the real struct itself. The `Version` impl of that struct sets the patch version to `None`.

An alternative to this option here would be to introduce a third variant to `CanonicalPatchVersion` which is called `NotApplicable`.

dandanlen: Never mind i see there's an example below.

dandanlen (approved):

utilities/src/without_std/generate_module.rs
MxmUrw: True, seems like a useful pattern. I think I'll do this in a follow-up PR though (after auto-generation for enums is done) because there are a few more macros that will be added. Issue exists: https://linear.app/chainflip/issue/PRO-2958/properly-re-export-all-used-imports-through-a-reexports-module

PR #6603, opened 2026-05-19
feat: Flip 2.1

ramizhasan111:
# Pull Request

Closes: PRO-2862, PRO-2863

## Summary

This feature changes the tokenomics from emissions and "buy and burn" model to fixed supply and "buy and distribute" model. The can be scheduled to be activated at any future epoch by governance (`Update_pallet_config` extrinsic).

- The onchain flip revenue (flip tx fees, slashing stc) is stored in in onchain reserve (`Reserve` storage item with `ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID` as reserve id)
- The offchain flip revenue (swap fees) is stored in `FlipToDistribute` storage item.
- The rewards for the whole epoch are distributed at the end of the epoch together to all validators and delegators.
- The offchain flip revenue is bridged in at the time of reward distribution and the subsequent transfer of these bridged flip from the vault to the State Chain gateway is attempted, and is scheduled for later if it cant be carried out immediately (due to the amount not big enough relative to how much eth tx fees it would cause).
- unit tests, integration tests and bouncer test was added to assert soundness of funds distribution and expected behaviour.

### RPCS

`cf_reward_distribution_estimate` rpc is added for product to be able to get a real time estimate of rewards accumulated for validators and delegators from the start of the epoch up until that moment.

### Extrinsics & Events

- A new `FlipDistributed` event was created which will be emitted at epoch boundary which will contain info about rewards that were distributed to validators and delegators.
- the `update_pallet_config` extrinsic in the flip pallet was modified to allow setting of the Flip 2.1 activation epoch.

---

* The PR is complete:
  * Tests:
    * [x] Bouncer test

### DIFF
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -572,7 +572,7 @@ impl<T: Config> Pallet<T> {
 		}
 
 		Self::deposit_event(Event::FlipDistributed { amount: flip_minted_list });
-		return flip_to_distribute.into();
+		flip_to_distribute.into()
 	}
 }
 
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -109,7 +109,7 @@ pub mod pallet {
 	/// A 4-byte identifier for different reserves.
 	pub type ReserveId = [u8; 4];
 
-	pub const ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID: ReserveId = *b"FEES";
+	pub(crate) const ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID: ReserveId = *b"FEES";
 
 	#[pallet::config]
 	#[pallet::disable_frame_system_supertrait_check]
@@ -607,6 +607,15 @@ impl<T: Config> Pallet<T> {
 			amount: flip_distributed_map.into_iter().collect(),
 		});
 		offchain_flip_bridged
+	}
+
+	/// Fee rewards pending distribution to authorities: the balance of the on-chain distribution
+	/// reserve plus any off-chain accumulated fees (which may be negative).
+	pub fn pending_rewards() -> i128 {
+		Reserve::<T>::get(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID)
+			.try_into()
+			.unwrap_or(i128::MAX)
+			.saturating_add(FlipToDistribute::<T>::get())
 	}
 
 	/// Whether FLIP 2.1 is active: fee rewards are accumulated for distribution to authorities
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -33,7 +33,8 @@ use scale_info::TypeInfo;
 pub use weights::WeightInfo;
 
 use cf_traits::{
-	AccountInfo, Bonding, DeregistrationCheck, FeePayment, FundingInfo, Issuance, Slashing,
+	AccountInfo, Bonding, DeregistrationCheck, FeePayment, FundingInfo, Issuance,
+	RewardsDistribution, Slashing,
 };
 use imbalances::{Deficit, ImbalanceSource, Surplus};
 
@@ -46,7 +47,7 @@ use frame_support::{
 	ensure,
 	pallet_prelude::*,
 	sp_runtime::{
-		traits::{AtLeast32BitUnsigned, Zero},
+		traits::{AtLeast32BitUnsigned, Hash as HashT, Zero},
 		DispatchError, Permill, RuntimeDebug,
 	},
 	traits::{Get, Imbalance, OnKilledAccount, SignedImbalance},
@@ -121,6 +122,12 @@ pub mod pallet {
 
 		type CallIndexer: CallIndexer<<Self as frame_system::Config>::RuntimeCall>;
 
+		/// An implementation of `RewardsDistribution` defining how to distribute rewards.
+		type RewardsDistribution: RewardsDistribution<
+			Balance = Self::Balance,
+			AccountId = Self::AccountId,
+		>;
+
 		/// Required in order to inject a HoldReason for impls in [substrate_impls]
 		type RuntimeHoldReason: Encode + TypeInfo + 'static;
 	}
@@ -167,6 +174,9 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type FeeScalingRate<T: Config> = StorageValue<_, FeeScalingRateConfig, ValueQuery>;
 
+	#[pallet::storage]
+	pub type FlipToDistribute<T: Config> = StorageValue<_, i128, ValueQuery>;
+
 	#[pallet::event]
 	#[pallet::generate_deposit(pub fn deposit_event)]
 	pub enum Event<T: Config> {
@@ -193,6 +203,9 @@ pub mod pallet {
 		BondUpdated {
 			account_id: T::AccountId,
 			new_bond: T::Balance,
+		},
+		FlipDistributed {
+			amount: Vec<(T::AccountId, T::Balance)>,
 		},
 	}
 
@@ -515,6 +528,51 @@ impl<T: Config> Pallet<T> {
 				amount: slash_amount,
 			})
 		}
+	}
+
+	pub fn trigger_flip_reward_distribution(authorities: Vec<T::AccountId>) -> T::Balance {
+		let flip_to_distribute = FlipToDistribute::<T>::take();
+		if flip_to_distribute <= Zero::zero() {
+			FlipToDistribute::<T>::put(flip_to_distribute);
+			return Zero::zero()
+		}
+		let flip_to_distribute: u128 =
+			flip_to_distribute.try_into().expect("checked for negative number above");
+
+		let count = authorities.len();
+		let count_u128 = count as u128;
+		let (per_authority_reward, remainder) =
+			(flip_to_distribute / count_u128, flip_to_distribute % count_u128);
+
+		let winner_authority = if remainder > 0 {
+			let parent_hash = frame_system::Pallet::<T>::parent_hash();
+			let seed =
+				<T as frame_system::Config>::Hashing::hash_of(&(parent_hash, b"flip_distribution"));
+			let random_num = u64::from_le_bytes(seed.as_ref()[..8].try_into().unwrap_or([0u8; 8]));
+			let winner_idx = (random_num % count as u64) as usize;
+			Some(&authorities[winner_idx])
+		} else {
+			None
+		};
+
+		let mut flip_minted_list = vec![];
+		for authority in &authorities {
+			T::RewardsDistribution::distribute(
+				if winner_authority.is_some_and(|wa| authority == wa) {
+					(per_authority_reward + remainder).into()
+				} else {
+					per_authority_reward.into()
+				},
+				authority,
+				|account, amount| {
+					Pallet::<T>::settle(account, Pallet::<T>::bridge_in(amount).into());
+					flip_minted_list.push((account.clone(), amount));
+				},
+			);
+		}
+
+		Self::deposit_event(Event::FlipDistributed { amount: flip_minted_list });
+		return flip_to_distribute.into();
 	}
 }
 
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -30,10 +30,12 @@ pub mod substrate_impls;
 
 pub mod weights;
 use scale_info::TypeInfo;
+use sp_runtime::traits::Saturating;
 pub use weights::WeightInfo;
 
+use cf_primitives::ACCUMULATE_REWARDS_EPOCH_START;
 use cf_traits::{
-	AccountInfo, Bonding, DeregistrationCheck, FeePayment, FundingInfo, Issuance,
+	AccountInfo, Bonding, DeregistrationCheck, EpochInfo, FeePayment, FundingInfo, Issuance,
 	RewardsDistribution, Slashing,
 };
 use imbalances::{Deficit, ImbalanceSource, Surplus};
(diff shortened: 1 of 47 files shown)


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -315,6 +315,8 @@ pub mod pallet {
 
 	/// A map between an epoch and the set of authorities (participating in this epoch).
 	#[pallet::storage]
+	#[pallet::getter(fn historical_authorities)]
```
dandanlen: nit: getters are deprecated.

dandanlen:
I think we might as well combine both PRs into one - there's not real need for them to be separate, neither is independent or complete without the other.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -923,7 +924,15 @@ pub mod pallet {
...
-						T::Flip::burn_offchain(amount.into());
+						if T::EpochInfo::epoch_index() >=
+							<T::Flip as FeePayment>::fee_rewards_activation_epoch()
+						{
+							T::Flip::bridge_in_and_add_to_onchain_flip_to_be_distributed(
+								amount.into(),
+							);
+						} else {
+							T::Flip::burn_offchain(amount.into());
+						}
```
dandanlen: We're still leaking this abstraction / repaeting this logic in multiple places. This can all be contained inside the FLIP pallet. Also instead of exposing `fee_rewards_activation_epoch()` we can add `is_flip_2_1_activated()` for the remaining places where we need it outside the flip pallet (pretty much just the swapping pallet). (see #6702 for a suggested change)

state-chain/runtime/src/chainflip/epoch_transition.rs
```diff
@@ -24,7 +24,22 @@ impl EpochTransitionHandler for ChainflipEpochTransitions {
...
 	}
-	fn on_new_epoch(_new: EpochIndex) {
+	fn on_new_epoch(new: EpochIndex) {
 		AssetBalances::trigger_reconciliation();
+
+		let activation_epoch = pallet_cf_flip::FeeRewardsActivationEpoch::<crate::Runtime>::get();
+		if new == activation_epoch {
+			Emissions::burn_and_broadcast_supply_update(
+				frame_system::Pallet::<crate::Runtime>::block_number(),
+			);
```
dandanlen: Note that we have a condition in the emissions pallet that will only trigger the butn_and_broadcast if the balance is high enough.

We might want to add a new argument and make it `burn_and_broadcast_supply_update(block, force)` so we can force the broadcast unconditionally.

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -694,7 +695,7 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type FlipToBeSentToGateway<T: Config> = StorageValue<_, AssetAmount, ValueQuery>;
 
-	/// Interval at which we buy FLIP in order to burn it.
+	/// Interval at which we buy FLIP in order from swap fee in USDC.
```
dandanlen: nit: not sure what this comment is trying to say

state-chain/pallets/cf-flip/src/tests.rs
dandanlen, line 1: We are testing the distribution, which is fine, but we aren't testing any of the interface methods (ie. to check that behaviour before/after actually switches, or that they write to the correct storage). Rather than writing directly to `FlipToDistribute` storage it would be better to use the fee payment trait for example.

state-chain/pallets/cf-flip/src/tests.rs
```diff
@@ -719,3 +719,153 @@ mod transfer {
...
+				let _ = Flip::mint(i64::MAX as u128).offset(Flip::bridge_out(i64::MAX as u128));
+
+				FlipToDistribute::<Test>::put(flip_to_distribute);
+				Reserve::<Test>::insert(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID, onchain_reserve);
+
+				let expected_bridged: u128 =
+					if flip_to_distribute > 0 { flip_to_distribute as u128 } else { 0 };
+				let total_reserve = onchain_reserve + expected_bridged;
+				let expected_per_authority = total_reserve / COUNT;
+				let expected_remainder = total_reserve % COUNT;
```
dandanlen, line 901: We are just testing that the implementation matches itself (ie. we're deriving expected values simply by repeating how we implemented it). This doesn't catch more subtle bugs if the implementation is *wrong*. A better (or additional) invariant might be conservation of funds: we add up all the validator balances and the reward pot before and after and make sure it's the same. 

Also fwiw we generally use proptest now rather than quickcheck. Proptest actually persists failure cases to guard against future regressions, which is nice to have.

dandanlen (commented):
Much better now. 

I added a branch (#6702 ) with some suggested refactors. 

All that's missing is the rpc call(s) for the product and some better testing of the transition. I think a bouncer test but also some unit/integration level testing would help (ie. bouncer test can check the high level mechanism, but lower-level tests can ensure that eg. each fee type transitions correctly).

state-chain/pallets/cf-funding/src/lib.rs
ramizhasan111: I dont think we are repeating logic anymore. Here its only about moving this logic from here in funding pallet to the flip pallet which is what your suggested refactor does. I deliberately left it here since it doesnt dedup anything. But I can see that we would rather have this in the flip pallet.

state-chain/runtime/src/chainflip/epoch_transition.rs
ramizhasan111: done.
I accidentally added the commit for this in #6702 . The commit will be included here when we merge that branch.

state-chain/pallets/cf-swapping/src/lib.rs
ramizhasan111: fixed

state-chain/pallets/cf-flip/src/tests.rs
ramizhasan111: I added the before and after. conservation of funds check. 

I used quickcheck since we already use it in this pallet for other balance integrity tests. Will think about using proptests for invariance tests in the future.

state-chain/pallets/cf-flip/src/tests.rs
ramizhasan111: added a commit that now uses interface methods instead of directly adding to storages. Also added a test that  tests the reward distribution trigger based on activation epoch as. much as is. possible within unit tests in this pallet.

dandanlen:
I just realised one thing that might be a little unintuitive: rewards are paid out when we start a *new* epoch. This means earnings don't compound until the following auction (ie. the earning from the current epoch are not taken into account during the auction). Worth discussing.

marcellorigotti:
When we call `trigger_flip_reward_distribution` on epoch transition we pass the list of validators for the old epoch, then we loop through them and call `distribute` for each of them, the implementation of `distribute` uses the current `epoch_index` though, and we use that to get the snapshot.

Couldn't there be a situation when an operator is part of the old snapshot but not the new one? maybe this is not possible but it is weird to rely on this to be sure that we credit correctly all the delegators as well.
Also this could possibly be gamed in some way if an operator can deregister as operator before we distribute rewards, causing the rewards to be distributed to himself only. Again not sure these scenarios are possible, just pointing it out just in case

marcellorigotti:
Also in `on_new_epoch` when transitioning epoch, we first bridge the funds on chain and then we try to egress them to the state-chain gateway, this egress can fail though if the amount after fee is lower than the threshold, at which point we don't really have these funds on chain. 
Moreover if the egress succeed we emit the event:
```
Self::deposit_event(Event::SentFlipToGateway {
						amount: (egress_amount.saturating_add(fee_withheld)),
						egress_id,
					});
```
we are keeping the fees into account which imo is wrong since we actually swapped them for the egress fee, this could create some discrepancies in what we have in the state-chain-gateway vs the vault.

I think it would be more correct to:
- first try to egress, and emit the event with only `egress_amount` if it succeed
- then use `egress_amount` - the amount of `FlipToBeSentToGateway` before we added `flip_distributed` (basically `flip_distributed - fee_withheld`), as a value to be distributed
- if the egress fail we postpone rewards distribution to the next epoch, this would probably never happen anyway

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -194,6 +223,9 @@ pub mod pallet {
 			account_id: T::AccountId,
 			new_bond: T::Balance,
 		},
+		FlipDistributed {
+			amount: Vec<(T::AccountId, T::Balance)>,
```
dandanlen: Can we rename this to amounts or distribution?

ramizhasan111:
> When we call `trigger_flip_reward_distribution` on epoch transition we pass the list of validators for the old epoch, then we loop through them and call `distribute` for each of them, the implementation of `distribute` uses the current `epoch_index` though, and we use that to get the snapshot.
> 
> Couldn't there be a situation when an operator is part of the old snapshot but not the new one? maybe this is not possible but it is weird to rely on this to be sure that we credit correctly all the delegators as well. Also this could possibly be gamed in some way if an operator can deregister as operator before we distribute rewards, causing the rewards to be distributed to himself only. Again not sure these scenarios are possible, just pointing it out just in case

nice catch! I've fixed it so that the distribute function takes in the epoch we want to distribute for. The fix commit is in the rewards PR (by accident, sorry for that).

marcellorigotti:
> Also in `on_new_epoch` when transitioning epoch, we first bridge the funds on chain and then we try to egress them to the state-chain gateway, this egress can fail though if the amount after fee is lower than the threshold, at which point we don't really have these funds on chain. Moreover if the egress succeed we emit the event:
> 
> ```
> Self::deposit_event(Event::SentFlipToGateway {
> 						amount: (egress_amount.saturating_add(fee_withheld)),
> 						egress_id,
> 					});
> ```
> 
> we are keeping the fees into account which imo is wrong since we actually swapped them for the egress fee, this could create some discrepancies in what we have in the state-chain-gateway vs the vault.
> 
> I think it would be more correct to:
> 
> * first try to egress, and emit the event with only `egress_amount` if it succeed
> * then use `egress_amount` - the amount of `FlipToBeSentToGateway` before we added `flip_distributed` (basically `flip_distributed - fee_withheld`), as a value to be distributed
> * if the egress fail we postpone rewards distribution to the next epoch, this would probably never happen anyway

If we don't want to make it like this and we want to distribute rewards every epoch, regardless of the egress to the gateway happening or not, we can return from `maybe_trigger_flip_to_gateway_egress` the `fee_withheld` which we will subtract from `FlipToDistribute`. 
aka. Fees are taken from the next epoch rewards.

EDIT: this addresses the fee problem but still it doesn't solve the problem of sending to the gateway the correct amount. `fee_withheld` should also be added back to `FlipToBeSentToGateway`, such that for instance:
- we distribute 100FLIP
- we egress 99FLIP to gateway, 1FLIP fee
- `FlipToDistribute` is set to -1 (next epoch rewards will pay for the egress)
- `FlipToBeSentToGateway` is set to 1, next egress we send 1FLIP more to properly offset between vault and gateway


IMO the first idea to first egress and then distribute the net amount is more correct and easier to reason about.

marcellorigotti:
> I just realised one thing that might be a little unintuitive: rewards are paid out when we start a _new_ epoch. This means earnings don't compound until the following auction (ie. the earning from the current epoch are not taken into account during the auction). Worth discussing.

I was thinking about this and while I like the idea to pay at the end of the epoch, the auction is resolved at the beginning of the rotation, swaps can still happen before the rotation ends, which could still increase earnings which would not be counted towards the auction, and we would end up in a situation where we have to pay rewards before the auction and when the rotation end/when the new epoch start.
I think this is just more confusing/not worth doing it, IMO it is not a big deal that one epoch rewards are not directly counted in the immediate next auction.

Still I think it is worth to pay the rewards before transitioning to the new epoch to avoid having to deal with cross epoch boundaries.

ramizhasan111:
> > Also in `on_new_epoch` when transitioning epoch, we first bridge the funds on chain and then we try to egress them to the state-chain gateway, this egress can fail though if the amount after fee is lower than the threshold, at which point we don't really have these funds on chain. Moreover if the egress succeed we emit the event:
> > ```
> > Self::deposit_event(Event::SentFlipToGateway {
> > 						amount: (egress_amount.saturating_add(fee_withheld)),
> > 						egress_id,
> > 					});
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > we are keeping the fees into account which imo is wrong since we actually swapped them for the egress fee, this could create some discrepancies in what we have in the state-chain-gateway vs the vault.
> > I think it would be more correct to:
> > 
> > * first try to egress, and emit the event with only `egress_amount` if it succeed
> > * then use `egress_amount` - the amount of `FlipToBeSentToGateway` before we added `flip_distributed` (basically `flip_distributed - fee_withheld`), as a value to be distributed
> > * if the egress fail we postpone rewards distribution to the next epoch, this would probably never happen anyway
> 
> If we don't want to make it like this and we want to distribute rewards every epoch, regardless of the egress to the gateway happening or not, we can return from `maybe_trigger_flip_to_gateway_egress` the `fee_withheld` which we will subtract from `FlipToDistribute`. aka. Fees are taken from the next epoch rewards.
> 
> EDIT: this addresses the fee problem but still it doesn't solve the problem of sending to the gateway the correct amount. `fee_withheld` should also be added back to `FlipToBeSentToGateway`, such that for instance:
> 
> * we distribute 100FLIP
> * we egress 99FLIP to gateway, 1FLIP fee
> * `FlipToDistribute` is set to -1 (next epoch rewards will pay for the egress)
> * `FlipToBeSentToGateway` is set to 1, next egress we send 1FLIP more to properly offset between vault and gateway
> 
> IMO the first idea to first egress and then distribute the net amount is more correct and easier to reason about.

This behaviour is basically unchanged. This logic is re-used from the emissions pallet, which will be removed after flip 2.1. I guess the assumption here is that when we account for the funds that need to be moved to SC gateway, we can already assume they are onchain. Even if the actual transfer of funds to the gateway happens with a delay, as long as the accounting of the funds is done correctly, it should be fine. It would only be a problem in the extreme case of if everyone tries to withdraw flip from the SC at the same time and there are not enough funds in the gateway since some funds have not yet been transferred to the gateway. This case is very unlikely, hence the assumption.

Regarding the fee withheld issue, I agree with the second approach to take the fee_withheld from the future flip to be distributed but setting it to `-fee_withheld` and then add the fee_withheld back to the `FlipTobeSentToGateway` since this also matches how we currently do it in the emissions pallet (talking the fee from the flip to burn)

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -509,13 +540,92 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	/// Fee rewards pending distribution to authorities: the balance of the on-chain distribution
+	/// reserve plus any off-chain accumulated fees (which may be negative).
+	pub fn pending_rewards() -> i128 {
+		Reserve::<T>::get(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID)
+			.try_into()
+			.unwrap_or(i128::MAX)
+			.saturating_add(FlipToDistribute::<T>::get())
+	}
```
marcellorigotti: When we add FlipToDistribute, we should do `FlipToDistribute::<T>::get().max(0)`, this matches the actual distribution path. We only bring FlipToDistribute if it is a positive value, this allows us to return u128 here and avoid a weird edge cases where the returned value is negative (right after rotation) and the estimate RPC wraps it returning a wrong amount, happening here:
```
let total_rewards = Flip::pending_rewards() as FlipBalance;
let per_authority_share = total_rewards / authority_count as FlipBalance;
```

marcellorigotti (approved):

PR #6604, opened 2026-05-20
fix: cf_account_info includes collateral in supply-only accounts

msgmaxim:
# Pull Request

## Summary

Added a free function `get_total_collateral_for_account` similar to `get_total_collateral` but does not require a loan account to exist. It is now used in `cf_account_info` which now correctly includes supplied funds as collateral in accounts that don't have loans. (I kept `get_total_collateral` as it is still useful in cases where we have already loaded the account).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes included in each Chainflip release will be documented in this file.
 
+## [2.2.2] - 2026-05-20
+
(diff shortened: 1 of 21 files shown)


GabrielBuragev (approved):
lgtm

dandanlen (approved):

dandanlen:
Cherry-picked this to release separately.

PR #6607, opened 2026-05-21
refactor: is oracle supported function for price feed api

j4m1ef0rd:
# Pull Request

Closes: PRO-2869

## Summary

Added a `is_oracle_supported` to the price feed api. It re-uses the logic from the get_price code.

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-swapping/src/lib.rs
+++ b/state-chain/pallets/cf-swapping/src/lib.rs
@@ -3371,30 +3371,32 @@ pub mod pallet {
 				// Get the price of both assets using oracles or simulated swap, both with fallback
 				// to hard coded prices
 				let get_usd_price = |asset, side: Side| -> Price {
-					if !T::PriceFeedApi::is_oracle_supported(asset) {
-						let asset_price =
-							Self::estimate_usdc_price_using_simulated_swap_or_fallback(asset, side); // USDC / Asset
+					if let Some(oracle) = T::PriceFeedApi::get_price(asset) {
+						// Apply a hard coded slippage in the correct direction.
+						let slippage_bps = if asset == STABLE_ASSET {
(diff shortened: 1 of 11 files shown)


state-chain/runtime/src/chainflip.rs
```diff
@@ -1224,6 +1224,10 @@ impl CcmAdditionalDataHandler for CfCcmAdditionalDataHandler {
 
 pub struct ChainlinkOracle;
 impl cf_traits::PriceFeedApi for ChainlinkOracle {
+	fn is_oracle_supported(asset: Asset) -> bool {
+		is_oracle_supported(asset)
```
MxmUrw: Maybe we could inline the local `is_oracle_supported()` function to here, wdyt?

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -3369,7 +3371,7 @@ pub mod pallet {
 				// Get the price of both assets using oracles or simulated swap, both with fallback
 				// to hard coded prices
 				let get_usd_price = |asset, side: Side| -> Price {
-					if asset == Asset::Flip || asset == Asset::Dot || asset == Asset::Trx {
```
MxmUrw, line 3372: Is there a reason why we can't do the following:
1. Try to call `get_price(asset)`
2. If that fails, estimate using a swap
3. If that fails, use hard-coded prices
?

j4m1ef0rd: I started to make that change already and thought it was wrong, but on second thought, you are correct, Its fine.

msgmaxim (approved):

dandanlen (approved):

PR #6617, opened 2026-05-26
refactor: cleanup oracle and inventory based ASS

j4m1ef0rd:
# Pull Request

Closes: PRO-2802

## Summary

- More efficient looping over strategies.
- No need to lookup price in oracle strategy logic. Uses the relative tick instead.
- Restrict inventory based strategy to only assets of same decimals (future proofing).
- Updated some weight amounts

---

## *Checklist 

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-trading-strategy/src/lib.rs
+++ b/state-chain/pallets/cf-trading-strategy/src/lib.rs
@@ -1078,18 +1078,35 @@ impl<T: Config> Pallet<T> {
 			}
 			*weight_used += limit_order_update_weight * CANCEL_ALL_ORDERS_WEIGHT_UNITS;
 
-			// Create the new desired orders
+			// Create the new desired orders.
 			let mut remaining_base_amount = total_base_asset;
 			let mut remaining_quote_amount = total_quote_asset;
+			let mut remaining_sell_orders =
+				new_orders.iter().filter(|order| order.side == Side::Sell).count();
+			let mut remaining_buy_orders = new_orders.len() - remaining_sell_orders;
 			new_orders.into_iter().for_each(
 				|StrategyLimitOrder { base_asset, side, order_id, tick, amount, .. }| {
 					let amount = if side == Side::Sell {
-						// Convert the base amount back from the quote denomination.
-						let base_amount = pricing.quote_to_base(amount).min(remaining_base_amount);
+						remaining_sell_orders -= 1;
+						let base_amount = if remaining_sell_orders == 0 {
+							// The last order on each side takes all
+							// of that side's remaining balance, so rounding never leaves dust in
+							// the free balance.
+							remaining_base_amount
+						} else {
+							// Amounts are sized in the quote denomination, so the
+							// sell side is converted back to the base asset.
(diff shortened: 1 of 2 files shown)


state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -63,6 +64,10 @@ pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(STORAGE_VERSION_
 const STRATEGY_ORDER_ID_0: OrderId = 0;
 const STRATEGY_ORDER_ID_1: OrderId = 1;
 
+// Weight constants for inventory-based and oracle-tracking strategies
+const CANCEL_ALL_ORDERS_WEIGHT_UNITS: u64 = 3;
```
msgmaxim: Is this the right name? The number is the upper bound on how many orders we expect to need to close, right? (Same for `OPEN_ORDERS_WEIGHT_UNITS`).

j4m1ef0rd: Yeah its a bad name, I'll change it, and I don't think we need 2 const's. Using the upper bound number of 4 for both cases is fine.

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -653,6 +666,56 @@ pub mod pallet {
...
+	/// Used by `OracleTracking`: `relative_price` is the oracle price of the base asset
+	/// denominated in the quote asset, with its matching `relative_tick`. Amounts are compared in
+	/// the quote denomination and orders are additionally refreshed whenever the oracle moves the
+	/// ticks.
+	Oracle { relative_tick: Tick, relative_price: Price },
+}
+
+impl PricingMode {
+	/// The tick offset applied to the configured tick ranges (zero for static strategies).
+	fn relative_tick(self) -> Tick {
```
msgmaxim: Would `tick_offset` be a better name?

state-chain/pallets/cf-trading-strategy/src/lib.rs
```diff
@@ -958,34 +1019,17 @@ impl<T: Config> Pallet<T> {
...
-				quote_balance_usd + base_balance_usd >= base_threshold_usd.min(quote_threshold_usd);
-			(update_due_to_balance, total_quote_usd, total_base_usd)
-		} else {
-			// For the inventory based strategy, we require both assets to be
-			// equivalent (have similar prices and the same number of decimals).
-			let update_due_to_balance =
-				quote_balance_asset + base_balance_asset >= base_threshold.min(quote_threshold);
-			(update_due_to_balance, total_quote_asset, total_base_asset)
-		};
+		let update_due_to_balance = quote_balance_asset
```
msgmaxim, line 1029: Would be nice to add a comment to make the intent clear. The name suggests the update is "due to balance", but not specifically what the condition is. I think this checks whether we have enough in free balance to make it worthwhile to update the orders?

msgmaxim (approved):

PR #6618, opened 2026-05-26
fix: assethub balance witnessing

dandanlen:
# Pull Request

Closes: PRO-2888

## Summary

This is a fix for the balance witnessing issue, where we could incorrectly witness eg. vesting funds.

The fix is quite inelegant but I don't think there's a better way:
- read the *liquid* account balance at the current and previous blocks
- check for any *outgoing* transfers in case there was a fetch
- derive the net inflow of funds to the address based on the net increase
- for each incoming transfer event, decrease the available liquid funds
- as long as the available funds are non-zero, generate a deposit

### DIFF
--- a/engine/src/witness/hub/hub_deposits.rs
+++ b/engine/src/witness/hub/hub_deposits.rs
@@ -128,7 +128,7 @@ fn address_and_details_to_addresses(
 
 /// The liquid budget for one (address, asset) over this block:
 ///
-///     budget = max(0, current_liquid - parent_liquid + outgoing_in_block)
+/// `budget = max(0, current_liquid - parent_liquid + outgoing_in_block)`
 ///
 /// i.e. the net new spendable funds for the pair, after compensating for funds the
 /// address sent out in the same block.
--- a/engine/src/witness/hub/hub_deposits.rs
+++ b/engine/src/witness/hub/hub_deposits.rs
@@ -36,6 +36,7 @@ use cf_primitives::{
 };
 use futures_core::Future;
 use pallet_cf_ingress_egress::{DepositChannelDetails, DepositWitness};
+use sp_runtime::traits::Saturating;
 use state_chain_runtime::AssethubInstance;
 use std::collections::BTreeMap;
 use subxt::events::Phase;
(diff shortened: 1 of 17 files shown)


engine/src/witness/hub/hub_deposits.rs
```diff
@@ -104,57 +116,141 @@ fn address_and_details_to_addresses(
...
+						ASSETHUB_USDT_ASSET_ID => HubAsset::HubUsdt,
+						ASSETHUB_USDC_ASSET_ID => HubAsset::HubUsdc,
+						_ => continue,
+					};
+					if monitored_addresses.contains(&from) {
+						*outgoing_balance_changes.entry((from, asset)).or_insert(0) += *amount;
+					}
+				},
+				_ => continue,
+			}
```
MxmUrw: nit: to reduce complexity, it might be worthwhile to factor out the Asset dispatch. Currently here and further down below there are two branches for `Balances` and for `Assets` that have effectively the same code inside them.

Also I'm wondering whether there's a nicer, less complex way of computing the final amount? It's pretty much the following, no?

`new_ingress = new_balance + outgoing - old_balance`

This might even be what's happening but it's a bit difficult to understand because the mutating caching logic is interleaved with the mutating computation...

dandanlen: Yeah that's basically it, i agree it's awkwardly written... 

I think it can be improved if I remove the 1-to-1 mapping of event-to-deposit, ie. just sum up all transfer-in and transfer-out events, and then credit a single deposit for `min(transfer_in, liquid_balance_diff + deposit_out)`. I'll refactor.

dandanlen: Actually I think what you're suggesting is even simpler: for incoming deposits, we can just ignore the event entirely (except that it acts as a marker for "there was a transfer") and use the balance diff plus any fetches that happened in the same block.

dandanlen: Actually - i decided against this (squashing multiple deposits into one) for now since it implies ignoring the deposit details for the tx. We an revisit.

engine/src/witness/hub/hub_deposits.rs
```diff
@@ -104,57 +116,141 @@ fn address_and_details_to_addresses(
...
-				_ => (),
+				_ => continue,
+			};
+			let available_balance = if let Some(balance) =
+				available_liquid_balances.get(&(deposit_address, asset))
+			{
+				*balance
+			} else {
+				let balance =
+					balance_increase(hub_client, deposit_address, asset, block_hash, parent_hash)
```
marcellorigotti: There is a bug, `balance_increase` saturates which can lead to an edge case.
The current code basically does `current.saturating_sub(parent) + outgoing`, the correct thing to do should be `(current + outgoing).saturating_sub(parent)`

Otherwise this scenario can happen:
```
parent liquid = 50
current liquid = 0
outgoing current block = 50
```
`0.saturating_sub(50) + 50 = 0 + 50 = 50`

this can be triggered by another incoming transfer (vested or not)

dandanlen: Yes ... I'm rewriting, will post a new version, it's nearly ready.

marcellorigotti (approved):
LGTM now!

MxmUrw (approved):
@dandanlen something is up with unsigned commits, I guess from the merged other PR?

PR #6619, opened 2026-05-26
test: bouncer checks for witnesser crashes

j4m1ef0rd:
# Pull Request

Closes: PRO-2827

## Summary

A simple test that runs after the fast & full bouncer tests that checks for the `WitnessingTaskRestarted` event.


MxmUrw (approved):
Nice!

Theoretically it would be cool if we could actually trigger a witnessing panic *in* this test to make sure that our event logic correctly catches it. Otherwise, e.g. if the event is renamed on the state-chain we'll probably forget to update this test and then it will always pass as a false-negative.

But I don't think there's a short nice way to do it, and it's pretty low probability so we're good I think.

albert-llimos (approved):

PR #6620, opened 2026-05-26
feat: add min to order amount

marcellorigotti:
# Pull Request

Closes: PRO-2846

## Summary

Enforce a minimum amount to limit orders

## API Changes

`set_limit_order` and `update_limit_order` now enforce a minimum amount, it is responsibility of the caller to ensure that the order size is above the minimum, if that's not the case a `BelowMinimumOrderAmount` error is returned.
When updating an existing order the check is performed on the resulting amount. The only scenario where an order can exist with amount smaller than minimum is if it fills a swap, then the remaining amount can be lower than the minimum, no check is performed there.

### Extrinsics & Events

Added new event `MinimumLimitOrderAmountSet`
Added new error `BelowMinimumOrderAmount`

---

* The PR is complete:
  * Tests:


msgmaxim:
Have we considered applying `max(amount, min_amount)` to the user's submitted amount instead of returning error? `min_amount` is small anyway, so I don't think any LPs would complain. Then LPs won't need to worry about potentially changing their code (right now they might submit the exact amount needed, which may be smaller than the min if it is fees etc).

The only downside is there is a chance the user might get a confusing `InsufficientBalance` error if their balance is higher than what they submitted, but lower than the minimum, but I think that's quite an edge case...

state-chain/pallets/cf-pools/src/migrations/set_minimum_limit_order_amounts.rs
```diff
@@ -0,0 +1,112 @@
...
+use sp_std::marker::PhantomData;
+
+#[cfg(feature = "try-runtime")]
+use frame_support::pallet_prelude::DispatchError;
+#[cfg(feature = "try-runtime")]
+use sp_std::vec::Vec;
+
+/// Initial minimum limit order amount per asset, in the asset's smallest unit (decimals noted
+/// per row). The minimum applies to the asset being *sold* by an order
+const MINIMUMS: [(Asset, AssetAmount); 18] = [
```
msgmaxim: Do we want to set this via the default value on `MinimumLimitOrderAmount` rather than via migration? Same way it is done for `LimitOrderAutoSweepingThresholds`, for example. Otherwise these values won't be set in cases the migration is not executed (e.g. localnet and backspin).

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1073,6 +1087,28 @@ pub mod pallet {
...
+		/// remain open. The minimum applies to the asset being sold (`base_asset` for
+		/// `Side::Sell`, `quote_asset` for `Side::Buy`). Set to `0` to disable the check
+		/// for an asset.
+		///
+		/// Requires Governance.
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::set_minimum_limit_order_amounts(minimums.len() as u32))]
+		pub fn set_minimum_limit_order_amounts(
+			origin: OriginFor<T>,
+			minimums: BoundedVec<(Asset, AssetAmount), ConstU32<20>>,
```
msgmaxim: Won't hurt to set the max size higher, say, to 100? We might decide to reduce/increase all values at once (i.e. if we deem them too big/small) and we are already approaching 20 total assets.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1720,13 +1779,27 @@ impl<T: Config> Pallet<T> {
...
+			// that we can validate the resulting amount against the per-asset minimum.
+			let existing_amount: AssetAmount = match existing_tick {
+				Some(tick) => pool
+					.pool_state
+					.limit_order(&(lp.clone(), id), side, tick)
+					.map(|(_, position)| position.amount.saturated_into::<AssetAmount>())
+					.unwrap_or(0),
+				None => 0,
+			};
+			let resulting_amount = match amount_change {
```
msgmaxim: Could we just check the new amount after we have applied the change (And rely on the extrinsic to rollback the changes on error)? Then you won't need to duplicate increase/decrease logic.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1952,3 +1952,284 @@ fn test_get_limit_orders() {
...
+			));
+		});
+	}
+
+	#[test]
+	fn set_limit_order_zero_always_allowed() {
+		new_test_ext().execute_with(|| {
+			setup_pool_with_minimums();
+
+			assert_ok!(LiquidityPools::set_limit_order(
```
msgmaxim: Nit: most of these fields are irrelevenant to the test (tick, dispatch_at etc). I would create a helper method for `set_limit_order` that only accepts the fields we do care about (assets, amount, side) and use some defaults for the rest. Right now it is very hard at a glance to tell which one of the provided fields is the important one here (i.e. the amount).

bouncer/generated/events/liquidityPools/minimumLimitOrderAmountSet.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from 'zod';
+import { cfPrimitivesChainsAssetsAnyAsset, numberOrHex } from '../common';
+import { defineEvent } from '@chainflip/processor/event';
+
+export const liquidityPoolsMinimumLimitOrderAmountSet = z.object({
+  asset: cfPrimitivesChainsAssetsAnyAsset,
+  amount: numberOrHex,
+});
+
+export const liquidityPoolsLimitOrderUpdatedEvent = defineEvent(
```
msgmaxim: AI pointed out that these are wrong (e.g. liquidityPoolsLimitOrderUpdatedEvent does not match the actual event `liquidityPoolsMinimumLimitOrderAmountSetEvent`)

marcellorigotti:
> Have we considered applying `max(amount, min_amount)` to the user's submitted amount instead of returning error? `min_amount` is small anyway, so I don't think any LPs would complain. 

I am not a big fan of this, I would rather my order not being placed at all. This is also how it works on other exchanges.

state-chain/pallets/cf-pools/src/lib.rs
marcellorigotti: We have to make the `inner_udpate_limit_order` transactional for this to work, cause this is not called only during extrinsic execution but also scheduled order dispatch. But I am fine with that 👌

state-chain/pallets/cf-pools/src/migrations/set_minimum_limit_order_amounts.rs
marcellorigotti: This I don't have strong opinion, it would probably be a nice to have. 
It would require switching the storage item to a `StorageValue` containing the BTreeMap though, like for `LimitOrderAutoSweepingThresholds`. 

If we are fine with that I can do it 👍

msgmaxim: Hm, fair point, I think we probably do want to keep `StorageMap`.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1073,6 +1087,28 @@ pub mod pallet {
...
+		/// remain open. The minimum applies to the asset being sold (`base_asset` for
+		/// `Side::Sell`, `quote_asset` for `Side::Buy`). Set to `0` to disable the check
+		/// for an asset.
+		///
+		/// Requires Governance.
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::set_minimum_limit_order_amounts(minimums.len() as u32))]
+		pub fn set_minimum_limit_order_amounts(
+			origin: OriginFor<T>,
+			minimums: BoundedVec<(Asset, AssetAmount), ConstU32<100>>,
```
j4m1ef0rd: nit: we have a type `AssetAndAmount` that could be used here.

state-chain/pallets/cf-pools/src/lib.rs
j4m1ef0rd: Maybe we should just make `inner_update_limit_order_at_tick` return the new position amount instead of the `sold_amount_change`, because we don't really use the `sold_amount_change` (seems unnecessary in the sweep).  This would save us another lookup below.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1952,3 +1952,258 @@ fn test_get_limit_orders() {
...
+
+			assert_matching_event_count!(
+				Test,
+				RuntimeEvent::LiquidityPools(Event::MinimumLimitOrderAmountSet { .. }) => 2
+			);
+		});
+	}
+
+	#[test]
+	fn governance_only() {
```
j4m1ef0rd: This test can be combined with the one above to keep things clean.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1952,3 +1952,258 @@ fn test_get_limit_orders() {
...
+					Side::Sell,
+					IncreaseOrDecrease::Increase(MIN_ETH - 1),
+				),
+				Error::<Test>::BelowMinimumOrderAmount,
+			);
+		});
+	}
+
+	#[test]
+	fn update_increase_to_min_succeeds() {
```
j4m1ef0rd: A lot of these tests are just one assert, We usually just combine them and leave a comment for each test case. In this case the comment would be the test name. Less boilerplate and setup this way.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1952,3 +1952,258 @@ fn test_get_limit_orders() {
...
+					Side::Sell,
+					IncreaseOrDecrease::Decrease(MIN_ETH + 1),
+				),
+				Error::<Test>::BelowMinimumOrderAmount,
+			);
+		});
+	}
+
+	#[test]
+	fn update_decrease_to_zero_allowed() {
```
j4m1ef0rd: I don't think you have a test that starts with an order that is already below the minimum (mostly filled) and then decreases it to 0.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1952,3 +1952,258 @@ fn test_get_limit_orders() {
...
+					last_event::<Test>(),
+					RuntimeEvent::LiquidityPools(Event::ScheduledLimitOrderUpdateDispatchFailure {
+						lp: ALICE,
+						order_id: 0,
+						error: Error::<Test>::BelowMinimumOrderAmount.into(),
+					})
+				);
+				// ...and the balance debit from the decrease was rolled back (no Eth credited
+				// back to ALICE), proving the transactional wrapper works on the scheduled path.
+				assert_eq!(MockBalance::get_balance(&ALICE, Asset::Eth), balance_after_open);
```
j4m1ef0rd, line 2218: Could be worth checking that the actual order is also untouched.

j4m1ef0rd (approved):

state-chain/pallets/cf-pools/src/lib.rs
marcellorigotti: That is actually used to mint a new order when the tick changes in `inner_update_limit_order`.
I updated `inner_update_limit_order_at_tick` to return both  `sold_amount_change ` and `resulting_position_amount`, which IMO makes sense for that function anyway.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1073,6 +1089,28 @@ pub mod pallet {
...
+		/// remain open. The minimum applies to the asset being sold (`base_asset` for
+		/// `Side::Sell`, `quote_asset` for `Side::Buy`). Set to `0` to disable the check
+		/// for an asset.
+		///
+		/// Requires Governance.
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::set_minimum_limit_order_amounts(minimums.len() as u32))]
+		pub fn set_minimum_limit_order_amounts(
+			origin: OriginFor<T>,
+			minimums: BoundedVec<AssetAndAmount<AssetAmount>, ConstU32<100>>,
```
dandanlen: Can we fold this into `update_pallet_config`? ie. we just make it a variant of PalletConfigUpdate, then we don't need a new event and extrinsic.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1861,6 +1930,8 @@ impl<T: Config> Pallet<T> {
 			},
 		};
 
+		let resulting_amount: AssetAmount = position_info.amount.saturated_into();
```
dandanlen: nit: can we rename this `remaining_amount`? I think it's clearer.

dandanlen (approved):
Sorry bit of a late review... A couple of minor suggestions, but LGTM otherwise.

msgmaxim (approved):

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Update: done.

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Update: changed it to `order_size`.

PR #6622, opened 2026-05-29
fix: treat AncientBirthBlock as recoverable in submission_watcher

msgmaxim:
# Pull Request

Closes: PRO-2886

## Summary

Now if the engine hits AncientBirthBlock, it will refetch latest finalised block and retry instead of panicking. (Panicking is problematic as it, for example, aborts all ceremonies, including keygen ceremonies leading to exclusion from authority set.)

### DIFF
--- a/engine/sc-client/src/extrinsic_api/signed/submission_watcher.rs
+++ b/engine/sc-client/src/extrinsic_api/signed/submission_watcher.rs
@@ -309,11 +309,8 @@ impl<'a, 'env, BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static>
 		request: &mut Request,
 		nonce: Nonce,
 	) -> Result<Result<H256, SubmissionLogicError>, anyhow::Error> {
-		// The era anchor used to sign the extrinsic. Initialised from the
-		// watcher's tracked finalized block, but may be refreshed locally on
-		// AncientBirthBlock without touching `self.finalized_block_*` — the
-		// watcher's bookkeeping must only advance through `on_block_finalized`
-		// to preserve the strictly-increasing invariant it asserts on.
+		// The era block hash and number used as anchor when signing the extrinsic. Stored as
+		// local copies so we can overwrite them on AncientBirthBlock failure.
 		let mut era_block_hash = self.finalized_block_hash;
 		let mut era_block_number = self.finalized_block_number;
 		loop {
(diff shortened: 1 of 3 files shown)


engine/sc-client/src/extrinsic_api/signed/submission_watcher.rs
```diff
@@ -391,6 +391,21 @@ impl<'a, 'env, BaseRpcClient: base_rpc_api::BaseRpcApi + Send + Sync + 'static>
...
+						ClientError::Call(obj)
+							if obj == invalid_err_obj(InvalidTransaction::AncientBirthBlock) =>
+						{
+							warn!(target: "state_chain_client", request_id = request.id, "Submission failed with AncientBirthBlock: {obj:?}. Refreshing finalized block reference.");
+							let new_finalized_block_hash =
+								self.base_rpc_client.latest_finalized_block_hash().await?;
+							let new_finalized_header =
+								self.base_rpc_client.block_header(new_finalized_block_hash).await?;
+							self.finalized_block_hash = new_finalized_block_hash;
+							self.finalized_block_number = new_finalized_header.number;
```
MxmUrw: Not sure whether we should do this here, we might want to let the submission watcher handle this as it does during normal operations.

For example there's this line in the code that handles newly received finalized blocks:
```rust
		assert_eq!(block.header.number, self.finalized_block_number + 1, "{SUBSTRATE_BEHAVIOUR}");
```

marcellorigotti: Agree with this

marcellorigotti: ~~I think this requires a little refactor so that we can retry submissions at the `on_block_finalized` level, currently it is not possible since `submit_extrinsic` expect an `Hash` to be returned~~.

I think, we can simply return the pre-computed `tx_hash` in case of `AncientBirthBlock` and invalidate the cached nonce, without tracking any submission, this will be detected on the next `on_block_finalized` and based on the submission policy it will get retried or not.

msgmaxim: OK, I agree, this seems more complicated than (Claude and) I originatlly thought.

MxmUrw: > I think, we can simply return the pre-computed tx_hash in case of AncientBirthBlock and invalidate the cached nonce, without tracking any submission, this will be detected on the next on_block_finalized and based on the submission policy it will get retried or not.

Why don't we want to use the `Err(...)` result of `submit_extrinsic_at_nonce`? Seems like this is where it should semantically fit? Also *it looks like* the handling outside already does the nonce invalidation in case of an error, right?

marcellorigotti: the `submit_extrinsic_at_nonce` Err path is handling nonce related errors, it invalidates the nonce but it retries immediately (tho code is wrapped in a `Ok(loop {...}`).
It would end up retrying that extrinsic indefinitely with the same `block_hash`, which we know it is invalid. 
`on_block_finalized` cannot advance until the that submission completes, which will never happen.

MxmUrw: Can't we have a new `SubmissionLogicError` variant which is handled by doing a `break` within the error handling after the `submit_extrinsic_at_nonce()` call?

marcellorigotti (approved):
LGTM now 👌

MxmUrw (approved):

dandanlen (approved):
I would prefer if we make the comments a little more concise before we merge this.

vdoflip (approved):

PR #6624, opened 2026-05-29
pick: latest changes from release/2.1

dandanlen:
# Pull Request

Cherry-pick of latest release/2.1 changes to main.


MxmUrw (approved):
Looks good. The `fix: make runtime spec-version check non-blocking` is intentional I guess?

Also we might want to rename the PR...

dandanlen:
> Looks good. The `fix: make runtime spec-version check non-blocking` is intentional I guess?
> 
> Also we might want to rename the PR...

Yes, this is visible anyway in the outcome of the CI run and prevents us from eg. releasing for current and previous versions in parallel.

PR #6625, opened 2026-05-29
refactor: bouncer event name in schema

j4m1ef0rd:
# Pull Request

## Summary

Uses a new [schema generator](https://github.com/chainflip-io/chainflip-product-toolkit/pull/1094).
View the commits separately to avoid the codegen noise.
- The new schema adds a new object to each event files that links the event data to the event name string.
- Converted all of the non-dynamic events in the bouncer to the new event descriptor. Removing the event name strings.
~~- Kept backwards compatibility for the name + schema so we can still support dynamic per chain events (for now) even though I didn't end up using it, all of the dynamic pallet names are still using the old ObserverEvent code. But just in-case we want to upgrade them.~~
- Refactored the boost test (and related setup, commands and utils) to only support Btc at 5bps. Seems the test did not fully support other assets anyway.
- Fixed an issue on the CI with the Deposit monitor using a different broker endpoint to the bouncer, causing a nonce collision because they use the same broker account. Now they point to a common endpoint for both normal and old-rpc's runs of the bouncer on CI.
- Changed all places in the bouncer where a tron address is used to convert to or from base58 as required.


bouncer/shared/perform_swap.ts
```diff
@@ -89,10 +93,13 @@ export async function requestNewSwap<A = []>(
     dcaParams,
   );
   const addressReady = await cf.expectEvent(
-    'Swapping.SwapDepositAddressReady',
-    swappingSwapDepositAddressReady.refine((event) => {
+    swappingSwapDepositAddressReadyEvent.refine((event) => {
+      const expectedAddress =
+        chainFromAsset(destAsset) === 'Tron'
+          ? hexToTronAddress(destAddress as `0x${string}`)
+          : destAddress;
```
MxmUrw, line 103: It's a bit unfortunate that we have to have a special case for Tron here, what would it take to not having to do that?

bouncer/shared/utils.ts
```diff
@@ -529,10 +530,22 @@ export async function observeSwapEvents(
...
+            ? (Object.keys(channel.depositAddress)[0] as string)
+            : undefined;
+          const eventDepositAddress = channel
+            ? (Object.values(channel.depositAddress)[0] as string)
+            : undefined;
+          const depositAddressMatches =
+            eventDepositAddress !== undefined &&
+            (eventAddressKind === 'Tron'
+              ? hexToTronAddress(eventDepositAddress as `0x${string}`) === depositAddress
+              : eventDepositAddress === depositAddress);
```
MxmUrw: Also here, why has Tron to be checked separately?

bouncer/tests/all_swaps.ts
```diff
@@ -81,12 +81,13 @@ export function testAllSwaps(timeoutPerSwap: number) {
   // TODO: properly include TRON and BSC assets once they are fully integrated
   // if we include Assethub swaps (HubDot, HubUsdc, HubUsdt) in the all-to-all swaps,
   // the test starts to randomly fail because the assethub node is overloaded.
-  const AssetsWithoutAssethub = Object.values(Assets).filter(
-    (id) => chainFromAsset(id) !== 'Assethub' && chainFromAsset(id) !== 'Bsc',
-  );
+  const AssetsForTesting = Object.values(Assets).filter((id) => {
+    const chain: string = chainFromAsset(id);
```
MxmUrw, line 85: Can `chain` be of type `Chain` here?

bouncer/pnpm-lock.yaml
```diff
@@ -40,7 +40,7 @@ importers:
         version: 6.19.0(prisma@6.19.1(typescript@5.9.3))(typescript@5.9.3)
       '@solana/kit':
         specifier: ^4.0.0
-        version: 4.0.0(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(ws@8.18.3(bufferutil@4.1.0)(utf-8-validate@5.0.10))
+        version: 4.0.0(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(ws@7.5.10(bufferutil@4.1.0)(utf-8-validate@5.0.10))
```
MxmUrw, line 43: It looks like we're downgrading `ws` here, is there a specific reason?

bouncer/shared/utils/chainflip_io.ts
```diff
@@ -47,6 +48,36 @@ async function toExtrinsicResult(
...
+/**
+ * Resolve an `expectedEvent` argument into a `{ name, schema }` pair. Accepts either a
+ * generated event descriptor (name and schema bundled), or the legacy `{ name, schema? }`
+ * object (schema optional, defaulting to `z.any()`). Both expose `name` and `schema`.
+ */
+function resolveExpectedEvent(
+  expectedEvent: EventDescriptor<string, z.ZodTypeAny> | { name: EventName; schema?: z.ZodTypeAny },
+): { name: EventName; schema: z.ZodTypeAny } {
+  return { name: expectedEvent.name as EventName, schema: expectedEvent.schema ?? z.any() };
+}
```
MxmUrw: I think we don't use the "legacy" way of specifying events anywhere now, so it would be even nicer if we wouldn't need code to support both of them.

Tried it out a bit, seems to pass: https://github.com/chainflip-io/chainflip-backend/pull/6656

MxmUrw (approved):
Very very nice!

bouncer/pnpm-lock.yaml
j4m1ef0rd: Not intentional, but claude says:
ws is a peer dependency of @solana/kit's rpc-subscriptions-channel-websocket, so pnpm resolves it to whatever ws is hoisted nearby rather than pinning it. The @chainflip/processor bump (2.1.3-bsc-dev.3 → 2.2.0-alpha.5, commit c69aeb66d) dropped its @chainflip/rpc dependency, which changed which ws@8.x was visible in that part of the graph — so pnpm re-resolved this peer down to 7.5.10 (the version @solana/web3.js v1 already uses). ws@8.18.3 is still in the tree for ethers/polkadot/smoldot, and 7.5.10 is what the Solana libs expect, so it's functionally fine.

bouncer/tests/all_swaps.ts
j4m1ef0rd: We can't do that until the above TODO is completed. Gets an error about no Bcs.

bouncer/shared/perform_swap.ts
j4m1ef0rd: Because the tron addresses in events are base58 instead of hex. I think this is intentional. @dandanlen  or @zoheb391 might be able to confirm that this is expected.

bouncer/tests/all_swaps.ts
MxmUrw: Right yeah I see.

bouncer/pnpm-lock.yaml
MxmUrw: Ok cool sounds fine, thanks for checking!

bouncer/shared/perform_swap.ts
MxmUrw: I've had a look, the reason seems to be that in the new `chainflip/processor` version which is generating the event schemas we're now properly parsing tron addresses into the `T...` format. But in the bouncer we "mostly" use the evm format. This means that now the parsed (!) event contains `T...` address, so we have to adjust the passed destination address (originally in evm format) to match that format.

Going forward we should definitely strive to improve our address handling in the bouncer. I've created an issue for how I think we should proceed: https://linear.app/chainflip/issue/PRO-2937/cleanup-address-handling-in-bouncer.

We probably don't want to do it in *this* PR though, so for the time being I think I'm ok with merging as is (with the special case handling for Tron in 2 places in the swapping testing flow).

@albert-llimos @dandanlen @nmammeri hope you are ok with such a procedure, if not let us know!

bouncer/shared/utils.ts
MxmUrw: Regarding *this* conversion, I don't think it's necessary. See https://github.com/chainflip-io/chainflip-backend/pull/6663.

bouncer/shared/perform_swap.ts
MxmUrw, line 103:
```suggestion
      // This currently has to be done because the `destAddress` passed to us is in evm format,
      // but event decoding for Tron addresses parses them into `T...` format.
      // TODO: See PRO-2937 for an improvement plan
      const expectedAddress =
        chainFromAsset(destAsset) === 'Tron'
          ? hexToTronAddress(destAddress as `0x${string}`)
          : destAddress;
```

MxmUrw (approved):
Here, maybe we could add also add this comment (I wrote it yesterday but forgot to hit submit).

Anyways I think we're good to merge!

PR #6628, opened 2026-06-03
feat: onchain, offchain rewards, trigger flip 2.1

ramizhasan111:
# Pull Request

Closes: PRO-xxxx

## Summary

## API Changes

### RPCS

### Extrinsics & Events

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -103,8 +103,8 @@ pub mod pallet {
 	/// A 4-byte identifier for different reserves.
 	pub type ReserveId = [u8; 4];
 
-	/// Reserve ID derived from blake2_256("OnchainFlipToDistribute")[..4].
-	pub const ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID: ReserveId = [0xa7, 0x3c, 0x2f, 0x67];
+	/// Reserve ID derived from blake2_256("OnchainFees")[..4].
+	pub const ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID: ReserveId = [0xc3, 0xee, 0x10, 0x5a];
 
 	#[pallet::config]
 	#[pallet::disable_frame_system_supertrait_check]
@@ -547,7 +547,7 @@ impl<T: Config> Pallet<T> {
 	}
 
 	pub fn trigger_flip_reward_distribution(authorities: Vec<T::AccountId>) -> T::Balance {
-		let onchain_flip_to_distribute = Reserve::<T>::take(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID);
+		let onchain_flip_to_distribute = Reserve::<T>::get(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID);
 
 		let count = authorities.len();
 		let count_u128 = count as u128;
@@ -581,7 +581,7 @@ impl<T: Config> Pallet<T> {
 			None
 		};
 
-		let mut flip_distrbuted_map = sp_std::collections::btree_map::BTreeMap::new();
+		let mut flip_distributed_map = sp_std::collections::btree_map::BTreeMap::new();
 		for authority in &authorities {
 			T::RewardsDistribution::distribute(
 				if winner_authority.is_some_and(|wa| authority == wa) {
@@ -592,7 +592,7 @@ impl<T: Config> Pallet<T> {
 				authority,
 				|account, amount| {
 					Pallet::<T>::settle(account, Pallet::<T>::bridge_in(amount).into());
-					flip_distrbuted_map
+					flip_distributed_map
 						.entry(account.clone())
 						.and_modify(|e: &mut T::Balance| *e = e.saturating_add(amount))
 						.or_insert(amount);
@@ -615,7 +615,7 @@ impl<T: Config> Pallet<T> {
 						)
 						.into(),
 					);
-					flip_distrbuted_map
+					flip_distributed_map
 						.entry(account.clone())
 						.and_modify(|e: &mut T::Balance| *e = e.saturating_add(amount))
 						.or_insert(amount);
@@ -624,7 +624,7 @@ impl<T: Config> Pallet<T> {
 		}
 
 		Self::deposit_event(Event::FlipDistributed {
-			amount: flip_distrbuted_map.into_iter().collect(),
+			amount: flip_distributed_map.into_iter().collect(),
 		});
 		offchain_flip_to_distribute.into()
 	}
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -823,7 +823,12 @@ impl<T: Config> OnKilledAccount<T::AccountId> for BurnFlipAccount<T> {
 		let dust = Pallet::<T>::total_balance_of(account_id);
 		Pallet::<T>::settle(
 			account_id,
-			Pallet::<T>::add_to_onchain_flip_to_be_distributed(dust).into(),
+			if T::EpochInfo::epoch_index() >= ACCUMULATE_REWARDS_EPOCH_START {
+				Pallet::<T>::add_to_onchain_flip_to_be_distributed(dust)
+			} else {
+				Pallet::<T>::burn(dust)
+			}
+			.into(),
 		);
 		Account::<T>::remove(account_id);
 		Pallet::<T>::deposit_event(Event::AccountReaped {
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -30,10 +30,12 @@ pub mod substrate_impls;
 
 pub mod weights;
 use scale_info::TypeInfo;
+use sp_runtime::traits::Saturating;
 pub use weights::WeightInfo;
 
+use cf_primitives::ACCUMULATE_REWARDS_EPOCH_START;
 use cf_traits::{
-	AccountInfo, Bonding, DeregistrationCheck, FeePayment, FundingInfo, Issuance,
+	AccountInfo, Bonding, DeregistrationCheck, EpochInfo, FeePayment, FundingInfo, Issuance,
 	RewardsDistribution, Slashing,
 };
 use imbalances::{Deficit, ImbalanceSource, Surplus};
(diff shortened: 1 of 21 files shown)


state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -603,17 +658,30 @@ impl<T: Config> FeePayment for Pallet<T> {
...
 		}
 	}
+
+	fn add_to_offchain_flip_to_be_distributed(amount: i128) {
+		FlipToDistribute::<T>::mutate(|flip| flip.saturating_accrue(amount));
+	}
+
+	fn add_to_onchain_flip_to_be_distributed(amount: Self::Amount) -> Deficit<T> {
+		Pallet::<T>::deposit_reserves(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID, amount)
+	}
```
dandanlen, line 684: This is IMO the wrong level of abstraction. We should have a single method (`burn_or_reserve`?) that branches on `if T::EpochInfo::epoch_index() >= ACCUMULATE_REWARDS_EPOCH_START`, ie. one location that needs to be updated. 

(Note how this if/else condition is repeated all over the place)

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -531,20 +547,30 @@ impl<T: Config> Pallet<T> {
...
 
 	pub fn trigger_flip_reward_distribution(authorities: Vec<T::AccountId>) -> T::Balance {
-		let flip_to_distribute = FlipToDistribute::<T>::take();
-		if flip_to_distribute <= Zero::zero() {
-			FlipToDistribute::<T>::put(flip_to_distribute);
-			return Zero::zero()
-		}
-		let flip_to_distribute: u128 =
-			flip_to_distribute.try_into().expect("checked for negative number above");
+		let onchain_flip_to_distribute = Reserve::<T>::take(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID);
```
dandanlen: This is likely a bug: `take` will erase the reserves, so below, `withdraw_reserves` will always withdraw zero.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -555,24 +581,52 @@ impl<T: Config> Pallet<T> {
...
+						.or_insert(amount);
+				},
+			);
+
+			T::RewardsDistribution::distribute(
+				if winner_authority.is_some_and(|wa| authority == wa) {
+					onchain_per_authority_reward + onchain_remainder
+				} else {
+					onchain_per_authority_reward
+				},
```
dandanlen, line 607: It would be better to compute a total reward per validator and credit it once rather than looping and crediting twice.

Basically, build the `flip_distrbuted_map` and then credit everything, rather than crediting and building the map as a side effect.

state-chain/pallets/cf-flip/src/lib.rs
dandanlen: This might also remove the need to add Deficit as an assoc type

state-chain/primitives/src/lib.rs
```diff
@@ -264,6 +264,10 @@ pub struct TxId {
 /// The very first epoch number
 pub const GENESIS_EPOCH: u32 = 1;
 
+/// From this epoch onward, functionality that accumulates rewards for off-chain distribution is
+/// enabled. Used to gate behaviour that should only activate after a protocol upgrade.
+pub const ACCUMULATE_REWARDS_EPOCH_START: EpochIndex = 3; // todo: set a reasonable value before releasing
```
dandanlen: How about we make this a delay? ie. apply it ~1 epoch after the runtime upgrade.

state-chain/runtime/src/chainflip/epoch_transition.rs
```diff
@@ -33,12 +33,18 @@ impl EpochTransitionHandler for ChainflipEpochTransitions {
...
-		Swapping::maybe_trigger_flip_to_gateway_egress(
-			Environment::state_chain_gateway_address(),
-			flip_distributed,
-		);
+		if new == ACCUMULATE_REWARDS_EPOCH_START {
+			Emissions::burn_and_broadcast_supply_update(
+				frame_system::Pallet::<crate::Runtime>::block_number(),
+			);
+		} else if new > ACCUMULATE_REWARDS_EPOCH_START {
+			let flip_distributed =
```
dandanlen, line 42: What about `new < ACCUMULATE_REWARDS_EPOCH_START`? Do we not want to continue with the existing behaviour up until the breakpoint epoch?

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -555,24 +581,52 @@ impl<T: Config> Pallet<T> {
 			None
 		};
 
-		let mut flip_minted_list = vec![];
+		let mut flip_distrbuted_map = sp_std::collections::btree_map::BTreeMap::new();
```
dandanlen: nit: typo

dandanlen (commented):
I know this is still in draft state but i added some early comments.

state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -101,11 +103,17 @@ pub mod pallet {
 	/// A 4-byte identifier for different reserves.
 	pub type ReserveId = [u8; 4];
 
+	/// Reserve ID derived from blake2_256("OnchainFlipToDistribute")[..4].
```
dandanlen: I would suggest we just use something easy to read like `"DIST"` or `"FEES"`.

state-chain/runtime/src/chainflip/epoch_transition.rs
ramizhasan111: yes exactly. As discussed, we would keep with the current behaviour until the trigger epoch. at the start of the trigger epoch we do one last supply sync.

state-chain/pallets/cf-flip/src/lib.rs
ramizhasan111: The issue is that the onchain flip and offchain flip are settled differently. If we do one distribute call, we would have to pass a complicated settle function that would be basically keep track how much offchain and onchain flip we have used to distribute the flip. I thinkthis is easier to read and follow even if calls the distribute function twice (which shouldnt be a performance issue since we just do this once per epoch). Also, the map is built by the distribute function which does the allocation calcuations. we cant build the map here.

state-chain/pallets/cf-flip/src/lib.rs
ramizhasan111: I didnt do it that way because the Issuance trait is to be removed anyways, after we activate flip 2.1 on mainnet, as part of code cleanup where we remove the conditions and the transition code. With no more Issuance trait, it makes sense to have the fee collection function on FeePayment trait which anyways has other fee taking functions. And since this function is not used in the pre-transition code paths,  we dont need a condition here (the condition is moved to on_charge_transaction.rs for example). The associate type is needed because of how state chain tx fees work (see on_charge_transaction.rs). Basically the fee is taken as a surplus which then needs a deficit and so we need to return a deficit from here. It was similar before when the burn function also returned the Deficit to be settled against the fee surplus.

state-chain/pallets/cf-flip/src/lib.rs
ramizhasan111: changed to "OnchainFees"

state-chain/primitives/src/lib.rs
ramizhasan111: we could but would be a bit messier. we would need to introduce a storage item and we would need a migration (which we were trying to avoid) that sets the storage item to the next one during the runtime upgrade.

state-chain/pallets/cf-flip/src/lib.rs
dandanlen: We can make it much simpler:

Remove the remainder handling. 
Move all the offchain rewards to the reserve via bridge_in.
Now we can distribute rewards from the reserve in one loop.
Any remainder will stay in the reserve until the next epoch. (anyhow it's at most 149e-18 FLIP).

state-chain/primitives/src/lib.rs
dandanlen: I think it will be simpler than having to predict the exact epoch at which we want to do the transition.

state-chain/pallets/cf-flip/src/lib.rs
dandanlen: What I meant was literally just `b"FEES": [u8; 4]`: there's no need to hash anything here. Hashing just obfuscates things.

state-chain/runtime/src/chainflip/epoch_transition.rs
dandanlen: Ok I see now, the diff is a bit misleading because it shows the diff against the flip-2.1 branch. We should combine them.

state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -923,7 +926,11 @@ pub mod pallet {
 						amount < MinimumFunding::<T>::get().into()
 					{
 						// Insufficient funds to create an account.
-						T::Flip::burn_offchain(amount.into());
+						if T::EpochInfo::epoch_index() >= ACCUMULATE_REWARDS_EPOCH_START {
+							T::Flip::add_to_onchain_flip_to_be_distributed(amount.into());
```
dandanlen, line 930: I think this should be offchain no onchain

state-chain/pallets/cf-flip/src/lib.rs
dandanlen: I don't see how this changes things: right now we have this if/else clause repeated multiple times. We can remove/refactor the Issuance trait already: rename burn to `burn_or_reserve_onchain/offchain`. Then we only have one place that needs updating (and reviewing) instead of >3. Also note not deduplicating this introduced a (minor) bug in the funding pallet.

ramizhasan111:
I am merging this into the base branch. Will address rest of the open comments there.

state-chain/primitives/src/lib.rs
ramizhasan111: added the storage item for activation epoch which is configurable via governance.

state-chain/pallets/cf-flip/src/lib.rs
ramizhasan111: done.

state-chain/pallets/cf-flip/src/lib.rs
ramizhasan111: The reason i used hashing was to be able to use a string of more than four bytes like "OnchainFees" and fit to 4 bytes, maintaining the uniqueness of the string. If we were to use a four byte string like "FEES", we wouldnt need hashing sure. its just that I wanted it to contain the info that these onchain fees. Anyways, "FEES" sounds  simpler. Changed to b"FEES".

state-chain/pallets/cf-funding/src/lib.rs
ramizhasan111: Actually, this should be onchain since when user transfers funds to the SC gateway they are effectively sending funds on chain. The bug here was that we needed to bridge in the funds. I have fixed this.

state-chain/pallets/cf-flip/src/lib.rs
ramizhasan111: we could dedup the checks without refactoring traits as the checks that can be deduped are all in flip pallet. I have deduped them by having a single burn_or_add_to_reserve function.

PR #6629, opened 2026-06-03
fix: assethub balance witnessing

dandanlen:
This is a cherry-pick of #6618 with the addition of assethub vault setup to the post-upgrade bouncer.


bouncer/commands/run_test.ts
nmammeri, line 80: I think is good to keep this check and make it such it tests a file exists

nmammeri (approved):

bouncer/commands/run_test.ts
dandanlen: I think there might have been a merge/display issue - I can't see what you were referring to.

PR #6631, opened 2026-06-04
feat: simplify lending network fee collection

msgmaxim:
# Pull Request

Closes: PRO-2850

## Summary

Network fees now use the same mechanism for handling 100% utilisation as broker fees. (Not intended to be merged into the 2.2 release).

Notes:

- Claude struggled a bit with runtime-api so I asked it to update claude.md such that it does a better job in the future.
- In the migration we remove `owed_to_network` field. As discussed before, if it is not 0, its removal will have the same effect at transerring uncollected fees from network to the pool (which seems like an acceptable fallback).

### RPCS

`cf_lending_pools`: returned lending pools no longer contain `owed_to_network` field

FYI @GabrielBuragev

---

## *Checklist*

* The PR is complete:
  * Tests:


state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1136,30 +1149,45 @@ impl<T: Config> GeneralLoan<T> {
...
 
-		let broker_interest_collected = Pallet::<T>::mutate_existing_pool(loan_asset, |pool| {
-			pool.record_pool_fee(pool_interest);
+		let (network_interest_collected, low_ltv_penalty_collected, broker_interest_collected) =
+			Pallet::<T>::mutate_existing_pool(loan_asset, |pool| {
+				pool.record_pool_fee(pool_interest);
 
-			let network_fees_collected = pool.record_and_collect_network_fee(fees_owed_to_network);
-			Pallet::<T>::credit_fees_to_network(loan_asset, network_fees_collected);
+				let network_collected = pool.collect_fee_from_available(network_interest);
```
j4m1ef0rd: nit: this pattern of `charge_fee_if_exceeds_threshold` -> `collect_fee_from_available`-> `accrue_uncollected` -> `owed_principal.saturating_accrue` is duplicated 3 times and easy to make a mistake or forget one of the steps. Would be nice to bundle it into a single function.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1415,19 +1443,16 @@ pub fn fund_loan<T: Config>(
...
 		pool.provide_funds_for_loan(principal).map_err(Error::<T>::from)?;
-
 		pool.record_pool_fee(origination_fee_pool);
 
-		let network_fee_collected = pool.record_and_collect_network_fee(origination_fee_network);
+		let network_fee_collected = pool.collect_fee_from_available(origination_fee_network);
+		// The pool must hold enough liquid funds to cover both the principal *and* the
+		// network portion of the origination fee.
+		ensure!(
+			network_fee_collected == origination_fee_network,
```
j4m1ef0rd, line 1460: nit: might be cleaner to have `collect_fee_from_available` return collected and uncollected amounts. Might help with the other locations is used as well

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: OK, added another closure to reuse most of the code between different fee types.

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Not sure about this one. Getting uncollected is just a matter of subtracting one value from the other, and I'd rather keep the interface simple. In this case we don't even need to subtract because we can compare against the full amount instead, which seems just as good.

state-chain/runtime/src/runtime_apis/custom_api.rs
```diff
@@ -346,6 +346,10 @@ decl_runtime_apis!(
 		fn cf_lending_pools(
 			asset: Option<Asset>,
 		) -> Vec<before_version_17::RpcLendingPool<AssetAmount>>;
+		#[changed_in(18)]
```
dandanlen: We should bump this to 19 instead of re-using 18 (which is already released).

state-chain/pallets/cf-lending-pools/src/general_lending.rs
```diff
@@ -1107,58 +1107,84 @@ impl<T: Config> GeneralLoan<T> {
...
-		let charge_fee_if_exceeds_threshold = |fee: &mut ScaledAmountHP| {
+		let charge_fee_if_exceeds_threshold = |pending: &mut ScaledAmountHP| {
 			let fee_taken = if let Some(PriceCacheAndThreshold { threshold_usd, price_cache }) =
 				price_cache_and_threshold
 			{
-				// If the threshold is provided, take fees only if they exceed it:
-				if price_cache.usd_value_of(loan_asset, fee.into_asset_amount())? > threshold_usd {
-					fee.take_non_fractional_part()
+				// If the threshold is provided, take fees only if they exceed it. Effectively
+				// noop if the vaule can't be determined
```
j4m1ef0rd:
```suggestion
				// noop if the value can't be determined
```

j4m1ef0rd (approved):

state-chain/runtime/src/runtime_apis/custom_api.rs
msgmaxim: True. I see that version 18 was added for `cf_supported_assets`. I do wonder if we could have reused version 17 for it instead? I.e. could we not have one version bump per runtime upgrade (on mainnet) or would have this created issues on perseverance?

state-chain/pallets/cf-lending-pools/src/general_lending.rs
msgmaxim: Don't you just miss imperfect hand-crafted non-AI comments? 😆

state-chain/runtime/src/runtime_apis/custom_api.rs
dandanlen: Could have created an issue. In the longer run we should transition to using the spec_version instead of incrementing this out of sync with the runtime, then we don't have to think about it.

state-chain/pallets/cf-lending-pools/src/migrations/drop_owed_to_network.rs
```diff
@@ -0,0 +1,132 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Drops the `owed_to_network` field from `LendingPool`.
+//!
+//! For any pool that still has `owed_to_network > 0` at upgrade time, we credit as much of
+//! it as `available_amount` can cover to `PendingNetworkFees` (and deduct it from
+//! `available_amount`). Any residue that exceeds `available_amount` is forgiven to the
+//! network and stays with the pool: the borrower's `owed_principal` already includes the
+//! corresponding fee, so when they repay, that cash flows into `available_amount` and —
+//! with no IOU left to drain — accrues to lenders rather than the network.
```
dandanlen: I guess the probability of this is low, but strictly speaking it would accrue to the lending pool, but nobody would be able to claim it. It would eventually just remain as dust, no?

dandanlen (approved):
One minor issue that the AI flagged is that we effectively forgive any pending interest when a loan is closed, but we could probably use whatever asset is at hand for the repayment to also settle the outstanding fees. 

I don't think this was introduced in the PR, it was already part of the broker-fee model that we have now adopted for the interest payments.

It's probably a small amount but might be worth closing this gap anyway? (Either way, can be a follow-up)

state-chain/pallets/cf-lending-pools/src/migrations/drop_owed_to_network.rs
msgmaxim: RIght, we also need to increase `total_amount` in this case, so it is correctly attributed to lenders. Updated.

msgmaxim:
> One minor issue that the AI flagged is that we effectively forgive any pending interest when a loan is closed, but we could probably use whatever asset is at hand for the repayment to also settle the outstanding fees.
> 
> I don't think this was introduced in the PR, it was already part of the broker-fee model that we have now adopted for the interest payments.
> 
> It's probably a small amount but might be worth closing this gap anyway? (Either way, can be a follow-up)

Created this: https://linear.app/chainflip/issue/PRO-2931/lending-settle-outstanding-interest-from-repayment-asset-on-loan-close

dandanlen (approved):

PR #6634, opened 2026-06-08
chore: skip ScUtils Ccm Deposits for clarity

albert-llimos:
# Pull Request

Closes: PRO-2901

---

## *Checklist*

* The PR is complete:

### DIFF
--- a/engine/src/witness/eth_elections.rs
+++ b/engine/src/witness/eth_elections.rs
@@ -330,15 +330,25 @@ impl WitnessClientForBlockData<EthereumChain, Vec<ScUtilsCall>>
 
 		let mut result: Vec<ScUtilsCall> = Vec::new();
 		for event in events {
+			// Events's signer, when emitted via CCM will have the `signer` field set to the
+			// zero address. That functionality was added for potential future use cases such
+			// as onboarding via any asset. That has now been implemented with special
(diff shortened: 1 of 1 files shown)


engine/src/witness/eth_elections.rs
```diff
@@ -350,16 +360,18 @@ impl WitnessClientForBlockData<EthereumChain, Vec<ScUtilsCall>>
...
+				ScUtilsEvents::DepositToVaultAndScCallFilter(DepositToVaultAndScCallFilter {
+					sender,
+					signer,
+					amount,
+					token,
+					sc_call,
+				}) => {
+					if signer.is_zero() {
+						continue;
+					}
```
dandanlen, line 374: So, signer = 0 means that it was triggered by CCM, and if that happens we want to ignore it to avoid any chance of abuse - is that right?

dandanlen (approved):

engine/src/witness/eth_elections.rs
albert-llimos: Yes exactly. And to be clear, there is no possible abuse to the protocol, it's just that the address (Vault address converted to a SC account) is a public address so a deposit could be accessed by anybody.

PR #6642, opened 2026-06-09
feat: don't refund rejection below dust egress limit

marcellorigotti:
# Pull Request

Closes: PRO-2887

## Summary

Perform the same `DustEgreeLimit` check we do in `schedule_egress` when performing a rejection refund. No fetch action is ever dispatched in this situation.

### Extrinsics & Events

Added a new event `TransactionRejectionRefundBelowDustLimit` emitted when a rejection is skipped because the post-fee amount was below the dust limit. The amount contained in the event is the gross amount, before fee deduction.

---

* The PR is complete:
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -226,15 +226,15 @@ enum FullWitnessDepositOutcome {
 
 enum RejectionRefundDidntSucceed {
 	RetryLater,
-	RecordFailureAndAbortRefund(RejectionRefundReason),
+	RecordFailureAndAbortRefund(RefundFailureReason),
 }
 #[derive(Clone, Debug, PartialEq, Encode, Decode, TypeInfo, DecodeWithMemTracking)]
-pub enum RejectionRefundReason {
+pub enum RefundFailureReason {
 	BelowDustLimit,
 	InvalidRefundAddress,
 	FailedToBuildRejectCall(RejectError),
 	FailedToBuildExecutexSwapAndCall(ExecutexSwapAndCallError),
-	SolanaCCMWithALTsNotSupported,
+	SolanaCcmWithAltsNotSupported,
 }
 
 mod deposit_origin {
@@ -1181,7 +1181,7 @@ pub mod pallet {
 		},
 		TransactionRejectionFailed {
 			tx_id: <T::TargetChain as Chain>::DepositDetails,
-			reason: RejectionRefundReason,
+			reason: RefundFailureReason,
 		},
 		UnknownBroker {
 			broker_id: T::AccountId,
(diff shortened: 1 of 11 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1193,6 +1194,11 @@ pub mod pallet {
 			account_id: T::AccountId,
 			deposit_address: TargetChainAccount<T, I>,
 		},
+		TransactionRejectionRefundBelowDustLimit {
+			tx_id: <T::TargetChain as Chain>::DepositDetails,
+			asset: TargetChainAsset<T, I>,
+			amount: TargetChainAmount<T, I>,
+		},
```
MxmUrw: Instead of creating a new event, what do you think about improving our `TransactionRejectionFailed` event and adding the failure reason there? We could extend `RecordFailureAndAbortRefund` to contain the abort reason and emit it as part of the event?

marcellorigotti: Yes I agree that's better

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -226,8 +226,15 @@ enum FullWitnessDepositOutcome {
...
-	AbortRefundBelowDustLimit,
+	RecordFailureAndAbortRefund(RejectionRefundReason),
+}
+#[derive(Clone, Debug, PartialEq, Encode, Decode, TypeInfo, DecodeWithMemTracking)]
+pub enum RejectionRefundReason {
+	BelowDustLimit,
+	InvalidRefundAddress,
+	FailedToBuildRejectCall(RejectError),
+	FailedToBuildExecutexSwapAndCall(ExecutexSwapAndCallError),
+	SolanaCCMWithALTsNotSupported,
```
dandanlen: nit: prefer `SolanaCcmWithAltsNotSupported`

The main reason being that if there's some automatic conversion of this to snake_case anywhere in the codegen pipeline, we want it to convert to `solana_ccm_with_alts_not_supported`, not `solana_c_c_m_with_a_l_ts_not_supported..`

dandanlen (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -226,7 +226,15 @@ enum FullWitnessDepositOutcome {
 
 enum RejectionRefundDidntSucceed {
 	RetryLater,
-	RecordFailureAndAbortRefund,
+	RecordFailureAndAbortRefund(RejectionRefundReason),
+}
+#[derive(Clone, Debug, PartialEq, Encode, Decode, TypeInfo, DecodeWithMemTracking)]
+pub enum RejectionRefundReason {
```
MxmUrw: I think this should rather be called `RefundFailureReason`, no?

MxmUrw (approved):

PR #6644, opened 2026-06-10
ci: add licence check to ensure all files have license header

dandanlen:
# Pull Request

Adds a CI job to the pre-checks that ensures all rust files have a license header.

### DIFF
--- a/.github/workflows/_01_pre_check.yml
+++ b/.github/workflows/_01_pre_check.yml
@@ -99,7 +99,7 @@ jobs:
 
       - name: Check licence headers 📜
         run: |
-          missing=$(grep -rL "SPDX-License-Identifier: Apache-2.0" --include="*.rs" . | grep -v "^./target/")
+          missing=$(grep -rL "SPDX-License-Identifier: Apache-2.0" --include="*.rs" . | grep -v "^./target/" || true)
           if [ -n "$missing" ]; then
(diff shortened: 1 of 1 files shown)


MxmUrw (commented):
It seems it failed for some reason?

Also, FYI, there are tools to automate checking the SPDX headers (and I think also adding them to files). One I've used is [this one](https://reuse.software/).

.github/workflows/_01_pre_check.yml
```diff
@@ -91,6 +91,22 @@ jobs:
         working-directory: bouncer
         run: pnpm tsc --noEmit
 
+  check-license-headers:
+    runs-on: namespace-profile-default-light
+    continue-on-error: true
```
albert-llimos: Are we fine to continue on error? I agree that this check is not super important but in the past using this ended up with continuously failing tests that noone bothered fixing.

dandanlen: I added this because there were some false positives but I think we can remove it now.

dandanlen:
I rebased on main and removed the commit making it non-blocking. The reason it was failing was fixed in [7c4b466](https://github.com/chainflip-io/chainflip-backend/pull/6644/commits/7c4b4665a267dd14e8c3056b9b59f67a183cae75). 

I should be ok now.

albert-llimos (approved):

PR #6646, opened 2026-06-10
chore: add transfer fallback bouncer check

albert-llimos:
# Pull Request

Closes: PRO-2904

## Summary

Add a low hanging fruit test to check that we have not witnessed a Transfer Fallback event nor build any transfer fallback calls for any of the tests during the normal bouncer runs. This is to prevent scenarios such as the one fixed in  #6627 .
It is implemented as a check at the end of the tests, same as for the Solana nonce checks. That is better than having to monitor for those events during all the concurrent tests.

---

## *Checklist*

* The PR is complete:


nmammeri (approved):
LGTM, one refactor suggestion instead of using prisma query directly inside the test it might be better to make a helper function in indexer.ts and use that instead.

PR #6651, opened 2026-06-11
chore: increase solana ledger size limit for localnets

dandanlen:
# Pull Request

This has been causing issues with the upgrade test recently.


albert-llimos (approved):

albert-llimos:
Fwiw, in the upgrade test the Solana client will be started with the last release (2.2.3 in this case) and will still run into this issue. We would need to merge it to the release branch if we want it to take effect immediately. Otherwise if it's flaky but doesn't happen often we can just let it be and it will apply after the next release.

PR #6660, opened 2026-06-16
chore: remove unnnecessary logging

albert-llimos:
# Pull Request

## Summary

Remove unnecessary logging of reverted transactions. These are transactions that are not even filtered for being related to our monitored addresses so it doesn't make sense to log them at all.
Also improved another log to see the full transaction id.

---

## *Checklist*

* The PR is complete:

### DIFF
--- a/engine/src/tron/rpc.rs
+++ b/engine/src/tron/rpc.rs
@@ -288,9 +288,9 @@ impl EvmRpcApi for TronRpcClient {
 #[async_trait::async_trait]
 impl TronRpcApi for TronRpcClient {
 	async fn get_transaction_info_by_id(&self, tx_id: H256) -> anyhow::Result<TransactionInfo> {
-		let tx_id = format!("{:x}", tx_id);
+		let tx_id_hex = format!("{:x}", tx_id);
 		let response = self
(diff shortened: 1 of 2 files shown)


engine/src/tron/rpc.rs
```diff
@@ -299,7 +299,7 @@ impl TronRpcApi for TronRpcClient {
 		if let Some(obj) = response.as_object() {
 			if obj.is_empty() {
 				return Err(anyhow!(
-					"Transaction info not available yet for tx_id: {}. The transaction may still be processing or does not exist.",
+					"Transaction info not available yet for tx_id: {:#x}. The transaction may still be processing or does not exist.",
 					tx_id
```
MxmUrw: There's a `format!()` just above there, I think we'd have to update that one instead.

Also there's a second instance of the same error printing in the same file, we might want to update that as well.

albert-llimos: I have pushed a commit. Is that what you meant?

MxmUrw: Yes!

MxmUrw (approved):

dandanlen (approved):

PR #6661, opened 2026-06-17
chore: migration: delete pre-allocated tron channels\

dandanlen:
# Pull Request

Fixes a small config issue with tron and removes incorrect storage.


nmammeri:
LGTM. It would be useful if we can verify this bu running try-runtime test against a snapshot from mainnet.

nmammeri (approved):

PR #6662, opened 2026-06-17
feat: bouncer improvements

nmammeri:
# Pull Request

## Summary

This PR introduces some improvements to bouncer tests:
* Adds a another chaifnflip io step to to the swaps test ensuring that either a BroadcastSuccess or BroadcastAborted events are observed making tests fail fast with meaningful errors.
* Format ccm metadata to not pollute logs 
* Increase the maxTimeRatio of the tron node config to avoid OUT_OF_TIME errors
* Increase the assethub pool limit to reduce assethub tests flakiness

## API Changes

### RPCS

### Extrinsics & Events

---

* The PR is complete:
  * Tests:


bouncer/shared/new_swap.ts
```diff
@@ -23,6 +23,20 @@ export type DcaParams = {
...
+  metadata: CcmDepositMetadata | undefined,
+  maxLength: number = 10,
+): string {
+  if (!metadata) {
+    return 'undefined';
+  }
+  return JSON.stringify({
+    ...metadata,
+    message:
+      metadata.message != null ? `${metadata.message.slice(0, maxLength)}...` : metadata.message,
```
MxmUrw: I *think* this can be simplified to
```suggestion
      `${metadata.message?.slice(0, maxLength)}...`,
```
The `?.` operator returns `undefined` for the expression around it if the field doesn't exist.

bouncer/shared/perform_swap.ts
```diff
@@ -360,6 +367,58 @@ async function waitForCcmExecution<A = []>(
...
+    `${destChain}IngressEgress.BatchBroadcastRequested`,
+    z
+      .object({
+        broadcastId: z.number(),
+        egressIds: z.array(z.tuple([cfPrimitivesChainsForeignChain, numberOrHex])),
+      })
+      .refine((event) =>
+        event.egressIds.some((e) => e[0] === egressId[0] && `${e[1]}` === `${egressId[1]}`),
+      ),
+  );
```
MxmUrw, line 229: It would be nice if we could use the auto generated event schemas, but I guess we can't because we have to be chain generic here.

bouncer/shared/perform_swap.ts
```diff
@@ -75,7 +82,7 @@ export async function requestNewSwap<A = []>(
   dcaParams?: DcaParams,
 ): Promise<SwapParams> {
   cf.debug(
-    `Requesting swap with sourceAsset ${sourceAsset}, destinationAsset ${destAsset}, destinationAddress ${destAddress} and metadata ${JSON.stringify(messageMetadata)}`,
+    `Requesting swap with sourceAsset ${sourceAsset}, destinationAsset ${destAsset}, destinationAddress ${destAddress} and metadata ${formatCcmDepositMetadata(messageMetadata)}`,
```
albert-llimos, line 80: While this is the right change to keep the same functionality, I am wondering if we need this at all. The message metadata is generally random and in almost no scenario it will help with the debugging, especially if we are anyway limiting the printed data. I think that in the spirit to avoid not useful code I would just remove the printing of the metadata and therefore the `formatCcmDepositMetadata`.

albert-llimos (approved):

bouncer/shared/perform_swap.ts
j4m1ef0rd: > guess we can't because we have to be chain generic 

Should be covered in one of my upcoming PR's.

nmammeri: I think we could make it use the generated event schemas it will just be a bit verbose. It is more concise this way because all of those events are similar they just have different names.
Let's with this and when @j4m1ef0rd PR is in we change it to be fully typed.

bouncer/shared/perform_swap.ts
MxmUrw: I think it's just a bit nicer to print a bit of the metadata. Then it's obvious that metadata exists and in case we do have to search for the swap somewhere we have a bit more of an "identifier".

nmammeri: I agree the ccm metadata is always random and printing it's content doesn't add much value. 
I don't have a strong opinion about this but I feel it is useful in differentiating CCM swaps from non ccm swaps. 
Also it is might be helpful to peek into the ccm data content somehow (in cases where there is a bug where the data is not random for some reason).

albert-llimos: I'm fine with either. Let's have the print for now.

PR #6668, opened 2026-06-19
feat: bouncer type safe client

j4m1ef0rd:
# Pull Request

Closes: PRO-2897

## Summary

Converts most of the bouncer to use a Dedot client instead of the polkadot client. This gives us solid types for extrinsics, queries, rpc's and errors. Added a new `generate_chaintypes` command that gets ran automatically just like the event schemas. 

**Note**: we don't use the dedot generated events. They overlap with our generated schema events, but because we want to use the indexer and the `event.refine()` system, we use our generated events instead.

- New dedot extrinsic submission code with mutex, nonce tracking, timeout and connection limit features.
- Replaced most of the extrinsic and gov extrinsic uses with the new dedot versions.
- Updated most tests to use the new dedot client and added any type conversions that they needed.
- Changed all storage queries to use the new dedot client.
- Changed most places that we match on an extrinsic errors to use type safe dedot errors.
- Fixed the `sendHubAsset` being flaky. It would double send the asset when it gets a `retracted` tx and re-tries. 

---

* The PR is complete:


bouncer/commands/force_sol_nonces.ts
```diff
@@ -7,8 +7,8 @@ import { submitGovernanceExtrinsic } from 'shared/cf_governance';
 async function forceRecoverSolNonce(nonceAddress: string, nonceValue: string) {
   await submitGovernanceExtrinsic(async (chainflip) =>
     chainflip.tx.environment.forceRecoverSolNonce(
-      decodeSolAddress(new PublicKey(nonceAddress).toBase58()),
-      decodeSolAddress(new PublicKey(nonceValue).toBase58()),
+      decodeSolAddress(new PublicKey(nonceAddress).toBase58()) as `0x${string}`,
+      decodeSolAddress(new PublicKey(nonceValue).toBase58()) as `0x${string}`,
```
MxmUrw: We might want to update `decodeSolAddress` to return the type `0x${string}` directly. I think the conversion from this to `string` is implicit so it would be compatible with all existing call sites.

bouncer/commands/lp_schedule_swap.ts
```diff
@@ -56,19 +50,19 @@ const swapEvent = observeEvent(logger, `swapping:CreditedOnChain`, {
...
-    .signAndSend(lp, { nonce }, handleSubstrateError(chainflip));
-});
+await signSendAndFinalize(
+  chainflip,
+  chainflip.tx.liquidityProvider.scheduleSwap(
+    BigInt(amountToFineAmount(amount.toString(), assetDecimals(inputAsset))),
+    inputAsset,
+    outputAsset,
+    0, // Retry duration
+    { minPrice: 0n, maxOraclePriceSlippage: undefined },
```
MxmUrw, line 60: It's interesting that the case changed here. Does this mean that we had a stale command implementation or does this mean the generated extrinsic interface accepts a different format?

bouncer/commands/observe_block.ts
```diff
@@ -13,9 +13,12 @@ import { getChainflipApi } from 'shared/utils/substrate';
 
 async function main(): Promise<void> {
   await using api = await getChainflipApi();
-  const expectedBlock = process.argv[2];
-  while ((await api.rpc.chain.getBlockHash(expectedBlock)).every((e) => e === 0)) {
+  const expectedBlock = Number(process.argv[2]);
+  // chain_getBlockHash returns the all-zero hash (or null) until the block exists.
+  let hash = await api.rpc.chain_getBlockHash(expectedBlock);
+  while (!hash || /^0x0+$/.test(hash)) {
```
MxmUrw: I think it's a bit more obvious to just write out `"0x0000000000000000000000000000000000000000000000000000000000000000"` in this case.

MxmUrw (commented):
This is so much better! (compared to our previous, untyped life) And it also looks pretty uninvasive in terms of overall changes.

Looks good in general, I've added a few random comments. I'll have a second deeper look in the coming days.

bouncer/commands/lp_schedule_swap.ts
j4m1ef0rd: I think it was broken here. In signed_runtime_call.ts we where using `minPrice: 1000000000000000`. Silently failed here because of 0 and null values. But this makes a 3rd example of places the bouncer was silently broken because of types.

bouncer/shared/utils/chainflip_io.ts
```diff
@@ -605,7 +555,7 @@ export class ChainflipIO<Requirements> {
 export async function newChainflipIO<Requirements>(logger: Logger, requirements: Requirements) {
   // find out current block height
   await using chainflipApi = await getChainflipApi();
-  const currentBlockHeight = (await chainflipApi.rpc.chain.getHeader()).number.toNumber();
+  const currentBlockHeight = Number(await chainflipApi.query.system.number());
```
MxmUrw: I think we don't need the type cast anymore.
```suggestion
  const currentBlockHeight = await chainflipApi.query.system.number();
```

bouncer/shared/setup_elections.ts
```diff
@@ -37,7 +38,7 @@ export async function setupElections<A = []>(cf: ChainflipIO<A>): Promise<void>
     response[2].safetyMargin = ingressSafetyMargin;
 
     // update election settings
-    await submitGovernanceExtrinsic((api) =>
+    await submitGovernanceExtrinsicPolkadot((api) =>
```
MxmUrw: Is there a reason why this uses the "legacy" method? I think this file is the only user so if we also migrated it we could remove the old `submitExistingGovernanceExtrinsicPolkadot` method.

bouncer/shared/utils/dedot.ts
```diff
@@ -0,0 +1,314 @@
...
+        queue.push(resolve);
+      });
+    }
+    try {
+      return await fn();
+    } finally {
+      release();
+    }
+  };
+}
```
MxmUrw: I *think* this is doing the same as what a semaphore does. We're already using them in `send_evm.ts` so it might be easier to simply use those instead of rolling our own? It would also keep the code more uniform.

bouncer/shared/utils/substrate.ts
```diff
@@ -153,7 +205,7 @@ export const { cachedDisposableFactory: getAssethubApi } = getCachedSubstrateApi
 );
 
 const apiMap = {
-  chainflip: getChainflipApi,
+  chainflip: getChainflipPolkadotApi,
   polkadot: getPolkadotApi,
   assethub: getAssethubApi,
 } as const;
```
MxmUrw, line 215: I assume eventually we'll be able to move everything over to use dedot, right? We can do that in separate PRs of course. Or are there major blockers?

bouncer/shared/fund_flip.ts
```diff
@@ -67,7 +69,20 @@ export async function fundFlip<A = []>(cf: ChainflipIO<A>, scAddress: string, fl
...
+  await using chainflip = await getChainflipApi();
+  for (let attempt = 0; attempt < 30; attempt++) {
+    const finalized = await chainflip.at((await chainflip.block.finalized()).hash);
+    if ((await finalized.query.flip.account(scAccount)).balance > 0n) {
+      return;
+    }
+    await sleep(1000);
+  }
+  throw new Error(`Funding of ${scAccount} confirmed via event but not finalized in time`);
 }
```
MxmUrw, line 88: It's a bit unfortunate that we have to deal with synchronization issues between the two. I don't see an immediate nice solution, unless dedot has a hidden configuration to check against unfinalized?

bouncer/shared/lending.ts
```diff
@@ -66,7 +66,7 @@ export async function addLenderFunds<A extends WithLpAccount>(
     extrinsic: (api) =>
       api.tx.lendingPools.addLenderFunds(
         asset,
-        amountToFineAmount(amount.toString(), assetDecimals(asset)),
+        BigInt(amountToFineAmount(amount.toString(), assetDecimals(asset))),
```
MxmUrw: Seem like we're now converting a lot between string and number / int when using `amountToFineAmount`. Maybe we can update the function eventually for that new interface if it turns out that this is how we mostly use it.

bouncer/shared/utils/substrate.ts
j4m1ef0rd: Yeah i'm working on converting the blockers in other PR's.

bouncer/shared/lending.ts
j4m1ef0rd: I already have an `amountToFineAmountBigInt` that would be better here (i'll do that now), but yes I agree, I will have to do some cleanup later and rethink some of the utilities.

bouncer/shared/setup_elections.ts
j4m1ef0rd: No, It gets removed in a later commit in this PR.

bouncer/shared/fund_flip.ts
j4m1ef0rd: No setting that AI could find. 🤷‍♂️ 😞

bouncer/shared/setup_elections.ts
j4m1ef0rd: oh, I thought I did. Guess I didn't remove it yet.But I just discovered a bug in Dedot when trying to write a gov call for a updatePalletConfig item. Its gets a `RangeError: Offset is outside the bounds of the DataView`. So we might want to keep the old polkadot code around as a fallback for now.

MxmUrw: > RangeError: Offset is outside the bounds of the DataView

Hmm interesting...

> So we might want to keep the old polkadot code around as a fallback for now.

Yeah I agree then.

Btw if this was for the setupElections: there is something funny going on with the typeinfo for them I think. Even with our old polkadot js library I had to submit data which wasn't actually of the correct type that I expected it should be based on rust types. Something like I had to submit a dictionary containing lists but I would have expected just a list or something like that.

bouncer/shared/utils.ts
```diff
@@ -361,6 +373,14 @@ export function chainGasAsset(chain: Chain): Asset {
   }
 }
 
+// JSON has no BigInt. Decode decimal-string values to BigInt so large u64/u128 amounts survive;
+// non-numeric strings (asset names, addresses) and JS numbers are left as-is.
+export const bigintReviver = (_key: string, value: unknown) =>
+  typeof value === 'string' && /^-?\d+$/.test(value) ? BigInt(value) : value;
+
+export const bigintReplacer = (_key: string, value: unknown) =>
+  typeof value === 'bigint' ? value.toString() : value;
```
MxmUrw, line 411: nit: These look a bit scary :)

I think bigintReviver is not used anywhere, so could be removed I assume? And bigintReplacer is only used in dedot to format strings that are displayed to the user AFAICT, so it might make sense to move it to the dedot file.

bouncer/shared/setup_elections.ts
j4m1ef0rd: No, not the elections, they seem fine, the bug was on some other code I'm working on. It couldn't encode `tradingStrategy.updatePalletConfig([{ MinimumDeploymentAmountForStrategy: { asset, amount: Some(u128) } }])`

MxmUrw (approved):
LGTM!

One last nit comment and also there seems to be some issue with the 'Bsc' chain type in the bouncer type check?

When things are green I think we can merge!

bouncer/shared/utils.ts
j4m1ef0rd: I put them here because I am using them in some other scripts that i'm working on that will be in an upcoming PR.

MxmUrw:
Getting this in the upgrade test:
```
No new blocks received for over 30000ms, triggering reconnection with endpoint switch...
Error: Timed out after 600s. Waiting on: upgrade_binaries.ts
   startTime: 2026-07-03T04:04:15.081Z
    at runWithTimeout (/opt/actions-runner/_work/chainflip-backend/chainflip-backend/bouncer/shared/utils.ts:425:15)
    at runWithTimeoutAndExit (/opt/actions-runner/_work/chainflip-backend/chainflip-backend/bouncer/shared/utils.ts:447:9)
    at <anonymous> (/opt/actions-runner/_work/chainflip-backend/chainflip-backend/bouncer/commands/upgrade_binaries.ts:50:7)
```

This looks like it might be a message from dedot, maybe it can't properly deal with a restart for some reason and we have to manually help it reconnect?

PR #6669, opened 2026-06-19
fix: emit log message when order fills fails to decode

dandanlen:
# Pull Request

This caused some confusion recently - the behaviour is ok but we should log a message when it happens.

### DIFF
--- a/state-chain/custom-rpc/src/order_fills.rs
+++ b/state-chain/custom-rpc/src/order_fills.rs
@@ -58,6 +58,35 @@ where
 	let prev_pools = StorageQueryApi::new(client)
 		.collect_from_storage_map::<pallet_cf_pools::Pools<_>, _, _, _>(header.parent_hash)?;
 
+	// Pools present now but missing from the previous block can't yield a fill delta, so they're
+	// skipped below. Log why rather than dropping them silently: across a runtime upgrade the
+	// parent block's pool storage may not decode under the current `Pool` type.
+	let missing_prev_pools =
+		pools.keys().filter(|pair| !prev_pools.contains_key(*pair)).copied().collect::<Vec<_>>();
+	if !missing_prev_pools.is_empty() {
+		// Key present in the parent but value missing from `prev_pools` => decode failure;
+		// key absent => pool newly created. Only scan keys in this rare branch.
+		let prev_pool_keys = StorageQueryApi::new(client)
+			.collect_keys_from_storage_map::<pallet_cf_pools::Pools<_>, _, _, HashSet<_>>(
+				header.parent_hash,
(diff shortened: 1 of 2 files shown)


state-chain/custom-rpc/src/order_fills.rs
```diff
@@ -52,11 +52,45 @@ where
...
+			)?;
+		for pair in missing_prev_pools {
+			if prev_pool_keys.contains(&pair) {
+				log::warn!(
+					"order_fills: previous pool state for {pair:?} failed to decode at block #{} \
+					 ({hash:?}); skipping its order fills for this block.",
+					header.number,
+				);
+			} else {
+				// This should be almost impossible: requires creating the pool, adding and
```
msgmaxim: nit: the comment implies that entering this line is rare, but in reality this will be reached every time we add a new pool? Presumably you just mean that it is safe to ignore order fills in this case?

state-chain/custom-rpc/src/order_fills.rs
```diff
@@ -52,11 +52,45 @@ where
...
 
-	let prev_pools = StorageQueryApi::new(client)
-		.collect_from_storage_map::<pallet_cf_pools::Pools<_>, _, _, _>(header.parent_hash)?;
+	let prev_pools: BTreeMap<_, _> = StorageQueryApi::new(client)
+		.collect_from_storage_map::<pallet_cf_pools::Pools<Runtime>, _, _, _>(header.parent_hash)?;
+
+	// Pools present now but missing from the previous block can't yield a fill delta, so they're
+	// skipped below. Log why rather than dropping them silently: across a runtime upgrade the
+	// parent block's pool storage may not decode under the current `Pool` type.
+	let missing_prev_pools = pools
```
msgmaxim: nit: IMO the name is misleading. To me it reads "previous pools that are now missing". Took me a while to understand (depsite the comment describing this correctly). I like one of the names Claude suggeted: `new_or_undecodable_pools ` -- makes it clear later on that we are trying to distinguish between the two cases.

state-chain/custom-rpc/src/order_fills.rs
dandanlen: Sorry I agree it's confusing. What I meant was that it should be almost impossible to enter this condition *and* to skip an order fill. ie. when we do enter this condition it means this is a new pool, and in that case it's highly unlikely that there will be any fills. But if there were, we would skip them. 

I've updated the comment and made it log this as in info - seems like useful context that it's a new pool, but I think we can ignore the edge case of possibly skipping a fill.

msgmaxim (approved):

PR #6675, opened 2026-06-22
feat: configurable per-chain refund multiplier

dandanlen:
# Pull Request

Adds a per-chain storage item for tx fee refund multiples.


marcellorigotti (approved):

marcellorigotti:
Do we want to set the values in a migration or are we fine in doing it via governance later on?

dandanlen:
> Do we want to set the values in a migration or are we fine in doing it via governance later on?

The migration is in #6676

albert-llimos (approved):
It needs to be formatted to pass the checks by lgtm!

PR #6684, opened 2026-06-23
feat: migrate AssetMap before v16 to new migration system

dandanlen:
# Pull Request

Some minor additions to the PR following review.


MxmUrw (approved):
Thanks! Maybe we can rename this PR to something like `feat: migrate AssetMap before v16 to new migration system`?

PR #6691, opened 2026-06-26
feat: bouncer aggregate/generic events

j4m1ef0rd:
# Pull Request

Closes: PRO-2938

## Summary

Using a simple code generating script that runs automatically, I generate an `aggregate` even file for each set of events that is "per chain". This gives us an indexable version of the event for use in the bouncer. This removes one of the last blockers for fully using the new schema event descriptor system.

```ts
// old (string building during runtime)
`${chain}IngressEgress.TransferFallbackRequested`,

// new (type safe, compile time)
ingressEgressTransferFallbackRequestedEvent[chain].name
```

- Refactored the gas limit ccm test
- Converted a few other places to use the new aggregate events
- Removed the `events_lookup` system

View individual commits to avoid code gen clutter.

---

* The PR is complete:
  * [x] Reviewers are assigned.


bouncer/commands/generate_generic_events.ts
```diff
@@ -0,0 +1,161 @@
...
+const uncap = (s: string): string => (s ? s[0].toLowerCase() + s.slice(1) : s);
+
+const chains = [...chainflipChains, ...legacyChainflipChains];
+
+// Some pallets are instanced per *cryptographic scheme* rather than per chain, so one generated
+// directory backs several chains via a non-chain prefix (e.g. `evmThresholdSigner`).
+const cryptoInstanceChains: Record<string, string[]> = {
+  evm: ['Ethereum', 'Arbitrum'],
+  polkadot: ['Assethub'],
+};
```
MxmUrw: Would it be possible to go the other way round and list generic pallet names here? So we would have `Broadcaster`, `IngressEgress`, ... Then all directories which end with that string are instances of this pallet. Then we won't have to treat chain-instantiated pallets differently from e.g. chaincrypto-instantiated pallets (like `Thresholdsigner`). Instead, we'll just take all the folders we can get for a given generic pallet and their prefixes will be the instance names.

Does this make sense?

bouncer/commands/generate_event_schemas.ts
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env -S pnpm tsx
 import generate from '@chainflip/processor/generate';
 import * as path from 'path';
+import generateAggregatedEvents from './generate_aggregated_events';
```
MxmUrw: nit: I don't think it matters here but usually we should not use `./` in module import paths. (Because this messes with how js instantiates global variables).

bouncer/tests/gaslimit_ccm.test.ts
```diff
@@ -65,27 +66,37 @@ function getChainMinFee(chain: Chain): number {
...
-
-  logger.debug(`${chain} fees: ${JSON.stringify(eventData)}`);
-
-  const { baseFee, priorityFee } = eventData.newChainState.trackedData as {
-    baseFee: number | undefined;
-    priorityFee: number;
-  };
-
-  return { baseFee: baseFee || 0, priorityFee };
 }
```
MxmUrw, line 81: I had a bit of a look at this. It would be nice if we didn't have to do a match on all chains here and do something more akin to the event name extrapolation we did previously.

Luckily, with the aggregated events, this works even in a type-safe way!

It did require updating ChainflipIO a bit, you can have a look here: https://github.com/chainflip-io/chainflip-backend/pull/6696

bouncer/tests/gaslimit_ccm.test.ts
```diff
@@ -128,36 +139,61 @@ async function executeAndTrackCcmSwap<A = []>(
...
-    ).event
-  ).data.transactionPayload;
-  cf.debug(`${tag} Found txPayload: ${txPayload}`);
-
-  return { tag, destAddress, broadcastId, txPayload };
+  // transactionPayload is chain-specific
+  switch (destChain) {
+    case 'Ethereum':
+    case 'Arbitrum':
+    case 'Bsc': {
```
MxmUrw: Would it be possible to also eliminate this match statement like for `getChainFees`?

bouncer/commands/generate_aggregated_events.ts
```diff
@@ -0,0 +1,161 @@
...
+    .map(({ chain, palletDir, eventBase }) => `  ${chain}: ${palletDir}${cap(eventBase)}Event,`)
+    .join('\n');
+
+  return `${imports}\n\nexport const ${constName} = {\n${entries}\n} as const;\n`;
+};
+
+export default async function generateAggregatedEvents(eventsDir: string): Promise<void> {
+  const aggregatedDir = path.join(eventsDir, 'aggregated');
+  // Clear the output directory
+  await fs.rm(aggregatedDir, { recursive: true, force: true });
```
MxmUrw: What do you think about calling these events "generic events" (and also the directory they're in). It feels like this would be a bit more descriptive (a generic pallet has generic event schemas generated for it).

MxmUrw (commented):
I like how things are coming together for a properly type-safe bouncer.

bouncer/commands/generate_generic_events.ts
j4m1ef0rd: We will still need to have a map evm = ['Ethereum', 'Arbitrum'], so the code knows what chains the generic pallet is mapped to. Assuming we want to keep the events chain indexed. 
Maybe I don't understand what saying here. But if the output is the same I don't think it really matters much.

MxmUrw: > We will still need to have a map evm = ['Ethereum', 'Arbitrum']

No we wouldn't, we would access the threshold signer event instance by actually using the ChainCrypto name, i.e. `thresholdSignerFailureReportProcessed['Evm']` and `thresholdSignerFailureReportProcessed['Bitcoin']`. This makes for a more consistent interface to the developer that doesn't have any magic involved (automatically mapping crypto names to chain names).

*If* we want to deal with these events on a chain-by-chain basis we can additionally have a record which maps chain to chain crypto:

```ts
const chainCrypto = {
  Bitcoin: 'Bitcoin',
  Ethereum: 'Evm',
  Arbitrum: 'Evm',
  ...
};
```

This way the event aggregator requires less domain knowledge about chainflip (it doesn't need to know about chains or chain crypto or any specific instances). It can do a very generic event aggregation and the rest is up to our actual bouncer codebase.

Also, fwiw, we currently don't generically refer to the threshold signer pallet at all in bouncer, so what I sketched above with `chainCrypto` isn't even necessary.

**EDIT**: another advantage of this approach would be that I suspect it would simplify the logic for gathering instances (.e.g. there would be no need for distinguishing between "exact" and "not exact" matches)

j4m1ef0rd: I think you are right. I made the change. Much cleaner code-gen and more robust.

MxmUrw (approved):
Very nice!

PR #6699, opened 2026-07-02
fix: limit witness cull batches

dandanlen:
# Pull Request

This adds a hard limit to the number of witness storage entries we cull during epoch rotations. This is to avoid/reduce spikes in memory and CPU during culling (which happens during rotations...).


MxmUrw (approved):

marcellorigotti (approved):

marcellorigotti:
I mistakenly pushed a commit to this branch, I reverted it back 👍

PR #6701, opened 2026-07-03
feat: support enums in migration system

MxmUrw:
# Pull Request

Closes: PRO-2954, PRO-2960, PRO-2987

## Summary

This PR adds support for enums in `#[cf_proc_macros::generate_module]`.

The core of the change is a new branch for the `generate_module!` declarative macro in `utilities/src/without_std/generate_module.rs`. But the implementation had to deal with a few complications along the way and a few more things had to be added.

New components:
1. Trait `HasTypeIntrospection`. This trait has two methods `is_empty_type()` (returns true iff `Self` is empty and cannot be constructed) and `sample_all_shapes()`. (returns a vector of sample values of type `Self` - e.g. for enums there's a sample value for every variant). There's also a new `#[derive(HasTypeIntrospection)]` macro.
2. Proc macro `cf_proc_macros::generic_modules!`. This wrapper parses all rust code that's inside as-is but also introduces a new syntax feature: anonymous modules that take a list of generic type parameters (`mod (A: Clone) (B: Debug) { // normal rust code }`). All definitions within that macro inherit these generic parameters without having to mention them.
3. Derive macro `#[derive(IntroElim)]`. This derives helper functions which are syntactically easier to work with than raw rust syntax. For enums it derives an `Enum::elim()` method which is like a `match` statement, but treats all variants the same way, independent of whether they are tuple-, struct- or empty variants. For structs it derives a `Struct::intro()` variant which is effectively a `new()` method that forwards all arguments.
4. Trait `IsEmptyType`. This trait describes the fact that a type cannot be constructed. It's useful as a bound on error types which describes that the error cannot actually occur.

Notes on why these became necessary:
 - (1.1): The default (derived) codec implementation for enums assigns a variant index to every variant. This is a problem in our setup because we want to create "historical versions" of enums by instantiating them with according variant types. In particular, assigning type `Never` to a variant should make it "disappear" in a historical version (simulating the fact that this variant was only introduced later). This is not possible with the default derived implementation, so a custom one had to be written. While en- or decoding it uses `HasTypeIntrospection::is_empty_type()` to know whether it should skip an index because for this particular instantiation this variant is marked as "non-existant".
 - (1.2): Previously testing of historical type encodings relied entirely on proptest's `derive(Arbitrary)` implementation. This works very reliably if we're only dealing with structs because for these it's enough to test a single example value to make sure that en-/decoding works. For enums, *every* variant has to be in the set of sampled values to verify compatibility. In particular with nested enums within e.g. structs it's not obvious how many proptest iterations are required for the tests to guarantee full coverage. The trait method `HasTypeIntrospection::sample_all_shapes()` is supposed to return all the different shapes that a type can take, such that they cover all possible cases that are relevant for codec compatibility.
 - (2) Rust declarative macros deal very badly with nested repetitions of meta variables (e.g. `$($name:ident $($more_names:ident)*)*`). This becomes a problem in particular with type bounds that have to be repeated within the context of a "repeated" list of enum variants. A very general solution to this problem is the `cf_proc_macros::generic_modules!` macro which enables a syntax where type bounds don't have to be repeated over and over and thus circumvents the aforementioned deficiency of macro meta variable repetitions. It also makes the main `generate_module!` macro a bit easier to read, hopefully.
 - (3) The intro/elim helpers from `#[derive(IntroElim)]` where mostly required because of `_phantom` fields. If a user-struct already contains a `_phantom` field, it's generic variant's own phantom field will be named `_phantom2`. This makes it impossible to e.g. construct that struct in a context where it's not clear whether there was a name collision or not.
 - (4): Migrations can now be fallible. If an enum variant is newly introduced, then there's no correct historical counterpart for this value when migrating backward. The possible errors that can occur are tracked through associated types in the Migration trait. This makes the interface to migrations a bit more cumbersome (there are now `try_backwards` and `try_forwards` instead of the cleaner `backwards()` and `forwards()` methods). To recreate the previous interface for infallible migrations, the `IsEmptyType` bound exists. If it is implemented for the error type it can be statically proven that the error cannot occur, and thus can be safely discarded.

Further notes:
 - All crates that use `#[cf_proc_macros::generate_module]` now have to have the `#[feat(lazy_type_alias)]` feature enabled. This required a few additional type annotations or visibility modifications in some of our existing pallets. 

## API Changes

None

---

* The PR is complete:
  * [ ] Reviewers are assigned.
  * Tests:


utilities/proc-macros/src/generic_modules.rs
```diff
@@ -0,0 +1,1079 @@
...
+			if !fields
+				.named
+				.iter()
+				.any(|field| field.ident.as_ref().is_some_and(|ident| ident == &candidate))
+			{
+				break candidate;
+			}
+
+			phantom_index += 1;
+		};
```
dandanlen, line 874: Re. point (3), would it not be simpler to choose another less-likely name for the field (`__generic_mod_phantom` for example) and then just emit a compiler error if there's a collision? Then we don't need to worry about collisions and maybe we don't need some of the `IntroElim` helper?

MxmUrw: Unfortunately that won't work because there are two "nested" locations where a phantom field is added by the same mechanism: which means that the mechanism will always run into a phantom field that was added by itself.

(This is because each enum variant produces an internal struct with the same fields (+ phantom), and then there's an additional call of `#[generate_module]` on this struct which itself generates a second phantom field into the generic representation struct of the enum-variant-struct.)

dandanlen (approved):

PR #6703, opened 2026-07-03
fix: account for db reads/writes in benchmark weight analysis

dandanlen:
Compare total effective weight (from_parts ref_time + reads/writes priced at ParityDbWeight).

Previously the script only compared the headline base/min weights.


MxmUrw (commented):
Did you run it? How does the output look like now?

MxmUrw (approved):
Did you run it? How does the output look like now?

PR #6708, opened 2026-07-06
feat: Asset withdrawal whitelist

msgmaxim:
# Pull Request

Closes: PRO-2872

## Summary

- Whitelist module in cf-asset-balances: stores current allowed entries, pending updates and the timelock config
- Added new extrinsics for updating whitelist + cli commands
- Whitelist is opt-in (no changes for existing users)
- Once enabled, it applies to all chains: otherwise it would be easy to circumvent it by swapping funds into a different asset.
- Migrated existing refund addreses to cf-asset-balances (refund addresses are implicitly whitelisted)
- Change in `ensure_has_refund_address_for_asset`: now also returns `true` if the refund address not set but pending (otherwise it would result in bad UX when adding a new chain)

Potentiall follow-up: a cli command or rpc to get the current state of the whitelist

### Extrinsics & Events

- update_whitelist
- set_withdrawal_timelock

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -196,10 +246,44 @@ pub mod pallet {
 	pub type RefundFeeMultiple<T> =
 		StorageMap<_, Twox64Concat, ForeignChain, u32, ValueQuery, ConstU32<100>>;
 
+	/// Per-account withdrawal whitelist: active external/internal allowlists, the timelock, and the
+	/// queue of timelocked changes awaiting activation.
+	#[pallet::storage]
+	pub type WithdrawalWhitelists<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		T::AccountId,
```
dandanlen:
```suggestion
		Identity,
		T::AccountId,
```

nit: this is a bit cheaper as long as we know the account is genuine.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -196,10 +246,44 @@ pub mod pallet {
 	pub type RefundFeeMultiple<T> =
 		StorageMap<_, Twox64Concat, ForeignChain, u32, ValueQuery, ConstU32<100>>;
 
+	/// Per-account withdrawal whitelist: active external/internal allowlists, the timelock, and the
+	/// queue of timelocked changes awaiting activation.
+	#[pallet::storage]
+	pub type WithdrawalWhitelists<T: Config> = StorageMap<
```
dandanlen: This will need to be cleaned up when an account is deleted. We can use `OnKilledAccount` for this.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -196,10 +246,44 @@ pub mod pallet {
...
+	pub type MaxPendingWhitelistUpdates<T> = StorageValue<_, u32, ValueQuery, ConstU32<16>>;
+
+	/// Maximum number of active allowlist entries per account (external addresses across all chains
+	/// plus internal accounts).
+	#[pallet::storage]
+	pub type MaxWhitelistEntries<T> = StorageValue<_, u32, ValueQuery, ConstU32<100>>;
+
+	/// The refund address registered by an account for each chain. Relocated from `cf-lp`: a
+	/// registered refund address is a trusted destination, so it is implicitly allowed by the
+	/// withdrawal allowlist (see [`Pallet::ensure_withdrawal_allowed_to`]).
```
dandanlen: nit: we don't need to comment about where things used to be... ("Relocated from cf-lp..")

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -214,12 +298,89 @@ pub mod pallet {
...
+				account_id,
+				change,
+				apply_at,
+			});
+
+			Ok(())
+		}
+
+		/// Set the caller's withdrawal timelock. Increasing/enabling applies immediately;
+		/// decreasing/disabling is itself delayed by the current timelock so a stolen key can't
```
dandanlen: On reflection I'm not sure it's necessarily a good thing to have this asymmetry - it's probably simpler and less surprising to users if the timelock just applies when it's set.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -214,12 +298,89 @@ pub mod pallet {
...
+		/// instantly remove the protection.
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::set_withdrawal_timelock())]
+		pub fn set_withdrawal_timelock(
```
dandanlen: nit: it's not a withdrawal timelock, it's a timelock on updating the whitelist (`whitelist_timelock`?)

state-chain/pallets/cf-asset-balances/src/whitelist.rs
```diff
@@ -0,0 +1,603 @@
...
+use cf_chains::{AccountOrAddress, ForeignChain, ForeignChainAddress};
+use frame_support::{pallet_prelude::*, DefaultNoBound};
+use sp_std::{
+	collections::{btree_map::BTreeMap, btree_set::BTreeSet},
+	vec::Vec,
+};
+
+/// A duration or a point in time, in seconds. The withdrawal timelock works in wall-clock time
+/// (via the pallet's `TimeSource`), matching cf-funding's redemption timelock.
+pub type DurationSeconds = u64;
```
dandanlen: Why not just `Seconds`? As the comment itself notes, sometimes it's not a duration, it's a point in time.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -398,6 +559,89 @@ impl<T: Config> Pallet<T> {
...
+		owner: &Self::AccountId,
+		dest: AccountOrAddress<&Self::AccountId, &ForeignChainAddress>,
+	) -> DispatchResult {
+		let now = Self::now_secs();
+		let mut whitelist = WithdrawalWhitelists::<T>::get(owner);
+		// Lazily fold in any matured pending updates so the active allowlists are current, writing
+		// back only if that changed anything.
+		if whitelist.apply_due_updates(now, MaxWhitelistEntries::<T>::get()) {
+			if whitelist.is_empty() {
+				WithdrawalWhitelists::<T>::remove(owner);
```
dandanlen: I guess this implicitly resets the timelock back to the default? Do we want this? 

Example: 
1. I set a timelock
2. I withdraw some funds => timelock is silently deleted

state-chain/pallets/cf-asset-balances/src/whitelist.rs
dandanlen: How about just `Seconds`? It's sometimes a duration, sometimes a point in time.

state-chain/pallets/cf-asset-balances/src/whitelist.rs
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	/// Schedules `value` to take effect `delay` seconds from `now` (folding any already-matured
+	/// change first). Returns the activation time.
+	fn schedule(
+		&mut self,
+		value: T,
+		now: DurationSeconds,
+		delay: DurationSeconds,
+	) -> DurationSeconds {
```
dandanlen: This always clobbers the existing value. This makes it trivially griefable, but also, any pending update is silently deleted. I would suggest we disallow updates if there is already one pending.

state-chain/pallets/cf-asset-balances/src/whitelist.rs
```diff
@@ -0,0 +1,603 @@
...
+		let apply_at = now.saturating_add(delay);
+		self.pending = Some((value, apply_at));
+		apply_at
+	}
+}
+
+/// Per-account withdrawal timelock, in seconds (`current == 0` = restriction off). Strengthening
+/// (longer/enabling) applies immediately; weakening (shorter/disabling) is delayed by the current
+/// timelock — the policy lives in [`WithdrawalWhitelist::set_timelock`].
+pub type TimelockState = TimelockedValue<DurationSeconds>;
```
dandanlen: nit: probably not worth adding as a type alias, it's only used once and the name doesn't really clarify anything.

state-chain/pallets/cf-asset-balances/src/whitelist.rs
```diff
@@ -0,0 +1,603 @@
...
+	///
+	/// The active allowlist is capped at `max_entries`: an `Allow` that would exceed it is applied
+	/// as a no-op. This is enforced here rather than at scheduling time to keep things simple (this
+	/// prevents abuse, while real users are unlikely to ever hit this).
+	pub(crate) fn apply_due_updates(&mut self, now: DurationSeconds, max_entries: u32) -> bool {
+		let mut changed = self.timelock.collapse_if_matured(now);
+		if self.pending.first_key_value().is_some_and(|(&apply_at, _)| apply_at <= now) {
+			// `pending` keeps the not-yet-due buckets; `due` takes the rest (apply_at <= now).
+			let not_due = self.pending.split_off(&now.saturating_add(1));
+			let due = core::mem::replace(&mut self.pending, not_due);
```
dandanlen: Note that this can end up applying updates "out of order". 

For example:
1. Timelock is 10 days
2. At T=0, Update(set_timelock(1)) => scheduled at 10
3. At T=9, Update(X) => scheduled at 19
4. Any changes scheduled between 10 and 18 will apply before Update(X)

Given that the updates are not idempotent, this could lead to some unexpected behaviour. 

I'm not sure what the right behaviour is here - I think maybe simplest would be clearing any pending changes while the timelock update is pending. Otherwise we could do something more complicated like check for conflicting changes, or update existing pending updates somehow.

state-chain/pallets/cf-asset-balances/src/lib.rs
dandanlen: Ok I see `is_empty` also checks the timelock. Can we rename it to `is_default` or `can_drop`?

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -398,6 +559,89 @@ impl<T: Config> Pallet<T> {
...
+					.as_ref()
+					.and_then(|refund| refund.effective(now)) ==
+					Some(address)
+			});
+		ensure!(allowed, Error::<T>::DestinationNotAllowed);
+		Ok(())
+	}
+}
+
+impl<T: Config> RefundAddressRegistry for Pallet<T> {
```
dandanlen, line 758: We moved refund addresses to this pallet, but we still only ever use them for LPs. The main reason to move it here would be to share both the whitelisting and refundAddress logic between LPs and brokers. It seems whitelisting is indeed shared, but for example brokers currently can't / don't need to register a refund address. We should probably plan to follow up with a migration of brokers' locked withdrawal addresses to this pallet.

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -398,6 +559,89 @@ impl<T: Config> Pallet<T> {
...
+
+	fn ensure_withdrawal_allowed_to(
+		owner: &Self::AccountId,
+		dest: AccountOrAddress<&Self::AccountId, &ForeignChainAddress>,
+	) -> DispatchResult {
+		let now = Self::now_secs();
+		let mut whitelist = WithdrawalWhitelists::<T>::get(owner);
+		// Lazily fold in any matured pending updates so the active allowlists are current, writing
+		// back only if that changed anything.
+		if whitelist.apply_due_updates(now, MaxWhitelistEntries::<T>::get()) {
```
dandanlen: We only ever call `apply_due_updates` during withdrawal, meaning even already-active pending changes sit in the pending queue forever if you don't withdraw. 

It probably makes sense to apply updates whenever we mutate the whitelist struct (ie. even when we *add* a pending update).

state-chain/pallets/cf-asset-balances/src/whitelist.rs
```diff
@@ -0,0 +1,603 @@
...
+		change: &WhitelistChange<AccountId, ForeignChainAddress>,
+		max_entries: u32,
+	) {
+		let (destination, allow) = match change {
+			WhitelistChange::Allow(destination) => (destination, true),
+			WhitelistChange::Remove(destination) => (destination, false),
+		};
+		if allow && self.active_entry_count() >= max_entries {
+			return;
+		}
```
dandanlen, line 145: We should probably emit an event for this instead of silently dropping the update?

state-chain/pallets/cf-asset-balances/src/lib.rs
```diff
@@ -398,6 +559,89 @@ impl<T: Config> Pallet<T> {
...
+	fn clear_refund_addresses(who: &Self::AccountId) {
+		let _ = RefundAddresses::<T>::clear_prefix(who, u32::MAX, None);
+	}
+
+	/// Make sure refund address is either set or will be set after timelock. (If we didn't allow
+	/// the latter, it would result in bad UX where upon adding a new chain the user would have to
+	/// wait the timelock period before they can use it. This should be OK because max
+	/// timelock is capped, so the address will become active soon.
+	fn ensure_has_refund_address_for_asset(who: &Self::AccountId, asset: Asset) -> DispatchResult {
+		ensure!(
```
dandanlen, line 792: This conflicts with the definition in `get_refund_address` (uses `effective`). 

This means that the rpc might show no registered address even when  `ensure_has_refund_address_for_asset` would pass.

dandanlen (commented):
This turned out to be more work than expected 🙈 

My honest feeling is that all the lazy-update machinery adds more trouble and edge cases than it's worth... 

We should consider falling back to tried-and-tested eagerly-applied on_idle update queues. These have their own issues of coure, but at least we are mostly aware of them, and they don't infect all the regular call-sites with side-effects (and we don't need custom rpcs to read the storage items). Let me know what you think.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: >  it's probably simpler and less surprising to users if the timelock just applies when it's set.

Not sure what you mean? Always apply it immediately? But then you can easily bypass all restrictions by setting the timelock to 0 and adding new addresses with no delay.

If you mean "all updates should have delay", yeah we could do it, but currently it a bit of extra code for slightly better UX, so seemed worth it. Not a strong opinion though.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: > The main reason to move it here would be to share both the whitelisting and refundAddress logic between LPs and brokers.

Not only that: it is helpful to have refund addresses in the same pallet so that `ensure_withdrawal_allowed_to` can be implemented entirely within this pallet (which is already called by pallets other than cf-lp e.g. cf-swapping).

> We should probably plan to follow up with a migration of brokers' locked withdrawal addresses to this pallet.

Sure, we can do this too. Would likely be a separate PR.

msgmaxim:
> This turned out to be more work than expected 🙈

Yeah, I did bring this up before starting to implement this...

> My honest feeling is that all the lazy-update machinery adds more trouble and edge cases than it's worth... We should consider falling back to tried-and-tested eagerly-applied on_idle update queues.

OK, I'm experimenting with a few approaches now. What I'm trying to avoid is having to scan multiple entires on every block (this could be an attack vector). This probably means that pending updates would need to be "flat" rather than per-account, but that comes with other tradeoffs (e.g. counting number of pending updates would be more expensive). Also, the nice thing about the current design is that most of the complexity is hidden behind the witnessing module. The with the eager approach we seem to lose this property (e.g. when pending updates are now a separate storage item).

state-chain/pallets/cf-asset-balances/src/lib.rs
dandanlen: Sorry I can see how thas was ambiguous. I meant the latter: we apply the timelock there's a timelock, otherwise we don't. 

I just think it's not obvious that one of increasing/decreasing is necessarily better/worse than the other.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: Yes. In fact I removed the method now directly comparing against `Default::default()`.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: Updated. Also seemed appropriate to add this to claude.md, so I updated that too.

state-chain/pallets/cf-asset-balances/src/whitelist.rs
msgmaxim: > This makes it trivially griefable

What do you mean here? If the key is compromised the attacker can cancel pending updates only, not what's already active. I would say being able to override an existing pending update is slightly better UX (in cases where the key is not compromised, which is the vast majority of the time).

state-chain/pallets/cf-asset-balances/src/whitelist.rs
msgmaxim: Deleted the whole struct under the new design.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: No longer relevant under the new design.

state-chain/pallets/cf-asset-balances/src/whitelist.rs
msgmaxim: Added an event

state-chain/pallets/cf-asset-balances/src/whitelist.rs
msgmaxim: After some updates here is how things work now:

- only one timelock change is pending (replaces current one on update), which keeps things simple
- the changes are applied "in order", always at `now + current delay at schedule`, simple and intuitive

Yeah there are edge cases where remove and then add account while also updating the timelock where the user might be surprised by the outcome, but I would argue it is a non-issue:
- it is rare and self inflicted: the user has to be doing something strange to trigger this, not really the expected use case
- if it does happen the consequences are minor (the user can just submit the update they actually wanted again).

We could try to do something more clever, but seems not worth it.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: Do you have suggestions for how to handle this? We did discuss the update to `ensure_has_refund_address_for_asset` and the consequences of not doing so. Not sure if updating `get_refund_address` too is the right move however.

dandanlen: Yeah I agree there's no good solution here. I think it's probably ok to leave as-is as long as the FE relies on the extrinsic error (ie. it doesn't make any independent refund address checks before allowing submission).

state-chain/pallets/cf-asset-balances/src/whitelist.rs
dandanlen: Ok I see what you're saying: the important property of the timelock is not that it's an apply-before-T cutoff time, but an apply-after-T safety buffer. Framed like that, what you say makes absolute sense.

state-chain/pallets/cf-asset-balances/src/lib.rs
msgmaxim: Update: created [PRO-2998](https://linear.app/chainflip/issue/PRO-2998/relocate-brokers-locked-withdrawal-addresses-to-cf-asset-balances)

dandanlen:
This looks good to me - only thing is that as a follow up I would like us to ensure that the timelock/scheduling mechanism can't be abused to fill up blocks. Right now the on_idle reads *all* pending changes, which can accumulate for up to 10 *days*. (I know it's limited per account but still)

We can use a similar approach to elsewhere (map of block or block range -> update). 

For example we could split it like this (thanks Claude):

```
// Replaces the BTreeMap's value
PendingByAccount: StorageMap<Identity, AccountId, Vec<(Seconds, PendingChange)>>

// Replaces the BTreeMap's key, the value is a ref into PendingByAccount
DueUpdate: StorageMap<Twox64Concat, Seconds, BTreeSet<AccountId>>

// Iteration cursor
NextDueUpdate: StorageValue<Seconds>
```

Then in the `on_idle` we iterate from the NextDueUpdate to current time and process as much of the DueUpdates as possible. If we fall behind we just don't update the cursor.

dandanlen (approved):

msgmaxim:
> This looks good to me - only thing is that as a follow up I would like us to ensure that the timelock/scheduling mechanism can't be abused to fill up blocks. Right now the on_idle reads _all_ pending changes, which can accumulate for up to 10 _days_. (I know it's limited per account but still)

Yeah, and that's the main reason I initially put all the logic in the LP's extrinsics. While working on this new version, I also considered ways of making it more resistant to intentional spam (honest usage isn't going to be a problem), but decided against additional complexity (extra storage items etc). The whole argument against the lazy approach was its complexity after all. I will create a linear [ticket](https://linear.app/chainflip/issue/PRO-3002/consider-making-withdrawal-whitelist-more-resistant-to-spam) for this and we can discuss when you are available again.

PR #6712, opened 2026-07-07
feat: rewards rpc

ramizhasan111:
# Pull Request

Closes: PRO-2899

## Summary

This adds an rpc call to provide all rewards related data for product to do apy calculations for the new flip 2.1 rewards system.

### RPCS

add a new rpc from api version 20 onwards. 

### Extrinsics & Events

---

* The PR is complete:
  * [x] Reviewers are assigned.


GabrielBuragev (approved):
👍

state-chain/runtime/src/runtime_apis/custom_api/types.rs
```diff
@@ -335,6 +338,154 @@ impl<A> OperatorInfo<A> {
...
+/// fee-reward distribution has not yet been activated for the current epoch.
+#[derive(Encode, Decode, Eq, PartialEq, TypeInfo, Clone, Debug, Serialize, Deserialize)]
+pub struct RewardDistributionEstimate<Amount> {
+	pub epoch_index: EpochIndex,
+	pub current_block: BlockNumber,
+	pub current_epoch_started_at: BlockNumber,
+	pub epoch_duration: BlockNumber,
+
+	/// The current Minimum Active Bid.
+	pub bond: Amount,
```
dandanlen: nit: if this is the MAB let's call it the MAB, not the bond. (I think the comment is wrong though, it's not the MAB)

state-chain/runtime/src/runtime_apis/custom_api/types.rs
```diff
@@ -335,6 +338,154 @@ impl<A> OperatorInfo<A> {
...
+	pub authority_count: u32,
+	/// `Reserve[ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID]`.
+	pub pool_onchain_amount: Amount,
+	/// `FlipToDistribute`, clamped to be non-negative.
+	pub pool_offchain_amount: Amount,
```
dandanlen: Is the onchain/offchain distinction important for outside observers? Can we not just combine it to one number?

state-chain/runtime/src/runtime_apis/custom_api/types.rs
```diff
@@ -335,6 +338,154 @@ impl<A> OperatorInfo<A> {
...
+	/// The commission rate in effect for this epoch's snapshot.
+	pub commission_bps: u32,
+	/// Pure commission earned so far, excluding the operator's own managed nodes' earnings.
+	pub operator_reward: Amount,
+	/// Sum of managed-validator bids (uncapped).
+	pub total_validator_stake: Amount,
+	/// Sum of delegator bids (uncapped).
+	pub total_delegator_stake: Amount,
+	pub num_authority_nodes: u32,
+	pub validators: Vec<AccountReward<Amount>>,
```
dandanlen: Should validators not get a `ValidatorRewardEstimate`? Why do we need two distinct structs for 'solo validators' and managed validators?

state-chain/runtime/src/runtime_apis/impl_api.rs
```diff
@@ -907,6 +908,137 @@ impl_runtime_apis! {
...
+									epoch_index
+								);
+							return None;
+						};
+
+						let total = per_authority_share.saturating_mul(num_authority_nodes as FlipBalance);
+						let rewards: BTreeMap<AccountId, FlipBalance> = snapshot
+							.distribute(total, bond)
+							.map(|(account, amount)| (account.clone(), amount))
+							.collect();
```
dandanlen: Are we not manually reimplementing a lot of `RewardsDistribution::distribute(..)`? Why not re-use that? 

Some variation of this:

```rust
let mut rewards: BTreeMap<AccountId, FlipBalance> = BTreeMap::new();
for authority in &authorities {
    pallet_cf_validator::distribute::<Runtime>(authority, per_authority_share, |account, amount| {
        *rewards.entry(account.clone()).or_default() += amount;
    });
}
```

state-chain/runtime/src/runtime_apis/impl_api.rs
```diff
@@ -907,6 +908,137 @@ impl_runtime_apis! {
...
+							epoch_index,
+							&operator,
+						) else {
+							log_or_panic!(
+									"Operator {:?} has {} authority slot(s) in epoch {} but no delegation snapshot found.",
+									operator,
+									num_authority_nodes,
+									epoch_index
+								);
+							return None;
```
dandanlen: Not sure if we can actually hit this condition, I don't think so - but if we do would it not be better to treat the operator as a solo validator?

state-chain/runtime/src/runtime_apis/impl_api.rs
```diff
@@ -907,6 +908,137 @@ impl_runtime_apis! {
...
+						let reward_of =
+							|account: &AccountId| rewards.get(account).copied().unwrap_or_default();
+						let role_of = |account: &AccountId| {
+							pallet_cf_account_roles::AccountRoles::<Runtime>::get(account)
+								.unwrap_or_default()
+						};
+						let validator_bonds = snapshot.validator_bond_distribution(bond);
+						let bond_of =
+							|account: &AccountId| validator_bonds.get(account).copied().unwrap_or_default();
+
```
dandanlen: Is the bond not already captured in the validator's FlipAccount.bond?

state-chain/runtime/src/runtime_apis/custom_api/types.rs
ramizhasan111: This is basically according to Gabe's specification. @GabrielBuragev?

dandanlen: OK i added a couple of comments in Linear.

state-chain/runtime/src/runtime_apis/impl_api.rs
ramizhasan111: i mean, its an unexpected case that, as you said should never happen. If this happens, then something is wrong in our runtime logic. I would, in that case, be reluctant to treat it as a solo. validator.

state-chain/runtime/src/runtime_apis/impl_api.rs
ramizhasan111: The `pallet_cf_validator::distribute` first checks whether we are distributing for managed validators or solo ones and then calls `snapshot.distribute`. Since here we are already in a code path where we know its a managed validator and we have the sanpshot so I thought I call the `snapshot.distribute` directly. I can see how I can use RewardDistribution::distrbute once we finalize the new schema in linear.

state-chain/runtime/src/runtime_apis/impl_api.rs
dandanlen: As discussed elsewhere, the reason for the distinction between solo/managed isn't clear and seems unrelated to reward estimation. 

Also this same case is already handled in the existing `pallet_cf_validator::distribute` logic so if we adopt that then this question is doubly irrelevant.

state-chain/runtime/src/runtime_apis/impl_api.rs
```diff
@@ -923,106 +924,117 @@ impl_runtime_apis! {
...
-					.into_iter()
-					.filter_map(|(operator, num_authority_nodes)| {
+					let reward_of =
+						|account: &AccountId| rewards.get(account).copied().unwrap_or_default();
+
+					for operator in operators_with_authorities {
+
+						reward_pool.push(AccountReward {
+							account: operator.clone(),
+							bid: Flip::balance(&operator),
```
dandanlen: This should be zero for operators, they don't bid.

dandanlen (commented):
I think the output shape looks good now. @GabrielBuragev please confirm. 

It would be good to add an rpc snapshot test to make sure the output format is as expected.

Been chatting to claude and I think it's worth factoring out and optimising the reward distribution. Previously we paid out individual authority rewards on each block, so i made sense to have `distribute(account, amount)` but now we only ever pay out in a batch, and if an operator has N validators we end up repeating the same distribution logic N times. 

The best solution IMO is to update the RewardDistribution trait with a `distribute_all` method that batches the work per operator. Then we can expose this to the RPC: that way we can be sure that we re-use the same payout logic in both. I'll propose a change.

Btw. if we're not keeping the two PRs isolated we might as well fix the conflicts and rebase this onto the other PR.

state-chain/pallets/cf-validator/src/delegation.rs
dandanlen, line 407: We probably need similar treatment for `bond` (it needs to be for the same epoch_index as the snapshot). `HistoricalBond::<T>::get(epoch_index)`.

PR #6715, opened 2026-07-10
fix: Correctly finalise boosted vault deposits when lost

msgmaxim:
# Pull Request

Closes: PRO-2965

## Summary

Now if a boosted vault depost is lost we will correctly finalise/acknowledge this like we already do for deposit channels.

- Added `BoostedVaultTransactionExpiry` storage to keep track of each boosted deposit's expiration (decided against merging it with `BoostedVaultTransactions` to avoid migrating storage and some other unwanted side effects).
- `on_idle` checks `BoostedVaultTransactionExpiry` to see if any boost "exire" and need to be cleaned up (boost liftetime is the same as in deposit channels, uses target chain's block height).
- Successful finalisation eagerly removes no longer requried entries from `BoostedVaultTransactionExpiry`.
- All codes paths are tested (boosted -> finalised, boosted -> expired).

---

## *Checklist*

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-ingress-egress/src/lib.rs
+++ b/state-chain/pallets/cf-ingress-egress/src/lib.rs
@@ -1007,6 +1007,18 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	/// Expirty blocks (in external chain's block height) of boosted vault transactions awaiting
+	/// full witnessing. Additionally stores the boosted asset required to finalise the boost.
+	///
+	/// Unlike deposit channels, vault swaps have no channel to recycle, so this is what guarantees
+	/// that a lost boosted deposit is correctly accounted for.
+	#[pallet::storage]
+	pub type BoostedVaultTransactionExpiry<T: Config<I>, I: 'static = ()> = StorageValue<
+		_,
+		BTreeMap<TransactionInIdFor<T, I>, (TargetChainBlockNumber<T, I>, TargetChainAsset<T, I>)>,
+		ValueQuery,
+	>;
+
 	#[pallet::storage]
 	pub(super) type PendingPrewitnessedDeposits<T: Config<I>, I: 'static = ()> = StorageMap<
 		_,
@@ -1247,6 +1259,16 @@ pub mod pallet {
 		fn on_idle(now: BlockNumberFor<T>, remaining_weight: Weight) -> Weight {
 			let mut used_weight = Weight::zero();
 
+			let external_height = match TargetChainOf::<T, I>::get() {
+				ForeignChain::Arbitrum |
+				ForeignChain::Bitcoin |
+				ForeignChain::Ethereum |
+				ForeignChain::Tron |
+				ForeignChain::Bsc => ProcessedUpTo::<T, I>::get(),
+				ForeignChain::Assethub | ForeignChain::Polkadot | ForeignChain::Solana =>
+					T::ChainTracking::get_block_height(),
+			};
+
 			// Approximate weight calculation: r/w DepositChannelLookup + w DepositChannelPool
 			let recycle_weight_per_address =
 				frame_support::weights::constants::ParityDbWeight::get().reads_writes(1, 2);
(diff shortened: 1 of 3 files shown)


state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -2765,6 +2851,16 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 				boost_fee,
 			) {
 				Ok(BoostOutcome { amounts, fees }) => {
+					// If it is a vault deposit, schedule its expiry so we can detect and clean up
+					// if the corresponding full witness never arrives. (Deposit channels
+					// don't need this — they are cleaned up via recycling.)
+					if let DepositOrigin::Vault { tx_id, .. } = &origin {
+						let expiry_height = Self::expiry_and_recycle_block_height().2;
```
j4m1ef0rd: nit: I know its not from this PR, but i'm not a fan of the return type of `expiry_and_recycle_block_height`. Leading to ".2". When all 3 of the returned values are the same type, it would be easy to use the wrong value here (and other 2 places its used).

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
```diff
@@ -815,6 +822,86 @@ mod vault_swaps {
...
+				assert!(BoostedVaultTransactions::<Test, Instance1>::contains_key(tx_id));
+				assert!(
+					BoostedVaultTransactionExpiry::<Test, Instance1>::get().contains_key(&tx_id)
+				);
+			})
+			.then_execute_at_next_block(|_| {
+				// The deposit is never fully witnessed. Raise the processed height to its expiry
+				// so that on_idle (run at the end of this block) deems it lost:
+				let recycle_block = EthereumIngressEgress::expiry_and_recycle_block_height().2;
+				set_eth_processed_up_to(recycle_block);
```
j4m1ef0rd: This would not catch a smaller expiry value being used by mistake. Would be better to first go to `recycle_block-1` and confirm it still exists.

j4m1ef0rd (approved):

MxmUrw (commented):
Hey, looks good on first glance. I'll have a deeper look tomorrow morning!

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1007,6 +1013,18 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	/// Expirty blocks (in external chain's block height) of boosted vault transactions awaiting
```
MxmUrw: nit: "Expiry"

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -1247,6 +1265,16 @@ pub mod pallet {
...
 
+			let external_height = match TargetChainOf::<T, I>::get() {
+				ForeignChain::Arbitrum |
+				ForeignChain::Bitcoin |
+				ForeignChain::Ethereum |
+				ForeignChain::Tron |
+				ForeignChain::Bsc => ProcessedUpTo::<T, I>::get(),
+				ForeignChain::Assethub | ForeignChain::Polkadot | ForeignChain::Solana =>
+					T::ChainTracking::get_block_height(),
+			};
```
MxmUrw, line 1278: Note that `ProcessedUpTo::<T, I>::get()` (which is set by the election based witnessing) is the highest bitcoin blockheight for which there are no ongoing or scheduled elections, minus the SAFETY_MARGIN, which is 8 for bitcoin.

In particular, I don't think it's guaranteed that it is monotonically increasing. In fact I think if there's a reorg it *will* go down because there will be a new witnessing election for a block that has been already witnessed.

I'll have to think a bit about whether this affects the deposit channel expiry, but I suppose it doesn't. It might be worth to eventually ensure monotonicity on `ProcessedUpTo` because that's somewhat implied by the name and what most use-cases of it would expect.

For the time being maybe maybe let's change the variable name `external_height` to be a bit more descriptive, so that it won't be used in a wrong context in the future.

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
MxmUrw: Agree. Also we've already bumped the eth block height to the deposit's `recycle_at-1` so I think we just have to advance the external chain height by 1 here, no?

Also FWIW we might want to be consistent and use a different word instead of "expiry" because it's linked to the current chain height's `recycle_at` and *not* to the `expire_at`.

MxmUrw (approved):
Had a second look, looks good indeed!

Wrt architecture, I'm wondering if there might be a way for us to treat channel deposits and vault deposits more uniformly, would be nice, but would also be a much bigger scope of course.

Some nits, otherwise LGTM.

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: > For the time being maybe maybe let's change the variable name external_height to be a bit more descriptive, so that it won't be used in a wrong context in the future.

Do you have name suggestions? I tried a few names (`cleanup_height`, `processed_external_height` etc), but all seem to come with some tradeoffs. IMO `external_height` is OK: it doesn't claim that's it is finalised in any way. It is just "a height of an external chain". Either way, I don't think we can concisely convey all of its properties through a name. Instead, how about we just add a comment to say that it is not finalised, nor the latest known height? (Edit: added a comment, please take a look).

MxmUrw: > (Edit: added a comment, please take a look).

I can't seem to find it, maybe you didn't push? Anyways, I'm totally fine with putting this info in a comment - anything that makes people aware that they have to be a bit careful will serve this purpose.

state-chain/pallets/cf-ingress-egress/src/tests/boost.rs
msgmaxim: Updated.

MxmUrw (approved):

state-chain/pallets/cf-ingress-egress/src/lib.rs
msgmaxim: It was in https://github.com/chainflip-io/chainflip-backend/pull/6715/changes/c8ef907d00324a0d29bda68c018a3859ee201eb5

PR #6718, opened 2026-07-10
Feat/distribute batch

dandanlen:
# Pull Request

This adds a distribute_all method to make reward distribution more efficient. (as mentioned on the base PR). 

I also updated the RPC to use a similar pattern, it now iterates over the snapshots instead of building a list of operators first. 

This also fixes a bug in the calculation where we were using `CurrentBond()` instead of `HistoricalBond(epoch)`.

### DIFF
--- a/state-chain/pallets/cf-flip/src/lib.rs
+++ b/state-chain/pallets/cf-flip/src/lib.rs
@@ -557,10 +557,7 @@ impl<T: Config> Pallet<T> {
 			.offset(Self::deposit_reserves(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID, amount));
 	}
 
-	pub fn trigger_flip_reward_distribution(
-		epoch_index: EpochIndex,
-		authorities: BTreeSet<T::AccountId>,
-	) -> T::Balance {
+	pub fn trigger_flip_reward_distribution(epoch_index: EpochIndex) -> T::Balance {
 		// Bridge in any pending offchain rewards and deposit them into the distribution reserve so
 		// that everything is distributed from a single source.
 		let offchain_flip_to_distribute = FlipToDistribute::<T>::take();
@@ -580,7 +577,6 @@ impl<T: Config> Pallet<T> {
 		T::RewardsDistribution::distribute_all(
 			epoch_index,
 			Reserve::<T>::get(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID),
-			&authorities,
 			|account, amount| {
 				// Skip zero-value settlements: `distribute_all` calls this for every reward-pool
 				// participant, including ones with nothing due; avoid the wasted withdrawal/event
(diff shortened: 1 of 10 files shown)


state-chain/pallets/cf-flip/src/lib.rs
```diff
@@ -580,32 +576,29 @@ impl<T: Config> Pallet<T> {
...
-					Pallet::<T>::settle(
-						account,
-						Pallet::<T>::withdraw_reserves(
-							ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID,
-							amount,
-						)
+		T::RewardsDistribution::distribute_all(
+			epoch_index,
+			Reserve::<T>::get(ONCHAIN_FLIP_TO_DISTRIBUTE_RESERVE_ID),
+			&authorities,
```
marcellorigotti: I think now that we pass the `epoch_index` we don't need to also pass the list of authorities.
We can fetch them inside `distribute_all`, thus avoiding relying on the caller passing the correct list of authorities for the correct epoch.

state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -357,6 +363,43 @@ where
...
+		total_amount: Self::Balance,
+		beneficiaries: &[Self::AccountId],
+		mut settle: impl FnMut(&T::AccountId, T::Amount),
+	) {
+		if beneficiaries.is_empty() {
+			return;
+		}
+		let per_beneficiary_amount = total_amount / (beneficiaries.len() as u32).into();
+
+		// `snapshot.validators` is exactly this operator's set of `epoch_index` authorities
```
marcellorigotti: Also this is not true, there is an edge case where the snapshot can contains validator which are not authorities.

when we build the snapshot we call:

```
pub fn maybe_optimize_bid(&mut self, auction_outcome: &AuctionOutcome<Account, Bid>) {
		while self.validators.len() > 1 && self.avg_bid() <= auction_outcome.bond {
			// in the case where the operator's nodes are at the boundary, maybe some of the
			// validators didnt make the set and so we can optimize further where we reduce one node
			// and increase the avg bid which would allow us to potentially add more of the
			// operator's nodes to the set thereby increasing the number of nodes in the set.
			if self.avg_bid() == auction_outcome.bond {
				if self.validators.iter().any(|(val, _)| !auction_outcome.winners.contains(val)) {
					self.move_lowest_validator_to_delegator();
				} else {
					break;
				}
			}
			// in case where all of operator's nodes are below bond, we increase the avg bid
			// sequentially until either the avg bid is equal to bond or greater.
			else {
				self.move_lowest_validator_to_delegator();
			}
		}
	}
```

`while self.validators.len() > 1`
So if the operator's entire pooled stake (own validators + all delegators combined) still averages below the bond with just one validator left, that validator loses the auction yet stays in `snapshot.validators`.

marcellorigotti:
Both my comments have been addressed in the two commits following them

ramizhasan111 (approved):

PR #6722, opened 2026-07-14
feat: Bouncer advanced commands

j4m1ef0rd:
# Pull Request

Closes: PRO-2946

## Summary

Note: targets my Dedot client PR
Adds new bouncer abilities for claude (and humans). You can now use the bouncer skill to do a lot of stuff that would usually need a custom command/script.
Uses the new Dedot client to get all the info and types to be able to do these advanced things. Uses a fresh dedot client, so you don't need to have compatible local files, it will download the metadata when connecting to the network.

```sh
# Query any statechain storage
/bouncer what is the minimum strategy amount?

# Can also check storage on mainnet and persa
/bouncer what is the network fee minimum on mainnet?

# Set any pallet update setting
/bouncer set the network fee rate for eth to 50bps

# Set any safe mode setting
/bouncer disable liquidations
/bouncer set all code green

# Get the oracle price of any or all assets
/bouncer what is the price of Eth on mainnet?

```
You don't need to use the "/bouncer" command, claude can pick up the intent and use the skill.
You can run all these without claude using the new commands. See the descriptions at the top of each file for instructions.

---

* The PR is complete:
  * [x] Reviewers are assigned.


MxmUrw (commented):
> TODO: next step is to remove all one off commands like set_minimum_network_fee.ts

Honestly I think it's useful to have scripts that run locally and don't require an internet connection and an AI subscription. So I wouldn't go as far as deleting the existing scripts.

Also, at least according to my own bouncer experience, I'm not really sure how much we gain by optimizing these bouncer commands - I've never ever used a single one of them, except the ones for sending localnet funds to a deposit channel (i.e. eth or btc or ...). 

What's your experience, do you use them a lot?

j4m1ef0rd:
> I think it's useful to have scripts that run locally and don't require an internet connection and an AI subscription

This PR is exactly that, it's just a set of scripts that you can run with or without AI. But they are now a more general "get/set anything" instead of a script for each "set thing".

> I'm not really sure how much we gain by optimizing these

I expect the main use to be testing new features. I bet that these commands would of been useful when testing lending. That's why commands like `set_boost_network_fee_percent.ts‎` existed (I deleted it in the dedot PR).
Saves us making one off scripts. That's less to maintain and quicker to test features.

> What's your experience, do you use them a lot?

I don't use them much either. But I expect that to change:
1) maybe we didn't use the commands before because commands for the things we actually want to do didn't exist, but now that we can do anything, it might actually get used. 
2) now that the product guys are starting to use localnets, they might find it useful to just "ask claude" to do stuff, instead of having to dig into statecain code.
3) It is now faster to ask claude what a value on mainnet is, then it is to check the chain state on the web ui. I have already been using this.
4) It's another tool that claude has for debugging bouncer tests (or actual mainnet issues).

MxmUrw:
> This PR is exactly that, it's just a set of scripts that you can run with or without AI. But they are now a more general "get/set anything" instead of a script for each "set thing".

Yeah you're right. One thing I'm a bit hesitant about is that this does add quite a bit of code, and some of it seems to be quite generic, so we theoretically should have a look about which of these things can be deduplicated amongst themselves and possibly be deduplicated against our existing code. Or how/whether this code can be useful in other places in bouncer.

> 2. now that the product guys are starting to use localnets, they might find it useful to just "ask claude" to do stuff, instead of having to dig into statecain code.
> 3. It is now faster to ask claude what a value on mainnet is, then it is to check the chain state on the web ui. I have already been using this.

These make sense, I've tried it out myself and it works nicely AFAICT.

j4m1ef0rd:
Can we merge this @MxmUrw ?

MxmUrw:
What's your opinion about this?

> One thing I'm a bit hesitant about is that this does add quite a bit of code, and some of it seems to be quite generic, so we theoretically should have a look about which of these things can be deduplicated amongst themselves and possibly be deduplicated against our existing code. Or how/whether this code can be useful in other places in bouncer.

j4m1ef0rd:
> One thing I'm a bit hesitant about is that this does add quite a bit of code, and some of it seems to be quite generic, so we theoretically should have a look about which of these things can be deduplicated amongst themselves and possibly be deduplicated against our existing code. Or how/whether this code can be useful in other places in bouncer.

I replaced the messy cli args code. I'll do another pass and try and factor more out. The code should work fine when we add new pallets and storage items, it won't break, so no upkeep should be required. It might be a large chunk of code, but we can just get AI to fix it if it breaks, it's not going to stop the bouncer tests or anything. So its low risk and with AI its low maintenance.

bouncer/shared/utils.ts
```diff
@@ -415,6 +415,10 @@ export const bigintReviver = (_key: string, value: unknown) =>
 export const bigintReplacer = (_key: string, value: unknown) =>
   typeof value === 'bigint' ? value.toString() : value;
 
+// Lower-case the first letter of a pallet name to get dedot's `client.tx`/`client.query`/`errors`
+// key (dedot camelCases the leading char). Used wherever we match a runtime-chosen pallet by key.
+export const lowercaseFirstLetter = (s: string): string => s.charAt(0).toLowerCase() + s.slice(1);
```
dandanlen, line 420: This is (almost?) always used in the same context, the comment even says as much: 

`(p) => lowercaseFirstLetter(p.name) === txPallet`

Maybe we can extract the full pattern into a (type-safe?) helper?

dandanlen (commented):
Sorry for taking so long to look at this. 

I like this idea (a single "swiss army knife" command that removes the need for small ad-hoc scripts). 

I can also see the concern with maintenance of this. I don't feel equipped to read and verify the code myself. My instinct is that the querying and set-safe-mode are good.

The pallet-config-updates stuff is something i'd like to take a closer look at (tomorrow). For example, why do we need list_pallet_config_updates when we already have the query tool? Can we not already just use that to get what we want?

bouncer/commands/list_pallet_config_updates.ts
```diff
@@ -0,0 +1,179 @@
...
+// `updatePalletConfig` takes an array of updates or a single one:
+//   {
+//     "swapping": {
+//       "txPallet": "swapping",
+//       "arity": "array",
+//       "updateType": "PalletConfigUpdate",
+//       "variants": [
+//         { "type": "SwapRetryDelay", "value": { "delay": "u32" } },
+//         { "type": "MaximumSwapAmount", "value": { "asset": "enum Asset: Eth|Flip|...", "amount": "Option<u128 (pass as string)>" } }
+//       ]
```
dandanlen: I took another look at this file. Is this not replicating what we already have in the generated type metadata? Why do we need it?

dandanlen (commented):
I think this is a good idea, especially the query tool. Getting information into / out of a localnet is a genuine annoyance and it's nice to be able to do it from the command line rather than the browser.

I'm a little less sure about the config update stuff: we seem to be working around some dedot issue (I think?) and therefore have to break this out of the usual type derivation tooling (types.d.ts). I'm not sure why this is necessary. Even if it is, why do we need two separate scripts, one to generate some JSON, another to read an submit it? 

Similar for safe mode... we're writing a dedicated tool for a single extrinsic. Can we not have a (governance) submission tool that can do this using the usual type inference? 

I also agree with Maxim that deduplicating the network utils would be worthwhile.

Otherwise, like I said, the idea is good! Thanks @j4m1ef0rd

bouncer/commands/list_pallet_config_updates.ts
j4m1ef0rd: No its different. The generated types are compile time only. We need a JSON catalog of the pallets types.

j4m1ef0rd:
> we seem to be working around some dedot issue 

No, the only issue was the pallet update extrinsic submission problem that needs a fallback due to a bug.

> have to break this out of the usual type derivation tooling (types.d.ts). I'm not sure why this is necessary.

Because the types are compile time only. We are using the runtime metadata with string matching from the command line args. We can't use type safety here.

> why do we need two separate scripts

because the AI does not know the shape of the pallet update at first, so it uses the first script to find the update and its shape. Then the second script to modify and submit it.

> we're writing a dedicated tool for a single extrinsic.

The tool covers every single pallet update extrinsic. They are all different.

> Can we not have a (governance) submission tool that can do this 

The safemode and pallet updates tools are quite different. It was better to keep them seperate. They still use common code for submitting the gov extrinsic. I think there is a little more I can deduplicate, but not a big change.

bouncer/commands/list_pallet_config_updates.ts
dandanlen: But this information is already contained in the generated types - it seems like we only need JSON because the other tool from this same PR expects JSON? Or is this because we're assuming that AIs like JSON? What am I missing?

dandanlen:
> Because the types are compile time only. We are using the runtime metadata with string matching from the command line args.

I still don't understand why we need *runtime* metadata.

Why can't I just say "Claude, use the dedot types to derive the right shape of extrinsic i need to submit if I want to update config item X".

bouncer/commands/list_pallet_config_updates.ts
j4m1ef0rd: Ooooh, I understand now what you're suggesting. I built it as a human usable tool, not just for AI to use. So I need it in JSON so I can iterate over it and present it to the user. A few points:
- The types.d.ts file is 23k lines, so a human using it is not recommended.
- Nested structs within the file will require several lookups instead of being presented all at once.
- We will need to document some stuff like how to use bigInt numbers in JSON and other things that trip up the AI. Stuff that the commands already handles.
- The types can be wrong with the localnet if it is not at the same commit. Not really an issue for localnets, but if we ever want to expand these commands to work on testnets like persa, it would be a problem.
- Using a command to do the heavy lifting fits the same flow as the other commands.

Seems like a downgrade just to get rid of a <200 line file.

dandanlen: > The types.d.ts file is 23k lines, so a human using it is not recommended

Humans also have other tools for dealing with big files, such as grep 😅 

Still - I don't understand, why would a human want a JSON-formatted list (and it's 977 lines btw) of config update type info? As a human *writing* code I can just use regular type inference. 

Hence my comment about this existing mainly as input for the updater tool. If that's the case, we should just combine the two into one. 

(FWIW there is already a [governance submission tool](https://github.com/chainflip-io/cf-gov-js) that works on localnet, it just isn't in the main repo)

---
In general, I think it makes sense to consolidate these as sub-commands in a single well-known self-documenting tool. Right now we are adding four new commands with differing conventions (eg. --network is not accepted by all).

Please consider merging at least the write-oriented commands into one. Then I think we can merge this to avoid further drift, and to start finding out if it's useful. 

We should however also actively try to shape it and not let AI go wild otherwise we're going to end up with dozens of special-case tools. There needs to be some coherent design behind all of this.

dandanlen (approved):
This is better. Let's get it merged and we can improve/evolve it based on actual usage.

PR #6724, opened 2026-07-14
feat: trigger rewards distribution before transitioning to the new epoch

marcellorigotti:
Distribute rewards before the epoch ends, still while transitioning to the new epoch but before calling:
- `initialise_new_epoch()`
- `on_new_epoch()`

### DIFF
--- a/bouncer/tests/flip_reward_activation.ts
+++ b/bouncer/tests/flip_reward_activation.ts
@@ -107,8 +107,9 @@ export async function testFlipRewardActivation(testContext: TestContext) {
   const epochAfterActivation = await cf.stepUntilEvent(validatorNewEpochEvent);
   assert.strictEqual(epochAfterActivation, activationEpoch + 1);
 
-  // The epoch following activation distributes accrued fee rewards to authorities. With real
-  // swap volume behind it, a non-zero amount should actually have been distributed.
+  // The fee rewards accrued during the activation epoch are distributed at its end, in the same
+  // block as the transition to the next epoch. With real swap volume behind it, a non-zero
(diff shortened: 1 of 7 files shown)


ramizhasan111:
I am not sure what this change achieves. distributing the rewards before `initialise_new_epoch()` or after doesn't make much difference since all of it is executed atomically, at the same time, so it doesn't matter whether we distribute rewards before or after the logic in `initialise_new_epoch()`.  And this change actually introducing a new trait function just to execute this logic which makes it harder to follow the logic.

marcellorigotti:
> I am not sure what this change achieves. distributing the rewards before `initialise_new_epoch()` or after doesn't make much difference since all of it is executed atomically, at the same time, so it doesn't matter whether we distribute rewards before or after the logic in `initialise_new_epoch()`. And this change actually introducing a new trait function just to execute this logic which makes it harder to follow the logic.

Calling an `on_epoch_ending` hook before changing epoch state establishes a stronger invariant: all current epoch-dependent state still belongs to the epoch being finalized. It also removes the need for callers to infer the previous epoch from the new one and gives the transition an explicit lifecycle: finalize the old epoch, initialize the new epoch, then run new-epoch handling. With this we could also avoid passing the `epoch_index` through multiple calls and always target `current_epoch` everywhere (this is not implemented yet but I think it would make the code much cleaner and easy to reason about). 
As we've seen it is easy to mistakenly target the wrong epoch somewhere, this is fixed now but it is easy for future changes to accidentally consult the new epoch’s authorities or bond.

ramizhasan111 (approved):

PR #6726, opened 2026-07-20
chore: increase local tron node maxtime ratio

nmammeri:
# Pull Request

## Summary

This PR improves localnet tron:
* Limits often CPU timeout errors caused by congested tron node by increasing MaxTimeRatio config
* Fixes decoding error response from estimate_gas rpc which returns a hex-encoded string

## API Changes

None

---

* The PR is complete:
  * [x] Reviewers are assigned.

### DIFF
--- a/engine/src/tron/retry_rpc.rs
+++ b/engine/src/tron/retry_rpc.rs
@@ -402,10 +402,13 @@ impl<Rpc: TronSigningRpcApi> TronRetrySigningRpcApi for TronRetryRpcClient<Rpc> 
 								}
 							},
 							_ => {
+								// Tron returns the message as hex-encoded UTF-8 bytes.
 								return Err(anyhow::anyhow!(
 									"Failed to estimate energy (code: {:?}, message: {:?})",
 									energy_estimate.result.code,
-									energy_estimate.result.message,
+									energy_estimate.result.message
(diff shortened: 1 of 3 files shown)


engine/src/tron/retry_rpc.rs
```diff
@@ -402,10 +402,13 @@ impl<Rpc: TronSigningRpcApi> TronRetrySigningRpcApi for TronRetryRpcClient<Rpc>
...
 							},
 							_ => {
+								// Tron returns the message as hex-encoded UTF-8 bytes.
 								return Err(anyhow::anyhow!(
 									"Failed to estimate energy (code: {:?}, message: {:?})",
 									energy_estimate.result.code,
-									energy_estimate.result.message,
+									energy_estimate.result.message
+									.and_then(|m| hex::decode(m).ok())
+									.and_then(|bytes| String::from_utf8(bytes).ok()),
```
MxmUrw: What do you think about falling back to the original message if the parsing fails?

nmammeri: makes sense, I've also made it into a function so that we can use for other rpc calls if we find out they suffer from the same problem.

MxmUrw (approved):
Nice!

PR #6727, opened 2026-07-21
feat: Validator max bid

msgmaxim:
# Pull Request

Closes: PRO-2663

## Summary

Validators can now set "max bid" similar to how this already works for delegators:
- By default validators continue to bid their full balance (no max bid is set)
- If validator sets max bid, their bid will be capped by that amount (we don't enforce max bid <= account's balance)
- Max bid can only be updated outside of auction phase (consistent with the existing behaviour where validators can't reduce their bids during the auction)
- Max bid cannot be set below `MinimumValidatorStake`
- Added `set-max-bid` cli command for validators (created a separate ticket to update docs: PRO-3012)

## API Changes

FYI @GabrielBuragev

### RPCS

- `cf_account_info`, `cf_accounts_info` and `cf_all_account_infos`: gained new `bid` and `max_bid` fields

### Extrinsics & Events

- new `set_validator_max_bid` extrinsic: sets max bid for a validator; None resets to default (no max bid)
- new `ValidatorMaxBidUpdated` event: emitted whenever max bid is changed for a validator

---

### Before Merging:

- [ ] Check with @GabrielBuragev whether `cf_operator_info` is expected to return validator balances or bids (as part of the `managed_validators` field)?

## *Checklist*

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -978,6 +984,41 @@ pub mod pallet {
...
+				);
+			}
+
+			ValidatorMaxBid::<T>::mutate_exists(&validator, |current_max_bid| {
+				if *current_max_bid != max_bid {
+					*current_max_bid = max_bid;
+					Self::deposit_event(Event::ValidatorMaxBidUpdated {
+						validator: validator.clone(),
+						max_bid,
+					});
```
MxmUrw, line 1015: I wonder if we could always emit this event, and not only if the bid was actually changed? I know this is actually a general question and there might be many more places in our code that behave the code here, but for the sake of talking about it:

For code that interacts with the state-chain it is important to know whether an extrinsic succeeded and the new value is what we want it to be. That is, from the PoV of the sender it is (usually) irrelevant what the old bid was. If we simply *always* emit the MaxBidUpdated event, then code interacting with the SC can be simpler: it doesn't have to care about whether the current max bid is already at what I want it to be. It simply sends the extrinsic and waits for this exact event.

So from this PoV I think it is nicer to be able to do:
```
submit_tx(new_bid);
wait_for_event();
```

Instead of:
```
let bid = get_current_bid();
if (bid != new_bid) {
    submit_tx(new_bid);
    wait_for_event();
}
```

Anyways I'm pretty sure we do the same thing in many other places so I'm not necessarily saying we should change it here. I'm more curious about your opinion on this long-term-wise.

MxmUrw (commented):
Looks good so far, I'll double-check tomorrow. Maybe @ramizhasan111 also wants to have a look.

**EDIT** I see you already added him as reviewer.

state-chain/pallets/cf-validator/src/lib.rs
msgmaxim: Feels like there ought to be a better (more general) way to detect that a call succeeded rather than require that some event got emitted? But likewise, not a strong opinion.

state-chain/runtime/src/runtime_apis/custom_api.rs
```diff
@@ -106,6 +106,8 @@ decl_runtime_apis!(
 		fn cf_account_flip_balance(account_id: &AccountId32) -> u128;
 		#[changed_in(7)]
 		fn cf_validator_info(account_id: &AccountId32) -> validator_info_before_v7::ValidatorInfo;
+		#[changed_in(19)]
```
ramizhasan111, line 109: The api version should be bumped from 19 to 20 on this trait and on this old method.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -2454,9 +2462,16 @@ where
 								is_online,
 								is_bidding,
 								apy_bp,
+								max_bid,
+								bid,
 								operator,
 								..
-							} = api.cf_validator_info(hash, &account_id)?;
+							} = if api_version < 19 {
```
ramizhasan111, line 2474: we should also bump to 20 here when we bump at the other place

state-chain/runtime/src/runtime_apis/custom_api.rs
msgmaxim: Version 19 has not been released anywhere, so why bump it again?

ramizhasan111: I was basically following the documentation of the trait which says that we should always bump the version when we change a method. My understanding was, therefore, that we bump everytime even if the version before was not actually released.

msgmaxim: I don't think we should follow it literally: maybe not a big difference in this case, but if you always require that every change needs a new version, then we will end up with spurious versions of structs baked into runtime that never had any reason to exist in the first place.

MxmUrw: I agree, I don't think we should bump it we if know that it hasn't been released yet. Still, IMO not really a problem if we do accidentally (which I'm pretty sure we have a bunch of examples of).

We might have a table mapping version numbers to api versions above the declaration. Or we might even align the api versions with the release versions (there isn't really a point in having different versioning schemes?).

state-chain/pallets/cf-validator/src/lib.rs
MxmUrw: Hmm what would that be? In general that would be nice, but I thought that this was the actual purpose of events: to know what happened during execution of an extrinsic.

state-chain/runtime/src/runtime_apis/custom_api.rs
ramizhasan111: yeah, sounds reasonable. We should probably update the comment to say this. And I think we should just bump it once right after a release as part of release process and the other devs wont have to worry about it. The releaser would take care of the bumping.

state-chain/pallets/cf-validator/src/lib.rs
msgmaxim: > Hmm what would that be? In general that would be nice, but I thought that this was the actual purpose of events: to know what happened during execution of an extrinsic.

My reasoning is that extrinsics and events are not a 1-1 relationship. Events could be triggered independently, for example, without a call. Some calls may emit multiple events. Using events to detect failure seems like the wrong mecahnism. It is true that we usually do emit at least one event, but that because calls usually change something that needs external observers need to be notified about, not because it is the only way to tell that the call didn't fail.

Clients sumbitting extinsics already have a way to tell success from failure independendly of events. In fact, the cli command for this (likely the only place this will be submitted from). Bouncer's `submitExtrinsic` also throws if it gets an error. In general, I think checking a call's success/error code is a better practice (more robustness) than require us to be disciplined to always emit an event, and potentially causing some unwanted side effects (UI displaying a change when it didn't happen).

state-chain/runtime/src/runtime_apis/custom_api.rs
msgmaxim: I imagine the argument for always bumping might be that it is the safest default option (when you don't know what's already been released). However, it would be a bit silly, for example, to require two version bumps (thus two versions of structs) if someone splits one large PR into two and merges them back to back. We could be a little flexible IMO.

state-chain/pallets/cf-validator/src/lib.rs
MxmUrw: Sure, there is a way to know if an extrinsic succeeded or failed, but this isn't actually the main thing I meant to talk about. The more interesting thing is to know what effects the execution of the extrinsic had. And the submitter of the extrinsic is the very first observer that would like to get notified about the effects of his extrinsic.

Maybe I try to restate what I meant:
- Event design number 1 (current code) gives an observer the following information if emitted: "value was changed to XYZ, it was something different previously". If it's not emitted the caller of the extrinsic has to either rely on the source code or on rpc calls to be certain that the value is what he wants it to be (and the event simply wasn't emitted because it was already at that value).
- Event design number 2 (my proposal) gives an observer the following information if emitted: "value is now set to XYZ value". It's always emitted so the caller doesn't need to consult any further sources to know that it's now at the value he wants it to be.

Anyways, not blocking or anything.

MxmUrw (approved):
Looks good. I retriggered CI runs, they failed for some probably unrelated reason.

ramizhasan111 (approved):

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: My 2 cents: I don't think there's a clear universal answer to this, but it *would* be nice to formalise and enforce some conventions. 

For example, when the extrinsic is idempotent, then I agree the event is redundant to some extent, but that doesn't imply that it's not valuable. Conditionally emitting the event in such cases can be confusing: the user needs to know the implementation detail that 'the event is only emitted on change', otherwise they might continue submitting waiting for the event. The user also needs to know which extrinsics/events are idempotent and which are conditional. There are some extrinsics that can 'succeed partially', and in that case the outcome is more than a binary success/failure.

msgmaxim: > the user needs to know the implementation detail

I think there is no way around it though (to an extent)? The user already needs to know which events may be emitted and under which conditions. Like you said, sometimes it might be multiple events. Sometimes it could be one but not the other depending on some condition. By extention, it is also reasonable that sometimes there will be 0 events.

In some cases calls *should* and do guarantee that some event is always emitted on success (e.g. when you request a deposit address), but that should be part of the call's contract. It makes sense there because the caller has no way of getting the result (the address) otherwise. However, when applying an idempotant change like `set_validator_max_bid`, the caller already has the `max_bid` (no need to read it back) and they shouldn't rely on the event just to know if the call was a success (the original argument made in this thread) because we have better machanisms for this already. This also means they don't need to know any implementation details in this case.

PR #6734, opened 2026-07-24
feat: Relocate bound broker addresses to cf-asset-balances

msgmaxim:
# Pull Request

Closes: PRO-2998

## Summary

- Moves restricted balances storage to cf-asset-balances (next to the withdrawal whitelists)
- Brokers can use the whitelist (existing functionality), but broker's immutable bindings (if set) take priority: both checks are enforced via the existing `ensure_withdrawal_allowed_to` api call.
- I left extrinsics, events in cf-swapping pallet (though perhaps backwards compatibility is not important here and we could move them to cf-asset-balances?)

## *Checklist*

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


j4m1ef0rd (approved):

state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -2789,7 +2795,10 @@ mod bound_broker_withdrawal {
 				OriginTrait::signed(BROKER),
 				bound_addr
 			));
-
+			MockWithdrawalAddressRestriction::restrict_to(
+				&BROKER,
+				vec![AccountOrAddress::ExternalAddress(ForeignChainAddress::Eth(bound_addr))],
+			);
```
MxmUrw: It might be worth merging the two traits `BrokerWithdrawalAddressRegistry` and `WithdrawalAddressRestriction`. Because `bind_broker_withdrawal_address()` in the first trait is used to write a restriction but `ensure_withdrawal_allowed_to()` in the second trait is used to check for restrictions, the two traits are not really seperable.

Then we would have a single mock for the unified trait and in this test it would be enough to only call `Swapping::bind_broker_fee_withdrawal_address()` instead of both restriction functions.

MxmUrw (commented):
Looks good, but I would suggest merging the two traits. WDYT?

state-chain/pallets/cf-swapping/src/tests.rs
msgmaxim: > Then we would have a single mock for the unified trait and in this test

Actualy it looks like the `enforce_withdrawal_address_restriction_for_usdc` test is now redundant:
  - we already check the more general "withdrawals check address restrictions" in `broker_withdraw_is_gated_by_the_whitelist` (just renamed it to `broker_withdraw_is_gated_by_withdrawal_restrictions`)
  - `can_bind_withdrawal_address` already checks that restrictions are created by the `bind_broker_fee_withdrawal_address` call
  
So I'm removing this test instead.

> It might be worth merging the two traits

You are right, I think these could be merged. Note that merging them means exposing LP pallet to functionality that it doesn't need (broker restirctions). It is true that one trait does not make sense without the other, but it is not true in reverse. However, I still think it probably makes sense to merge them (I will experiment).

Update:
- Combined the traits and mocks.
- Removed `broker_withdrawal_address` method from the trait (it was only used in `bind_broker_fee_withdrawal_address` extrinsic to check whether BrokerAlreadyBound should be returned, but I made `bind_broker_withdrawal_address` return this error directly, which is a better design.
- Removed `enforce_withdrawal_address_restriction_for_usdc` as mentioned above.
- Removed `no_restriction_for_other_chains` test <- cf-asset-balances already tests this (and is the right place for this test now).

@MxmUrw let me know if you are still happy to merge this with these changes.

MxmUrw: Yes, I like the new design!

MxmUrw (approved):

PR #6738, opened 2026-07-28
feat: enforce lp account role for delegation

ramizhasan111:
# Pull Request

Closes: PRO-3025

## Summary

- makes it a requirement for delegators to have an lp account.
- registers the the delegators as lp if they have no account role (delegating for the first time).
- when the undelegate everything, we try to deregister that LP role which would only succeed if there is no other LP related state (like open limit orders etc).

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-account-roles/src/lib.rs
+++ b/state-chain/pallets/cf-account-roles/src/lib.rs
@@ -283,14 +283,13 @@ impl<T: Config> AccountRoleRegistry<T> for Pallet<T> {
 		account_role: AccountRole,
 	) -> DispatchResult {
 		T::DeregistrationHooks::check(account_id).map_err(Into::into)?;
+		T::DeregistrationHooks::on_deregistered(account_id);
 		AccountRoles::<T>::try_mutate(account_id, |role| {
 			role.replace(AccountRole::Unregistered)
 				.filter(|r| *r == account_role)
 				.ok_or(Error::<T>::UnknownAccount)
 		})?;
 		<frame_system::Pallet<T>>::dec_consumers(account_id);
-
-		T::DeregistrationHooks::on_deregistered(account_id);
 
 		Self::deposit_event(Event::AccountRoleDeregistered {
 			account_id: account_id.clone(),
--- a/state-chain/pallets/cf-account-roles/src/lib.rs
+++ b/state-chain/pallets/cf-account-roles/src/lib.rs
@@ -30,7 +30,9 @@ use sp_std::boxed::Box;
 use cf_traits::Chainflip;
 
 use cf_primitives::AccountRole;
-use cf_traits::{AccountRoleRegistry, DeregistrationCheck, SpawnAccount, VanityName};
+use cf_traits::{
+	AccountRoleRegistry, DeregistrationCheck, RefundAddressRegistry, SpawnAccount, VanityName,
+};
 use frame_support::{
 	dispatch::GetDispatchInfo,
 	error::BadOrigin,
@@ -80,6 +82,10 @@ pub mod pallet {
 			AccountId = <Self as frame_system::Config>::AccountId,
 		>;
 		type WeightInfo: WeightInfo;
+		/// Used to clean up a Liquidity Provider's refund addresses on deregistration.
+		type RefundAddressRegistry: RefundAddressRegistry<
+			AccountId = <Self as frame_system::Config>::AccountId,
+		>;
 	}
 
 	#[pallet::pallet]
@@ -296,6 +302,17 @@ impl<T: Config> AccountRoleRegistry<T> for Pallet<T> {
 		});
 
 		Ok(())
+	}
+
+	/// Deregister a Liquidity Provider, clearing their refund addresses first.
+	///
+	/// Overrides the trait default so that this cleanup happens for every caller (e.g.
+	/// `cf-lp`'s `deregister_lp_account` extrinsic, or `cf-validator`'s auto-deregistration on
+	/// full undelegation) rather than being duplicated at each call site.
+	#[frame_support::transactional]
+	fn deregister_as_liquidity_provider(account_id: &T::AccountId) -> DispatchResult {
+		T::RefundAddressRegistry::clear_refund_addresses(account_id);
+		Self::deregister_account_role(account_id, AccountRole::LiquidityProvider)
 	}
 
 	fn set_vanity_name(
--- a/state-chain/pallets/cf-account-roles/src/lib.rs
+++ b/state-chain/pallets/cf-account-roles/src/lib.rs
@@ -30,9 +30,7 @@ use sp_std::boxed::Box;
 use cf_traits::Chainflip;
 
 use cf_primitives::AccountRole;
-use cf_traits::{
-	AccountRoleRegistry, DeregistrationCheck, RefundAddressRegistry, SpawnAccount, VanityName,
-};
+use cf_traits::{AccountRoleRegistry, DeregistrationHooks, SpawnAccount, VanityName};
 use frame_support::{
 	dispatch::GetDispatchInfo,
 	error::BadOrigin,
@@ -54,7 +52,7 @@ pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(STORAGE_VERSION_
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
-	use cf_traits::DeregistrationCheck;
+	use cf_traits::DeregistrationHooks;
 	use frame_support::{
 		dispatch::{DispatchResultWithPostInfo, PostDispatchInfo},
 		pallet_prelude::*,
@@ -65,7 +63,7 @@ pub mod pallet {
 	#[pallet::config]
 	pub trait Config: frame_system::Config + cf_traits::Chainflip {
 		type EnsureGovernance: EnsureOrigin<Self::RuntimeOrigin>;
-		type DeregistrationCheck: DeregistrationCheck<
+		type DeregistrationHooks: DeregistrationHooks<
 			AccountId = <Self as frame_system::Config>::AccountId,
 		>;
 		type RuntimeCall: Parameter
@@ -82,10 +80,6 @@ pub mod pallet {
 			AccountId = <Self as frame_system::Config>::AccountId,
 		>;
 		type WeightInfo: WeightInfo;
-		/// Used to clean up a Liquidity Provider's refund addresses on deregistration.
-		type RefundAddressRegistry: RefundAddressRegistry<
-			AccountId = <Self as frame_system::Config>::AccountId,
-		>;
 	}
 
 	#[pallet::pallet]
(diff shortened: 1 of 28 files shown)


dandanlen:
@ramizhasan111 please update the PR description to at least reference the Linear issue.

marcellorigotti:
Mostly looks good to me, just a couple of things:  

- I guess the migration will be done in a separate PR?
- benchmarks are not updated
- it feels weird to deregister the LP when they undelegate, I can see why this is done there, but maybe we can improve in general the flow. Instead of requiring an account to be deregister when redeeming everything we could implicitly deregister it when they try to redeem everything, and perform all the necessary checks there, such that we never end up with account holding FLIP but having no role. (this can eventually be done in a separate PR) I don't know what others think about this, and if there is a particular reason why it is the way it is today.  @dandanlen @MxmUrw

ramizhasan111:
> @ramizhasan111 please update the PR description to at least reference the Linear issue.

The issue for this didn't exist. I created the issue and linked it now.

dandanlen:
>     * it feels weird to deregister the LP when they undelegate, I can see why this is done there, but maybe we can improve in general the flow. Instead of requiring an account to be deregister when redeeming everything we could implicitly deregister it when they try to redeem everything, and perform all the necessary checks there, such that we never end up with account holding FLIP but having no role. (this can eventually be done in a separate PR) I don't know what others think about this, and if there is a particular reason why it is the way it is today.  @dandanlen @MxmUrw

I agree. I think the underlying issue is that there are two competing user flows and we're using the same role for both: 

1. 'Retail' delegation flow: users don't know anything about account roles, they send funds to an address and everything else is magic. Requiring them to then 'deregister' is confusing, and making this automatic is sensible. 
2. Actual LP users who have been through the LP onboarding process and may need to close out open orders etc. before 

Now, on the one hand we're combining these into the same 'role', but there are nonetheless different user experiences that need to be considered. 

I'm not quite sure how to reconcile this, it needs some thought. I agree moving towards an atomic deregister-and-close-account rather than the current two-step process seems like the ideal solution. I'll review this PR and hopefully it will give me some more context/ideas.

state-chain/pallets/cf-account-roles/src/lib.rs
```diff
@@ -298,6 +304,17 @@ impl<T: Config> AccountRoleRegistry<T> for Pallet<T> {
...
+	/// Deregister a Liquidity Provider, clearing their refund addresses first.
+	///
+	/// Overrides the trait default so that this cleanup happens for every caller (e.g.
+	/// `cf-lp`'s `deregister_lp_account` extrinsic, or `cf-validator`'s auto-deregistration on
+	/// full undelegation) rather than being duplicated at each call site.
+	#[frame_support::transactional]
+	fn deregister_as_liquidity_provider(account_id: &T::AccountId) -> DispatchResult {
+		T::RefundAddressRegistry::clear_refund_addresses(account_id);
+		Self::deregister_account_role(account_id, AccountRole::LiquidityProvider)
+	}
```
dandanlen: I think a better way to implement this is with a new method on the `DeregistrationCheck` hook trait. (ie. we can rename it `DeregistrationHooks` and add a `fn on_deregistered(account_id: &Self::AccountId, role: AccountRole)` method.

The current implementation means that the account roles pallet needs to know about the LP invariants, which is the wrong way round.

dandanlen: But also: we already delete the refund addresses in `OnKilledAccount` in the asset-balances pallet. So this actually duplicates that. Is it necessary at all? 

*edit: I think it's worth keeping , but updating to use the same `clear_refund_addresses` method. We can evaluate separately if we really need both cleanups. I also opened PRO-3031 to review the whole role lifecycle question*

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1363,6 +1371,16 @@ pub mod pallet {
...
 
+			// An account with an entry in `DelegationChoice` can only have got there via
+			// `delegate`, which requires (and enforces) the Liquidity Provider role.
+			ensure!(
+				T::AccountRoleRegistry::has_account_role(
+					&delegator,
+					AccountRole::LiquidityProvider
+				),
+				Error::<T>::NotLiquidityProvider
+			);
```
dandanlen: Why not simply use `ensure_liquidity_provider` above instead of doing `ensure_signed` and then checking separately?

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1393,6 +1411,13 @@ pub mod pallet {
 					operator: current_operator,
 					max_bid: current_max_bid,
 				});
+
+				// Mirrors the auto-registration in `delegate`. Best-effort: accounts that also
+				// hold other LP state (open orders, balances, etc.) fail the deregistration
+				// check and simply remain registered as a Liquidity Provider. The
+				// `#[transactional]` override rolls back any partial cleanup on failure, so
+				// discarding the error here is safe.
```
dandanlen: There are a lot of slightly redundant (presumably AI-generated) comments. Some just explain what the code does. I think we can reduce the verbosity a bit and keep the info to important details like safety assumptions.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1270,14 +1274,18 @@ pub mod pallet {
...
-			);
+			// Delegation is only available to Liquidity Providers. Accounts that haven't yet
+			// registered a role (e.g. freshly-funded accounts delegating via the SC Utils EVM
+			// contract) are implicitly registered as Liquidity Providers on first delegation.
+			match T::AccountRoleRegistry::account_role(&delegator) {
+				AccountRole::Unregistered =>
+					T::AccountRoleRegistry::register_as_liquidity_provider(&delegator)?,
+				AccountRole::LiquidityProvider => {},
+				other => {
+					log::error!("DEBUG role was: {:?}", other);
```
dandanlen: Did you mean to remove this?

dandanlen:
Regarding @marcellorigotti 's comment, I opened PRO-3031 to review the overall lifecycle of accounts/account roles. I agree we can make it more intuitive / user friendly.

dandanlen:
Btw. I also agree there should be a migration in this PR to update the roles for existing delegators.

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: I wanted to have this message here to explain that it would be a best effort deregistration and it would remain registered if there is other lp realted state. simplified the message a little.

state-chain/pallets/cf-validator/src/migrations/assign_lp_role_to_delegators.rs
```diff
@@ -0,0 +1,147 @@
...
+							delegator
+						);
+					}
+				},
+				AccountRole::LiquidityProvider => {},
+				// it was technically possible to delegate with a broker account role and an edge
+				// case where you could have operator or validator role after delegating. Its not
+				// immediately clear if we should convert it to LP here. We should log it at least,
+				// to surface these cases
+				other => cf_runtime_utilities::log_or_panic!(
```
dandanlen: nit: Can we please `log::warn!` here instead of `log_or_panic`ing?

state-chain/pallets/cf-account-roles/src/lib.rs
ramizhasan111: added the `on_registered` method and renamed to `DeregistrationHooks`.

I am assuming you meant using `clear_refund_addresses` in the `OnKilledAccount` method since we already use `clear_refund_addresses` in the deregistration. Also, just to clarify: this PR didnt introduce the `clear_refund_addresses` call on deregistration. It was already called before on lp deregistration. I just moved it to the account-role pallet to have common deregistration code called from lp and validator pallets.

dandanlen: Thanks. Yes, I meant using clear_refund_adresses in OnKilledAccount too.

I realise it wasn't introduced here, my point was that we should follow the pre-existing hook pattern for this instead of coupling the account-roles pallet to role-specific concerns.

state-chain/runtime/src/chainflip/deregistration_hooks.rs
```diff
@@ -62,4 +63,10 @@ impl DeregistrationCheck for RuntimeDeregistrationCheck {
 			},
 		}
 	}
+
+	fn on_deregistered(account_id: &Self::AccountId) {
+		if AccountRoles::account_role(account_id) == AccountRole::LiquidityProvider {
```
dandanlen: This is called *after* the account role is set to deregistered, so reading the role back again won't work. We need to pass the role as a parameter. We should also write a test that catches this.

state-chain/runtime/src/chainflip/deregistration_check.rs
```diff
@@ -16,13 +16,13 @@
...
 use cf_primitives::AccountRole;
-use cf_traits::{AccountRoleRegistry, DeregistrationCheck};
+use cf_traits::{AccountRoleRegistry, DeregistrationHooks, RefundAddressRegistry};
 use frame_support::sp_runtime::DispatchError;
 use pallet_cf_flip::Bonder;
 
 pub struct RuntimeDeregistrationCheck;
 
-impl DeregistrationCheck for RuntimeDeregistrationCheck {
+impl DeregistrationHooks for RuntimeDeregistrationCheck {
```
dandanlen: We should also rename `RuntimeDeregistrationCheck` to `RuntimeDeregistrationHooks` (and the module name too I suppose).

state-chain/pallets/cf-validator/src/migrations/assign_lp_role_to_delegators.rs
```diff
@@ -0,0 +1,147 @@
...
+				AccountRole::LiquidityProvider => {},
+				// it was technically possible to delegate with a broker account role and an edge
+				// case where you could have operator or validator role after delegating. Its not
+				// immediately clear if we should convert it to LP here. We should log it at least,
+				// to surface these cases
+				other => log::warn!(
+					"Delegator {:?} has incompatible role {:?}; expected Unregistered or LiquidityProvider",
+					delegator,
+					other
+				),
```
dandanlen, line 69: What we *could* do is keep `undelegate` as `ensure_signed` instead of gating it with `ensure_lp`. This way any remaining non-LP accounts an still undelegate.

state-chain/runtime/src/chainflip/deregistration_hooks.rs
ramizhasan111: I fixed this by calling the on_deregistered before removing the role. Also added a test

state-chain/pallets/cf-validator/src/migrations/assign_lp_role_to_delegators.rs
ramizhasan111: yeah, that sounds reasonable. will do.

state-chain/runtime/src/chainflip/deregistration_hooks.rs
dandanlen: The suggested fix would have been better: it makes it harder to get wrong. Also, the name `on_deregistered` (past tense) suggests that it's being called after. So now it's more like `before_deregistering` or something. So please either move it back and make it an argument (better, makes it harder to get wrong) or rename the hook accordingly.

ramizhasan111: done

dandanlen (approved):

PR #6740, opened 2026-07-29
feat: assethub elections

MxmUrw:
# Pull Request

Closes: PRO-3013

## Summary

This PR migrates Assethub to election-based witnessing. The design looks as follows:
 - Assethub has 5 electoral systems: BHW, Deposit Channel BW, Egress BW, Fee Tracking and Liveness. It does not have vault swap support, so there's no BW for that.
 - Since Assethub blocktimes are 2s on mainnet, witnessing is batched (`WITNESS_PERIOD = 3`). This means the functionally closest chain is Arbitrum. But, there is no analogue of evm's `get_logs_range` rpc call, so all 3 blocks in a range have to be queried individually by the engine.
 - As in all election-based witnessing implementations, WS subscriptions are not used by the engine.
 - Since Assethub supports deterministic finality there is no need to deal with reorgs. All witnessing processes only use finalized blocks. The safety_margin is 0, the safety_buffer contains a single block.
 - FeeTracking elections run every 10 SC blocks.

Storage changes:
 - Assethub DepositDetails were just `ExtrinsicIndex`. Now they are `(BlockHeight, ExtrinsicIndex)`. The reason is that the deposit details are part of the `DepositWitness` struct and since deposits are now witnessed in batches of 3 blocks it could have been the case that 2 deposits in sequential blocks would get the same deposit witness and the second deposit would have been deduplicated.

Notes:
 - This PR caused the generated WASM to go above the MAX allocation size of subxt's internally used wasm executor in the `build.rs` script of `cf-node-client`. This value is hardcoded, so can't be adjusted easily. A workaround is possible by generating the metadata ourselves and passing it to subxt (which then doesn't have to invoke any wasm executor at all).

### Question for the reviewers
 - [x] Previously, there was a function `proxy_added_witnessing` which added all extrinsics indices of extrinsics that resulted in `ProxyAdded` events. But also, later in the pipeline, these indices were merged with the indices of *all* successful extrinsics. So it looks like the first step was redundant (unless a ProxyAdded event can happen and be effective in an extrinsic which isn't marked as successful?). We now skip the `proxy_added_witnessing` and only extract all successful extrinsic indices before processing. Is this enough?
 - [x] Would be nice if you can double-checked the migrations. It seems to be a bit unconventional because I'm migrating only a single pallet instance, but I have to change the storage version of all the instances.
 - [x] Do we have to do something with the old witnesser pallet that's now unused? For now I kept it as is, as I think it will deal itself with cleaning up its own storage once the current epoch runs out. And upgraded engines won't vote through the witnesser pallet anymore so it would be effectively inactive. Is that enough?

 
### RPCS

 - `cf_ingress_egress_events` and `cf_ingress_events` now support `Assethub` as chain parameter and return the deposits currently known by the witnessing system. Since `safety_margin` is 0, these have been ingressed already and are available in the election storage only for a single block.

### Extrinsics & Events

 - `Assethub::DepositDetails` changes from `ExtrinsicIndex` to `(BlockHeight, ExtrinsicIndex)`. This affects all events that contain this type. Most of them are never emitted (boost, transaction rejection). The only actively used events is `DepositFinalised`. See [this commit](https://github.com/chainflip-io/chainflip-backend/pull/6740/changes/897cea5a4c7788718f126ee008821b3bdaafdf44#diff-a5eaa6d9d1d9e68e158c4b596a35fa5800bf0ef5473e1b32d95a84312d657e0e) for all event schema changes.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


dandanlen:
> Previously, there was a function proxy_added_witnessing which added all extrinsics indices of extrinsics that resulted in ProxyAdded events. But also, later in the pipeline, these indices were merged with the indices of all successful extrinsics. So it looks like the first step was redundant (unless a ProxyAdded event can happen and be effective in an extrinsic which isn't marked as successful?). We now skip the proxy_added_witnessing and only extract all successful extrinsic indices before processing. Is this enough?

This sounds accidental. I vaguely remember we previously detected events only and ran detection for each event type sequentially, ignoring the extrinsics. Later, we added scanning of extrinsics, probably because we determined that events alone didn't carry enough information (or maybe this was for broadcast success witnessing, I can't remember). So the 'new' implementation sounds correct. 

> Would be nice if you can double-checked the migrations. It seems to be a bit unconventional because I'm migrating only a single pallet instance, but I have to change the storage version of all the instances.

👌 will do

> Do we have to do something with the old witnesser pallet that's now unused? For now I kept it as is, as I think it will deal itself with cleaning up its own storage once the current epoch runs out. And upgraded engines won't vote through the witnesser pallet anymore so it would be effectively inactive. Is that enough?

Yes I think that's enough for now, and we *do* want the clean-up. We can decommission it fully in a follow-up release.

engine/src/witness/hub/hub_deposits.rs
```diff
@@ -185,13 +85,16 @@ fn event_as_transfer(
...
+		block_hash,
+		parent_block_hash,
+		block_height,
+		events,
+	}: &AssethubBlockHeader,
 	hub_client: &Client,
-	monitored_addresses: Vec<PolkadotAccountId>,
-	events: &Vec<(Phase, EventWrapper)>,
-) -> Vec<DepositWitness<Assethub>> {
+	monitored_addresses: &[PolkadotAccountId],
```
dandanlen, line 96: Not really related to this PR but we could return early here in the (common) case where `monitored_addresses` is empty?

engine/src/witness/hub.rs
```diff
@@ -166,214 +132,77 @@ fn extract_state_chain_signature(raw_extrinsic: &[u8]) -> Option<PolkadotSignatu
...
-		+ Send
-		+ Sync
-		+ Clone
-		+ 'static,
-	ProcessingFut: Future<Output = ()> + Send + 'static,
-{
-	let ((events, mut extrinsic_indices), monitored_egress_data) = header.data;
+pub async fn process_egresses_in_block(
+	hub_client: &impl DotRetryRpcApiWithResult,
+	pending_tx_signatures: &[PolkadotSignature],
```
dandanlen, line 157: I wonder if the same applies here: maybe we can return early and save a few rpc round trips if `pending_tx_signatures` is empty, which is the common case.

state-chain/pallets/cf-ingress-egress/src/lib.rs
```diff
@@ -147,25 +147,25 @@ impl<T: Config<I>, I: 'static> BoostStatusLookup<T, I> {
...
-	block_height: TargetChainBlockNumber<T, I>,
-	amount: TargetChainAmount<T, I>,
-	asset: TargetChainAsset<T, I>,
-	deposit_details: <T::TargetChain as Chain>::DepositDetails,
-	deposit_address: Option<TargetChainAccount<T, I>>,
-	action: ChannelActionForDeposit<T::AccountId, TargetChainAccount<T, I>>,
-	boost_fee: u16,
-	channel_id: Option<u64>,
-	origin: DepositOrigin<T, I>,
+pub struct PendingPrewitnessedDeposit<T: Config<I>, I: 'static> {
```
dandanlen, line 159: I can't immediately see why this was made `pub`? For the migrations?

state-chain/runtime/src/chainflip/witnessing/assethub_elections.rs
```diff
@@ -0,0 +1,453 @@
...
+	Hooks<
+		AssethubBlockHeightWitnesserES,
+		AssethubDepositChannelWitnessingES,
+		AssethubEgressWitnessingES,
+		AssethubFeeTracking,
+		AssethubLiveness,
+	> for AssethubElectionHooks
+{
+	fn on_finalize(
+		(block_height_witnesser_identifiers, deposit_channel_witnessing_identifiers, egress_identifiers, fee_identifiers, liveness_identifiers): (
```
dandanlen, line 283: `fee_identifiers` and `liveness_identifiers` are mixed up here. The types line up but not sure if there's any (potential) impact.

state-chain/runtime/src/migrations/assethub_deposit_details.rs
```diff
@@ -0,0 +1,539 @@
...
+use sp_runtime::DispatchError;
+
+const OLD_STORAGE_VERSION: u16 = 30;
+const NEW_STORAGE_VERSION: u16 = 31;
+
+pub type Migration = (
+	VersionedMigration<
+		OLD_STORAGE_VERSION,
+		NEW_STORAGE_VERSION,
+		NoopRuntimeUpgrade,
```
dandanlen, line 49: Yeah this is the correct way to do it. (There's an `instanced_migrations` macro for this too)

state-chain/runtime/src/migrations/assethub_deposit_details.rs
```diff
@@ -0,0 +1,539 @@
...
+		tx_id: old.tx_id,
+		broker_fee: old.broker_fee,
+		affiliate_fees: old.affiliate_fees,
+		refund_params: old.refund_params,
+		dca_params: old.dca_params,
+		boost_fee: old.boost_fee,
+	}
+}
+
+fn clear_deferred_witness_calls() -> bool {
```
dandanlen, line 278: Do you think we should activate witnesser safe mode during the migration? (This is the only way the deferred calls storage can get populated)

state-chain/runtime/src/migrations/safe_mode.rs
```diff
@@ -14,16 +14,24 @@
...
+use crate::{safe_mode::RuntimeSafeMode, Runtime};
 
 pub struct SafeModeMigration;
 
 use crate::runtime_apis::custom_api::types::before_version_19::RuntimeSafeMode as OldRuntimeSafeMode;
 
 impl OnRuntimeUpgrade for SafeModeMigration {
 	fn on_runtime_upgrade() -> Weight {
+		let storage_key = pallet_cf_environment::RuntimeSafeMode::<Runtime>::hashed_key();
+		if unhashed::get_raw(&storage_key)
```
dandanlen, line 29: Is there a reason for adding this early return?

state-chain/runtime/Cargo.toml
```diff
@@ -273,6 +273,7 @@ try-runtime = [
 	"pallet-authorship/try-runtime",
 	"sp-runtime/try-runtime",
 ]
+skip-wasm-build = []
```
dandanlen, line 276: What is this for, just as a convenience?

engine/src/dot/http_rpc.rs
```diff
@@ -284,6 +300,8 @@ impl DotRpcApi for DotRpcClient {
 		block_hash: PolkadotHash,
 		parent_hash: PolkadotHash,
 	) -> Result<Option<Events<PolkadotConfig>>> {
+		let _events_guard = self.events_lock.lock().await;
```
dandanlen, line 309: Maybe we can hold this guard only for as long as it takes to check the metadata? (ie. release it after cloning the client, before fetching the events).

engine/src/dot/http_rpc.rs
```diff
@@ -147,6 +148,9 @@ pub struct DotRpcClientBuilder {
 pub struct DotRpcClient {
 	online_client: OnlineClient<PolkadotConfig>,
 	rpc_methods: LegacyRpcMethods<PolkadotConfig>,
+	// OnlineClient clones share mutable metadata, so we need this lock to ensure
+	// that a metadata update and event reads don't interfere with each other between tasks.
+	events_lock: Arc<tokio::sync::Mutex<()>>,
```
dandanlen, line 159: Is this required because of the election tasks running in parallel vs. the legacy witnesser being sequenced? Or is it something we should have had all along?

dandanlen (approved):
Just some minor comments/questions. Good work 🚀

state-chain/runtime/src/migrations/safe_mode.rs
MxmUrw: Yes, this is the fix for the non idempotent try-runtime that I mentioned in discord. The `::translate()` function uses `decode()` under the hood which succeeds even if not all of the input bytes were consumed. For some reason the post-migration SafeMode definition decodes correctly into the pre-migration SafeMode definition (with left-over bytes). The early return makes the upgrade idempotent by not migrating if the `decode_all` fails.

state-chain/runtime/Cargo.toml
MxmUrw: This was required to fix the wasm file corruption problem (me complaining about namespaces runners being broken), which was caused by the `build.rs` script of `cf-node-client` now depending on a `state-chain-runtime` build. It actually only needs the natively compiled runtime, and shouldn't trigger another runtime wasm build. This feature flag is used to tell the runtime build script to not build wasm.

engine/src/dot/http_rpc.rs
MxmUrw: It's actually required to hold this until after fetching the events. The reason is that "cloning the client" doesn't do anything, as it's all `Arc<>` underneath. If there are concurrent calls to `events()` for blocks with different metadata, we have to ensure that setting the metadata and fetching events is done atomically.

engine/src/dot/http_rpc.rs
MxmUrw: It's now much more likely to happen across runtime upgrade boundaries. Previously it could only happen during phases of "catching up" which would also have allowed for concurrent event queries.

state-chain/runtime/src/migrations/assethub_deposit_details.rs
MxmUrw: No, I don't think so. We should disable deposit channels, then wait a few hours, then we should be good. This part of the migration was more of an additional precaution.

state-chain/runtime/src/chainflip/witnessing/assethub_elections.rs
MxmUrw: Yeah they're actually mixed up also in Eth and Arb elections (where I copied this from). The order of the arguments is correct, only their stated types are switched. I thought we can have a follow-up PR which corrects this for all affected ESs, but actually I can do it just as well here :)

state-chain/pallets/cf-ingress-egress/src/lib.rs
MxmUrw: Yeah it's for the migration.

engine/src/witness/hub/hub_deposits.rs
MxmUrw: Not sure this is very necessary: if `monitored_addresses` is empty, the code doesn't do any rpc querying anyways. It still *does* iterate through events but I would expect this to be negligible.

I think we can keep it as is - wdyt?

engine/src/witness/hub.rs
MxmUrw: This one yes, I've added the early return.

PR #6741, opened 2026-07-30
fix: resubmit pending Assethub broadcasts and re-enable Assethub in AllSwaps

j4m1ef0rd:
Split off from [this PR](https://github.com/chainflip-io/chainflip-backend/pull/6509).
The engine was reporting a "tx already in pool" error as a failure. This caused the bouncer to fail when the assethub node is fully loaded with swaps. Instead it now handles the nonce gap properly and is able to recover.
This allows us to add the assethub bouncer swaps back into the all swaps bouncer test


marcellorigotti:
Looks fine to me, although I am not understanding what exact problem is this solving.

> Instead it now handles the reties that are coming from the bouncer properly and is able to recover.

What do you mean by retries coming from the bouncer? From my understanding a validator is chosen to broadcast a tx only once so there shouldn't be a scenario where it hits `TxAlreadyImported`.

So the scenario I can see is, validator is selected, it broadcast the tx and it fails because the node is overloaded, in that case we end up with a nonce gap that never gets filled since we don't retry. With this fix we retry all the previous nonce tx so it increases the chance to get included (and maybe this is the case where it hits `TxAlreadyImported`?).

Regardless I think re-submitting all the previously pending tx is a good practice if we need nonce sequentiality!

engine/src/dot/http_rpc.rs
```diff
@@ -332,19 +338,21 @@ impl DotRpcApi for DotRpcClient {
...
-		Ok(success.extrinsic_hash())
+		);
+		match tx.submit().await {
+			// Check if the pool already knows this transaction and treat it as a success.
+			Err(subxt::Error::Rpc(RpcError::ClientError(subxt_rpcs::Error::User(user_error))))
+				if [POOL_TEMPORARILY_BANNED, POOL_ALREADY_IMPORTED, POOL_TOO_LOW_PRIORITY]
+					.contains(&user_error.code) =>
+				Ok(tx.hash()),
+			result => Ok(result?),
+		}
```
MxmUrw, line 355: Yes I think this is good. We shouldn't wait for finalization here. For example on mainnet Assethub finalization time is around 30-40s...

MxmUrw (approved):
Quoting from the other PR:

> The behaviour and storage read is copied from the above code for BtcTxBroadcastRequest.

True, I wasn't aware that we're doing this for bitcoin! Looks good then, also the change to the submission makes sense.

j4m1ef0rd:
> What do you mean by retries coming from the bouncer? 

Sorry I was confusing the two.

> So the scenario I can see is,

Yes you are correct.
Assethub broadcasts go out through a proxy account whose nonces are allocated in broadcast-id order. A transaction with nonce N can't execute until every nonce < N is in the pool.
The old submit_raw_encoded_extrinsic did submit_and_watch().wait_for_finalized_success() — it required the tx to finalize.
When the node was overloaded with the swap burst, a lower-nonce broadcast wasn't accepted into the pool in time. The higher-nonce tx then got parked in the "future" queue and never finalized → that broadcast "failed" (timed out).
The engine's retrier resubmitted the same tx — and now the pool rejected the resubmission with temporarily banned (1012) / already imported (1013), because it had already seen it. The engine surfaced that as a failure, which is what made the bouncer fail.

engine/src/dot/http_rpc.rs
```diff
@@ -332,19 +338,21 @@ impl DotRpcApi for DotRpcClient {
...
-		.await?
-		.wait_for_finalized_success()
-		.await?;
-
-		Ok(success.extrinsic_hash())
+		);
+		match tx.submit().await {
+			// Check if the pool already knows this transaction and treat it as a success.
+			Err(subxt::Error::Rpc(RpcError::ClientError(subxt_rpcs::Error::User(user_error))))
+				if [POOL_TEMPORARILY_BANNED, POOL_ALREADY_IMPORTED, POOL_TOO_LOW_PRIORITY]
```
msgmaxim, line 351: Are we certain that POOL_TEMPORARILY_BANNED should be treated as success?

j4m1ef0rd: As discussed, it seems like yes. Its means that the tx is either in the pool or recently in a block.

MxmUrw: Yeah I also looked a bit into it:

It looks like this error can be returned even if the tx won't be mined by this node - e.g. if the mempool (or future queue) is overfull, all txs that don't fit into it are banned (very questionable decision if you ask me). So this message does not actually mean success.

But this error is also returned if a tx was included in a block recently.

So with our architecture, and in particular for localnet, I agree that it's better to treat this as success, instead of erroneously erroring out and failing bouncer. The mainnet case where the tx wasn't actually included will still be handled by the "we can't witness the egress -> we reschedule the broadcast" loop.

marcellorigotti (approved):

MxmUrw:
The other "reduce number of swaps" PR got merged! What's the status of this one, we also want to merge it, right?

msgmaxim (approved):

j4m1ef0rd:
The Deposit monitor needs a fix before this PR can be merged. Got claude to open a [PR](https://github.com/chainflip-io/chainflip-deposit-monitor/pull/186). Then we will need to update the DM image on CI again. @MxmUrw

PR #6744, opened 2026-07-31
chore: upgrade polkadot assethub localnet to v2.3.2

MxmUrw:
# Pull Request

Closes: PRO-3014

## Summary

This upgrades the polkadot and assethub runtime and node used in localnet to the latest version 2.3.2. To make such updates easier in the future it includes a new script `generate-polkadot-assethub-chainspecs.sh` which automates the steps required to assemble the final chainspecs.

Assethub blocktimes are now 3s. Changing them to 2s (as on mainnet) would require running 3 polkadot validators in localnet due to how the work is distributed between nodes. Currently we're already required to have 2 validators and there's really no good reason to run yet another node that uses ~1GB RAM. 

It also:
- updates the chainflip localnet chainspec to include the new assethub (and polkadot) genesis hashes
- regenerates chainflip backspin chainspec (even though it is probably unused at the moment)
- deletes old assethub and polkadot chainspecs
- removes the old workflow to build a custom chainflip polkadot node (it was based on polkadot-v1.0.0 which is quite far in the past and isn't used anymore)

CC @dandanlen, I've marked this as pick-to-release. Then the upgrade test for the actual assethub-elections PR will go through more smoothly.

## API Changes

Assethub genesis hash changed on localnet and backspin. Block time is now 3s.

---

* The PR is complete:
  * [x] Reviewers are assigned.


dandanlen:
Note we will need to regenerate the backspin chainspecs too. 

Are you planning to keep the `.txt` wasm files? Are they used for anything? Also maybe we can delete the old chainspecs if they are now unused. We can always recover them from an old commit if needed.

MxmUrw:
> Note we will need to regenerate the backspin chainspecs too.

Yeah true, on it.

**EDIT**: actually, regarding this one: are you sure that the backspin chainspec is even used? I regenerated it and the diff contains lots of things (e.g. arbitrum elections, solana usdt support), so if backspin worked without these I would think it doesn't consume the chainspec at all?

> Are you planning to keep the `.txt` wasm files? Are they used for anything?

No they aren't. Their value is injected into the final polkadot chainspec. I agree, makes sense to make them temporary.

> Also maybe we can delete the old chainspecs if they are now unused. We can always recover them from an old commit if needed.

Agree. I'll delete the old ones.

dandanlen:
> CC @dandanlen, I've marked this as pick-to-release. Then the upgrade test for the actual assethub-elections PR will go through more smoothly.

Good idea. We will need to remember re-tag and then update the OLD_VERSION on this branch too.

dandanlen (approved):
Hmmm maybe backspin already generates a fresh chainspec on every deploy (this would make sense actually, it needs to pick up recent changes to the chainspec). 

Thanks for the script! Will make updating much easier in the future. 

LGTM.

PR #6745, opened 2026-07-31
chore: dont return apy in rpc after flip 2.1 activation

ramizhasan111:
# Pull Request

Closes: PRO-3017

## Summary

dont return apy in account info rpc after flip2.1 activation.

This PR also fixes an edge case in redemption where liquid balance is less than redemption fee. We now dont allow these kind of redemptions and in general 0 flip redemptions are not executed.

### RPCS

`cf_validator_info` rpc changes but not breaking

### Extrinsics & Events

The `RedemptionAmountZero` has now been removed since we now error on 0 flip redemption attempts.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


MxmUrw (approved):

GabrielBuragev (approved):

ramizhasan111:
added an unrelated fix for an edge case in redemption. The edge case was breaking a test in this PR. tagging @dandanlen to review the fix.

EDIT: I also basically removed the successful execution of 0 flip redemptions which i think doesnt make sense to execute.

dandanlen (changes requested):
Can we please separate the rpc fix into a separate PR from the redemption change?

Also, I'm not sure we want to return an error. The existing behaviour was presumably implemented on purpose: we want to process the account deletion etc even if there are no remaining funds to redeem. 

If we make it impossible to execute a redemption for small amounts, then it becomes impossible to delete the account.

ramizhasan111:
> Can we please separate the rpc fix into a separate PR from the redemption change?
> 
> Also, I'm not sure we want to return an error. The existing behaviour was presumably implemented on purpose: we want to process the account deletion etc even if there are no remaining funds to redeem.
> 
> If we make it impossible to execute a redemption for small amounts, then it becomes impossible to delete the account.

I separated the 0 flip redemption in another PR. we can discuss there. I still kept a fix where I remove a debug_assert that is just genuinely a bug, which breaks a test in this PR. Instead of trying to the test pass by workaround, I think its simpler to just fix the bug, especially because its simple.

dandanlen:
> Instead of trying to the test pass by workaround, I think its simpler to just fix the bug, especially because its simple.

The bug wasn't fixed, you just removed the code 😅 

There may have been unit tests relying on this. 

I'll add a commit with a fix.

dandanlen (approved):

ramizhasan111:
> > Instead of trying to the test pass by workaround, I think its simpler to just fix the bug, especially because its simple.
> 
> The bug wasn't fixed, you just removed the code 😅
> 
> There may have been unit tests relying on this.
> 
> I'll add a commit with a fix.

the existence of debug_assert WAS the bug I was talking about thats why removing it was the fix. In the case where you try to redeem max and your liquid balance is less than the redemption fee, we currently redeem 0 and take the whole liquid balance as fee (which I was arguing doesnt make sense) and so in that case the debug_assert will fail even though it is a valid scenario in production and will execute successfully. So if we allow this in production, we shouldn't have the debug assert.

dandanlen:
> In the case where you try to redeem max and your liquid balance is less than the redemption fee, we currently redeem 0 and take the whole liquid balance as fee (which I was arguing doesnt make sense) and so in that case the debug_assert will fail even though it is a valid scenario in production and will execute successfully. 

Please, try to write a test that demonstrates this. I'm pretty sure it's fine.

ramizhasan111:
> > In the case where you try to redeem max and your liquid balance is less than the redemption fee, we currently redeem 0 and take the whole liquid balance as fee (which I was arguing doesnt make sense) and so in that case the debug_assert will fail even though it is a valid scenario in production and will execute successfully.
> 
> Please, try to write a test that demonstrates this. I'm pretty sure it's fine.

I added the test for the edge case in another PR which is indeed failing: https://github.com/chainflip-io/chainflip-backend/pull/6787

dandanlen:
> I added the test for the edge case in another PR which is indeed failing: #6787

If that's the case, we should look at what assumptions have been broken and why, rather than simply deleting the check. Presumably it was added for a reason: it's asserting that the difference between the amount debited and the amount actually redeemed should be equal to the fee applied. Why does this no longer hold?

ramizhasan111:
It was always the case and was not changed in this PR 

> > I added the test for the edge case in another PR which is indeed failing: #6787
> 
> If that's the case, we should look at what assumptions have been broken and why, rather than simply deleting the check. Presumably it was added for a reason: it's asserting that the difference between the amount debited and the amount actually redeemed should be equal to the fee applied. Why does this no longer hold?

I already explained it above. The assumption that was wrong is that the difference between redeem amount and debit amount is always the redemption fee. As I explained above, thats not always the case as there is a genuine case that is possible in production where your liquid balance is less than the redemption fee itself. in that case your redeem amount will be 0 and the debit amount will be your total liquid balance (which is less than the redemption fee) and so the difference is less (not equal) to redemption fee. Since this is a valid case that executes successfully in mainnet, we shouldnt be asserting this since the assertion itself is not correct and so the fix is to remove the assertion. Or we dont allow redemptions which result in 0 redeem amount (which I was stressing earlier and impl'ed a fix for) in which case the assert will work.

PR #6746, opened 2026-07-31
fix: lp deregistration checks in-progress swap

marcellorigotti:
# Pull Request

Closes: PRO-3016

## Summary

Lp deregistration needs to check for ongoing swap to ensure funds don't get credited to a de-registered account.

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/pallets/cf-swapping/src/tests.rs
```diff
@@ -1530,6 +1565,51 @@ fn broker_deregistration_checks_private_channels() {
 	});
 }
 
+#[test]
+fn pending_swap_deregistration_check_blocks_account_directed_swaps() {
```
msgmaxim: There are effectively 3 "tests" here. It is fine to put all three in one, but it would help to leave a comment next to each to make it clear what condition is being tested. Looks like they are: (1) unrelated swap does not block deregistration, (2) swap into an account blocks it deregistration and (3) swap with account's address as refund prevents deregistration?

Alternatively, you could crate an inline mod (e.g. `deregistration_checks`) and put this as three separate tests. Then you won't need to clean up in between (like you do when removing swap 2) and you could move helper functions there too.

msgmaxim (approved):

PR #6747, opened 2026-07-31
fix: Bump transitive `event-listener` dependency to fix `RUSTSEC-2026-0221` and ignore `RUSTSEC-2026-0222`.

marcellorigotti:


MxmUrw (approved):
Nice, thanks!

marcellorigotti:
FYI, also added RUSTSEC-2026-0222 to the audit ignore list!

dandanlen:
It looks like this is no longer needed? (There is no diff apart from a comment in the config file). 

Closing.

PR #6749, opened 2026-08-03
ci(bouncer): use ephemeral CPX52 runner in Hetzner

vdoflip:
# Pull Request

## Summary

Replace the always-running Hetzner bouncer runner with an ephemeral CPX52 created for each CI bouncer job and deleted afterward.

  - Provision a uniquely labelled, repository-scoped GitHub runner.
  - Register it as ephemeral so it accepts one job only.
  - Apply a Hetzner firewall.
  - Destroy the server, and runner registration after the job.
  - Add hourly cleanup for resources older than four hours.
  - Upload runner diagnostics before teardown.

### DIFF
--- a/.github/workflows/_40_post_check.yml
+++ b/.github/workflows/_40_post_check.yml
@@ -71,11 +71,21 @@ jobs:
 
   bouncer:
     needs: [provision-bouncer-runner]
-    runs-on:
-      [self-hosted, linux, x64, "${{ needs.provision-bouncer-runner.outputs.runner-label }}"]
+    # Run even when provisioning fails, so that this job always reports under the name
+    # required by branch protection. A skipped job counts as a passing required check,
+    # which would hide provisioning failures.
+    if: ${{ !cancelled() }}
+    # The ephemeral runner label is unique per run, so it identifies the runner on its own.
+    # Fall back to a hosted runner when there is none to target, otherwise the job waits
+    # on a label that matches nothing until it times out.
+    runs-on: ${{ needs.provision-bouncer-runner.result == 'success' && needs.provision-bouncer-runner.outputs.runner-label || 'namespace-profile-default-light' }}
     timeout-minutes: ${{ inputs.timeout-minutes }}
     continue-on-error: ${{ inputs.continue-on-bouncer-error }}
     steps:
+      - name: Fail if runner provisioning failed 🚨
+        if: ${{ needs.provision-bouncer-runner.result != 'success' }}
+        run: exit 1
+
       - name: Setup ngrok ✨
         if: inputs.ngrok
         run: |
--- a/.github/workflows/_40_post_check.yml
+++ b/.github/workflows/_40_post_check.yml
@@ -50,8 +50,10 @@ jobs:
     runs-on: namespace-profile-default-light
     timeout-minutes: 15
     outputs:
+      firewall-id: ${{ steps.provision.outputs.firewall-id }}
       runner-label: ${{ steps.provision.outputs.runner-label }}
       runner-name: ${{ steps.provision.outputs.runner-name }}
+      server-id: ${{ steps.provision.outputs.server-id }}
     steps:
       - name: Checkout chainflip-backend 🛒
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-development.yml
```diff
@@ -95,6 +95,9 @@ jobs:
   test-benchmarks:
     needs: [build-benchmarks]
     uses: ./.github/workflows/_41_test_benchmarks.yml
+    secrets:
+      CF_GITHUB_RUNNERS_MANAGEMENT_TOKEN: ${{ secrets.CF_GITHUB_RUNNERS_MANAGEMENT_TOKEN }}
+      CF_HETZNER_API_TOKEN: ${{ secrets.CF_HETZNER_API_TOKEN }}
```
MxmUrw, line 100: This looks a bit strange: it looks like we don't do this step for `post-check`, we just immediately inject the secrets into the "provision" and "destroy" steps, so why is it required here for `test-benchmarks`?

**EDIT**: Appearently we don't have to do this for post-check because there we use `secrets: inherit` - why don't we also use it here?

.github/workflows/ci-main-merge-queue.yml
```diff
@@ -30,6 +30,9 @@ jobs:
   test-benchmarks:
     needs: [build-benchmarks]
     uses: ./.github/workflows/_41_test_benchmarks.yml
+    secrets:
+      CF_GITHUB_RUNNERS_MANAGEMENT_TOKEN: ${{ secrets.CF_GITHUB_RUNNERS_MANAGEMENT_TOKEN }}
+      CF_HETZNER_API_TOKEN: ${{ secrets.CF_HETZNER_API_TOKEN }}
```
MxmUrw, line 35: Same?

.github/workflows/release-sisyphos.yml
```diff
@@ -96,6 +96,9 @@ jobs:
   test-benchmarks:
     needs: [build-benchmarks]
     uses: ./.github/workflows/_41_test_benchmarks.yml
+    secrets:
+      CF_GITHUB_RUNNERS_MANAGEMENT_TOKEN: ${{ secrets.CF_GITHUB_RUNNERS_MANAGEMENT_TOKEN }}
+      CF_HETZNER_API_TOKEN: ${{ secrets.CF_HETZNER_API_TOKEN }}
```
MxmUrw, line 101: Same?

MxmUrw (approved):
Nice, looks good! One question:

.github/workflows/ci-development.yml
vdoflip: they are not inherited workflow-wide. `secrets: inherit` applies only to the specific reusable-workflow job where it appears, and secrets are not automatically forwarded transitively.

We could replace those mappings with `secrets: inherit`, but that would expose every available repository secret to the benchmark workflow. Explicitly passing only these two is the safer option.

.github/workflows/_40_post_check.yml
```diff
@@ -299,6 +332,25 @@ jobs:
         run: |
           docker compose -f localnet/docker-compose.yml -p "chainflip-localnet" down --volumes --remove-orphans
 
+  destroy-bouncer-runner:
```
nmammeri, line 345: Same as above. It looks like there is a lot of duplication between `destroy-bouncer-runner`, `destroy-benchmarks-runner` and `destroy-upgrade-runner`. Can't we somehow shared logic that can be called with different params.

.github/workflows/_40_post_check.yml
```diff
@@ -46,8 +46,33 @@ permissions:
   contents: read
 
 jobs:
+  provision-bouncer-runner:
```
nmammeri, line 49: It looks like there is a lot of duplication between `provision-bouncer-runner`, `provision-benchmarks-runner` and `provision-upgrade-runner`.  Can't we somehow shared logic that can be called with different params.

Provisioning a runner is a fundamental step and this way we'll duplicate this provisioning logic whenever we add a new job type. It's better if we figure a way to make it shred.

.github/workflows/_40_post_check.yml
MxmUrw: Yeah I thought the same.

I tried asking AI to do this, the result was 4 lines more of code than it started with. The reason is that if we extract it into it's own "reusable workflow", defining the input parameters of that workflow takes some lines, and also actually calling that extracted workflow takes up some space.

And when we look at what we gain it's not that much honestly, because the duplicated code here is mostly just collecting config and forwarding to the actual script anyway.

So unfortunately I think in this case it's actually more concise to *not* extract the duplicated code.

vdoflip: yes I don't think we need to add yet another workflow for this

.github/workflows/upgrade-test.yml
```diff
@@ -36,8 +36,31 @@ permissions:
   contents: read
 
 jobs:
+  provision-upgrade-runner:
```
dandanlen, line 39: We are always provisioning the server even if the `run-job` input is false (and then the job does nothing). There seems to be some some reason for this no-op job, there is a comment:

```
      # Allows us to create a workflow call that doesn't run the job
      # so that branch protection rules can skip when running on push, but not in merge queue.
```

I'm not quite sure what this means in detail but it seems that we simply need the job to exist and return success. Previously this was ok, but it seems that we are now provisioning some Hetzner resources that sit idle. My understanding is that this would happen on every run of the ci-development checks.

.github/workflows/_40_post_check.yml
nmammeri: If it ends up with more code there is no point in creating a separate workflow then

dandanlen (approved):

PR #6755, opened 2026-08-06
chore: update storage analysis script:

dandanlen:
# Pull Request

- Update dependency versions
- Better analytics for storage size
- Cargo fmt/clippy

The script now generates a storage-report file (gitignored) containing information about storage patterns and usage, as well potential orphaned storage that can be removed.

Orphan storage is identified by building a list of removed `pub type X` declarations from a history of git diffs and matching against those hashes.

I decided to keep this crate separate from the workspace to avoid building and running it with every CI run. The cost is that we won't auto-detect when dependencies go out of date. We can revisit later if this becomes an issue.

### DIFF
--- a/state-chain/runtime-tests/src/tests.rs
+++ b/state-chain/runtime-tests/src/tests.rs
@@ -13,11 +13,8 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-use std::collections::BTreeMap;
-
 use crate::StateChainBlock;
 use frame_remote_externalities::RemoteExternalities;
-use frame_support::StorageHasher;
 
 pub type Ext = sp_state_machine::TestExternalities<sp_runtime::traits::BlakeTwo256>;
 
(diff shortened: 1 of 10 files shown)


dandanlen:
[storage-report-0x47fcfec41800356c6e8cddfa60e6aad4409e2a8cdaacd6f212ded71c40ef422b.md](https://github.com/user-attachments/files/30786526/storage-report-0x47fcfec41800356c6e8cddfa60e6aad4409e2a8cdaacd6f212ded71c40ef422b.md)

This is an example of the generated report.

state-chain/runtime-tests/src/tests.rs
```diff
@@ -32,63 +29,48 @@ pub trait RuntimeTest: Default {
 pub mod auction_resolution;
 pub mod rotation_breakdown;
 pub mod rotation_on_initialize;
+pub mod storage_analysis;
 pub mod swap_rate;
 pub mod witnesser_cull;
 
-pub fn run_all(ext: RemoteExternalities<StateChainBlock>) -> anyhow::Result<()> {
+/// Set to skip the runtime tests and only produce the storage report.
+fn storage_analysis_only() -> bool {
```
nmammeri: I think we need document how this env variable is used to filter other test and only run the storage analysis test in  the README.

Also, no need to have this check in a function, it only used once.

nmammeri (approved):
LGTM, the script uses many heuristics to assign names to metadata unknown storage prefixes,I think this is fine for analysis purposes.

Another minor issue, when I run the test binary I got an error ` Error: fs::write failed` which was requiring a `snapshots` directory in the calling dir.

dandanlen:
I made the following changes:

- Added help (more likely to stay in sync than a README IMO). 
- Changed the default snapshot directory under the OS-standard cache directory (for example on a mac this would be `~/Library/Caches/chainflip/runtime-tests/snapshots`). This way it doesn't pollute the repo and can also be shared between multiple checkouts of the repo. 
- Inlined the function as you suggested

dandanlen:
About the heuristics, I think they make sense: It filters the repo and git history for anything that looks like a storage declaration and calculates the storage hashes associated with this in order to match them against unknown storage keys.

PR #6756, opened 2026-08-06
feat: perf improvements

nmammeri:
# Pull Request

Closes: PRO-3035

## Summary

* This PR includes some basic performance improvements improving block execution time
* The changes are kept minimal, any other optimization that require big changes are included in followup PRs
* The improvements in this PR are for the `vote` and `onFinalize` functions mainly reducing storage accesses
* Each commit contains a self contained improvement to make it easy to review

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
@@ -206,32 +206,34 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
 		safety_buffer: usize,
 	) {
 		//--------- calculate new events ---------
-		let new_events: Vec<_> = self
-			.blocks_data
-			.iter_mut()
-			.flat_map(|(block_height, block_info)| {
-				let new_next_age_to_process = ChainBlockNumberOf::<T::Chain>::steps_between(
-					block_height,
-					&seen_heights_below,
-				)
-				.0;
-				let age_range: Range<u32> =
-					(block_info.next_age_to_process)..new_next_age_to_process as u32;
+		let mut new_events = Vec::new();
+		for (block_height, block_info) in self.blocks_data.iter_mut() {
+			let new_next_age_to_process =
+				ChainBlockNumberOf::<T::Chain>::steps_between(block_height, &seen_heights_below).0;
+			let age_range: Range<u32> =
+				(block_info.next_age_to_process)..new_next_age_to_process as u32;
 
-				block_info.next_age_to_process = new_next_age_to_process as u32;
+			block_info.next_age_to_process = new_next_age_to_process as u32;
 
-				self.debug_events.run(BlockProcessorEvent::ProcessingBlockForAges {
-					height: *block_height,
-					ages: age_range.clone(),
-				});
+			// Nothing has aged, so the rules would produce nothing: skip before cloning the block
+			// data for them. Common, since this runs once per state machine step.
+			if age_range.is_empty() {
+				continue;
+			}
 
+			self.debug_events.run(BlockProcessorEvent::ProcessingBlockForAges {
+				height: *block_height,
+				ages: age_range.clone(),
+			});
+
+			new_events.extend(
 				self.rules
 					.run((age_range, block_info.block_data.clone(), block_info.safety_margin))
 					.into_iter()
 					.filter(|event| !self.processed_events.contains_key(event))
-					.map(|event| (*block_height, event))
-			})
-			.collect();
+					.map(|event| (*block_height, event)),
+			);
+		}
 
 		//--------- execute new events ---------
 		self.execute.run(new_events);
--- a/state-chain/pallets/cf-elections/src/electoral_systems/state_machine/state_machine_es.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/state_machine/state_machine_es.rs
@@ -205,7 +205,7 @@ impl<ES: StatemachineElectoralSystemTypes> ElectoralSystem for StatemachineElect
 
 		// step for each election that reached consensus
 		log::debug!("ESSM: stepping for each election with consensus ({:?})", election_identifiers);
-		let mut elections = Vec::with_capacity(election_identifiers.len());
+		let mut identifiers_and_properties = Vec::with_capacity(election_identifiers.len());
 		for election_identifier in election_identifiers {
 			let election_access = ElectoralAccess::election_mut(election_identifier);
 			let properties = election_access.properties()?;
@@ -214,7 +214,7 @@ impl<ES: StatemachineElectoralSystemTypes> ElectoralSystem for StatemachineElect
 				log::debug!("ESSM: stepping with input {input:?}");
 				step(Either::Right((properties.clone(), input)))?;
 			}
-			elections.push((election_identifier, properties));
+			identifiers_and_properties.push((election_identifier, properties));
 		}
 
 		// gather the queries after all state transitions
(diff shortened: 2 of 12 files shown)


dandanlen:
@nmammeri can we create and link a Linear issue if there isn't one already?

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -2065,6 +2067,10 @@ pub mod pallet {
 			unique_monotonic_identifier: UniqueMonotonicIdentifier,
 			authority: &T::ValidatorId,
 			authority_index: AuthorityCount,
+			// Whether `authority` is in `ContributingAuthorities`. An argument rather than a
```
dandanlen: FWIW i think we should keep comments to a minimum. AI likes to add lots of multiline comments but eg. here it's not adding anything very useful (the justification for *why* this was added is captured in git history, the fact that it's an argument is trivial, etc..).

state-chain/runtime/src/chainflip/witnessing/arbitrum_elections.rs
```diff
@@ -201,9 +209,10 @@ impl BlockWitnesserInstance for ArbitrumVaultDepositWitnessing {
 	}
 
 	fn election_properties(
-		_block_height: pallet_cf_elections::electoral_systems::block_height_witnesser::ChainBlockNumberOf<Self::Chain>,
-	) {
+		block_heights: &[pallet_cf_elections::electoral_systems::block_height_witnesser::ChainBlockNumberOf<Self::Chain>],
+	) -> Vec<Self::ElectionProperties> {
 		// Vault address doesn't change, it is read by the engine on startup
+		block_heights.iter().map(|_| ()).collect()
```
dandanlen, line 215: Does this need to return anything? The old version didn't...

dandanlen: This is related to the `zip`ing of results I suppose?

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
```diff
@@ -206,32 +206,34 @@ impl<T: BWProcessorTypes> BlockProcessor<T> {
...
+				ChainBlockNumberOf::<T::Chain>::steps_between(block_height, &seen_heights_below).0;
+			let age_range: Range<u32> =
+				(block_info.next_age_to_process)..new_next_age_to_process as u32;
+
+			block_info.next_age_to_process = new_next_age_to_process as u32;
+
+			// Nothing has aged, so the rules would produce nothing: skip before cloning the block
+			// data for them. Common, since this runs once per state machine step.
+			if age_range.is_empty() {
+				continue;
```
dandanlen, line 221: Are we guaranteed that `run` does nothing if the `age_range` is empty? It might be worth documenting that requirement on method.

state-chain/runtime/src/chainflip/witnessing/arbitrum_elections.rs
dandanlen: There seem to be a few places where we do this (iterate with `|_| ()`) to return 'nothing'. Maybe there's a way to skip this if we don't need it?

dandanlen (commented):
LGTM! Just a few minor comments. I'd like @MxmUrw to take a look too if he hasn't already. 

Did you measure the improvement for each change individually?

nmammeri:
> LGTM! Just a few minor comments. I'd like @MxmUrw to take a look too if he hasn't already.
> 
> Did you measure the improvement for each change individually?

No I didn't measure individual improvements, all vote optimization measured together.

state-chain/runtime/src/chainflip/witnessing/arbitrum_elections.rs
nmammeri: election_properties should return one entry per height, as you said the caller usually zip the results. I don't think there is perf loss we can write it as `vec![(); block_heights.len()]`

dandanlen: I don't think `vec![..]` would make much difference tbh. It's more that I'm wondering if there was a way to short-circuit this if we don't need the output. Not a big deal.

MxmUrw: I think it makes sense as is. The `election_properties()` function now has to return a vector that's exactly as long as the input, so it makes sense to follow this invariant even if we return a bunch of `()`.

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
MxmUrw: Yes, the two (generic) implementations of the rules hook (`PrewitnessImmediatelyAndWitnessAtSafetyMargin` and `JustWitnessAtSafetyMargin`) both guarantee it.

I've written up the full requirements & guarantees of the rules hook as a comment in this PR: https://github.com/chainflip-io/chainflip-backend/pull/6762 which targets this branch.

It's a bit unfortunate that the dispatch is a bit indirect, which means that there isn't really a neat way to make the comments show up in the IDE when calling `rules.run()` here, because the `run` function is a generic method on `Hook`.

state-chain/runtime/src/chainflip/witnessing/arbitrum_elections.rs
```diff
@@ -154,17 +154,25 @@ impl BlockWitnesserInstance for ArbitrumDepositChannelWitnessing {
...
+				let height = height.root();
+				ArbitrumIngressEgress::filter_active_deposit_channels(
+					&channels,
+					height.saturating_forward(ARBITRUM_MAINNET_SAFETY_BUFFER as usize),
+					*height,
+				)
+				.into_iter()
+				.map(|deposit_channel_details| deposit_channel_details.deposit_channel)
+				.collect()
+			})
```
MxmUrw, line 174: Previously we did a single call into the ingress egress pallet, now we do two: one for getting channels and one for filtering them. I think we should either move more of the logic to the ingress pallet or move all of it here.

This would require changing a few function signatures so maybe it's fine as is for now.

state-chain/pallets/cf-elections/src/electoral_systems/state_machine/state_machine_es.rs
```diff
@@ -205,13 +205,16 @@ impl<ES: StatemachineElectoralSystemTypes> ElectoralSystem for StatemachineElect
 
 		// step for each election that reached consensus
 		log::debug!("ESSM: stepping for each election with consensus ({:?})", election_identifiers);
-		for election_identifier in &election_identifiers {
-			let election_access = ElectoralAccess::election_mut(*election_identifier);
+		let mut elections = Vec::with_capacity(election_identifiers.len());
```
MxmUrw: nit: could we rename this, e.g. to `identifiers_and_properties`?

Below we call `let election = ElectoralAccess::election_mut(election_identifier);` to get an actual "election" so it is confusing if we already have a list called `elections` that doesn't actually contain elections.

MxmUrw (approved):

dandanlen (approved):

state-chain/pallets/cf-elections/src/electoral_systems/block_witnesser/block_processor.rs
MxmUrw: Pushed the comment commit to this branch.

PR #6768, opened 2026-08-11
fix: scale StablecoinDefaults to each asset's decimals

j4m1ef0rd:
## Summary

StablecoinDefaults was used for BscUsdt but did not support different decimal assets. This would make the default values for the localnet wrong for BscUsdt. Simple fix to make the default now specified in USD (instead of fine units) and scale as needed.

## API Changes

None.

### RPCS

None.

### Extrinsics & Events

None.

---

* [x] Unit tests for the rescaling and the resulting per-asset defaults.

### DIFF
--- a/state-chain/primitives/src/lib.rs
+++ b/state-chain/primitives/src/lib.rs
@@ -475,29 +475,6 @@ impl ChainflipNetwork {
 	}
 }
 
-#[test]
-fn is_more_recent_semver() {
-	fn ver(major: u8, minor: u8, patch: u8) -> SemVer {
-		SemVer { major, minor, patch }
-	}
-
-	fn ensure_left_is_more_recent(left: SemVer, right: SemVer) {
-		assert!(left.is_more_recent_than(right));
-		// Additionally check that the inverse is false:
-		assert!(!right.is_more_recent_than(left));
-	}
-
-	assert!(!ver(0, 1, 0).is_more_recent_than(ver(0, 1, 0)));
-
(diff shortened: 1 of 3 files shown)


state-chain/primitives/src/lib.rs
```diff
@@ -574,12 +551,28 @@ pub struct DcaParameters {
...
+/// A default amount for every USD stablecoin, from an amount `N` expressed in the fine units of
+/// [`STABLE_ASSET`]. Each entry is rescaled to that asset's own decimals so that they all represent
+/// the same nominal USD amount.
 pub struct StablecoinDefaults<const N: u128>();
 impl<const N: u128> Get<BTreeMap<Asset, AssetAmount>> for StablecoinDefaults<N> {
 	fn get() -> BTreeMap<Asset, AssetAmount> {
 		Asset::all()
 			.filter(|asset| asset.is_usd_stablecoin())
-			.map(|asset| (asset, N))
+			.map(|asset| (asset, rescale_fine_amount(N, STABLE_ASSET, asset)))
```
msgmaxim: Could we change the meaning of N to be "how many $" and then here we would just multiply by 10^asset.decumals()? It seems strange that 6 decimals has a special meaning now and we need to compute "relative" scaling for that.

dandanlen: ☝️ I think this makes sense. This way we wouldn't need `rescale_fine_amounts`.

msgmaxim (approved):

dandanlen (approved):

PR #6769, opened 2026-08-11
fix: retry assethub vault creation extrinsic on startup

MxmUrw:
# Pull Request

## Summary

Since the localnet assethub has been upgraded to 2.3.2 (3s block times) it seems that the failure mode of bouncer where assethub vault creation fails on startup has become much more frequent.

I think this is because of polkadot reorgs affecting inclusion of assethub blocks. Thus not all successfully submitted assethub extrinsics are finalised.

We don't currently have a retry mechanism for the vault creation extrinsic. This PR extracts the recently added retry mechanism from `sendHubAsset()` and makes it available for all extrinsic submissions.

To add a bit more detail:
 - the `createPolkadotVault(api)` method which existed to share code between assethub and polkadot setup is now removed and the content inlined into `createAssethubVault`, as we don't need polkadot setup anymore
 - the extracted extrinsic submission logic lives in a new function called `submitHubExtrinsic` and `sendHubAsset` calls into it. It follows a similar pattern as `ChainflipIO.submitExtrinsic` and allows to specify the expected event. It is a bit less sophisticated of course.

## API Changes

None

---

* The PR is complete:
  * [ ] Reviewers are assigned.

### DIFF
--- a/bouncer/shared/send_hubasset.ts
+++ b/bouncer/shared/send_hubasset.ts
@@ -71,51 +71,46 @@ export async function submitHubExtrinsic(
     resolve: (result: { txHash: string; eventData?: unknown }) => void,
     reject: (error: Error) => void,
   ) => {
-    const unsubscribe = await extrinsic(assethubApi).signAndSend(
-      alice,
-      { nonce, tip },
-      (result) => {
-        if (result.dispatchError !== undefined) {
-          if (result.dispatchError.isModule) {
-            const decoded = assethubApi.registry.findMetaError(result.dispatchError.asModule);
-            const { docs, name, section } = decoded;
-            unsubscribe();
-            reject(new Error(`${section}.${name}: ${docs.join(' ')}`));
-          } else {
-            unsubscribe();
-            reject(new Error('Error: ' + result.dispatchError.toString()));
-          }
-        }
-        if (result.status.isFinalized) {
+    const tx = extrinsic(assethubApi);
+    const txHash = tx.hash.toString();
+    const unsubscribe = await tx.signAndSend(alice, { nonce, tip }, (result) => {
+      if (result.dispatchError !== undefined) {
+        if (result.dispatchError.isModule) {
+          const decoded = assethubApi.registry.findMetaError(result.dispatchError.asModule);
+          const { docs, name, section } = decoded;
           unsubscribe();
+          reject(new Error(`${section}.${name}: ${docs.join(' ')}`));
+        } else {
+          unsubscribe();
+          reject(new Error('Error: ' + result.dispatchError.toString()));
+        }
+      }
+      if (result.status.isFinalized) {
+        unsubscribe();
 
-          if (expectedEvent) {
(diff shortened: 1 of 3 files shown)


dandanlen:
> Thus not all successfully submitted assethub extrinsics are finalised.

This seems quite insane... so we can have an extrinsic that is included in an assethub block, but that block is not included by the relay chain??

MxmUrw:
> This seems quite insane... so we can have an extrinsic that is included in an assethub block, but that block is not included by the relay chain??

I think so: if there is a problem with including an authored assethub block #n on the relay chain, the assethub node will author another block for that block height. I'm not sure if it's that surprising: assethub block inclusion depends on posting updates to the relay-chain, and that one has reorgs because we have to run at least two nodes.

dandanlen:
> > This seems quite insane... so we can have an extrinsic that is included in an assethub block, but that block is not included by the relay chain??
> 
> I think so: if there is a problem with including an authored assethub block #n on the relay chain, the assethub node will author another block for that block height. I'm not sure if it's that surprising: assethub block inclusion depends on posting updates to the relay-chain, and that one has reorgs because we have to run at least two nodes.

But doesn't this imply that our submissions from the engine should fail from time to time on mainnet? Shouldn't the Assethub node re-author a block that includes the missing transaction? It seems very strange!

bouncer/shared/send_hubasset.ts
```diff
@@ -128,5 +130,35 @@ export async function sendHubAsset(
...
+    result = await submitHubExtrinsic(
+      logger,
+      (api) => api.tx.balances.transferKeepAlive(address, planckAmount),
+      `balances.transferKeepAlive(${address}, ${planckAmount})`,
+    );
+  } else if (asset === 'HubUsdc' || asset === 'HubUsdt') {
+    result = await submitHubExtrinsic(
+      logger,
+      (api) => api.tx.assets.transferKeepAlive(getHubAssetId(asset), address, planckAmount),
+      `transferKeepAlive(${asset}, ${address}, ${planckAmount})`,
```
dandanlen: Just to remain consistent with the above, should this be `assets.transferKeepAlive(..)`?

MxmUrw:
> But doesn't this imply that our submissions from the engine should fail from time to time on mainnet?

I don't understand what the difference is to just normal reorgs, say on bitcoin?

> Shouldn't the Assethub node re-author a block that includes the missing transaction? It seems very strange!

Yes, I think parity thought the same and created the fork-aware mempool!

Anyways I agree that all in all this makes assethub a bit of a hassle to deal with. On mainnet we have the advantage that we retry up to 150 times to submit an extrinsic and the likelyhood is high that it goes actually through eventually.

dandanlen (commented):
I guess what we're discussing above maps to the `Retracted` state?

dandanlen (approved):

MxmUrw:
> I guess what we're discussing above maps to the Retracted state?

Yes that seems to be the state where an already included (but not finalized) tx left the canonical chain.

PR #6772, opened 2026-08-12
feat: inventory strategy support for different decimal assets

j4m1ef0rd:
# Pull Request

## Summary

BcsUsdt has 18 decimals (all other stable coins have 6) and does not have an oracle (oracle supports any decimal assets already), so it was unable to be used for any strategy. This PR changes how the Inventory based strategy works so it can support different asset decimals. Same behaviour, just using a different relative tick. No external/UI change needed.

Added a `into_nearest_tick` function because the 12 decimal difference price lands on a tick of 0.9, making the relative price almost 1 tick off without the new rounding. This will affect the oracle strategy as well, but it's for the better and will not be noticable.
Note: BcsUsdt will still need to be enable using governance on mainnet. Or we can add a migration to this PR to enable it if we like. @dandanlen 

---
* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/state-chain/amm-math/src/lib.rs
+++ b/state-chain/amm-math/src/lib.rs
@@ -919,8 +919,10 @@ mod test {
 			);
 		}
 
-		// The ratio between assets with a 12 decimal difference, as used to price stablecoin pairs.
-		// Truncating lands 10 ticks (~1 bps) below the true price.
+		// For the price of assets with a 12 decimal difference (Usdc and BscUsdt), the true tick is
+		// fractional, and flooring lands almost a full tick (~1 bps) below it.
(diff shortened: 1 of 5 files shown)


dandanlen (approved):
LGTM! Thanks for the quick turnaround

dandanlen:
Note that the tests were failing because this actually changes the output ticks for the oracle pricing strategy because of `into_nearest_tick`. 

I think it's a fair change to make, the argument applies to order updates just as much as it does to pricing estimates. I'll update the tests with the new values so CI can pass.

msgmaxim (approved):

PR #6777, opened 2026-08-12
fix: update Solana localnet connection to use 127.0.0.1

dandanlen:
# Pull Request

This was causing issues on my local machine. 

`0.0.0.0` is a bind address, not a target. Linux is lenient and treats it as localhost. Mac OS refuses it (on my machine). 

Changed it to the more correct 127.0.0.1

### DIFF
--- a/bouncer/shared/utils.ts
+++ b/bouncer/shared/utils.ts
@@ -895,9 +895,9 @@ export function getWeb3(chain: Chain): Web3 {
 }
 
 export function getSolConnection(): Connection {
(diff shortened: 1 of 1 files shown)


MxmUrw (approved):
Strange though as it must have been working before, and seems to be working for others...

Anyways, 127.0.01 seems to be the proper address anyways.

dandanlen:
> Strange though as it must have been working before, and seems to be working for others...
> 
> Anyways, 127.0.01 seems to be the proper address anyways.

Yes, strange! I saw this on a fresh clone of the repo - clearly my 'normal' clone has some lingering context that (maybe) isn't reflected when the repo is cloned? Either way - this is better.

PR #6779, opened 2026-08-12
fix: gate generated Arbitrary impls on the proptest feature

dandanlen:
# Pull Request

## Summary

The `test` annotation activates only for crate-local unit tests.

The correct approach is to activate only on the proptest feature, otherwise the code can be emitted
in a context where proptest isn't available.

This was causing test build failures when testing certain crates/pallets in isolation (eg. cf-swapping).


MxmUrw (approved):
I was under the impression that this was required somewhere (e.g. when running `cargo test` inside a create where the tests actually need the proptest implementations), but that might be outdated - and tests seem to pass.

It's cleaner this way so let's hope it works everywhere!

PR #6790, opened 2026-08-14
fix: Ensure Flip is credited to account when opening channel with FLIP

dandanlen:
## Summary

Fixes: PRO-3058

When opening a deposit channel with FLIP from the vault, we weren't scheduling a transfer of those funds to the gateway.

FWIW this is an accounting issue, not a vulnerability (eg. double spend).

### DIFF
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -34,7 +34,7 @@ pub use weights::WeightInfo;
 mod tests;
 
 use cf_chains::{evm::Address as EthereumAddress, RegisterRedemption};
-use cf_primitives::{chains::assets::eth::Asset as EthAsset, AssetAmount};
+use cf_primitives::{chains::assets::eth::Asset as EthAsset, Asset, AssetAmount};
 use cf_traits::{
 	impl_pallet_safe_mode, AccountInfo, AccountRoleRegistry, Broadcaster, Chainflip, FeePayment,
 	FundAccount, Funding, FundingSource, GetMinimumFunding, MoveFlipToGateway, RedemptionCheck,
@@ -1205,7 +1205,13 @@ impl<T: Config> FundAccount for Pallet<T> {
 				},
 			// These funds are backed by Flip in the Vault rather than in the Gateway, so an
 			// equivalent amount must be moved across to keep the Gateway fully backed.
-			FundingSource::FreeBalance =>
+			//
+			// The same applies to a Flip deposit at account creation, which is credited straight
+			// from the Vault. Account creation with any other asset instead reaches the account
+			// via a swap, and the swap output action earmarks the full output itself - so
+			// earmarking here as well would double-count it.
+			FundingSource::FreeBalance |
+			FundingSource::InitialFunding { asset: Asset::Flip, .. } =>
 				T::MoveFlipToGateway::add_flip_to_be_sent_to_gateway(amount.into()),
 			FundingSource::Swap { .. } | FundingSource::InitialFunding { .. } => {},
 		}
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -34,7 +34,7 @@ pub use weights::WeightInfo;
 mod tests;
 
 use cf_chains::{evm::Address as EthereumAddress, RegisterRedemption};
-use cf_primitives::{chains::assets::eth::Asset as EthAsset, Asset, AssetAmount};
+use cf_primitives::{chains::assets::eth::Asset as EthAsset, AssetAmount};
 use cf_traits::{
 	impl_pallet_safe_mode, AccountInfo, AccountRoleRegistry, Broadcaster, Chainflip, FeePayment,
 	FundAccount, Funding, FundingSource, GetMinimumFunding, MoveFlipToGateway, RedemptionCheck,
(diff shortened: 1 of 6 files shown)


state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -1205,7 +1205,13 @@ impl<T: Config> FundAccount for Pallet<T> {
...
 			// These funds are backed by Flip in the Vault rather than in the Gateway, so an
 			// equivalent amount must be moved across to keep the Gateway fully backed.
-			FundingSource::FreeBalance =>
+			//
+			// The same applies to a Flip deposit at account creation, which is credited straight
+			// from the Vault. Account creation with any other asset instead reaches the account
+			// via a swap, and the swap output action earmarks the full output itself - so
+			// earmarking here as well would double-count it.
+			FundingSource::FreeBalance |
+			FundingSource::InitialFunding { asset: Asset::Flip, .. } =>
```
MxmUrw: Honestly this seems a bit brittle. I would rather have both flip and non-flip paths trigger the flip sending in a "parallel" way.

What I mean is, the code in `perform_channel_action` currently does:
```
let is_flip = matches!(asset, Flip)
fund_account(FundingSource::InitialFunding { asset, ... })

// other steps

if is_flip {
  // transfer of flip should happen here, I think
  amount
else {
  trigger_swap { output_action: SwapOutputAction::CreditFlipAndTransferToGateway { ... } }
}
```
So for flip the scheduling happens in fund_account based on the fact that the provided asset was flip, while for other assets it happens later on inside one if branch due to the output action. So there are now two places that are invisibly linked and the correctness guarantee depends on their individual triggers being mutually exclusive + cover all cases.

I think it would be more straight-forward if the flip transfer happened in the first branch of the `is_flip` check.

dandanlen: I agree it's not ideal and I want to consolidate the decision of whether or not to trigger the transfer in the fund_account method, otherwise it's the caller's responsiblity to reason about this (IMO it shouldn't be). See #6791 

(Arguably we should *always* trigger a transfer but that's another discussion)

Maybe something like this would be better:

```
if is_flip {
  // transfer of flip should happen here, I think
  fund_account(amount, FundingSource::FreeBalance);
else {
  fund_account(MINIMUM, FundingSource::Deferred);
  trigger_swap { output_action: SwapOutputAction::CreditFlipAndTransferToGateway { ... } }
}

// .. Later in the swapping pallet
fund_account(swap_output - MINIMUM, FundingSource::SwapOutput)
```

However note that the choice of variants in FundingSource was made not because the information is needed by the funding pallet, but because the information is emitted in an event that is consumed in the product.

dandanlen: Actually looking at it, I think it makes sense to include that change in this PR. Will add it.

MxmUrw: > Maybe something like this would be better: ...

Yes I like this much more already!

> However note that the choice of variants in FundingSource was made not because the information is needed by the funding pallet, but because the information is emitted in an event that is consumed in the product.

Yes by itself that's fine - just the behaviour of `fund_account` should ideally not have two branching layers: it can branch on the funding source, but then not again on the asset that's within one of the variants.

dandanlen: > but then not again on the asset that's within one of the variants

What's wrong with that?

dandanlen: Anyway the new version doesn't do that :)

dandanlen: Since `asset` actually has nothing to do with `amount`, which is always denominated in FLIP.

MxmUrw (approved):
Looks good, I like the new design!

**EDIT**: will have to be rebased (seems that stacks are not magic...)

state-chain/pallets/cf-funding/src/lib.rs
MxmUrw: > > but then not again on the asset that's within one of the variants
> 
> What's wrong with that?

It's more difficult to reason about because now one has to think "in two layers" about the possible code execution paths. And the fact that a match inside a match is required hints at the possibility that the outer enum might be a wrong abstraction for what the function does. So either the input type should be updated such that a single layer match is enough or the function should do something different.

> Anyway the new version doesn't do that :)

And it's much nicer, isn't it!

ramizhasan111:
LGTM to me too 👍 
Can have a final quick look once its rebased since the code from the base branch is creating noise.

dandanlen:
Unstacked... what a terrible feature... 

(rebased manually)

state-chain/pallets/cf-swapping/src/lib.rs
```diff
@@ -2660,11 +2660,6 @@ pub mod pallet {
 													total.saturating_reduce(deficit);
 												});
 											}
```
ramizhasan111, line 2663: we dont need to do this in this PR but I think that now that we dont have a need to have `FlipToBeSentToGateway` in this pallet, we should move it to the pallet where it is actually modified (funding pallet) instead of the `fund_account` function in the funding pallet calling back to this pallet to modify the storage. Its needless indirection.

The `maybe_trigger_flip_to_gateway_egress` function can also move along with the storage since the only reason that function exists here is because the storage is here.

ramizhasan111 (approved):

state-chain/pallets/cf-swapping/src/lib.rs
dandanlen: Yes, there's a follow up PR for this: #6791

PR #6801, opened 2026-08-18
feat: runtime profiling feature

nmammeri:
# Pull Request

Closes: PRO-3035

## Summary

This PR implements:
* `runtime-tracing` feature which enables `sp_tracing` spans in the runtime for per-pallet/per-extrinsic timing. Off by default and never enabled for production: with it off every span expands to nothing, so there is zero overhead on the on-chain runtime.
 * `#[cf_runtime_utilities::instrument]` can be to annotate pallet functions. It opens a span named after the function, with no feature or dependency plumbing needed in the annotated crate. Use (pallet) to record the instance name of generic pallets (e.g. BitcoinElections) as a field — otherwise an instantiable pallet's  instances are indistinguishable. 
* Both `vote` and `onFinalize` election pallet functions are annotated with: `#[cf_runtime_utilities::instrument(pallet)]`
* Also `state-chain/scripts/spans-to-profile.py` script is implemented to allow for rebuilding the span tree from a benchmark execution log containing tracing log lines. Help is documented in "Profiling runtime execution" in the README.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,7 +713,7 @@ version = "3.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
 dependencies = [
- "event-listener 5.4.1",
+ "event-listener 5.4.2",
  "event-listener-strategy",
  "pin-project-lite",
 ]
@@ -742,7 +742,7 @@ dependencies = [
  "async-task",
  "blocking",
  "cfg-if",
- "event-listener 5.4.1",
+ "event-listener 5.4.2",
  "futures-lite",
  "rustix",
 ]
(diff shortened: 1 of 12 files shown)


dandanlen (approved):
Very cool!

MxmUrw (commented):
Is there a reason we don't want to use the official `#[instrument]` command from https://docs.rs/tracing-attributes/0.1.31/tracing_attributes/attr.instrument.html ?

nmammeri:
> Is there a reason we don't want to use the official `#[instrument]` command from https://docs.rs/tracing-attributes/0.1.31/tracing_attributes/attr.instrument.html ?

Mainly because you need to add it as a dependency wherever you want to use it and also it doesn't get the pallet name of generic pallets. We get the pallet instance name using `::cf_runtime_utilities::__reexports::frame_support::traits::PalletInfoAccess>::name()`

Cargo.lock
```diff
@@ -713,7 +713,7 @@ version = "3.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
 dependencies = [
- "event-listener 5.4.2",
```
MxmUrw, line 716: Are we downgrading this on purpose?

MxmUrw:
> Mainly because you need to add it as a dependency wherever you want to use it and also it doesn't get the pallet name of generic pallets.

We discussed in office, conclusion:

I think there'd be a way to extract the pallet name from the type parameters (and also add any other possibly required information) and use it through the official tracing-attributes crate. But I get that the current implementation makes it a bit simpler. And for any proper in-depth investigation we'd have to rely on custom investigation specific annotations anyways.

So good to go!

MxmUrw (approved):

Cargo.lock
nmammeri: When rebasing the 3f50f6a commit carried with it the stale  5.4.1 on the top. Had to manually run `cargo update -p event-listener@5.4.1 --precise 5.4.2`

PR #6802, opened 2026-08-19
fix: report bad bitcoin RPC responses instead of decode errors (PRO-1083)

dandanlen:
# Pull Request

Closes: PRO-1083

Linear issue: https://linear.app/chainflip/issue/PRO-1083/bitcoin-call-rpc-raw-inner-function-to-better-handle-bad-responses

## Summary

`call_rpc_raw` deserialised the HTTP response straight into `Vec<serde_json::Value>` via
`reqwest::Response::json()`, so anything that isn't a JSON-RPC batch reply came back as an opaque
`Error::Transport("error decoding response body")` — the status code and the actual body were both
lost. That covers exactly the cases in the issue:

* a `502 Bad Gateway` (or similar) html page from a proxy in front of the node,
* an empty body, e.g. a `401 Unauthorized` from the node itself,
* a `400 Bad Request` where bitcoin core rejects the request as a whole and replies with a *single*
  JSON-RPC error object instead of a batch — this was decoded as an `RpcError`, failed, and turned
  into a decode error too.

The change splits transport from parsing: the body and status are read separately and handed to a
new pure `parse_response`, which

* reports anything that isn't JSON as a new `Error::BadResponse { status, body }`, with the body
  truncated to 512 bytes so a large error page doesn't flood the logs;
* unwraps a single reply object through the same path as batch replies, so a whole-request
  rejection surfaces as the `Error::Rpc` the node actually sent;
* keeps the existing behaviour for batch replies: the reply-count check and per-reply error
  extraction. A reply that isn't a JSON-RPC reply at all — not a json object, or an object carrying
  neither `result` nor `error` — is now reported as a `BadResponse` (with a warning) rather than as
  a nonsensical attempt to decode it as an `RpcError`.

### Judgement calls

* **The status is not used to decide whether a response is an error.** Bitcoin core replies with a
  non-success status *and* a well-formed JSON-RPC body for genuine RPC errors (e.g. `500` for a
  failed call), so rejecting on status alone would have replaced good error messages with worse
  ones. The body is parsed first; the status is only used for reporting.
* A reply carrying **both** a `result` and an `error` is malformed, but is reported as the error it
  names rather than as a `BadResponse`: the node's own message says more about what went wrong than
  a dump of the raw body would.
* A `result` that is present but `null` is a valid result (some methods return one) and is
  preserved; a reply carrying *neither* `result` nor `error` yields `None` from `result_from_reply`,
  leaving the caller to report it.
* The 512-byte cap applies only to a body being quoted into a `BadResponse`; a body that parses is
  never truncated.
* Making `parse_response` a free function over `(status, body)` is what makes the behaviour unit
  testable without standing up an HTTP server.

## API Changes

None. `Error` is only used within `engine/src/btc`; the new variant is additive and the only
existing `match` on it (`send_raw_transaction`, on `Error::Rpc`) is unaffected.

### RPCS

No changes to any RPC interface. This is the engine's *client* for an external bitcoin node; only
the error reported for a malformed response changes.

### Extrinsics & Events

None.

## Verification

Commands run locally on the current head (`954c656`):

* `cargo test -p chainflip-engine --lib btc::rpc::tests` — **passed**: `12 passed; 0 failed; 1 ignored`
  (the ignored one is the pre-existing `test_btc_async`, which requires a local node). This includes
  the 9 new tests: `proxy_error_page_reports_status_and_body`, `empty_response_reports_status`,
  `reported_body_is_truncated`, `rejected_request_reports_rpc_error`, `batch_replies_are_unwrapped`,
  `batch_reply_error_is_reported`, `incomplete_batch_reply_is_rejected`, `null_result_is_preserved`,
  `reply_with_neither_result_nor_error_is_rejected`.
* `cargo clippy -p chainflip-engine --all-targets -- -D warnings -A deprecated -W clippy::allow_attributes`
  (the lint flags from the `cf-clippy` alias, scoped to this package) — **passed** (exit 0).
* `cargo fmt --all --check` — **clean**.

## Not verified

* The full workspace build and test suite locally — only `chainflip-engine` was compiled and only
  its `btc::rpc` tests were run. CI covers the rest and is green (see below).
* Any test against a real or mocked bitcoin node: the new behaviour is covered only by unit tests
  over `parse_response`, not by an end-to-end request. The change to the request/response plumbing
  in `call_rpc_raw` itself (reading the body as bytes before parsing) is therefore compile-checked
  but not exercised against a live node.
* Runtime benchmarks and migrations — not applicable, no runtime code is touched.

## CI notes

Green on the current head (`954c656`) except one job, which fails for a reason outside this branch:

* **`post-check-turbo / bouncer`** — infra, fails before any test runs: the ngrok authtoken is
  rejected (`ERR_NGROK_107`), so no tunnel is created, `ci/scripts/get_ngrok_urls.py` dies with
  `IndexError: list index out of range`, and vault setup then fails with tronweb `NO_CONNECTION`.
  This is a repository-secret problem (`NGROK_AUTHTOKEN`) that will hit any PR running the turbo
  bouncer. It has now reproduced identically across three heads (2026-08-28, 2026-08-31 and
  2026-09-03), so it isn't a flake. It is not a required check — the PR is `unstable`, not
  `blocked` — so it doesn't gate this merge, but it does need the token refreshed. Re-running failed
  jobs is denied to this integration (403).

Everything else is green on this head: the other two bouncer jobs (`post-check / bouncer`,
`post-check-old-rpcs / bouncer`), both unit-test variants, clippy, fmt, license headers, coverage,
benchmarks, upgrade-check, `pre-check / audit` and `check-semantic-pr`.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:
    * [ ] Proptests for important invariants. — n/a
    * [ ] Integration tests for runtime-wide behaviours. — n/a, engine-only change
    * [ ] Bouncer tests for E2E features involving external chains. — n/a
  * [x] Benchmarks: none needed, no runtime code touched.
  * [x] Migrations: none.

Opened autonomously by a scheduled Claude Code routine, and reviewed by @dandanlen before being
marked ready.

https://claude.ai/code/session_01Et49AeAbPCABxKQ76EfvvD


engine/src/btc/rpc.rs
```diff
@@ -172,45 +184,69 @@ async fn call_rpc_raw(
...
+		_ => Err(bad_response()),
+	}
+}
+
+/// The result of a single JSON-RPC reply, or the error it reported.
+fn result_from_reply(reply: &Map<String, serde_json::Value>) -> Result<serde_json::Value, Error> {
+	match reply.get("error") {
+		Some(error) if !error.is_null() =>
+			Err(Error::Rpc(serde_json::from_value(error.clone()).map_err(Error::Json)?)),
+		_ => Ok(reply.get("result").cloned().unwrap_or_default()),
```
dandanlen: This is consistent with the old behaviour but it seems odd to return `Ok(..)` if the response contains neither `error`nor `result`.

dandanlen: Good catch — fixed in eb347ab. `result_from_reply` now returns `BadResponse` when a reply object carries neither `result` nor `error`. A present-but-`null` `result` — which some methods do return — is still preserved, and there's a new test for each case.

---

dandanlen:
`post-check-old-rpcs / bouncer` failed with an Assethub broadcast abort in the `AllSwaps` test, unrelated to this diff:

> `Broadcast 8 to Assethub was ABORTED (egress ["Assethub","5"]). All authorities failed to broadcast — likely a stuck/burned nonce.`
> `test: AllSwaps > Swap 50: HubUsdt to HubUsdc (Swap)` — [job log](https://github.com/chainflip-io/chainflip-backend/actions/runs/32387662451/job/96489376190)

Just before that, the Assethub node logs show `assets.transferKeepAlive` submissions failing with "Transaction not included within 30s of submission" and one "Transaction is invalid" — this looks like a localnet Assethub flake during test setup that then made the HubUsdt→HubUsdc swap unbroadcastable.

The diff is a pure engine-side error-parsing refactor in `engine/src/btc/rpc.rs` — it doesn't touch any Assethub, broadcast, or nonce code path. The same job (`post-check-old-rpcs / bouncer`) passed on the previous head (`8f02113`), and the other two bouncer jobs on this head (`post-check / bouncer`, `post-check-turbo / bouncer`) are green. Not pushing a change for this. Re-running failed jobs is denied to this integration (`403 Resource not accessible by integration`), so I'll leave the re-run to a human if they want it.

---

engine/src/btc/rpc.rs
```diff
@@ -172,45 +184,76 @@ async fn call_rpc_raw(
...
+/// that a `result` that is present but `null` is a valid result, which some methods do return.
+fn result_from_reply(
+	reply: &Map<String, serde_json::Value>,
+	bad_response: impl Fn() -> Error,
+) -> Result<serde_json::Value, Error> {
+	match (reply.get("error"), reply.get("result")) {
+		(Some(error), _) if !error.is_null() =>
+			Err(Error::Rpc(serde_json::from_value(error.clone()).map_err(Error::Json)?)),
+		(_, Some(result)) => Ok(result.clone()),
+		_ => Err(bad_response()),
```
dandanlen: Strictly speaking `Some(_), Some(_)` is also a bad response but would resolve as an error here. I think returning error is probably a decent compromise, at least it preserves some of the response.

engine/src/btc/rpc.rs
```diff
@@ -37,6 +37,10 @@ use anyhow::{anyhow, Result};
 // https://github.com/bitcoin/bitcoin/blob/fb7b5293844ea6adc5dcf5ad0a0c5890b4495939/src/rpc/protocol.h#L58
 const RPC_VERIFY_ALREADY_IN_CHAIN: i32 = -27;
 
+/// How much of an unexpected response body to keep for reporting. Enough to identify the source
+/// of a proxy error page without flooding the logs.
+const MAX_REPORTED_BODY_LEN: usize = 512;
```
dandanlen, line 43: Does this not risk truncating valid (but large) responses?

engine/src/btc/rpc.rs
dandanlen: Agreed, and that's the intent — a reply with both is malformed, but the node's own error message says more about what went wrong than a dump of the raw body would, so the error arm is checked first deliberately. Written down in abfc029 so the next reader doesn't have to reconstruct it.

---

engine/src/btc/rpc.rs
dandanlen: No — the cap only applies to the body quoted inside a `BadResponse` error. Responses are parsed from the full slice (`serde_json::from_slice::<Value>(body)`), and the truncation happens in the `bad_response` closure, which is only ever reached for a body that failed to parse or a reply that isn't a JSON-RPC reply. Nothing that gets returned to a caller is truncated, so a large valid response is unaffected.

The doc comment on the constant said "unexpected response body", which clearly wasn't explicit enough — reworded in 0d1fabb to say the cap applies only to a body that failed to parse, never to one we return.

---

dandanlen:
`post-check-turbo / bouncer` failed on this head ([job log](https://github.com/chainflip-io/chainflip-backend/actions/runs/33162486040/job/98822052427)), and it's an infra failure that happens before any test runs — the ngrok authtoken is rejected:

```
ERROR:  authentication failed: The authtoken you specified is properly formed, but it is invalid.
ERROR:  ERR_NGROK_107
```

With no tunnel established, `ci/scripts/get_ngrok_urls.py` then dies with `IndexError: list index out of range` (exit 1), and vault setup fails right after with tronweb `error: 'NO_CONNECTION'` (exit 255) while inserting the Tron key.

This is a repository-secret problem (`NGROK_AUTHTOKEN`), not something this branch can fix — and worth a look independently of this PR, since it will hit any PR that runs the turbo bouncer. The two commits on this head (`abfc029`, `0d1fabb`) are doc comments only, so nothing here can have caused it.

Re-running failed jobs is still denied to this integration (`403 Resource not accessible by integration`), so a human will need to kick it once the token is refreshed. I'll keep watching the PR — `post-check / bouncer` and `post-check-old-rpcs / bouncer` are still running on this head, and I'll report where they land.

---

marcellorigotti (approved):

engine/src/btc/rpc.rs
```diff
@@ -172,45 +185,79 @@ async fn call_rpc_raw(
...
+/// The result of a single JSON-RPC reply, or the error it reported.
+///
+/// A reply carrying neither is not a JSON-RPC reply, so it is reported through `bad_response`. Note
+/// that a `result` that is present but `null` is a valid result, which some methods do return.
+///
+/// A reply carrying *both* is malformed too, but is reported as the error it names rather than as a
+/// `bad_response`: the node's own message says more about what went wrong than the raw body does.
+fn result_from_reply(
+	reply: &Map<String, serde_json::Value>,
+	bad_response: impl Fn() -> Error,
```
j4m1ef0rd: Seems unnecessary to give this function bad_response only for it to just hand it back to the caller with no added information. Would be better to just return an Option<Result<...>> instead.

j4m1ef0rd (approved):

engine/src/btc/rpc.rs
dandanlen: Agreed — done in 954c656. `result_from_reply` now returns `Option<Result<serde_json::Value, Error>>` and takes only the reply; `None` means "not a JSON-RPC reply", and the caller decides how to report it, which it already did for a reply that isn't a json object at all. That also let the two arms in `parse_response` collapse onto the same `bad_response()` path, so the warning now says "was not a valid json-rpc reply" — accurate for both cases, where before it claimed "not a valid json object".

Verified on the new head: `cargo test -p chainflip-engine --lib btc::rpc::tests` (12 passed, 1 ignored), clippy with the `cf-clippy` lint flags scoped to this package, and `cargo fmt --all --check`.

---

j4m1ef0rd (approved):

PR #6804, opened 2026-08-20
fix: Bouncer, de-flake boost test

j4m1ef0rd:
### What
Raise Bitcoin's `WitnessSafetyMargin` to 6 for the duration of the `BoostingForAsset` test (restored afterwards), instead of tuning any chain's block time.

### Why
The boost test is timing-sensitive: a deposit is boosted only if its block is pre-witnessed before it ages past the ingress safety margin. If the tracked Bitcoin height advances several blocks in one processing step — possible when the Bitcoin block time is ≤ the state chain's (e.g. turbo) or on a transient witnessing hiccup — the prewitness and finalise events collapse into the same step and the boost is dropped in favour of finalisation. A larger safety margin gives pre-witnessing enough headroom, so the test no longer depends on chain block-time configuration.

Per the [review discussion](https://github.com/chainflip-io/chainflip-backend/pull/6804): preferred over changing block times, and over retrying deposits (which could mask a real boost regression).

Note: the runtime syncs `WitnessSafetyMargin` into the ingress block-witnesser election settings every block, so it's the authoritative knob (the existing `setup_elections.ts` write to the election settings directly is overwritten by that sync — worth a separate cleanup).

### DIFF
--- a/bouncer/tests/boost.ts
+++ b/bouncer/tests/boost.ts
@@ -34,6 +34,23 @@ import { bitcoinIngressEgressInsufficientBoostLiquidityEvent } from 'generated/e
 import { submitGovernanceExtrinsic } from 'shared/cf_governance';
 import { boostPoolFee } from 'shared/setup_boost_pools';
 
+// A generous Bitcoin ingress safety margin applied only for the boost test. Boosting is timing-
+// sensitive: a deposit is boosted only if its block is pre-witnessed before it
+// ages past the ingress safety margin. When both fire in the same processing step the boost is
+// dropped in favour of finalisation. That happens whenever the tracked Bitcoin height advances
+// several blocks at once — which is possible when the Bitcoin block time is at or below the state
+// chain's and combined with a transient witnessing hiccup.
+const BOOST_TEST_BTC_SAFETY_MARGIN = 6n;
+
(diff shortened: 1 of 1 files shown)


MxmUrw (commented):
Interesting, good catch!

Regarding how to fix this: it would be nicer if we could change the test such that it doesn't depend on timing anymore instead of trying to change timing constants to make it pass... Ideally, the full bouncer suite should always pass independently of the block time configurations of *any* of the chains.

Could we, for example, retry another deposit if the first one is ingressed without being boosted?

j4m1ef0rd:
> Could we, for example, retry another deposit if the first one is ingressed without being boosted?

This was my first instinct too. But I'm not so sure its a good idea. Because there is no way of knowing why the boost didn't happen. So retrying the test because it failed could hide an actual intermittent failure with boost. I guess you could assume that if the bouncer didn't see a prewitness event, then it must of been because of slow witnessing.

One other option would be to set the `WitnessSafetyMargin` of bitcoin to 6 blocks just for the boost test. For some reason I thought this was a const and ruled it out as a option.

MxmUrw:
> One other option would be to set the WitnessSafetyMargin of bitcoin to 6 blocks just for the boost test. For some reason I thought this was a const and ruled it out as a option.

Haha yeah I also didn't think about it for some reason. Yes I like this option more.

MxmUrw (approved):

PR #6805, opened 2026-08-20
feat: PRO-2996 add `cf_withdrawal_restrictions` rpc and cli command

dandanlen:
# Pull Request

Closes: [PRO-2996](https://linear.app/chainflip/issue/PRO-2996/add-lp-cli-command-to-read-the-state-of-withdrawal-whitelist)

## Summary

PRO-2872 gave LPs a way to *update* their withdrawal whitelist, but no way to read it back. This adds an RPC that answers "what restricts where this account can withdraw to?", and a CLI command that consumes it.

Scope grew twice in review, both times for good reason:

1. From reading storage in the CLI to a **node RPC**, so LP tooling and the frontends can read this too.
2. From whitelist-only to **all the restrictions** `ensure_withdrawal_allowed_to` actually applies, and from an LP-scoped CLI command to a general one, since brokers have restrictions too.

`cf_withdrawal_restrictions(account_id, at)` returns:

```json
{
  "account_role": "LiquidityProvider",
  "whitelist": { "timelock_secs": 86400, "allowed": [ … ] },
  "pending": [ … ],
  "refund_addresses": [ { "chain": "Ethereum", "address": "0xcf12…abcd" } ],
  "bound_broker_withdrawal_address": null
}
```

Approach and judgement calls:

- **Why all three sources.** The whitelist alone can't answer the question: a registered refund address is allowed *without* being whitelisted, and a bound broker withdrawal address rejects every other Ethereum destination *before* the whitelist is consulted. Reporting only the whitelist meant the CLI printed claims that were sometimes false.
- **The broker binding is reported separately** from the allowed destinations. It's a restriction that overrides the whitelist, so listing it as an allowance would misrepresent it.
- **`account_role` is included** — it determines which restrictions can apply, and saves callers a second `cf_account_info` round-trip.
- **Renamed rather than extended.** "Whitelist" stops being an honest name once the response carries the other two sources.
- **Runtime API versioning**: `#[changed_in(21)]` plus version dispatch in `custom-rpc` that returns `UnsupportedRuntimeApiVersion` below 21. Note main independently bumped `CustomRuntimeApi` to 21 for `cf_penalties`/`cf_suspensions`; this method now shares that version, which is fine — one version can cover several changes. There's an **open question** on the thread about whether to return a default unrestricted value instead of an error for older blocks.
- **CLI command is top-level**: `chainflip-cli get-withdrawal-restrictions`, not under `lp`.
- `WithdrawalWhitelist::{timelock, external, internal}` are made `pub` so the runtime API impl can read the entries; the struct's fields stay private.

### RPCS

New: `cf_withdrawal_restrictions(account_id, at)`. Additive; no existing method changes.

`bouncer/generated/chaintypes` regenerated. `WhitelistDestinationRpc` gains `PartialEq, Eq` derives — additive, no wire change.

### Extrinsics & Events

None. No extrinsic, event, or storage changes — this is read-only. Event schemas regenerate byte-identical.

CLI surface: one new top-level command, `get-withdrawal-restrictions`.

## Verification

Exercised end-to-end against a dev node built from this branch (the `chainflip-backend-bin` artifact from this PR's CI), **including the non-empty path**:

- `rpc_methods` lists `cf_withdrawal_restrictions`; the old name is gone.
- Registered a real LRA via `lp register-liquidity-refund-address`, then saw it come back through the runtime API, the RPC (`"refund_addresses":[{"chain":"Ethereum","address":"0xcf12…abcd"}]`) and the CLI.
- Role dispatch checked against both an LP account and a broker account.
- `get-withdrawal-restrictions` runs and prints correctly, which also covers the clap wiring for the renamed command.

| Command | Result |
| --- | --- |
| `cargo clippy` (CI's flags) over the touched crates | passes |
| `cargo test -p custom-rpc` | 62 passed, 0 failed (post-merge, includes main's tests and the updated snapshot) |
| `cargo test -p pallet-cf-asset-balances --lib whitelist::` | 27 passed, 0 failed |
| `cargo fmt --all` | applied |
| `generate_chaintypes.ts` / `generate_event_schemas.ts` | run against the node; chaintypes committed, event schemas unchanged |

## Not verified

- An account with **whitelist entries** and one with a **pending timelocked change** are covered by the snapshot test, not against a chain. The refund-address path *is* now covered live.
- The chaintypes were regenerated before the latest main merge, which brought API changes of its own; if CI's regeneration disagrees I'll push an update.
- No integration, bouncer, or benchmark runs locally.
- `post-check-turbo / bouncer` has failed intermittently on this branch (4 of 5 runs) at an identical point in `setup_vaults` before any test runs, including on a commit that changed only `println!` strings. Unrelated to this change; see the comment thread.


dandanlen:
CI status on `703dfd8` — two red checks, neither of which I'm pushing a fix for. Details so a reviewer doesn't have to re-derive them:

**`post-check-turbo / bouncer` — failed, not caused by this change.**

It died 26s into the test step, inside `setup_vaults` (localnet bring-up), before any test ran. The bouncer log artifact ends right after "Arbitrum key inserted" and the Solana key-insertion tx landing at slot 566; the localnet logs show no errors — `solana-test-validator` was producing slots normally and the engine was voting on elections at the moment of death.

The same test suite on the same commit passed in both other variants: `post-check / bouncer` ✅ (21 min) and `post-check-old-rpcs / bouncer` ✅. And this diff cannot reach vault setup — it adds three `pub` getters on a struct, a `QueryApi` method nothing in the runtime calls, and a CLI subcommand. Worth a re-run when someone with the permission gets to it.

**`check-semantic-pr` — stale.**

It ran against the original PR title, which lacked a conventional-commit prefix. The title has since been corrected to `feat: PRO-2996 ...`, but that workflow is `on: pull_request` without `edited` in its types, so it only re-runs on push. It should go green on the next commit to this branch.

Everything else is green, including `cargo-clippy`, `cargo-fmt`, both unit-test jobs, coverage, `build-try-runtime`, `test-benchmarks`, chainspec compatibility on all three networks, and the upgrade test.

---

dandanlen (commented):
I'm not sure I like this solution. I would rather we add an rpc call and then route the cli command through that.

dandanlen:
That's fair, and the argument for it is stronger than taste: reading storage from `QueryApi` only helps someone running the CLI against a node. An RPC makes the whitelist readable by LP tooling and the frontends too, which is probably where an LP actually wants to see it.

There are two shapes for this in the codebase and they cost very different amounts, so I'd rather you pick than guess:

**1. Node custom RPC — `cf_withdrawal_whitelist(account_id, at)`.** Same shape as `cf_free_balances`: implemented in `impl_api.rs`, and both the CLI and `chainflip-lp-api` call it through `raw_rpc_client`. Per the header comment in `custom_api.rs` this means bumping `#[api_version(20)]` to 21, adding the dummy method annotated `#[changed_in(21)]`, handling it in the version dispatch in `custom-rpc/src/lib.rs`, an rpc type plus a serialization snapshot test, and probably regenerating the bouncer typegen. This is the general answer and the one I'd lean towards.

**2. LP API method — `lp_get_whitelist`.** Same shape as `lp_get_open_swap_channels`: the read stays in `QueryApi` (roughly what's here now) and `chainflip-lp-api` exposes it, with the CLI sharing the code path. No runtime-API versioning, much smaller diff — but it only serves LP-API consumers, not arbitrary RPC users.

Option 1 puts this well past the 1-point estimate on [PRO-2996](https://linear.app/chainflip/issue/PRO-2996), so that ticket probably wants re-pointing if you want it done that way.

Happy to implement either — say which and I'll rework the branch. Leaving it as-is meanwhile rather than pushing a versioned-API change on a guess.

---

dandanlen:
Do Option 1. I've re-weighted the issue accordingly.

dandanlen:
Second data point on `post-check-turbo / bouncer`, since it failed again on `6580277`: it is the **same failure, in the same place**, and it now looks like a property of that variant rather than of this branch.

| | `703dfd8` (CLI-only diff) | `6580277` (runtime + RPC diff) |
|---|---|---|
| `post-check-turbo / bouncer` | ❌ died 26s into step 17 | ❌ died 25s into step 17 |
| `post-check / bouncer` | ✅ | ✅ |
| `post-check-old-rpcs / bouncer` | ✅ | ✅ |

Both turbo runs stop at the identical point in `setup_vaults` — last lines are the Solana aggregate-key funding and `Arbitrum key inserted`, before any test executes. Both bouncer log artifacts are ~2.9KB where a passing run's is ~320KB. In both, the localnet logs show no errors: `solana-test-validator` producing slots normally, engine voting on elections.

Two very different diffs producing a byte-identical failure signature, while the other two variants pass on the same commits, points at the turbo localnet setup rather than at either change. Possibly one for [PRO-2510](https://linear.app/chainflip/issue/PRO-2510) — happy to open a separate ticket with these artifacts if that's useful, but I don't think it should block this PR.

Everything else on `6580277` is green: clippy, fmt, audit, `check-generated-bouncer-dependencies`, both unit-test jobs, coverage, test-benchmarks, chainspec compatibility on all three networks, the upgrade test, packaging and the docker builds.

---

state-chain/custom-rpc/src/lib.rs
```diff
@@ -836,6 +837,85 @@ pub use ingress_egress_tracker::{
...
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Debug, Clone)]
+pub enum RpcWhitelistUpdate {
+	Allow(RpcWhitelistDestination),
+	Remove(RpcWhitelistDestination),
+	Timelock(u64),
+}
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Debug, Clone)]
+pub enum RpcWhitelistDestination {
```
msgmaxim: Is this a duplicate of an existing `WhitelistDestinationRpc`?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1886,6 +1972,27 @@ where
...
+	fn cf_withdrawal_whitelist(
+		&self,
+		account_id: state_chain_runtime::AccountId,
+		at: Option<state_chain_runtime::Hash>,
+	) -> RpcResult<RpcWithdrawalWhitelist> {
+		self.rpc_backend.with_versioned_runtime_api(at, |api, hash, version| {
+			// The withdrawal whitelist itself only exists from version 21 onwards, so there is
+			// nothing sensible to return for older blocks.
+			if version < 21 {
+				Err(CfApiError::ErrorObject(call_error(
```
msgmaxim, line 2035: Could alternatively just return some default value (unrestricted whitelist).

state-chain/custom-rpc/src/lib.rs
dandanlen: Yes — it was, thanks. Fixed in `a5e59a6`: the response now uses the existing `WhitelistDestinationRpc`.

It needed no new dependency (`cf_rpc_apis::lp` already re-exports `cf_rpc_types::lp::*`, and custom-rpc already depends on `cf-rpc-apis`), and it's the better type anyway — it carries `AddressString` rather than the bare `String` I'd used, matching the rest of custom-rpc. The only change on that side was adding `PartialEq, Eq` derives, which the response types need. The `From` became a free function since both types are now foreign to custom-rpc.

The snapshot test passes against the **unchanged** committed snapshot, so the wire shape is byte-identical — `AddressString` is a newtype over `String` and serialises the same.

I kept `RpcWhitelistUpdate` separate from `WhitelistChangeRpc`: they mirror two different pallet enums, and `WhitelistUpdate` has the `Timelock` variant that `WhitelistChange` doesn't. Happy to collapse them if you'd rather.

---

state-chain/custom-rpc/src/lib.rs
dandanlen: It hinges on one thing I can't establish from here: **can a block exist with `api_version` 20 and a non-empty whitelist?**

- If no — the whitelist ships in the same runtime upgrade as `api_version` 21 — then "unrestricted" is factually true for every older block, and your suggestion is strictly nicer: callers don't have to special-case an error for historical queries.
- If yes — whitelists went out in an earlier upgrade and only this RPC is new — then returning the default would report "withdrawals to any destination are allowed" for an account that actually has restrictions. That's the case I was trying to avoid: a wrong answer is worse than an explicit "can't answer".

What I can see in-repo points at the first case: `refund_addresses_to_asset_balances` is still in the pending `UpgradeSet` and its own docs describe refund addresses as feeding the withdrawal whitelist, which suggests the feature hasn't been through a runtime upgrade yet. But this container has a shallow clone (59 commits), so I can't check the release history properly, and you'd know it directly.

Left as the error for now, since that's the safe direction if I've got it backwards. If you confirm whitelists can't predate version 21, say so and I'll switch it to the default — it's a one-line change and the type already has the right shape for it (`active: None`, `pending: []`).

---

api/bin/chainflip-cli/src/main.rs
```diff
@@ -435,6 +442,53 @@ async fn bind_executor_address(api: Arc<impl OperatorApi + Sync>, eth_address: &
...
+
+async fn get_withdrawal_whitelist(api: QueryApi) -> Result<()> {
+	let RpcWithdrawalWhitelist { active, pending } =
+		api.get_withdrawal_whitelist(None, None).await?;
+
+	match active {
+		Some(RpcActiveWithdrawalWhitelist { timelock_secs, allowed }) => {
+			println!("Withdrawal whitelist timelock: {timelock_secs} seconds.");
+			if allowed.is_empty() {
+				println!("No destinations are whitelisted, so withdrawals are blocked.");
```
msgmaxim: This would be misleading since it doesnt take refund addresses into account. We need to decide whether we want to simply report the state of the whitelist (and not make false generalisations) or a comprehensive report of what's allowed taking refund addresses (and probably broker restricted addreses) into account.

dandanlen: Agreed, and it's worse than the line you flagged — I read `ensure_withdrawal_allowed_to` again and **all three** of my messages were making claims the whitelist can't support:

1. `"No destinations are whitelisted, so withdrawals are blocked"` — false when a refund address is registered; that address stays allowed.
2. `"Withdrawals are allowed to: <list>"` — incomplete, same reason.
3. `"...no withdrawal whitelist: withdrawals to any destination are allowed"` — false when `BoundBrokerWithdrawalAddress` is set. That check runs *before* the whitelist and rejects any Ethereum address other than the bound one, so "any destination" is wrong regardless of whitelist state. Your "probably broker restricted addresses" is confirmed.

I've taken option 1 in `b0d1d1e` — report the whitelist, don't generalise:

```
Withdrawal whitelist timelock: 0 seconds.
Whitelisted destinations:
  Ethereum address 0xcfcf…

Note: this reports the whitelist only. A registered refund address is also allowed without
being whitelisted, and a bound broker withdrawal address blocks Ethereum withdrawals to
anywhere else, whitelisted or not.
```

I stopped there rather than building option 2, for two reasons. It needs data the RPC doesn't return yet — `RefundAddresses` (per chain) and `BoundBrokerWithdrawalAddress` — so it grows both the runtime API and this PR, which has already outgrown its original point. And option 1 is a prerequisite either way: whatever we decide, the CLI shouldn't be asserting things that aren't true in the meantime.

If you want option 2, I'd suggest it as a follow-up, and I'd argue it belongs in the RPC rather than assembled client-side — an "what's allowed" view computed where `ensure_withdrawal_allowed_to` lives, so the answer can't drift from the rule. Happy to do it in this PR instead if you'd rather have it in one go — say which and I'll pick it up.

---

dandanlen:
Correction to my earlier comment about `post-check-turbo / bouncer`: **it passed on `b0d1d1e`**, and the run is fully green.

`b0d1d1e` changes `println!` strings and nothing else, so it can't have affected localnet vault setup. That means the right description is **intermittent**, not "broken" as I characterised it after three consecutive identical failures — please don't file a ticket claiming it fails deterministically on the strength of my earlier table.

What the four runs actually support:

| commit | diff | turbo bouncer |
|---|---|---|
| `703dfd8` | CLI only | ❌ |
| `6580277` | runtime + RPC | ❌ |
| `a5e59a6` | rpc type reuse | ❌ |
| `b0d1d1e` | println strings only | ✅ |

The conclusion that holds up is the one that matters for review: the failures were never caused by this branch — three unrelated diffs failed at an identical point in `setup_vaults` before any test ran, and the run that flipped it green changed no executable behaviour. A ~75% intermittent failure rate on that variant still looks worth someone's attention, but as flakiness ([PRO-2510](https://linear.app/chainflip/issue/PRO-2510) territory) rather than a hard break.

Everything on `b0d1d1e` is green: clippy, fmt, audit, both unit-test jobs, coverage, `check-generated-bouncer-dependencies`, all three bouncer variants, test-benchmarks, chainspec compatibility, the upgrade test, packaging and docker.

---

api/bin/chainflip-cli/src/main.rs
msgmaxim: I wonder if the real @dandanlen has an opinion here.

dandanlen: I think it makes sense if the RPC does more than just reading storage: it should do what is non-trivial for the user (aka. "it should earn its keep" in AI parlance). In this case, that includes listing the emergency withdrawal address / broker redeem address depending on the account role. 

We should adapt the RPC: it should not be `lp_` scoped, it should be a general `cf_` rpc (since it will also serve brokers). I think it makes sense to expand the scope of this PR to encompass this. I don't want us to add an RPC that will be superseded immediately.

dandanlen: Done — expanded in `f5e8c3b`, with the chaintypes and a main merge in `50d2796`.

`cf_withdrawal_whitelist` is now **`cf_withdrawal_restrictions`**, returning everything `ensure_withdrawal_allowed_to` actually consults:

```json
{
  "account_role": "LiquidityProvider",
  "whitelist": { "timelock_secs": 86400, "allowed": [ … ] },
  "pending": [ … ],
  "refund_addresses": [ { "chain": "Ethereum", "address": "0xcf12…abcd" } ],
  "bound_broker_withdrawal_address": null
}
```

On the `lp_` scoping: the RPC itself was already `cf_`-prefixed, so I read your point as being about the surface around it. What was LP-scoped was the **CLI command**, which is now top-level `chainflip-cli get-withdrawal-restrictions` rather than `lp get-whitelist`, since a broker needs it too. Shout if you meant something else.

Judgement calls worth checking:

- **Renamed rather than added a field.** "Whitelist" would have been a lie once the response carries refund addresses and the broker binding. `api_version` 21 is new in this PR and unreleased, so the rename is free — no second version bump, and the `#[changed_in(21)]` dummy just moves with it.
- **`account_role` is in the response.** It's what decides which restrictions can apply, and it saves every caller a second `cf_account_info` round-trip to render this sensibly.
- **The broker binding is reported separately, not merged into the allowed list.** It's a restriction that *overrides* the whitelist for Ethereum, so folding it in with the allowances would misrepresent it.

Verified against a dev node built from this branch, including the non-empty path that had only been snapshot-tested until now: registered an LRA via `lp register-liquidity-refund-address`, then saw it come back through the runtime API, the RPC and the CLI. Role dispatch checked on both an LP and a broker account. Clippy with CI's flags, the pallet tests and the snapshot test all pass; chaintypes regenerated and stable.

One thing to flag: the first regeneration picked up `broadcastAttemptCount` from main, because CI builds the artifact from the PR *merge* ref. Rather than carry someone else's generated code in this diff I merged main properly — it arrived with its own regenerated `query.d.ts`, so the branch is now up to date and my diff only contains my three files.

The open question above about pre-version-21 blocks is unaffected by all this and still needs a call.

---

j4m1ef0rd (approved):

PR #6807, opened 2026-08-21
fix: only treat AlreadyImported as broadcast-submit success on Assethub

j4m1ef0rd:
Closes: PRO-3071

### What
Narrows the Assethub/Polkadot broadcast submitter to swallow **only** `1013 POOL_ALREADY_IMPORTED` as a success. [PR #6741](https://github.com/chainflip-io/chainflip-backend/pull/6741) also swallowed `1012 POOL_TEMPORARILY_BANNED` and `1014 POOL_TOO_LOW_PRIORITY`; this reverts those two so they fall through to the normal error path.

### Why
Only `1013` actually guarantees the pool holds our *exact* transaction and will include it, so treating it as a no-op success is correct. The other two don't carry that guarantee (verified against the pinned polkadot-sdk source):
- **`1012 "Transaction is temporarily banned"`** — the hash is in the pool's *ban list* (`PoolRotator`), i.e. the pool is actively *refusing* it. Substrate itself flags it `is_retriable() == true` ("not currently actionable, retry later"). A tx banned after being *dropped/invalidated* (not included) will never land until the ban expires (default **30 min**).
- **`1014 "Priority is too low: (old vs new)"`** — the tx was *rejected* on a priority/space contest (pool full, or a different tx already holds that nonce). Our tx may not be in the pool at all.

Swallowing them as success can report a broadcast that never lands, waiting on a witness that never arrives.

### Follow-up
Once this is deployed, Assethub's `BroadcastTimeout` can be restored to its normal value.


dandanlen:
> Once this is deployed, Assethub's BroadcastTimeout can be restored to its normal value — it was extended to absorb the slow phantom-success recovery, which no longer happens now that these submissions fail fast.

What we actually did was *reduce* the timeout by about 1/2. 

---

It would be good to dig a bit deeper and find out *why* we are seeing these 1012/1013/1014 errors. 

What I would assume is something like this:

### 1012 Temporarily banned 

=> We are trying to submit a transaction that expired

The only scenario I can imagine is that one node submits successfully, but witnessing is stalled or finality is lagging. Then we would not witness success before the timeout, and we would request the next node to submit. *However* I also think that our transactions are immortal so I have no idea how we would be reaching expiry.

### 1013 Already in Pool

=> This is the only one that makes sense to me

This one is saying "this exact transaction is already in the pool". This is what I would expect if there's a duplicate submission. 

### 1014 Priority too low

=> This is an odd one

This suggests that we are trying to submit a *different* transaction at the same nonce. This shouldn't be possible under our protocol rules: the transaction is built and signed once. The submitter simply forwards it unaltered. So I'm surprised that we would ever even see this. 



We might also want to consider handling 1010 (Nonce Stale): this is what we should see when we submit a transaction that is already included in a block.


TLDR: I would not expect us to ever see 1012 or 1014 errors so I'm ok with *not* treating these as silent success outcomes. But I'd be surprised if this changes the behaviour. We should go through our logs / request some from operators to get some more info.

dandanlen (approved):

PR #6808, opened 2026-08-21
feat: ensure broadcasts are retried a minimum number of times

dandanlen:
# Pull Request

Related to: PRO-3071

## Summary

Adds `const MIN_RETRY_ATTEMPTS = 10` to the broadcast pallet and changes the retry behaviour thus:

We retry using all validators, and if we haven't reached the minimum number of attempts, we clear the failed validators list and retry with all validators. 

This required introducing a new storage item to track the total attempt count, and updating the signer nomination mock.

### DIFF
--- a/bouncer/generated/chaintypes/chainflip-node/query.d.ts
+++ b/bouncer/generated/chaintypes/chainflip-node/query.d.ts
@@ -2613,6 +2613,16 @@ export interface ChainStorage extends GenericChainStorage {
     failedBroadcasters: GenericStorageQuery<(arg: number) => Array<AccountId32>, number>;
 
     /**
+     * The number of failed attempts for each broadcast. This is *not* the same as the size of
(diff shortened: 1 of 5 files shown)


MxmUrw (approved):

j4m1ef0rd (approved):

j4m1ef0rd:
The Insufficient Gas CCM tests deliberately create a CCM whose gas budget can't cover the call, and wait for EthereumBroadcaster.BroadcastAborted. Now that a solo-node retries 10 times, it's hitting the timeout on the test. Not too sure what to do about that..

dandanlen:
Yes. I'm going to make it configurable by governance.

dandanlen:
Actually, I changed my mind. It was a much smaller change to just update the ccm test. Now it doesn't wait for an aborted broadcast, it waits for the first retry. 

The full aborted flow is checked in the unit tests.

PR #6809, opened 2026-08-21
feat: auto deregister lp when redeem all

nmammeri:
# Pull Request

Closes: PRO-3075

## Summary

* Redeeming an account's entire FLIP balance required the account to already be unregistered (AccountMustBeUnregistered), so LPs had to call deregister_lp_account first. This PR makes it easier by auto-deregistering
* Only implemented for LPs because it is the source of most reported issues
* This is a first PR in a series of account lifecycle improvements
*   Moved LP storage cleanup (refund addresses, pending withdrawal changes) out of deregister_lp_account and into the runtime's on_deregistered hook, so every deregistration path — explicit or automatic — clears the same state.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/state-chain/pallets/cf-funding/src/lib.rs
+++ b/state-chain/pallets/cf-funding/src/lib.rs
@@ -34,7 +34,7 @@ pub use weights::WeightInfo;
 mod tests;
 
 use cf_chains::{evm::Address as EthereumAddress, RegisterRedemption};
-use cf_primitives::{chains::assets::eth::Asset as EthAsset, AssetAmount};
+use cf_primitives::{chains::assets::eth::Asset as EthAsset, AccountRole, AssetAmount};
 use cf_traits::{
 	impl_pallet_safe_mode, AccountInfo, AccountRoleRegistry, Broadcaster, Chainflip, FeePayment,
 	FundAccount, Funding, FundingSource, GetMinimumFunding, MoveFlipToGateway, RedemptionCheck,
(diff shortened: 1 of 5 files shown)


state-chain/pallets/cf-funding/src/lib.rs
```diff
@@ -730,6 +744,8 @@ pub mod pallet {
 					expiry_time: contract_expiry,
 				});
 			} else {
+				// The fee may have emptied the account, in which case it needs reaping here
+				Self::kill_account_if_zero_balance(&account_id);
```
dandanlen: Do we really need this? Is this in case the fee happened to be *exactly* the same as the balance?

ramizhasan111, line 748: as discussed, we should delete this line. The account is killed when the redemption is successfully executed.

nmammeri: Actually I looked into it again, because the amount is 0 it seems that no `PendingRedemptions` is inserted and we deposit `RedemptionAmountZero` event instead. In this case `redeemed()` is not called at all and since `redeemed()`  rely on `PendingRedemptions` storage, we end up with an account that has a balance of 0 but not deleted.

ramizhasan111 (approved):

dandanlen (approved):

PR #6810, opened 2026-08-21
feat: Include Chain in FailedToBroadcastTransaction

marcellorigotti:
# Pull Request

Closes: PRO-2700

## Summary

Include `ForeignChain` in `FailedToBroadcastTransaction` penalty.
This is a quality of life improvement to help node operators easily detect which chain their validators are having problem to submit transaction to, and also it simplifies debugging this problem from our side.

Also added backwards compatibility for monitoring RPCs which was missing.

### RPCS

Rpcs returning the offence have been updated to be backwards compatible, since the previous penalty was general for every chain I opted for defaulting every `FailedToBroadcastTransaction` to `FailedToBroadcastTransaction(Ethereum)` in case an RPC is querying for historical values.
I don't think these RPCs will ever be used for historical queries but if that can be a problem maybe it can be changed (not sure how tho).

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/state-chain/custom-rpc/src/lib.rs
+++ b/state-chain/custom-rpc/src/lib.rs
@@ -97,15 +97,16 @@ use state_chain_runtime::{
 		custom_api::CustomRuntimeApi,
 		elections_api::ElectoralRuntimeApi,
 		types::{
-			AuctionState, BoostPoolDepth, BrokerInfo, CcmData, ChainAccounts, DelegationSnapshot,
-			DispatchErrorWithMessage, EncodedNonNativeCall, EncodedNonNativeCallGeneric,
-			EncodingType, EvmCallDetails, FailingWitnessValidators, FeeTypes, IngressEvents,
-			LendingPosition, LiquidityProviderBoostPoolInfo, LiquidityProviderInfo, NetworkFees,
-			NonceOrAccount, OpenedDepositChannels, OperatorInfo, RewardDistributionEstimate,
-			RpcAccountInfoCommonItems, RpcLendingConfig, RpcLendingPool, RuntimeApiAccountInfo,
-			RuntimeApiPenalty, ShouldSweep, SimulateSwapAdditionalOrder, SimulatedSwapInformation,
-			TradingStrategyInfo, TradingStrategyLimits, TransactionScreeningEvents, ValidatorInfo,
-			VaultAddresses, VaultSwapDetails,
+			before_version_21, AuctionState, BoostPoolDepth, BrokerInfo, CcmData, ChainAccounts,
+			DelegationSnapshot, DispatchErrorWithMessage, EncodedNonNativeCall,
+			EncodedNonNativeCallGeneric, EncodingType, EvmCallDetails, FailingWitnessValidators,
+			FeeTypes, IngressEvents, LendingPosition, LiquidityProviderBoostPoolInfo,
+			LiquidityProviderInfo, NetworkFees, NonceOrAccount, OpenedDepositChannels,
+			OperatorInfo, RewardDistributionEstimate, RpcAccountInfoCommonItems, RpcLendingConfig,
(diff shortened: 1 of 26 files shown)


state-chain/pallets/cf-broadcast/src/lib.rs
```diff
@@ -78,7 +78,7 @@ pub type AttemptCount = u32;
 	MaxEncodedLen,
 )]
 pub enum PalletOffence {
-	FailedToBroadcastTransaction,
+	FailedToBroadcastTransaction(ForeignChain),
```
MxmUrw, line 81: Could we alternatively make this an `Option`?

marcellorigotti: You mean to solve the problem of backwards compatibility? yeah that's an option, although it feels a bit weird since we always have a chain we are broadcasting to.

IMO it is fine to hardcode it to `Ethereum` since I don't expect these RPC to be used for historical reason. Another option would be to add new RPCs altogether smth like `_v2` and deprecate the old one, but not sure I like this option either

state-chain/runtime/src/chainflip/signer_nomination.rs
```diff
@@ -84,17 +84,17 @@ fn eligible_authorities(
 ///
 /// Signers serving a suspension for any of the offences in ExclusionOffences are
 /// excluded from being nominated.
-pub struct RandomSignerNomination;
+pub struct RandomSignerNomination<C = ()>(PhantomData<C>);
```
dandanlen: nit: it might have been less invasive if we added the constraint only to the nominate_broadcaster method rather than to the whole trait.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1795,6 +1783,47 @@ where
 		cf_auction_state() -> RpcAuctionState [map: Into::into],
 	}
 
+	fn cf_penalties(
```
dandanlen, line 1786: Since cf_penalties never took a hash argument, it's arguable that we can keep it this way (ie. it's an RPC that is for current state not historical state) and therefore we don't need to support historical queries. Do we know how this is used?

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1795,6 +1783,47 @@ where
...
+							RuntimeApiPenalty { reputation_points, suspension_duration_blocks },
+						)| {
+							(offence, RpcPenalty { reputation_points, suspension_duration_blocks })
+						},
+					)
+					.collect()
+			})
+	}
+
+	fn cf_suspensions(&self, at: Option<state_chain_runtime::Hash>) -> RpcResult<RpcSuspensions> {
```
dandanlen, line 1815: Same for suspensions.

WDYT?

dandanlen (commented):
Some minor suggestions. 

FWIW as a simple alternative in the migrations we could just forgive all existing offences...

state-chain/pallets/cf-broadcast/src/lib.rs
dandanlen: See my suggestion re backwards compatibility - I'm not sure it's very important.

marcellorigotti: I am all down to not keep it since also imo it is not useful, it is mainly useful for monitoring, so that we don't have to do remember to not update the binaries until the runtime upgrade is released.

Nevertheless removing this would mean removing a bunch of code which effectively won't be used anymore after the upgrade, so not a big deal for me

dandanlen: Ok I missed that this was for monitoring. In that case I think it makes sense to handle it.

dandanlen (approved):

PR #6811, opened 2026-08-24
fix: bouncer, recover landed-but-lost Assethub extrinsics instead of failing

j4m1ef0rd:
### What
In bouncer, On any Assethub submit error, if the pinned nonce has already been consumed on-chain, stop resubmitting and recover the result (including any expected event) from block history instead of failing.

### Why
A deposit/transfer can be reorged out of the block it landed in right as the inclusion deadline fires; the pinned-nonce resubmission then bounces `1010 Invalid Transaction: Transaction is outdated` (Stale) and the whole submission is reported as failed even though the original landed. Seen as a spurious `HubDot → ArbUsdt` swap failure on a solo-node CI run. Resubmitting can never recover this — the real tx already finalized in a past block — so we look it up and read its event from that block.


dandanlen (approved):

j4m1ef0rd:
@dandanlen We might want to merge this fix into the branch that the CI uses for the bouncer upgrade test. Otherwise the upgrade test will still flake during setup of the assethub vault.

PR #6813, opened 2026-08-25
test: PRO-2303 optimistic elections resolving out of order

dandanlen:
# Pull Request

Closes: [PRO-2303](https://linear.app/chainflip/issue/PRO-2303/test-optimistic-elections-resolve-out-of-order)

## Summary

Adds `multiple_optimistic_elections_resolved_out_of_order` to the block witnesser tests, covering the case PRO-2303 asks about: more than one optimistic election in flight, resolving out of order.

The existing coverage stops short of this — `elections_resolved_out_of_order_has_no_impact` runs with `max_optimistic_elections: 0`, and `optimistic_election_result_saved_and_used_or_discarded_correctly` runs with `max_optimistic_elections: 1`, where only one optimistic election exists at a time so out-of-order resolution between them is impossible. Every production config currently sets `max_optimistic_elections: 1`, so this test describes behaviour that is not live yet; the motivation for pinning it down is the note on PRO-2303 that TRON's `SingleBlockQuery` semantics may want 2.

The scenario, with `max_optimistic_elections: 2`:

1. BHW reports headers 11 and 12 → by-hash elections for 11, 12 and optimistic elections for 13, 14.
2. 11, 12 and the optimistic election for **14** reach consensus; **13** does not. Closing 14 pushes `seen_heights_below` to 15, so `start_more_elections` drops the still-ongoing optimistic election for 13 and restarts optimistic elections at 15, 16.
3. BHW reports headers 13 and 14 → 14 is served from the optimistic cache (hash matches) and processed; 13 was never cached, so it is queued and elected by hash.
4. The by-hash election for 13 resolves.

Net result: every height 11–14 is witnessed exactly once, nothing is skipped or duplicated, and the block processor ends up with data for all four. That is the answer to the issue's question.

### Judgement calls

- **`processed_up_to` runs ahead, and the test asserts that.** After step 2, `lowest_in_progress_height()` finds nothing in progress (optimistic elections are excluded from it), so it falls back to `seen_heights_below` and the hook reports `processed_up_to = 14` — even though 13 has never been witnessed. It falls back to 12 on the next block once 13 is queued. `ProcessedUpTo` drives deposit-channel recycling and boost timeouts in `cf-ingress-egress`, so a channel could in principle be recycled over an unwitnessed block height in this window. It is not reachable today (`max_optimistic_elections` is 1 everywhere), so I have not changed any production code — I have asserted the current values with a `NOTE` comment at the call site flagging that this needs addressing before the setting is raised above 1. **Please look at this closely**: if you'd rather the test assert the correct-in-principle value and fail until it's fixed, say so and I'll flip it.
- Added one new registered check, `last_processed_up_to_is`, to make the above observable.
- The final `expect_consensus_multi` needs its entries in election-id order (optimistic elections carried over from the previous block come before the newly created by-hash election), not height order. Commented in place, since it is not obvious.
- Test-only change; no production code touched.

## API Changes

None.

### RPCS

None.

### Extrinsics & Events

None.

## Verification

Commands run and observed to pass on this branch:

- `cargo test -p pallet-cf-elections --lib multiple_optimistic_elections_resolved_out_of_order` — 1 passed.
- `cargo test -p pallet-cf-elections --lib block_witnesser` — 24 passed, 0 failed (includes the BW proptests `prop_tracker_ops_hold_invariants` and `test_bw_statemachine`).
- `cargo test -p pallet-cf-elections --lib` — 106 passed, 0 failed.
- `cargo clippy -p pallet-cf-elections --all-targets` — no warnings for this crate (only the pre-existing `trie-db` future-incompat note).
- `cargo fmt --all` — applied, diff is clean afterwards.

## Not verified

- No full-workspace build or test run; only `pallet-cf-elections` was compiled and tested.
- No runtime integration tests, bouncer/E2E tests, benchmarks, or try-runtime checks (none apply to a test-only change in this pallet, but they were not run).
- The `processed_up_to` concern above is reasoned from the code (`lowest_in_progress_height` → `ProcessedUpTo` → channel recycling in `cf-ingress-egress`) and observed in this unit test; it has not been reproduced end-to-end.

---

## *Checklist*

* The PR is complete:
  * [ ] Reviewers are assigned.
  * Tests:
    * [ ] Proptests for important invariants — existing BW proptests still pass; none added.
    * [ ] Integration tests for runtime-wide behaviours — n/a.
    * [ ] Bouncer tests for E2E features involving external chains — n/a.
  * [x] Benchmarks: no placeholders.
  * [x] Migrations: none.
  * [x] Unrelated changes are isolated — single commit, test file only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NGrmrtauJoT3pTCtLW3d3y)_

### DIFF
--- a/state-chain/pallets/cf-elections/src/electoral_systems/tests/block_witnesser.rs
+++ b/state-chain/pallets/cf-elections/src/electoral_systems/tests/block_witnesser.rs
@@ -182,6 +182,13 @@ register_checks! {
 				let (n, election_type) = election;
 				assert_eq!(get_election(post, n), election_type, "election should be of type {:?}", election_type);
 			}
+		},
+		last_processed_up_to_is(_pre, post, height: ChainBlockNumber) {
+			assert_eq!(
+				post.unsynchronised_state.processed_up_to.call_history.last().cloned(),
+				Some(height),
+				"the block processor should have last reported processed-up-to {height:?}",
+			);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


dandanlen:
`check-semantic-pr` failed on the original title (`PRO-2303: test optimistic elections resolving out of order` — no conventional-commit prefix). Retitled to `test: PRO-2303 optimistic elections resolving out of order`, which the action accepts.

The red check run is stale: `ci-pr.yml` triggers on `pull_request` with the default activity types (`opened`, `synchronize`, `reopened`), so a title edit doesn't re-run it, and I don't have permission to re-run the job. It'll go green on the next push to this branch. Everything else in the run was green or still in progress at the time of writing.

---

dandanlen:
`post-check-turbo / bouncer` failed. I don't believe it's this PR's, and I'm not pushing anything for it:

- This diff adds a `#[cfg(test)]`-only test to `pallet-cf-elections`. It changes no runtime, engine, or bouncer code, so it can't affect a localnet E2E run.
- The same job is red on unrelated recent PRs: [#6812](https://github.com/chainflip-io/chainflip-backend/pull/6812) (`post-check-turbo / bouncer` failed while `post-check` and `post-check-old-rpcs` both passed) and [#6808](https://github.com/chainflip-io/chainflip-backend/pull/6808) (`post-check-turbo` and `post-check-old-rpcs` both failed). Likely related to the flaky-bouncer tracking issue, PRO-2510.
- I don't have permission to re-run the job (`rerun-failed-jobs` returns 403), so I can't confirm with a re-run.

Everything else on this commit is green, including the full workspace build, `pre-check / cargo-clippy`, `cargo-fmt`, `test / unit-tests` (both variants), `coverage / unit-tests`, `test-benchmarks`, `upgrade-check / upgrade_test`, and chainspec-compatibility for all three networks. `post-check / bouncer` and `post-check-old-rpcs / bouncer` were still running when I wrote this; I'm watching them and will follow up if either fails in a way that looks connected to this change.

Happy to be told otherwise if this turbo failure is known to be real right now.

---

dandanlen:
Following up on the bouncer question: `post-check / bouncer` and `post-check-old-rpcs / bouncer` both **passed** on this commit. Only `post-check-turbo / bouncer` is red, which is the variant also failing on unrelated PRs (#6812, #6808) — so nothing here points at this change.

Now that this is out of draft, the two red marks blocking it are:

1. **`check-semantic-pr`** — stale. It failed on the original title; the title has since been corrected to `test: PRO-2303 …`, which the action accepts. `ci-pr.yml` triggers on `pull_request` default types (`opened`, `synchronize`, `reopened`), so the title edit didn't re-run it, and my token can't re-run jobs (403 on both `rerun-failed-jobs` and `rerun`). It needs either a maintainer re-run or any push to the branch.
2. **`post-check-turbo / bouncer`** — the flaky one above.

Both need someone with write access; there's nothing I can push that would legitimately fix either.

---

MxmUrw (commented):
I'm not really comfortable with merging this without having thought through all the consequences myself, and without them being properly included in the proptest coverage... The optimistic elections were a very tricky part and even though the possibility to configure `> 1` optimistic elections is part of the config it was never tested or modeled in any sufficient manner.

I would rather we don't necessarily touch this at the moment - or do it properly.

It might all be just fine, but it will take some time and investigations to convince myself.

dandanlen:
Understood — that's your call and I'm not pushing anything further at it. @dandanlen, over to you whether this gets closed or parked.

Since you'd want the behaviour in the proptests rather than a hand-written scenario, I spent a bit of time poking at whether the existing ones could get there. Three findings, from local scratch edits that are **not** in this branch (the diff here is still the single test-only commit):

**1. The `> 1` config is already generated.** `generate_settings()` in `test_bw_statemachine` samples `max_optimistic_elections: 0..3u8`, so 2 optimistic elections is explored today. The generator isn't the gap.

**2. But forward-only chain progress is never generated.** `generate_context()` sets, for any progress with at least one header:

```rust
progress.removed = Some(first_header.block_height..=seen_heights_below.saturating_backward(1));
```

and the headers themselves always start in `[seen_heights_below - STATIC_SAFETY_BUFFER_FOR_TESTS, seen_heights_below)`. So every non-empty `ChainProgress` the proptest produces is a reorg that rewinds to `seen_heights_below - 1`. I checked this by asserting `removed.is_some()` on every `Left(Some(progress))` — the only counterexample it found was `all_headers: []`, i.e. the no-op.

That matters here because the scenario in this PR lives entirely on the *forward* path: BHW reports new headers, the optimistic cache is consumed, `seen_heights_below` advances past a height whose optimistic election never resolved. The proptest can't reach that shape today regardless of `max_optimistic_elections`.

**3. Any invariant on the processed-up-to report has to be reorg-aware.** I first tried plain monotonicity of `lowest_in_progress_height()` (which is what `processed_up_to` is derived from). It fails in ~18s — but on a *legitimate* reorg: `285 -> 281` with `removed: Some(281..=284)`, where going backwards is correct. Making it reorg-aware then makes it vacuous, because of (2) — everything is a reorg.

So "doing it properly" looks like: teach `generate_context()` to emit forward-only progress as well, then add a reorg-aware invariant on the processed-up-to report. That's a real piece of modelling work on the part you're describing as tricky, well beyond this issue's 2-point estimate, and I don't think it's mine to do unsupervised — but it does mean your instinct that this isn't sufficiently modelled is right, and points at where.

Happy for this PR to be closed if you'd rather start from the proptest end.

---

dandanlen:
Sounds like we now have some additional context at least... will close.

PR #6817, opened 2026-08-25
feat: add adversarial property tests for key-handover ceremony

dandanlen:
# Pull Request

Closes: PRO-3085

## Summary

Adds some proptests for the key handover ceremony.

Opening as draft because I haven't really looked at the tests in much detail so I'm not sure how sensible/exhaustive they are.

### DIFF
--- a/engine/multisig/src/client/keygen/tests.rs
+++ b/engine/multisig/src/client/keygen/tests.rs
@@ -1642,4 +1642,446 @@ mod key_handover {
 		let (new_key, _new_shares) = ceremony.complete();
 		assert_eq!(new_key, initial_key);
 	}
+
+	/// Adversarial property tests over the key-handover ceremony.
+	///
+	/// The existing handover tests each pin one hand-picked malicious behaviour to
+	/// one expected outcome. These instead enumerate a *space* of malicious
+	/// behaviours and assert the invariants that must hold across all of them,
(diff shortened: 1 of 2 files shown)


engine/multisig/src/client/keygen/tests.rs
```diff
@@ -1642,4 +1642,446 @@ mod key_handover {
 		let (new_key, _new_shares) = ceremony.complete();
 		assert_eq!(new_key, initial_key);
 	}
+
+	/// Adversarial property tests over the key-handover ceremony.
```
msgmaxim, line 1646: I suspect that some of these are duplicates of the test we already have (some of them likely in https://github.com/chainflip-io/chainflip-backend/pull/6818). I don't mind reframing them as "property tests" if this improves coverage, but ideally we should either keep one or the other (while making sure we don't reduce coverage). Claude confirmed that these are mostly duplicates (I didn't dig deeper since this is a draft).

dandanlen: Yes it's likely that these are (mostly) duplicating existing tests but in general property tests are superior because they explore more edge cases, and also allow encoding known past regressions etc. Many of our existing tests use a handle up of specific curated inputs. 

I was mainly curious to see if we can express our assumptions in this way and whether it's useful. 

I need to have a closer look myself to form a stronger opinion.

Issue #6814, opened 2026-08-25
Tron USDT egress fee inflated ~17x (≈$4 → ≈$72): simulated-swap TRX price accepts a dust-pool fill


CumpsD:
## Summary

The protocol-level **egress fee for USDT on Tron is currently \~17x too high**: **\~71.61 USDT**, against a real on-chain broadcast cost of **10.7757 TRX (~$3.68)**.

The TRX-denominated gas model is fine (\~12.668 TRX vs 10.776 TRX actually burned, +17.6%). The inflation happens entirely in the **TRX → USDT price conversion**, which for TRX is sourced from a rolled-back 20-USDC simulated swap against Chainflip's own TRX/USDC pool. That pool is currently dust, so the probe returns a price of **~$5.65/TRX against a market price of \~$0.3416** — and the only guard on that value is `.filter(|v| *v > 0)`.

## Impact

- Users swapping to USDT on Tron are overcharged \~$68 per swap. On the swap below the egress fee was **2.45% of the output**.
- Fill-or-kill does not protect against it (see "Secondary finding").
- Tron **ingress** is hit by the same path: `ingress_fees.Tron.USDT` ≈ **63.52 USDT**.
- Any asset lacking a Chainlink feed and egressing a non-gas asset is exposed to the same failure mode.

## Reproduction (live, mainnet)

```bash
curl -s https://mainnet-archive.chainflip.io -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"cf_environment","params":[]}' \
  | jq '.result.ingress_egress.egress_fees.Tron
        | map_values(ltrimstr("0x") | ascii_downcase | explode
          | reduce .[] as $c (0; . * 16 + (if $c >= 97 then $c - 87 else $c - 48 end)))'
# => { "TRX": 6174174, "USDT": 71613002 }    # 6.174174 TRX, 71.613002 USDT
#    (drifts by a few hundred units block-to-block; the USDT figure has sat at \~71.61 since Aug 23)
```

The pool the price probe reads:

```bash
# 20 USDC -> TRX  : succeeds (this is exactly ESTIMATION_AMOUNT_USDC)
# 25 USDC -> TRX  : "DispatchError: Simulated swap failed on the output leg: swap leg failed."
curl -s https://mainnet-archive.chainflip.io -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"cf_swap_rate_v3","params":[
[...]

PR #6821, opened 2026-08-27
feat: circuit breaker for fee price estimates

dandanlen:
# Pull Request

Fixes: PRO-3094

Updates the fallback prices for all assets to current values (rounded). 

Adds a tolerance factor that is applied to the pool-based price estimate: if the estimate is more than 4x away from the fallback price, we use the fallback price. This prevents excessive drift and lessens the potential for manipulation.

Mitigates #6814 pending addition of new price oracles.

### DIFF
--- a/bouncer/commands/go_bananas.ts
+++ b/bouncer/commands/go_bananas.ts
@@ -264,13 +264,16 @@ async function playSwapper<A = []>(cf: ChainflipIO<A>) {
   }
 }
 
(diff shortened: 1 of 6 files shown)


MxmUrw (approved):
Looks good to me, also to copilot :)

I think maybe we'd want to have storage values for the fallback prices eventually?

marcellorigotti (approved):

PR #6822, opened 2026-08-28
feat: bouncer seedable ccm

j4m1ef0rd:
Makes CCM metadata generation reproducible and fixes a flaky AllSwaps failure.

- Thread a seeded `rng` through the CCM message/metadata builders (defaults to `Math.random` for other callers); each AllSwaps swap gets a deterministic per-swap seed derived from the AllSwaps seed, so a failing run reproduces from the logged seed.
- Raise the default HubDot swap amount `50 → 200`. HubDot is low-value (~2 USDC), so a 50-DOT swap could egress below a CCM egress's dust limit once the (message-driven) gas budget is deducted, ignoring the egress.

### DIFF
--- a/bouncer/shared/new_sol_address.ts
+++ b/bouncer/shared/new_sol_address.ts
@@ -4,3 +4,13 @@ import { sha256 } from 'shared/utils';
 export function newSolAddress(seed: string): string {
   return Keypair.fromSeed(sha256(seed)).publicKey.toBase58();
 }
+
+// A Solana keypair derived deterministically from `rng`, so generated account addresses are
+// reproducible from the seed rather than cryptographically random.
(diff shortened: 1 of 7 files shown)


dandanlen (commented):
Can we use a config object with named params so we don't need to thread `undefined, undefined, rng` through everything?

```ts
type CcmMetadataOptions = {
  message?: string;
  additionalData?: string;
  rng?: () => number;
};
```

Also we should make the `rng` required on private methods and only default it for public methods. Otherwise it's too easy to forget to thread it through. 

---
But also: we're using the rng() to decide the *length* of the ccm data but not the content - so it's still not really determined from the seed (might be currently irrelevant for validation but we might as well make it fully reproducible).

dandanlen (approved):
Thanks

PR #6824, opened 2026-08-28
feat: multi operator delegation

ramizhasan111:
# Pull Request

Closes: PRO-3089

## Summary

This PR adds multi operator delegation support. The previous delegate/undelegate interface is kept as is and will be usable for cases where the delegator has delegated only to a single operator. A new `delegate_multi` extrinsic is added to support declarative multi delegation plan which overwrites any existing plan.

- The `DelegationChoice` storage entry is changed to allow storing multi-operator delegation. Migration was added to migrate the storage.

### RPCS

`cf_common_account_info` rpc is changed and the shape of returned `RpcAccountInfoCommonItems` changed to include a map of operator->amount for current and upcoming delegations.

### Extrinsics & Events

new extrinsic called `delegate_multi` was added that emits a new `DelegationPlanUpdated` event.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:

### DIFF
--- a/bouncer/tests/multi_delegate_flip.ts
+++ b/bouncer/tests/multi_delegate_flip.ts
@@ -1,19 +1,55 @@
-import type { AccountId32 } from 'dedot/codecs';
-import { amountToFineAmountBigInt, defaultAssetAmounts } from 'shared/utils';
-import { getIsoTime } from 'shared/utils/logger';
+import type { SubmittableResult } from '@polkadot/api';
+// eslint-disable-next-line no-restricted-imports
+import type { KeyringPair } from '@polkadot/keyring/types';
 import { fundFlip } from 'shared/fund_flip';
 import { AccountRole, setupAccount } from 'shared/setup_account';
 import { newChainflipIO, partialAccountFromUri } from 'shared/utils/chainflip_io';
+import { getChainflipPolkadotApi } from 'shared/utils/substrate';
+import { getIsoTime } from 'shared/utils/logger';
+import { amountToFineAmountBigInt, defaultAssetAmounts } from 'shared/utils';
 import { TestContext } from 'shared/utils/test_context';
-import { validatorDelegationPlanUpdatedEvent } from 'generated/events/validator/delegationPlanUpdated';
 
-// The generated `PalletCfValidatorDelegationDelegatorRelations` type is strict about
-// `AccountId32` (unlike top-level call params, which accept the more permissive
-// `AccountId32Like`), so SS58 addresses from a `KeyringPair` need this cast.
-const accountId = (address: string): AccountId32 => address as unknown as AccountId32;
+async function submitDelegateMulti(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  polkadotApi: any,
+  delegator: KeyringPair,
+  operators: Map<string, bigint>,
+) {
+  return new Promise<SubmittableResult['events']>((resolve, reject) => {
+    polkadotApi.tx.validator
+      .delegateMulti({ operators })
+      .signAndSend(delegator, (result: SubmittableResult) => {
+        if (result.dispatchError) {
+          reject(new Error(`delegateMulti: dispatch error ${result.dispatchError.toString()}`));
+        } else if (result.status.isInBlock || result.status.isFinalized) {
+          resolve(result.events);
+        }
+      })
+      .catch(reject);
+  });
+}
+
+function findEventData(events: SubmittableResult['events'], section: string, method: string) {
+  const found = events.find(({ event }) => event.section === section && event.method === method);
+  if (!found) {
+    throw new Error(`Event ${section}.${method} not found among: ${JSON.stringify(events)}`);
+  }
+  return found.event.data;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function operatorsOf(planUpdatedEventData: any): [string, bigint][] {
+  const operators = planUpdatedEventData.plan.operators;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return [...operators.entries()].map(([account, amount]: [any, any]) => [
+    account.toString(),
+    amount.toBigInt(),
+  ]);
+}
 
 export async function testMultiDelegate(testContext: TestContext) {
   const cf = await newChainflipIO(testContext.logger, []);
+  await using polkadotApi = await getChainflipPolkadotApi();
 
   // Account names have to be unique across bouncer runs, since if the test is run a second
   // time for accounts that are already registered/funded, expected events won't be re-emitted.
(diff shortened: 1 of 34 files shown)


state-chain/pallets/cf-validator/src/delegation.rs
```diff
@@ -139,6 +139,34 @@ pub struct OperatorSettings {
...
+	Decode,
+	DecodeWithMemTracking,
+	TypeInfo,
+	Serialize,
+	Deserialize,
+)]
+pub struct DelegatorRelations<Account: Ord, Amount> {
+	pub operators: BTreeMap<Account, Amount>,
+}
+
```
dandanlen, line 163: The spec document specifies this:

```rust
#[derive(...)]
pub enum DelegationPlan<Amount, N> {
    /// Absolute caps per operator. Σ ≤ balance. At most one entry may be `Max`.
    Fixed(BoundedVec<(AccountId, DelegationAmount<Amount>), N>),
    // v2, appended later:
    // /// Proportional split of the whole balance. Σ = 100%. Auto-compounds.
    // Proportional(BoundedVec<(AccountId, Perbill), N>),
}

// new call index (20 is the last used, `report_witnessing_task_restart`)
pub fn set_delegation(origin: OriginFor<T>, plan: DelegationPlan<T::Amount, T::MaxOperatorsPerDelegator>) -> DispatchResult;
```

Why did we deviate from that? (Both the naming - not sure why we need a separate name for the same thing - and using a struct instead of an enum).

Using a struct means any addition (such as adding proportions rather than fixed amounts, as discussed) is a breaking change.

dandanlen: Also if we're going to create a dedicated type it might as well be useful: things like encoding invariants, validation etc.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -401,13 +401,17 @@ pub mod pallet {
...
-	///
-	/// The max bid determines how much of the delegator's balance can be used
-	/// used by the operator when bidding for an authority slot.
+	/// Maps a delegator to its live relations: the set of operators it delegates to and the max
+	/// bid pledged to each. The sum of all of a delegator's max bids is capped at its funding
+	/// balance. The key is always removed entirely once a delegator's relations become empty.
 	#[pallet::storage]
-	pub type DelegationChoice<T: Config> =
-		StorageMap<_, Identity, T::AccountId, (T::AccountId, T::Amount), OptionQuery>;
+	pub type DelegationChoices<T: Config> = StorageMap<
```
dandanlen, line 408: Not sure why needed to rename the storage... seems to just create more surface in the diff than necessary.

state-chain/pallets/cf-validator/src/delegation.rs
dandanlen: Another benefit: with an enum we can express `NoDelegation` or `Unique(operator)` which allows more explicit handling of these edge cases (instead of inferring from things like `.empty()`)

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1456,16 +1459,96 @@ pub mod pallet {
...
+		/// scaled down proportionally so the total exactly matches the balance -- the sum of a
+		/// delegator's relations can never exceed what they actually hold. The (possibly
+		/// scaled-down) total must be at least the minimum funding amount if `plan` is
+		/// non-empty; individual entries may be smaller, only the total is checked.
+		#[pallet::call_index(24)]
+		#[pallet::weight(T::ValidatorWeightInfo::delegate_multi())]
+		pub fn delegate_multi(
+			origin: OriginFor<T>,
+			plan: DelegatorRelations<T::AccountId, T::Amount>,
+		) -> DispatchResult {
```
dandanlen, line 1490: I think validation of the delegation plan here should be focused on basic validity: Things like number of operators (currently unbounded), or total balance checks that cause the submission to fail or succeed. 

What we *don't* want to alter the plan that the user submits. Any *adjustment* against balances should be resolved at auction time, not submission time.

dandanlen: This is also mentioned in the design doc btw.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1456,16 +1459,96 @@ pub mod pallet {
...
+				Self::ensure_delegator_role(&delegator)?;
+
+				for operator in new_relations.keys() {
+					Self::ensure_operator_accepts_delegator(&delegator, operator)?;
+				}
+
+				// The sum of a delegator's relations can never exceed what they actually hold --
+				// scale every entry down proportionally to fit, rather than rejecting the plan
+				// outright.
+				let raw_total: T::Amount = new_relations.values().copied().sum();
```
dandanlen, line 1518: This is trivially attackable: the map is unbounded and so are the values. We need to limit the size of the map and use saturating math.

A proptest would have caught this.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1456,16 +1459,96 @@ pub mod pallet {
...
+					&delegator,
+					DelegatorRelations { operators: new_relations.clone() },
+				);
+				new_relations
+			};
+
+			Self::deposit_event(Event::DelegationPlanUpdated {
+				delegator: delegator.clone(),
+				plan: DelegatorRelations { operators: new_relations.clone() },
+			});
```
dandanlen, line 1547: We need to discuss with @GabrielBuragev & co. whether a new event makes sense here, or whether we want to emit existing Delegated / Undelegated events instead/as well here.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1456,16 +1459,96 @@ pub mod pallet {
...
+		/// delegator's relations can never exceed what they actually hold. The (possibly
+		/// scaled-down) total must be at least the minimum funding amount if `plan` is
+		/// non-empty; individual entries may be smaller, only the total is checked.
+		#[pallet::call_index(24)]
+		#[pallet::weight(T::ValidatorWeightInfo::delegate_multi())]
+		pub fn delegate_multi(
+			origin: OriginFor<T>,
+			plan: DelegatorRelations<T::AccountId, T::Amount>,
+		) -> DispatchResult {
+			let delegator = ensure_signed(origin)?;
```
dandanlen, line 1491: Do we not require LP now?

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1456,16 +1459,96 @@ pub mod pallet {
...
+
+			ensure!(
+				CurrentRotationPhase::<T>::get() == RotationPhase::Idle,
+				Error::<T>::RotationInProgress
+			);
+
+			let new_relations: BTreeMap<T::AccountId, T::Amount> =
+				plan.operators.into_iter().filter(|(_, amount)| !amount.is_zero()).collect();
```
dandanlen, line 1499: We should probably disallow zero-value entries. Absence of an entry is equivalent to zero. Adding an entry with zero value is effectively just spam.

state-chain/runtime/src/runtime_apis/impl_api.rs
```diff
@@ -813,12 +813,24 @@ impl_runtime_apis! {
...
-				.map(|(operator, max_bid)| DelegationInfo { operator, bid: core::cmp::min(flip_account.total(), max_bid) });
-			let current_delegation_status = pallet_cf_validator::DelegationSnapshots::<Runtime>::iter_prefix(Validator::current_epoch())
-				.find_map(|(operator, snapshot)| snapshot.delegators.get(account_id).map(|&bid|
-					DelegationInfo { operator, bid }
-				));
+			// Operator -> bid, for every operator this account currently delegates to (a
+			// delegator may hold relations to multiple operators simultaneously).
+			let upcoming_delegation_status: BTreeMap<AccountId, FlipBalance> =
+				pallet_cf_validator::DelegationChoices::<Runtime>::get(account_id)
+					.map(|relations| {
```
dandanlen, line 820: Do we not need to resolve the auction-equivalent bids here? Otherwise this will return something different than the actual outcome.

state-chain/pallets/cf-validator/src/lib.rs
```diff
@@ -1124,12 +1138,14 @@ pub mod pallet {
...
 							allowed.clear();
 							// Any existing delegators need be added to the allowed list.
-							for (delegator, (assigned_operator, _)) in DelegationChoice::<T>::iter()
-							{
-								if assigned_operator == operator {
-									allowed.insert(delegator.clone());
-								}
-							}
+							allowed.extend(
+								Self::get_all_associations_by_operator(
```
dandanlen, line 1142: This call iterates over a storage map. The plan in Linear mentioned adding a reverse index to make this more efficient.

bouncer/tests/multi_delegate_flip.ts
dandanlen, line 1: What is this test actually testing? From what I can tell: extrinsic submission, event emission on the happy path. 

Do we really need a bouncer test for this?

dandanlen (commented):
This still needs work... it doesn't really follow the design proposal.

bouncer/tests/multi_delegate_flip.ts
ramizhasan111: Yes, this is testing the end to end happy path flow. This is similar to the bouncer test we have for single operator delegation which also tests the happy path.

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: I am not sure if that is a significant optimization, since the realistic number of delegators and operators is low and so iterating over them is very fast and so effeciency gains would be trivial. Also, by adding the reverse index we are introducing overhead of an extra storage read and write since two storages will now be involved so it is not clear if it brings net benefits. In any case, I dont have a strong opinion, I have added this change as an another PR on top of this (https://github.com/chainflip-io/chainflip-backend/pull/6895) that has the suggested change. We can merge that if we want to.

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: added the bound (now using bounded vec) and validity checks. Changed to erroring if the plan adds to more than balance instead of scaling.

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: we do, check a bit further down. the lp check is further down where check it if the plan is non-empty. this allows users with no role to remove existing delegation.

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: this part of the code is removed now.

state-chain/pallets/cf-validator/src/delegation.rs
ramizhasan111: added the suggested change

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: reverted

state-chain/pallets/cf-validator/src/lib.rs
ramizhasan111: we are filtering them out here. If someone tries to spam this way, they pay tx fees for extrinsic execution.

state-chain/runtime/src/runtime_apis/impl_api.rs
ramizhasan111: yes, updated.

ramizhasan111:
> This still needs work... it doesn't really follow the design proposal.

the design proposal wasn't supposed to be final and set in stone (you yourself confirmed that). We can discuss individual points here in this PR and come to a final version.

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: I don't think we necessarily need to error: this will cause unexpected failures. For example:
- User reads balance, X FLIP
- User submits 'delegate `X`'
- Submitting the transaction consumes 123 flipperinos, so their balance is now `X - 123`
- Error

We should instead accept the Plan, where each apportioned bid is understood to be the target bid. Then we scale at bid resolution time (auction).

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: I thought we were requiring LPs now and they were all being migrated?

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: Adding `Max` only complicates thing unnecessarily IMO

state-chain/pallets/cf-validator/src/delegation.rs
dandanlen: We should make the type us `BoundedBTreeMap` instead of converting back and forth.

dandanlen: And then  define some methods on the enum instead of exposing the inner structure. Otherwise we lose the benefit of abstracting this out in the first place. 

For example if/when we add proportional allocation, we will need to rewrite all the code where we just extract the raw values. 

Also it makes it testable in isolation.

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: It's also about providing an unambiguous API.

state-chain/pallets/cf-validator/src/lib.rs
dandanlen: I think it's an obvious improvement when we need to iterate (which is what we usually need to do). 

I would suggest putting the smaller change on main first and putting this change on top of the smaller one. (Since the smaller change is less likely to drift and cause conflicts)

bouncer/tests/multi_delegate_flip.ts
dandanlen: The difference is that the single-operator test is testing the integration with *Ethereum* via the SC Utils contract, so a bouncer test was a requirement: it covers the contract, engine witnessing, SC Api decoding. None of this is a requirement (yet) for multi delegation.

dandanlen:
> > This still needs work... it doesn't really follow the design proposal.
> 
> the design proposal wasn't supposed to be final and set in stone (you yourself confirmed that). We can discuss individual points here in this PR and come to a final version.

Yes we can discuss later. It's fine to diverge from the proposal if there's a conscious and justifiable choice to do so. The point is that the design proposal was there and following it would have avoided a lot of the issues that were raised, but it seems it was just ignored or forgotten. 

Also btw. there are still no proptests.

dandanlen (commented):
This is better thanks. 

Before you continue too much with the proptests: I think we can still improve the interface of the `DelegationPlan` type by making the methods more domain-specific. Currently, besides resolve_bids, they are very close to just exposing the inner map (we have try_insert/remove/pop/get/total). 

As a concrete example: `pop()` is only ever used in the context where we expect a single entry (`ensure!(plan.len() <= 1, Error::<T>::MultiOperatorDelegator); ..`). Better to fold that into the method and rename it `only_member()` or something. 

`resolve_bids` is a good example of the kind of method this type *should* have. 

With a cleaner interface, I think the proptest properties/invariants should become clearer.

Also i thought again about validation: i think we should reject if the total is above the current balance (maybe we allow a small diff to avoid the race condition). This is less ambiguous. 

We can discuss in-person tomorrow.

PR #6825, opened 2026-08-31
fix: engine runner split args

0xSleipnir:
# Pull Request

## Summary

The engine runner filters CLI arguments that are unsupported by the fallback engine when switching between CFE versions.

The existing filtering handled arguments passed as `--flag=value`, but not the equivalent split form:

```text
--flag
value
```

For example, when BSC configuration is passed as:

```text
--bsc.rpc.http_endpoint
http://localhost:8549
```

the runner removed `--bsc.rpc.http_endpoint` but left its value in `argv`. The fallback engine then interpreted the remaining URL as a positional argument and failed to start.

This change separates the version-specific list of incompatible arguments from the filtering logic. Value-taking incompatible arguments are now removed correctly in both forms:

```text
--flag=value
```

and:

```text
--flag value
```

Compatible arguments continue to pass through unchanged.

The existing engine-runner test now uses split BSC arguments to cover the fallback path, with additional unit coverage for split arguments, inline arguments, compatible arguments, and valueless flags. 

The incompatible argument list is explicit rather than prefix-based, so CLI compatibility differences stay tied to the specific old/new engine version pair and can be reviewed deliberately during future version bumps.

## API Changes

None.

### RPCS

No RPC interface changes.

### Extrinsics & Events

None.

---

## Checklist

* The PR is complete:

  * [ ] Reviewers are assigned.
* Tests:

    * [x] Proptests for important invariants — not applicable.
    * [x] Integration coverage for the engine fallback behaviour.
    * [ ] Bouncer tests for E2E features involving external chains — not applicable.
  * [x] Benchmarks — not applicable.
  * [x] Migrations — not applicable.
  * [x] Bouncer typegen — not applicable.

  * [x] Unrelated changes are isolated.
  * [x] Anything non-obvious is documented in the `Summary`.
* [x] Performed a local review of the change.
* [x] Read the code carefully after the local review.

### DIFF
--- a/engine-upgrade-utils/src/lib.rs
+++ b/engine-upgrade-utils/src/lib.rs
@@ -40,6 +40,10 @@ struct IncompatibleArg {
 	takes_value: bool,
 }
 
+/// Arguments that OLD_VERSION cannot parse.
+///
+/// This list must be reviewed whenever OLD_VERSION or NEW_VERSION changes so it
+/// reflects only the CLI differences between the two engine versions.
 const INCOMPATIBLE_WITH_OLD: &[IncompatibleArg] = &[
 	IncompatibleArg { name: "--bsc.rpc.http_endpoint", takes_value: true },
 	IncompatibleArg { name: "--bsc.backup_rpc.http_endpoint", takes_value: true },
--- a/engine-upgrade-utils/src/lib.rs
+++ b/engine-upgrade-utils/src/lib.rs
@@ -33,16 +33,52 @@ pub const ENGINE_LIB_PREFIX: &str = "chainflip_engine_v";
 pub const ENGINE_ENTRYPOINT_PREFIX: &str = "cfe_entrypoint_v";
 
 // Sometimes we need to adapt arguments between the new and old versions while both CFEs can be run
-// by the upgrade runner. The old engine (OLD_VERSION) and the new engine (NEW_VERSION) currently
-// share the same CLI argument schema, so this is a passthrough. If a future version bump changes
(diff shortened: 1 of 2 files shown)


dandanlen (approved):

engine-upgrade-utils/src/lib.rs
```diff
@@ -33,18 +33,58 @@ pub const ENGINE_LIB_PREFIX: &str = "chainflip_engine_v";
...
+
+/// Arguments that OLD_VERSION cannot parse.
+///
+/// This list must be reviewed whenever OLD_VERSION or NEW_VERSION changes so it
+/// reflects only the CLI differences between the two engine versions.
+const INCOMPATIBLE_WITH_OLD: &[IncompatibleArg] = &[
+	IncompatibleArg { name: "--bsc.rpc.http_endpoint", takes_value: true },
+	IncompatibleArg { name: "--bsc.backup_rpc.http_endpoint", takes_value: true },
+	IncompatibleArg { name: "--bsc.private_key_file", takes_value: true },
+];
```
MxmUrw, line 51: nit: Since this is only used once, I would rather we inline it directly in `args_compatible_with_old()`. Then that function also would become more useful rather than being a small shim around `filter_incompatible_args()`.

**EDIT**: actually, the separation is required for tests, so nevermind.

MxmUrw (commented):
Thanks for the PR!

dandanlen:
Closing this in favour of #6827 (commits needs to be signed and CI needs an authorised user).

PR #6835, opened 2026-09-01
feat: AMM fixed pool removal

j4m1ef0rd:
# Pull Request

Closes: PRO-2813 & PRO-2156

## Summary

The problem is that the fixed pools where an optimization for swap execution but slower at placing an order. This PR flips that around and makes placing an order faster and only computes the tick pools during swap execution. But that is not the main reason why we are doing this. It removes a lot of complexity from the AMM, solves the orphaned fixed pools and dust issues and opens up opportunities to do more complex limit orders (eg. lending/supply limit orders).

- Fixed pools deleted. FixedPool, FloatBetweenZeroAndOne and the percent_remaining / pool_instance bookkeeping are gone.
- No more collecting or sweeping Limit orders. Proceeds are credited to the LP by the swap that fills the order.
- Each order's share is computed at fill time. Removes the rounding dust the old float product accumulated.
- New LimitOrderFilled event 
- remove_fixed_pools migration: settles every outstanding order's earnings, drops orders bought in full, writes off liquidity no order can claim, rebuilds limit_orders_cache, kills the thresholds. Verified against a mainnet snapshot — 16 pools, 18 orders settled, 3263 base / 1447 quote written off, no errors.
- Order-fills RPC subscription now reads LimitOrderFilled events rather than diffing pool state before and after.
- `PriceImpactLimitExceeded` error is new. Price-impact rejection used to surface as `InsufficientLiquidity`.
- The weight of the on_finalize increased, but we are not accounting for it in any pallets. Created issue PRO-3109 to track.
- The per-price liquidity cap (`MAX_LIQUIDITY_PER_PRICE`) is removed. It existed so the swap could convert a price's whole liquidity without overflowing; those conversions now saturate instead, which is exact in both directions. A mint that would previously have failed with `MaximumGrossLiquidity` now succeeds.

### Weight changes

**Swap** — a swap crossing the whole book

| ticks | 1 order/tick | 4 | 16 | 100 | 1000 |
|---|---|---|---|---|---|
| 1 | 2.08 → 2.38 (1.14×) | 2.13 → 2.75 (1.29×) | 2.17 → 4.50 (2.1×) | 2.17 → 15.6 (7.2×) | 2.17 → 135 (62×) |
| 10 | 12.5 → 17.8 (1.42×) | 13.1 → 22.6 (1.72×) | 13.0 → 42.7 (3.3×) | 12.6 → 189 (14.9×) | — |
| 100 | 119 → 170 (1.43×) | 123 → 225 (1.83×) | — | — | — |
| 1000 | 1.22ms → 1.83ms (1.50×) | — | — | — | — |

**Placing** — marginal cost of opening one more order into a book that already holds *n*.

| orders held | same price: main → new | distinct prices: main → new |
|---|---|---|
| 0 | 167ns → 83ns (2.0× faster) | 166ns → 83ns (2.0× faster) |
| 100 | 167ns → 42ns (4.0× faster) | 375ns → 167ns (2.2× faster) |
| 1000 | 208ns → 42ns (5.0× faster) | 459ns → 250ns (1.8× faster) |

Placing is now flat in the depth of the book: inserting into the price's map is all that happens. It stays flat well past any realistic size — at 100,000 orders on one price the marginal cost is still 42ns, against 625ns on `main`, and a 100,000 order burst into a single price takes 7.3ms against 31.9ms.

### RPCS

- collected_fees field in LimitOrderUpdated is now always 0. Lp-api returns this deprecated 0 as well.
- `cf_lp_get_order_fills` returns no limit-order fills for pre-upgrade blocks
- `cf_pool_orders` filled_orders flag is deprecated. (does nothing now)

### Extrinsics & Events

- new `LimitOrderFilled` event.
- `LimitOrderUpdated` even now always returns 0 for the `bought_amount`

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -171,7 +162,9 @@ impl<T: Config> LimitOrderUpdate<T> {
 					None, // Dispatch now
 				);
 				let result = if let Err(err) = result {
-					if err == Error::<T>::OrderDoesNotExist.into() {
+					if err == Error::<T>::OrderDoesNotExist.into() ||
+						err == Error::<T>::UnspecifiedOrderPrice.into()
+					{
```
msgmaxim, line 167: Do you need to update the comment here?

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1608,30 +1610,7 @@ impl<T: Config> Pallet<T> {
...
-						&mut pool,
-						lp,
-						&asset_pair,
-						assets.sell_order(),
-						*id,
-						*tick,
-					)?;
-				}
-			}
+			// Limit orders have nothing to sweep: a swap pays their proceeds out as it fills them.
```
msgmaxim: Would it make sense to rename this to "sweep_range_orders"? Then the comment probably won't be necessary.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -2388,18 +2311,18 @@ impl<T: Config> Pallet<T> {
 						},
 					)
 					.filter_map(|(lp, id, tick): (T::AccountId, OrderId, Tick)| {
-						let (collected, position_info) = pool
+						let position_info = pool
 							.pool_state
 							.limit_order(&(lp.clone(), id), asset.sell_order(), tick)
 							.unwrap();
-						if filled_orders || !position_info.amount.is_zero() {
+						if !position_info.amount.is_zero() {
```
msgmaxim: As dicussed, I don't think this check is necessary since you now always remove orders that are fully consumed.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -2465,7 +2388,6 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Process changes to limit order:
-	/// - Payout collected `fee` and `bought_amount`
 	/// - Update cache storage for Pool
 	/// - Deposit the correct event.
 	fn process_limit_order_update(
```
msgmaxim: As dicussed, worth inlining.

state-chain/amm/src/limit_orders.rs
```diff
@@ -554,353 +270,142 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
+						SD::output_amount_floor(amount, price).unwrap_or(available),
+						available,
+					),
+					amount,
+				),
+			};
 
-				let swapped_amount = amount;
+			// Cannot underflow as swapped_amount is bounded by amount in both cases above.
+			amount -= bought_amount;
```
msgmaxim: I see that it holds, but a more defensive approach would be to use `saturating_reduce`?

state-chain/amm/src/limit_orders.rs
```diff
@@ -341,22 +155,25 @@ pub struct Collected {
 pub struct PositionInfo {
```
msgmaxim: This is now exactly the same struct as Position. Worth removing/merging?

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -166,6 +166,8 @@ impl<T: Config> LimitOrderUpdate<T> {
 						err == Error::<T>::UnspecifiedOrderPrice.into()
 					{
 						// Ignore the error if the order doesn't exist, as this is expected.
+						// UnspecifiedOrderPrice is expected if the order was fully filled and
```
msgmaxim, line 169: I don't think this comment makes sense:`update_limit_order` is called with price explicitly `None` hardcoded. In fact `OrderDoesNotExist` is *unexpected* here, because if you call this when no order is in the cache, you will always get `UnspecifiedOrderPrice`.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1736,7 +1736,7 @@ impl<T: Config> Pallet<T> {
 		amount_change: IncreaseOrDecrease<AssetAmount>,
 	) -> DispatchResult {
 		let asset_pair = asset_pair_try_from::<T>(base_asset, quote_asset)?;
-		Self::inner_sweep(Select::Single(lp))?;
+		Self::sweep_range_orders(Select::Single(lp))?;
```
msgmaxim, line 1739: Do we still need to sweep here considering that this method is only about limit orders and sweeping only applies to range orders now?

state-chain/amm/src/limit_orders.rs
```diff
@@ -289,74 +126,33 @@ pub enum DepthError {
...
-#[derive(
-	Default, Debug, PartialEq, Eq, TypeInfo, Encode, Decode, DecodeWithMemTracking, MaxEncodedLen,
-)]
-pub struct Collected {
-	/// The amount of assets purchased by the LP using the liquidity in this position since the
-	/// last collect.
-	pub bought_amount: Amount,
-	/// The amount of assets sold by the LP using the liquidity in this position since the
-	/// last collect.
+/// A swap buying into a single limit order. The proceeds are owed to the LP as of the swap, the
```
msgmaxim, line 136: > The proceeds are owed to the LP as of the swap, the pool does not hold on to them.

This reads a bit odd if you don't know the pervious design. IMO we could just say `bought_amount` needs to be credited to the LP.

state-chain/amm/src/limit_orders.rs
```diff
@@ -289,8 +274,7 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
 				),
 			};
 
-			// Cannot underflow as swapped_amount is bounded by amount in both cases above.
-			amount -= bought_amount;
+			amount = amount.saturating_sub(bought_amount);
```
msgmaxim, line 277: IMO we should still say that saturation should not happen and that this is a defensive measure only. Otherwise someone may "optimise" this away in the future. (AI flagged this as "regression" since the removed comment was useful.)

state-chain/amm/src/limit_orders.rs
```diff
@@ -911,73 +399,1032 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
+	let mut remaining_available = available;
+	let mut remaining_sold = sold_amount;
+	let mut remaining_bought = bought_amount;
+
+	orders.retain(|lp, position| {
+		// Calculate share against remaining amount instead of total amount to avoid rounding
+		// errors. The divisions cannot fail: every order holds a non-zero amount, so the
+		// remaining liquidity is non-zero for as long as there is an order left to fill.
+		let sold = mul_div_floor_checked(remaining_sold, position.amount, remaining_available)
+			.unwrap_or(remaining_sold);
```
msgmaxim, line 447: AI recommends `unwrap_or(0)` and I think I agree: if remaining_available is 0, then the position must be 0 too, so it didn't sell anything. Same below.

Also, I know this is supposed to be unreachable, but AI also recommends to test a 0 amount order. We have a defensive mechanism againt it here, but it is not tested.

state-chain/amm/src/limit_orders.rs
```diff
@@ -911,73 +399,1032 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
+		.values()
+		.fold(Amount::zero(), |total, position| total.saturating_add(position.amount))
+}
+
+/// Buys `sold_amount` of the orders' liquidity for `bought_amount` of the other asset, splitting
+/// both across them in proportion to the liquidity each of them provides, and appending the result
+/// to `fills`. Orders left with nothing to sell are removed.
+///
+/// `available` must be the total liquidity the orders provide, and `sold_amount` no more than it.
+fn fill_orders<LiquidityProvider: Ord + Clone>(
```
msgmaxim, line 427: Apparently all tests call with `sold == bought` (tick == 0), so we if we mix them up a test wouldn't catch it.

state-chain/pallets/cf-pools/src/migrations/remove_fixed_pools.rs
```diff
@@ -0,0 +1,643 @@
...
+			"Wrong number of pools after migration"
+		);
+
+		for (asset_pair, before) in before {
+			let pool = Pools::<T>::get(asset_pair)
+				.ok_or(DispatchError::Other("A pool was lost during the migration"))?;
+
+			// The range order half is not this migration's business, so it must come through byte
+			// for byte.
+			frame_support::ensure!(
```
msgmaxim, line 257: might as well import/use `frame_support::ensure!` so you can simply use `ensure!`

state-chain/pallets/cf-pools/src/migrations/remove_fixed_pools.rs
```diff
@@ -0,0 +1,643 @@
...
+	fn post_upgrade(state: Vec<u8>) -> Result<(), DispatchError> {
+		let before = <Vec<(AssetPair, PoolBefore<T>)>>::decode(&mut &state[..])
+			.map_err(|_| DispatchError::Other("Failed to decode the pre-upgrade state"))?;
+
+		frame_support::ensure!(
+			!old::LimitOrderAutoSweepingThresholds::<T>::exists(),
+			"The auto sweeping thresholds should have been removed"
+		);
+		frame_support::ensure!(
+			Pools::<T>::iter_keys().count() == before.len(),
```
msgmaxim, line 247: AI flagged this: if old pools aren't decoded correctly, you might get `0 == 0` and this will incorrectly pass. It is safer for the old pool to just count the keys to guard against this: `Pools::<T>::iter_keys().count()`.

state-chain/amm/src/limit_orders.rs
```diff
@@ -911,73 +399,1032 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
 pub mod migration_support {
 	use super::*;
+	use crate::common::Pairs;
+	use sp_core::U512;
+
+	/// A number in `0.0..1.0`, as the old representation stored it: a normalised mantissa, so the
+	/// precision does not decay as the value shrinks, and a 256 bit negative exponent, so the
+	/// running product could shrink across effectively unbounded swaps.
+	#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
+	pub struct FloatBetweenZeroAndOne {
```
msgmaxim, line 501: nit: I suspect the PR would have been much smaller (and easier to reivew) if we keep old structs where they were, potentially marking some struct for removal later.

state-chain/amm/src/limit_orders.rs
```diff
@@ -911,73 +399,1032 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
+
+	type Lp = cf_primitives::AccountId;
+	type OldPoolState = PoolStateV9<Lp>;
+
+	fn lp(id: u8) -> Lp {
+		Lp::from([id; 32])
+	}
+
+	/// A price nothing has been bought from.
+	fn untouched() -> FloatBetweenZeroAndOne {
```
msgmaxim, line 1135: nit: the name and the comment are both confusing: the name is too vague and the comment says this is "price" when in reality it is a fraction...

state-chain/pallets/cf-pools/src/migrations/remove_fixed_pools.rs
```diff
@@ -0,0 +1,643 @@
...
+					let (amount_before, original_before) = orders_before
+						.get(&(tick, lp))
+						.ok_or(DispatchError::Other("The migration invented an order"))?;
+
+					frame_support::ensure!(
+						!position_info.amount.is_zero(),
+						"An order with no liquidity left survived the migration"
+					);
+					frame_support::ensure!(
+						position_info.amount <= *amount_before,
```
msgmaxim, line 286: Maybe check that the values are at least somewhat close? Otherwise migrating to all 0 would pass the test too (below for `available_before` as well).

state-chain/amm/src/limit_orders/tests.rs
```diff
@@ -600,41 +357,204 @@ fn swap() {
...
+	assert_ok!(pool_state.mint::<QuoteToBase>(&lp(2), far, 1000.into()));
+
+	// Enough to take the whole of the nearest price and nothing else.
+	let (_output, _remaining, fills) = swap::<QuoteToBase>(&mut pool_state, 2000.into(), None);
+
+	assert_eq!(fills.len(), 2);
+	assert!(fills.iter().all(|fill| fill.remaining_amount.is_zero() && fill.tick == near));
+
+	assert_matches!(
+		pool_state.position::<QuoteToBase>(&lp(0), near),
```
msgmaxim, line 549: Is it strange that requesting a position that does not exist returns an error rather than `None`? I know at some level higher up an error might make sense (like the user doing something to a non-existing order), but I feel like at this level it should simply return Option.

state-chain/amm/src/limit_orders/tests.rs
```diff
@@ -658,24 +574,22 @@ fn boundary_tick_limit_order_consumed_without_price_limit() {
 
 #[cfg(feature = "slow-tests")]
 #[test]
-fn maximum_liquidity_swap() {
+fn every_price_in_the_range_can_be_swapped_out() {
+	// A realistic ceiling for one price, and low enough that the totals stay below the point
+	// where the conversions saturate.
+	const LIQUIDITY_PER_PRICE: Amount = U256([u64::MAX, u64::MAX, 0, 0] /* little endian */);
```
msgmaxim, line 580: Worth being precise that this is essentially `u128::max` (which is exactly the largest possible value an order can be).

state-chain/amm/src/lib.rs
```diff
@@ -303,40 +321,49 @@ impl<LiquidityProvider: Clone + Ord> PoolState<LiquidityProvider> {
...
-			Pairs::Base => self.limit_orders.collect_and_mint::<QuoteToBase>(lp, tick, sold_amount),
-			Pairs::Quote =>
-				self.limit_orders.collect_and_mint::<BaseToQuote>(lp, tick, sold_amount),
+			Pairs::Base => self.limit_orders.mint::<QuoteToBase>(lp, tick, sold_amount),
+			Pairs::Quote => self.limit_orders.mint::<BaseToQuote>(lp, tick, sold_amount),
 		}
 	}
 
-	pub fn collect_and_burn_limit_order(
+	pub fn burn_limit_order(
```
msgmaxim, line 371: Nit: I know this is pre-existing code, but while we are working here it would be nice to document what the returned values are.

state-chain/custom-rpc/src/lib.rs
```diff
@@ -1031,7 +1031,6 @@ pub trait CustomApi {
 		base_asset: Asset,
 		quote_asset: Asset,
 		lp: Option<state_chain_runtime::AccountId>,
-		filled_orders: Option<bool>,
```
msgmaxim, line 1034: We could consider keeping this backwards compatible by keeping `filled_orders` as a dummy parameter.

state-chain/custom-rpc/src/order_fills.rs
```diff
@@ -99,112 +99,84 @@ where
 	})
 }
 
-fn order_fills_for_pool<'a>(
+/// Limit order fills are reported by the pool as they happen, since a swap pays the proceeds out
+/// there and then rather than leaving them on the order to be collected.
```
msgmaxim: > rather than leaving them on the order to be collected

IMO this is just noise (no need to mention how things worked in the past, especially since paying out immeiately is the natural choice).

state-chain/pallets/cf-pools/src/lib.rs
j4m1ef0rd: Yes because It's possible that the limit order could use the funds that are swept from the range order in the following update. Unlikely situation, but still valid I think.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -1106,29 +1106,35 @@ impl<T: Config> SwappingApi for Pallet<T> {
...
+					output_amount,
+					remaining_input_amount,
+					limit_order_fills,
+				} = pool.pool_state.swap(side, input_amount, None);
+
+				// Any leftover amount means we ran out of liquidity. It'a all or nothing at this
+				// point.
+				ensure!(remaining_input_amount.is_zero(), Error::<T>::InsufficientLiquidity);
+
+				// The pool doesn't hold the proceeds of the limit orders it just bought into, so
```
msgmaxim: Again, no need to referece the old design where the pool held the proceeds.

state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -2387,22 +2314,18 @@ impl<T: Config> Pallet<T> {
...
-								lp: lp.clone(),
-								id: id.into(),
-								tick,
-								sell_amount: position_info.amount,
-								fees_earned: Default::default(),
-								original_sell_amount: collected.original_amount,
-							})
-						} else {
-							None
+							.expect("Cache is always in sync with the pool");
```
msgmaxim: Should we `log_or_panic` here instead? The fallback would be to skip the entry.

state-chain/pallets/cf-pools/src/migrations.rs
```diff
@@ -18,6 +18,7 @@ use crate::{Pallet, STORAGE_VERSION_U16};
 use cf_runtime_utilities::PlaceholderMigration;
 use frame_support::migrations::VersionedMigration;
 
+mod remove_fixed_pools;
 mod set_minimum_limit_order_amounts;
```
msgmaxim: Do we want to remove this migration (it should run during 2.3 upgrade)?

state-chain/pallets/cf-pools/src/lib.rs
msgmaxim: This error is now unreachable.

msgmaxim, line 549: Worth leaving a comment to say that `collected_fees` and `bought_amount` are only here for historical reasons and will always be 0.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1322,331 +1332,144 @@ fn handle_zero_liquidity_changes_set_range_order() {
...
-			// This swap should take us over the threshold for ASSET:
-			assert!(LiquidityPools::swap_single_leg(ASSET, STABLE_ASSET, 2_000).is_ok());
-		})
-		.then_execute_at_next_block(|_| {
-			assert_eq!(get_balance(&ALICE), (5000, 15_063));
-		});
-}
-
-#[test]
-fn cancel_all_limit_orders_for_account() {
```
msgmaxim, line 1411: Did you remove this intentionally?

state-chain/pallets/cf-pools/src/migrations/remove_fixed_pools.rs
j4m1ef0rd: Its expected for it to be smaller than amount before because it could have filled any amount. We check that its not 0 above because an empty order should have been removed, but any other value <= the amount before is correct.

state-chain/pallets/cf-pools/src/tests.rs
```diff
@@ -1322,331 +1332,144 @@ fn handle_zero_liquidity_changes_set_range_order() {
...
+		assert_eq!(get_balance(&BOB), (0, 3_334));
 
-		// All Alice's orders must be closed, Bob's order is untouched:
-		assert_eq!(count_orders(ASSET_1, ALICE), (0, 0));
-		assert_eq!(count_orders(ASSET_2, ALICE), (0, 0));
-		assert_eq!(count_orders(ASSET_1, BOB), (1, 0));
-	});
-}
-#[test]
-fn can_update_all_config_items() {
```
msgmaxim, line 1474: Instead of removing this test, I think we should make it work on `SetMinimumLimitOrderAmount`. I see that we already have `minimum_limit_order_amount_config_governance_only_and_emits_event`, but I think we should keep the convention of using `can_update_all_config_items` like we do in other pallets.

j4m1ef0rd: We use the update_pallet_config with the only item (SetMinimumLimitOrderAmount) int he setup for the other tests, so it is covered. But i agree we should keep the convention and explicitly test it.

state-chain/amm/src/limit_orders.rs
j4m1ef0rd: Yep, much cleaner. You will need to look at all changes instead of the commit to see the diff with main because I didn't want to rebase after the review.

state-chain/pallets/cf-pools/src/migrations.rs
j4m1ef0rd: yep, just needed to merge main.

state-chain/pallets/cf-pools/src/migrations/remove_fixed_pools.rs
msgmaxim: My point is, there is nothing in `post_upgrade` that makes sure that "funds don't disappear".

dandanlen:
I did a first AI-assisted review of this and it came up with one significant issue: 

In `fill_orders`, flooring both the sell and buy amounts independently means that the rounding on both sides might have different magnitudes, and this in turn means that the effective price of the fill can be below the LP's limit price, or conversely that an LP could get more than their fair share of the swap. 

Say you're and LP selling BTC and buying USDC: rounding down by 1 satoshi is worth much more than rounding down by 1microUSD. So the leading LPs in the queue sell less than they should for the awarded buy amount. The last LP in the queue ends up compensating for this and basically eats the difference. This can even cause the final fill fall below the limit price.

After digging a bit deeper it appeared that there's an incentive to take advantage of this: an LP could generate an ID that ensures they would sort first in the Map, then enter lots if 10$ orders with different IDs. A test with 500x10$ orders and 1x5000$ order against a single 10k$ swap showed that this approach allows the exploiter to gain a 25bp advantage over the 'honest' single order.

The solution (or at least, an improvement) seems to be to derive the bought amount from the sold amount rather than deriving it independently from the inputs. I'll push a commit with this fix, and some tests that target this edge case.

state-chain/pallets/cf-pools/src/migrations/remove_fixed_pools.rs
```diff
@@ -230,21 +234,48 @@ impl<T: Config> UncheckedOnRuntimeUpgrade for Migration<T> {
...
 	#[cfg(feature = "try-runtime")]
 	fn post_upgrade(state: Vec<u8>) -> Result<(), DispatchError> {
-		let before = <Vec<(AssetPair, PoolBefore<T>)>>::decode(&mut &state[..])
-			.map_err(|_| DispatchError::Other("Failed to decode the pre-upgrade state"))?;
+		use frame_support::ensure;
+
+		type BalancesBefore<T> =
+			BTreeMap<(<T as frame_system::Config>::AccountId, Asset), AssetAmount>;
+
+		let (count_before, before, balances_before) =
```
msgmaxim: nit: for some reason AI defaults to overly short names. IMO `pools_before` as a variable name would be a lot more helpful.

dandanlen:
@j4m1ef0rd thank you for this, it's quite a chunk of work. 

I pushed one other adjustment to the fill apportionment algorithm: instead of giving the dust to the last LP, it just doesn't give it to anyone, it stores it as a protocol surplus in the balances pallet. I think this is cleanest: nobody has any incentive either way. The dust might build up over time but at least this way it's accounted-for somewhere. 

Before putting this on mainnet I would like us to put some thought into how address the weight question. Previously, weight per tick was effectively constant, but now it scales with the number of limit orders per tick, which opens up the possibility of DoS. 

It would be good to quantify how much headroom we have here, also considering that we re-execute a swap if slippage limits are hit, for example. How many orders is too many? I can't see much change to the benchmarks - where did the numbers quoted above come from? What machine?

I'm not sure we can account for the weight in the on-initialize function. At that point we don't know how many limit orders there will be come execution time. I had a think with the AI and so far here are a couple of ideas:
- Set a cap on the number of orders per tick, ejecting the lowest-value order if capacity is reached. <- This is the simplest solution but might require also restricting orders to 1 per LP per tick to prevent trivial sybil attacks
- Adopt a hybrid model where we keep the old accounting model for swap execution and then 'lazily' emit the fill info (basically, we always eagerly execute, but settlement can be deferred if the block is full). <- this is much more involved ofc

j4m1ef0rd:
> but now it scales with the number of limit orders per tick, which opens up the possibility of DoS.

That's why we needed the minimum order amount protection before this was viable. I think that should be good enough for now. Just need to make sure it's set to a reasonable value for every pool.

> where did the numbers quoted above come from? What machine?

I ran it all on my m1. So not the most accurate numbers but should be close enough to reason with.

dandanlen:
We need some basis for deciding if it's enough. Right now it's a bit hand-wavy, I would prefer if we quantify and state the assumptions more precisely. Based on the numbers above, it looks like 10k orders would be enough to fill up the block budget. But this was on an M1, and real hardware is likely to perform worse. With a minimum order size of ~5 USD that would  be < 50k$ - is that ok? I'm not sure and I'd like to think of alternatives. 

I'll try to run the numbers on representative hardware.

j4m1ef0rd:
The one major optimization I think we should do for DOS defense is to separate out the limit order storage from the amm pool. We read and decode the entire pool of orders whether we are going to use them or not, so if we store them in a map and read them via tick, then we will only read/write the orders that we use as we use them. This means that spam orders at unrealistic ticks will have no effect on the block weight for swap execution. And if the spam orders are at a usable tick, then they will get consumed and not be a lingering problem. 
With that optimization, I think a $10 minimum would be safe. I'm still trying to get some solid numbers to prove that but I think it is already safer than what is currently on mainnet where no minimum means anyone could bloat the amm pool decoding.

PR #6836, opened 2026-09-01
feat: skip searching for keys an election can never have

nmammeri:
# Pull Request

Closes: PRO-3036

## Summary

This PR uses component storage kind to skip `IndividualComponents::iter_prefix` for for the bitmap-only election systems such as the block-height witnesser, all block witnessers which count for most election systems.

* `VoteStorage::component_storage_kind` drops its &PartialVote argument  which was only used by composites.  The type of storage components a vote occupies is a property of the election system, not of the vote itself.
*  New ElectoralSystemRunner::election_component_storage_kind(election_identifier). CompositeRunner reads the member electoral system off the identifier's extra and asks that member's storage.
*  Three call sites use it: check_election_consensus (skips whichever storage is unusable), do_vote (previously asked the storage with the vote — same value, as check_vote_mismatch guarantees the vote's variant matches the identifier's), and delete_vote (previously hardcoded Both).
* Similarly optimises `clear_election_votes` by skippping searching the trie for keys an election can never have.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/pallets/cf-elections/src/vote_storage.rs
```diff
@@ -209,8 +220,12 @@ mod tests {
 		);
 	}
 
+	/// The composite cannot narrow to a member - the identifier names it, not the storage - so it
+	/// has to cover both, and every member's components must fall inside that. The narrowing is
+	/// tested against the real runtime composites in
+	/// `state_chain_runtime::chainflip::witnessing::bitcoin_elections`.
```
MxmUrw, line 226: Is the composit's `component_storage_kind()` every used?

MxmUrw (commented):
Looks good! I'll have a second look tomorrow but doing this totally makes sense.

state-chain/pallets/cf-elections/src/vote_storage.rs
nmammeri: No it is not used,  we mainly use the the election system `election_component_storage_kind`.

marcellorigotti:
It would be nice to keep the governance `clear_election_votes` extrinsic clear both components regardless of the election type.

nmammeri:
> It would be nice to keep the governance `clear_election_votes` extrinsic clear both components regardless of the election type.

Good idea, I've updated the clear_election_votes function to take it a flag to force cleaning all components

marcellorigotti (approved):
LGTM! 
But if possible let's also add a test for the full pipeline, voting -> consensus -> cleanup/deletion, that verifies that everything works as expected

PR #6838, opened 2026-09-02
fix: PRO-3076 report the underlying error when the retrier gives up

dandanlen:
# Pull Request

Closes: [PRO-3076](https://linear.app/chainflip/issue/PRO-3076/retryrpc-ignores-inner-error)

## Summary

When a request submitted to `RetrierClient` reached its retry limit — or when no client was
available to serve it — the retrier simply dropped the response channel. The caller's `rx.await`
then failed with a `RecvError`, which `inner_to_return_type` turned into a generic
`"Maximum attempt of ... reached for request ..."` error. The actual RPC error was only ever
written to the logs, so diagnosing a failed request meant searching back through the log history
for the right retrier name and message format.

This changes the response channel from `oneshot::Sender<BoxAny>` to
`oneshot::Sender<Result<BoxAny>>`, and sends the error of the last attempt back to the caller when
we stop retrying:

- the retry-delay future now carries the `anyhow::Error` of the attempt that caused the delay, so
  it is still available at the point where we give up;
- reaching the retry limit sends `Err(... Last error: {last_error})`;
- both "no clients available" paths send an error instead of silently dropping the request.

The public API is unchanged: `request_with_limit` still returns `Result<T>` (now with a useful
error) and `request` still returns `T`. `NoRetryLimit` requests are never abandoned by the
retrier, so that impl keeps the existing "this cannot happen" `expect`.

### Judgement calls

- The issue suggested two possible directions: include the error in the retrier error, or add a
  unique id to every retrier log message so the logs can be filtered. This PR does the first only
  — it addresses the title of the issue and keeps the diff small. The log-filtering idea is
  independent and can be done separately if still wanted.
- The last error is embedded in the message (`... Last error: {e}`) rather than attached with
  `anyhow::Context`, so that it shows up for callers that format the error with `{}` rather than
  `{:#}`/`{:?}`.

## API Changes

None. This is engine-internal; no state chain, RPC, extrinsic or event encodings are affected.
The only externally visible difference is the text of the error returned by
`RetrierClient::request_with_limit`, which now includes the underlying failure.

### RPCS

No changes.

### Extrinsics & Events

No changes.

## Verification

Commands run on this branch, on an x86-64 Linux container:

- `cargo check -p chainflip-engine --lib` — passed (`Finished dev profile ... EXIT:0`).
- `cargo check -p chainflip-engine --lib --tests` — passed, so the test code compiles too.
- `cargo clippy -p chainflip-engine --all-targets -- -D warnings -A deprecated -W clippy::allow_attributes` — passed with no warnings.
- `cargo test -p chainflip-engine --lib retrier::` — 12 passed, 0 failed, 1 ignored
  (`request_always_fails` is `#[ignore]`d in the existing code because it runs forever).
  This includes `using_the_request_with_limit_fails_after_some_attempts`, which this PR extends to
  assert that the error returned to the caller contains the underlying request error, and
  `return_error_when_retry_limit_if_no_client_ready`, which covers the "no clients available" path.
- `cargo fmt --all` — produced no changes.

## Not verified

- The full workspace build and test suite (`cargo nextest run`) — only the `chainflip-engine`
  package was built and only its `retrier::` tests were run.
- Any other package's clippy/check.
- Integration, bouncer, or any end-to-end tests; nothing was run against a localnet.
- Behaviour against a real node: the change was not exercised on a running engine, so the improved
  error text has only been observed in unit tests.
- Runtime benchmarks/weights — not applicable, no runtime code is touched.

---

## *Checklist*

* The PR is complete:
  * [ ] Reviewers are assigned.
  * Tests:
    * [ ] Proptests for important invariants. — n/a
    * [ ] Integration tests for runtime-wide behaviours. — n/a, engine-only change
    * [ ] Bouncer tests for E2E features involving external chains. — n/a
  * [x] Benchmarks: none needed, no runtime code touched.
  * [x] Migrations: none needed.
  * [x] Unrelated changes are isolated in dedicated commits (single commit).

https://claude.ai/code/session_01Shpu2sKzy9nXV7UmYibQxK

---
_Generated by [Claude Code](https://claude.ai/code/session_01Shpu2sKzy9nXV7UmYibQxK)_


dandanlen:
CI note: `post-check-turbo / bouncer` failed on 2a4fc3c, in the `Run HeuteLeiderNicht.einfach.exe 🦺` step. I don't believe this failure belongs to this PR:

- It died in `setup_for_test.sh` → `commands/setup_vaults.ts`, ~15s in, while inserting the new agg keys into the external-chain contracts/programs (the bouncer log's last line is `Arbitrum key inserted`, right after `Funding Solana new aggregate key`). No bouncer test body ran.
- `post-check / bouncer` **passed** on the same commit (11 minutes, same `fast_bouncer.sh`, same setup path), as did `post-check-old-rpcs`'s sibling checks, `cargo-clippy`, `cargo-fmt`, both unit-test jobs, coverage, the upgrade check and all builds.
- The engine logs uploaded from the failing run contain no errors or panics.
- This PR only changes how `RetrierClient` reports an error that it already returned before (`engine/src/retrier.rs`); the external-chain key insertion in the failing step doesn't involve the engine at all.

I'll re-run that job once when the workflow run finishes. If it fails again in the same place I'll dig in rather than assume it's environmental.

---

dandanlen:
Follow-up: I can't do the re-run myself — `rerun-failed-jobs` returns `403 Resource not accessible by integration` for me. So `post-check-turbo / bouncer` stays red on 2a4fc3c until someone with write access re-runs that job.

Everything else on this commit is green, including the other two bouncer jobs (`post-check / bouncer` and `post-check-old-rpcs / bouncer`), which ran the same `fast_bouncer.sh` to completion. Making the localnet vault-setup step more robust would be unrelated to this diff, so I'm not widening the PR to do it.

---

dandanlen:
Correction to my two comments above: no re-run is needed. `post-check-turbo` is invoked with `continue-on-bouncer-error: true` in `ci-development.yml`, so that bouncer job is non-blocking by design — the `Release Chainflip Development` run for 2a4fc3c concluded **success** despite it. Sorry for the noise.

State as of now: no merge conflict with `main` (7 commits ahead, none touching `engine/src/retrier.rs`), CI green, waiting on review.

---

engine/src/retrier.rs
```diff
@@ -400,10 +414,10 @@ impl RetryLimitReturn for u32 {
 	}
 
 	fn inner_to_return_type<T: Send + 'static>(
-		inner: Result<BoxAny, tokio::sync::oneshot::error::RecvError>,
+		inner: Result<RequestResponse, tokio::sync::oneshot::error::RecvError>,
 		log_message: String,
 	) -> Self::ReturnType<T> {
-		let result: BoxAny = inner.map_err(|_| anyhow::anyhow!("{log_message}"))?;
+		let result: BoxAny = inner.map_err(|_| anyhow::anyhow!("{log_message}"))??;
```
MxmUrw: aren't we still just dropping the error?

dandanlen: Fair question — the line looks almost unchanged, but the error being discarded there is a different one.

`inner` is now `Result<Result<BoxAny, anyhow::Error>, RecvError>`:

- the **outer** `Result` is the oneshot channel. Its error is `RecvError`, which carries nothing but "the sender was dropped" — that's what `map_err` was swallowing, and it's also the case that should no longer happen, because every give-up path now sends an error instead of dropping the sender.
- the **inner** `Result` is the request outcome, and the second `?` (that's the new one — `??` rather than `?`) is what propagates the actual RPC error, e.g. ``Retrier eth_rpc: Reached maximum attempts of `5` for request `eth_getTransactionReceipt(0x…)` with id `12`. Last error: <the node's error>``.

That said, the appearance was doing no one any favours, so ecd427c makes it explicit:

- `map_err(|e| ...)` now includes the `RecvError` rather than dropping it, with a comment saying why it carries no detail;
- the fallback message no longer claims the retry limit was reached — that case now comes through the inner error — and instead says the retrier dropped the request without returning a result, which is the only way that branch is reachable.

Re-ran `cargo test -p chainflip-engine --lib retrier::` on ecd427c: 12 passed, 1 ignored (the pre-existing `request_always_fails`).

---

MxmUrw (approved):

dandanlen:
CI on ecd427c: 58 jobs green, 3 red — none of them from this diff, and all three need a re-run I can't trigger (`rerun-failed-jobs` still returns `403 Resource not accessible by integration` for me).

**`post-check / chainspec-compatibility (berghain)` and `post-check-turbo / chainspec-compatibility (berghain)`**

`ci/scripts/check_node_syncing.sh` starts `chainflip-node` against the network's chainspec and fails if `system_health` reports 0 peers for 60 seconds. Both berghain jobs sat at `⏩ Warping, Downloading finality proofs … (0 peers)` for the full minute and timed out; the sisyphos and perseverance matrix legs passed in the same run. That's mainnet bootnode reachability from the runner, on a 60-second budget.

It can't be this PR: the binary under test is `chainflip-node`, and this diff only changes `chainflip-engine`'s retrier — no node code is linked into it. The same check passed on the previous head (2a4fc3c); the only code change since is a closure binding and a message string.

**`post-check-turbo / bouncer`** — same as [before](https://github.com/chainflip-io/chainflip-backend/pull/6838#issuecomment-5503671175): non-blocking by design (`continue-on-bouncer-error: true`), and `post-check / bouncer` plus `post-check-old-rpcs / bouncer` both passed on this commit.

Raising the peer-wait tolerance in `check_node_syncing.sh` would be unrelated to this PR, so I'm not touching it. Someone with write access re-running the two chainspec jobs should be enough.

---

PR #6844, opened 2026-09-03
feat: batch election instances votes into one engine extrinsic call

nmammeri:
# Pull Request

Closes: PRO-3081

## Summary

* This PR adds an optimisation where it allows engines to batch all votes for all elections instances into one extrinsic call to the state chain. Now every election instance (currently 8) submits its own vote and incurring the full extrinsic submission fixed overhead (signature verifications,  SignedExtension pipeline nonce, tip/fee charging, mortality check)

 **State chain**
* Split Elections::vote into an instance-agnostic authorise_voter (now cf_traits::elections, returning a VoterContext) and a per-instance Pallet::do_vote. The existing vote extrinsic is unchanged in behaviour — it just calls both.
* New Environment::submit_elections_votes extrinsic taking votes for all instances at once (AllElectionInstancesVotes, one Option field per instance). It authorises once, then records each instance's votes.

 **Engine**
* New VoteBatcher, shared by every Voter: collects votes off a channel and submits one submit_elections_votes. Voters all tick off the same block stream, so a short quiet period (block_time/40) catches the whole set, with a hard cap (block_time/5) so a slow instance delays only its own votes by a block.
* Overflow beyond MAXIMUM_VOTES_PER_EXTRINSIC goes into a second batch rather than forcing the gathered one out early. send never blocks a voter — a full channel drops the votes and the election is voted in again next block.

**Config switch**
* Environment::ElectionVoteBatchingDisabled, set via update_pallet_config (governance). The engine watches it on unfinalized blocks so it takes effect without a restart, and falls back to the per-instance Elections::vote path — deliberately not a single-instance batch, since the point is to leave the batched dispatch path entirely.

 * block without bathcing 
Screenshot 2026-09-03 at 12 17 36

* block with batching on
Screenshot 2026-09-03 at 12 17 57

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/traits/src/elections.rs
```diff
@@ -0,0 +1,92 @@
...
+	/// The weight of [`Self::vote_all`], summed over the instances `votes` actually target.
+	fn vote_all_weight(votes: &Self::Votes) -> Weight;
+
+	/// Record `votes` in every instance they target.
+	///
+	/// Instances are independent: one failing must neither abort the rest nor roll back their
+	/// storage, which is what a caller submitting a separate extrinsic per instance gets today.
+	/// Returns the failures - each paired with an implementation-defined index identifying the
+	/// instance - for the caller to report, rather than failing the whole call.
+	fn vote_all(context: &VoterContext<T>, votes: Self::Votes) -> Vec<(u32, DispatchError)>;
```
MxmUrw: Could we use `ForeignChain` instead of `u32` for the error identification? That's how we compute the index, right?

state-chain/traits/src/elections.rs
```diff
@@ -0,0 +1,92 @@
...
+use sp_std::prelude::*;
+
+/// Everything recording a vote needs about the caller that does *not* depend on which elections
+/// instance is being voted in.
+pub struct VoterContext<T: Chainflip> {
+	pub epoch_index: EpochIndex,
+	pub authority: <T as Chainflip>::ValidatorId,
+	pub authority_index: AuthorityCount,
+	pub block_number: BlockNumberFor<T>,
+}
```
MxmUrw, line 36: Yes, I like this.

state-chain/pallets/cf-elections/src/lib.rs
```diff
@@ -1368,28 +1371,37 @@ pub mod pallet {
...
-			// Box to avoid RuntimeCall size
-			authority_votes: Box<AuthorityVotes<T, I>>,
+		pub fn do_vote(
+			context: &VoterContext<T>,
+			authority_votes: AuthorityVotes<T, I>,
 		) -> DispatchResult {
-			let (epoch_index, authority, authority_index) = Self::ensure_can_vote(origin)?;
+			let VoterContext { epoch_index, authority, authority_index, block_number } = context;
+			let (epoch_index, authority_index, block_number) =
+				(*epoch_index, *authority_index, *block_number);
```
MxmUrw: Is this line necessary?

engine/src/elections/vote_batcher.rs
```diff
@@ -0,0 +1,386 @@
...
+
+/// How long to wait after the most recent votes before deciding no more are coming.
+///
+/// Every voter ticks off the same block stream at the same offset, so their votes arrive within
+/// a few milliseconds of each other and a short quiet period is enough to catch the whole set.
+/// This is what keeps the common case fast: the voters submit half a block in, timed to land in
+/// the next block, so holding a batch for the full window would eat into that. A quiet period
+/// also adapts to however many instances have something to say, which is not always all of them.
+///
+/// Derived from the block time, like [`BATCHING_WINDOW`], so the two keep their 8:1 ratio if the
```
dandanlen: nit: if the 8:1 ration is important, we could also define `QUIET_PERIOD = BATCHING_WINDOW / 8`.

engine/src/elections/vote_batcher.rs
```diff
@@ -0,0 +1,386 @@
...
+			.storage_value::<pallet_cf_environment::ElectionVoteBatchingDisabled<Runtime>>(
+				block.hash,
+			)
+			.await
+		{
+			Ok(disabled) => {
+				let enabled = !disabled;
+				// Only report transitions - this runs every block and is otherwise silent.
+				if batching_enabled.swap(enabled, Ordering::Relaxed) != enabled {
+					tracing::info!(
```
dandanlen: Here, when disabling, should we not clear the pending batches? Otherwise if enabled again, it will submit stale votes, no?

engine/src/elections.rs
```diff
@@ -186,15 +191,21 @@ where
 						})
 					}))
 				}).chunks(MAXIMUM_VOTES_PER_EXTRINSIC as usize).for_each_concurrent(None, |votes| {
-					let state_chain_client = &self.state_chain_client;
 					async move {
 						for (election_identifier, _) in votes.iter() {
 							self.log(Level::DEBUG, &format!("Submitting vote for election: '{:?}'", election_identifier));
 						}
-						// TODO: Use block hash you got this vote tasks details from as the based of the mortal of the extrinsic
```
dandanlen, line 194: Was this comment deliberately deleted or should we keep track of this in an issue somewhere?

engine/src/elections.rs
```diff
@@ -186,15 +191,21 @@ where
...
-						state_chain_client.submit_signed_extrinsic(pallet_cf_elections::Call::<state_chain_runtime::Runtime, Instance>::vote {
-							authority_votes: Box::new(BTreeMap::from_iter(votes).try_into().unwrap(/*Safe due to chunking*/)),
-						}).await;
+						let votes = BTreeMap::from_iter(votes).try_into().unwrap(/*Safe due to chunking*/);
+
+						if self.vote_batcher.batching_enabled() {
```
dandanlen: We're leaking the batching_enabled abstraction. The `VoteBatcher` is responsible for maintaining the batching-enabled flag, but it doesn't use it. This task uses the flag but has no control over it. 

I feel like we could isolate this better, for example by putting it all in the batcher (which I guess then becomes a general `VoteSubmitter`). Then we have all the logic in one place.

engine/src/elections/vote_batcher.rs
```diff
@@ -0,0 +1,386 @@
...
+///
+/// Separated from the task so the batching decisions can be tested against a clock that is
+/// passed in, with no channel or chain client involved.
+#[derive(Default)]
+struct PendingBatches {
+	batches: Vec<AllElectionInstancesVotes>,
+	/// When the batches go out regardless, set by the first votes gathered.
+	hard_cap: Option<tokio::time::Instant>,
+	/// When to send: `QUIET_PERIOD` after the latest votes, but never past `hard_cap`.
+	deadline: Option<tokio::time::Instant>,
```
dandanlen: nit: these are always synchronised (both None or both Some)

We could instead store `Option<(Instant, Instant)>`.

engine/src/elections/vote_batcher.rs
```diff
@@ -0,0 +1,386 @@
...
+		match state_chain_client
+			.storage_value::<pallet_cf_environment::ElectionVoteBatchingDisabled<Runtime>>(
+				block.hash,
+			)
+			.await
+		{
+			Ok(disabled) => {
+				let enabled = !disabled;
+				// Only report transitions - this runs every block and is otherwise silent.
+				if batching_enabled.swap(enabled, Ordering::Relaxed) != enabled {
```
dandanlen: nit: We could use a watch channel instead - this exposes a `changed()` method and we don't need to implement it ourselves. Then this becomes a process that simply reports the current value, and the logic of when we transition can be handled in `submit` and we can handle the case where there's a transition by dropping queued batches if needed.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -865,6 +899,53 @@ pub mod pallet {
...
+		}
+
+		/// Apply a list of configuration updates to the pallet.
+		///
+		/// Requires Governance.
+		#[pallet::call_index(15)]
+		#[pallet::weight(<T as frame_system::Config>::SystemWeightInfo::set_storage(updates.len() as u32))]
+		pub fn update_pallet_config(
+			origin: OriginFor<T>,
+			updates: BoundedVec<PalletConfigUpdate, ConstU32<100>>,
```
dandanlen: nit: we can be more conservative: we'll never need 100 updates here, 10 would be plenty.

state-chain/pallets/cf-environment/src/lib.rs
```diff
@@ -865,6 +899,53 @@ pub mod pallet {
...
+
+		/// Record a validator's election votes across every `pallet-cf-elections` instance in
+		/// one extrinsic.
+		#[pallet::call_index(14)]
+		#[pallet::weight((
+			T::ElectionInstances::authorise_voter_weight()
+				.saturating_add(T::ElectionInstances::vote_all_weight(votes)),
+			DispatchClass::Operational,
+		))]
+		pub fn submit_elections_votes(
```
dandanlen, line 910: Why is this in the environment pallet :( 

I guess I can see why: we don't want to have this repeated across every elections instance...

state-chain/runtime/src/chainflip.rs
```diff
@@ -1302,3 +1307,163 @@ impl cf_traits::ChainflipNetworkInfo for ChainflipNetworkProvider {
...
+/// `Environment::submit_elections_votes` extrinsic.
+///
+/// Each field is optional so a validator can target any subset of instances - in practice it
+/// sends whichever ones produced votes this block.
+///
+/// Fields are ordered by pallet instance: the chain-agnostic `generic` instance first, then one
+/// field per chain in `ForeignChain` order. Adding a chain means appending a field here and a
+/// `vote_in_weight!` and `vote_in!` line below, keyed by its `ForeignChain` index.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Encode, Decode, DecodeWithMemTracking, TypeInfo)]
+pub struct AllElectionInstancesVotes {
```
dandanlen: Can this be defined inside the macro?

dandanlen: Also, nit: can we put this under the elections module instead of just tacking it on to this (already quite large) file?

state-chain/runtime/src/chainflip.rs
```diff
@@ -1302,3 +1307,163 @@ impl cf_traits::ChainflipNetworkInfo for ChainflipNetworkProvider {
...
+
+	fn vote_all_weight(votes: &Self::Votes) -> Weight {
+		macro_rules! vote_in_weight {
+			($field:ident, $instance:ty) => {
+				votes.$field.as_ref().map_or(Weight::zero(), |v| {
+					pallet_cf_elections::Pallet::<Runtime, $instance>::do_vote_weight(v.len() as u32)
+				})
+			};
+		}
+		vote_in_weight!(generic, ())
```
dandanlen: This macro (and `vote_in!` macro below) both have quite similar structure to the `election_instances` macro, we can probably combine all three in a single macro.

dandanlen (commented):
Looking good!

engine/src/elections/vote_batcher.rs
nmammeri: make sense, I don't like Option<(Instant, Instant)>. Will encapsulate into small struct BatchTiming that owns the timing logic along with the constants BATCHING_WINDOW, and QUIET_PERIOD

engine/src/elections/vote_batcher.rs
nmammeri: better to tie the 2 together to keep the ratio

state-chain/pallets/cf-environment/src/lib.rs
nmammeri: exactly we just need one, it can be in a different pallet but I guess it can live in the environment pallet next to the `batch` extrinsic

engine/src/elections/vote_batcher.rs
nmammeri: It might seem so, but pending batches are always submitted at the hard deadline (the first vote arrived + maximum BATCHING_WINDOW) regardless of the switch. Let's say if there is a pending batch and governance disables batching, the pending batch would be submitted in batch mode at the deadline. Any new votes after that would not create any pending batches and will be submitted individually

engine/src/elections/vote_batcher.rs
nmammeri: The change detection here is only gating the `info` log line, because we don't want to log every block. Will remove log line and the atomic swap can be replaces a simple atomic store.

I don't think there is any need to detect changes please see the response for https://github.com/chainflip-io/chainflip-backend/pull/6844#discussion_r3988224294

engine/src/elections.rs
nmammeri: I agree since we are already threading through all instances.  I'll rename it to VoteSubmitter and let it handle batching/non-batching detail.

state-chain/traits/src/elections.rs
nmammeri: Yes, but I don't want to tie it to a ForeignChain, we might want to add another election instance that is unrelated to a chain. u32 is not good either may be an enum representing the instance_name is clearer. It is also carried in the `ElectionInstanceVotesRejected` event.

engine/src/elections/vote_batcher.rs
dandanlen: Ok yeah I can see that now, makes sense. Thanks.

state-chain/pallets/cf-elections/src/lib.rs
nmammeri: removed

engine/src/elections.rs
nmammeri: restored

state-chain/traits/src/elections.rs
MxmUrw: > may be an enum representing the instance_name

sounds good.

state-chain/runtime/src/chainflip.rs
nmammeri: Agree combining them into 1 macro is cleaner.

state-chain/runtime/src/chainflip.rs
nmammeri: `AllElectionInstancesVotes` struct also moved inside the macro and moved all of this election related code inside a new elections module

dandanlen (approved):
LGTM!

dandanlen:
Just to confirm this is intentional: the storage item (`ElectionVoteBatchingDisabled`) defaults to false, so batching will become active with the upgrade. Yes?

nmammeri:
> Just to confirm this is intentional: the storage item (`ElectionVoteBatchingDisabled`) defaults to false, so batching will become active with the upgrade. Yes?

Yes, this is intentional batching should be on by default.

PR #6846, opened 2026-09-04
feat: add bsc as a new source of oracle prices

marcellorigotti:
# Pull Request

Closes: PRO-3088

## Summary
- Extend generic election to support querying prices from BSC
- Add missing assets: BNB, TRX, wBTC, cbBTC, DOT
- Update bouncer commands/utilities to support new assets and chain
- Update localnet to rely on the new published images, which now deploy chainlink feeds on bsc

---

* The PR is complete:
  * [x] Reviewers are assigned.


state-chain/custom-rpc/src/lib.rs
```diff
@@ -3761,15 +3765,44 @@ where
...
+					return Ok(Vec::new());
+				};
+				let Ok(quote_asset) = try_migrate_to_historical_type(v20300, quote_asset) else {
+					return Ok(Vec::new());
+				};
+
+				Some((base_asset, quote_asset))
+			} else {
+				None
+			};
```
MxmUrw: Does this function have to return `Ok(_)` in case a migration fails? Maybe it's more appropriate to actually say "error" if the requested data isn't available? If a user requested the data for a specific asset which didn't exist, will he be happy if he instead gets data for all assets *except* the one he requested?

Also, I think this code can be done simpler, because migrations map into Options and tuples transparently:

```suggestion
			let historical_query = try_migrate_to_historical_type(v20300, base_and_quote_asset)
                .ok_or_else(|| Vec::new()) // this preserves current behaviour, but we maybe should return the error here...
```

state-chain/runtime/src/chainflip/witnessing/generic_elections.rs
```diff
@@ -285,19 +289,29 @@ pub fn initial_state(
...
-		(ChainlinkAssetpair::UsdcUsd, Seconds(60 * 5)),
-		(ChainlinkAssetpair::UsdtUsd, Seconds(60 * 5)),
+		(ChainlinkAssetpair::UsdcUsd, Seconds(60)),
+		(ChainlinkAssetpair::UsdtUsd, Seconds(60)),
+		(ChainlinkAssetpair::SolUsd, Seconds(60)),
+		(ChainlinkAssetpair::TrxUsd, Seconds(60)),
+		(ChainlinkAssetpair::DotUsd, Seconds(60)),
+		(ChainlinkAssetpair::CbbtcUsd, Seconds(60 * 5)),
 	]
 	.into();
```
MxmUrw: Why is Cbbtc special here?

state-chain/runtime/src/migrations/oracle_price_bsc.rs
```diff
@@ -0,0 +1,337 @@
...
+			pallet_cf_elections::ElectionConsensusHistoryUpToDate::<Runtime>::clear(u32::MAX, None);
+		let _ = pallet_cf_elections::ElectoralSettings::<Runtime>::clear(u32::MAX, None);
+
+		pallet_cf_elections::ElectoralSettings::<Runtime>::insert(
+			UniqueMonotonicIdentifier::default(),
+			initial.settings,
+		);
+
+		log::info!("Reset generic oracle elections with the BSC price source enabled");
+		Weight::zero()
```
MxmUrw, line 60: Can't we delete all data and then call `elections::initialize()`? In fact we could provide a function to delete all election data and then the migration would be `delete(); initialize();`, wdyt?

state-chain/runtime/src/migrations/oracle_price_bsc.rs
```diff
@@ -0,0 +1,337 @@
...
+				H160(hex_literal::hex!("C333eb0086309a16aa7c8308DfD32c8BBA0a2592")),
+			],
+		},
+		// Perseverance
+		ChainflipNetwork::Testnet => perseverance_and_sisyphos_oracle_feeds(),
+		// Sisyphos
+		ChainflipNetwork::TestnetDev => perseverance_and_sisyphos_oracle_feeds(),
+		// Local development
+		ChainflipNetwork::Development => OracleFeeds {
+			arbitrum: [
```
MxmUrw, line 209: nit: why do persa and sisy get their own function and localnet doesn't?

state-chain/runtime/src/migrations/oracle_price_bsc.rs
```diff
@@ -0,0 +1,337 @@
...
+				// Usdc
+				H160(hex_literal::hex!("09635F643e140090A9A8Dcd712eD6285858ceBef")),
+				// Usdt
+				H160(hex_literal::hex!("c5a5C42992dECbae36851359345FE25997F5C42d")),
+				// Wbtc (BTC feed)
+				H160(hex_literal::hex!("a85233C63b9Ee964Add6F2cffe00Fd84eb32338f")),
+				// Bnb (ETH feed)
+				H160(hex_literal::hex!("4A679253410272dd5232B3Ff7cF5dbB88f295319")),
+				// Dot (SOL feed)
+				H160(hex_literal::hex!("7a2088a1bFc9d81c55368AE168C2C02570cB814F")),
```
MxmUrw: Shouldn't these assets get their own feed?

state-chain/runtime/src/chainflip/witnessing/generic_elections.rs
```diff
@@ -285,19 +289,29 @@ pub fn initial_state(
...
 	let maybe_stale_timeout_overrides: BTreeMap<_, _> = [
-		(ChainlinkAssetpair::UsdcUsd, Seconds(60 * 5)),
-		(ChainlinkAssetpair::UsdtUsd, Seconds(60 * 5)),
+		(ChainlinkAssetpair::UsdcUsd, Seconds(60)),
+		(ChainlinkAssetpair::UsdtUsd, Seconds(60)),
+		(ChainlinkAssetpair::SolUsd, Seconds(60)),
+		(ChainlinkAssetpair::TrxUsd, Seconds(60)),
+		(ChainlinkAssetpair::DotUsd, Seconds(60)),
+		(ChainlinkAssetpair::CbbtcUsd, Seconds(60 * 5)),
 	]
```
MxmUrw: Now that we have different heartbeat frequencies on different chains we should probably have separate overrides for them. Let's discuss!

state-chain/custom-rpc/src/lib.rs
marcellorigotti: Yes agree, returning error is better, fixing this 👌

state-chain/runtime/src/chainflip/witnessing/generic_elections.rs
marcellorigotti: I just followed the same (old) behaviour for Usdc/Usdc, we used 5m extra instead of 1m.
Don't have a strong opinion on this one tbh.

state-chain/runtime/src/migrations/oracle_price_bsc.rs
marcellorigotti: Yes we could, but idk not sure about it, we should properly reset the pallet completely to be sure everything works. 
i.e. `initialize` does this check:
`ensure!(Status::<T, I>::get().is_none(), Error::<T, I>::AlreadyInitialized);`

so it becomes more `reset()` -> `initialize()`.

I don't have a strong opinion here as well, maybe it is worth to have this reusable if we ever need this kind of migration again?

state-chain/runtime/src/migrations/oracle_price_bsc.rs
marcellorigotti: Just deduplicating code, localnet uses different addresses while sisy and persa share them

state-chain/runtime/src/migrations/oracle_price_bsc.rs
marcellorigotti: I deployed localnet feeds only on the bsc image following the available asset on bsc mainnet. So new assets available only on arb/eth don't have their contract deployed.
I actually didn't think this would be a problem but can do that if we think this is the case, for reference on sisy/persa we don't have all the assets either because chainlink deploys only some feeds on testnet.

state-chain/runtime/src/migrations/oracle_price_bsc.rs
MxmUrw: > so it becomes more `reset() -> initialize()`

Yes, this is what I had in mind.

> I don't have a strong opinion here as well, maybe it is worth to have this reusable if we ever need this kind of migration again?

I think this isn't the first time we're doing this kind of migration, and I remember having the same thought last time already. So I think it would be useful for the future, then we wouldn't have to review the same deletion/reset code over and over.

state-chain/runtime/src/migrations/oracle_price_bsc.rs
MxmUrw: > I deployed localnet feeds only on the bsc image following the available asset on bsc mainnet. So new assets available only on arb/eth don't have their contract deployed.

Yes I think it makes sense to have separate feeds for separate assets. Then we can target prices for certain assets and don't change another asset's prices as a side-effect.

For the currently undeployed ones, is it possible to copy the mainnet situation and deploy these on arb/eth?

> [..] for reference on sisy/persa we don't have all the assets either because chainlink deploys only some feeds on testnet.

Yeah true, but since for localnet we have the option to do it, we should.

state-chain/runtime/src/chainflip/witnessing/generic_elections.rs
MxmUrw: For the record, we discussed in office and agreed to add separate overrides for the separate target chains. The reasoning is that the timeouts (and resulting `UpToDate`/`MaybeStale`/`Stale` status is per-chain and is meant to reflect chainlink price submissions to that chain - *not* whether the price is stale or not from the consumers' point of view.)

state-chain/runtime/src/chainflip/witnessing/generic_elections.rs
MxmUrw: Discussed in this comment: https://github.com/chainflip-io/chainflip-backend/pull/6846#discussion_r3987317208 .

PR #6871, opened 2026-09-10
feat: Additional debug logs for multisig

msgmaxim:
# Pull Request

## Summary

While doing some p2p/multisig work I noticed a couple of improvements to the logs:

1. ZMQ logs used incorrect subscriber (due to them being in a separate OS threads unlike most of the resf of the code). This means no p2p/zmq logs ever surfaced in logs. I added `spawn_with_tracing` so the thread now use the same subscriber as the main thread.
2. Added additional debug logs to the broadcast stage to help us debug any reachability issues between nodes:
  - for each node we print the number of peers that didn't received their message (omitted if all received);
  - for our own node we also print the exact parties that we couldn't reach.
Currently reachability issues are mostly undetectable since broadcast verification recovers from them (via redundant echo messages), only logging on failures to recover.

- 

## *Checklist*

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


engine/multisig/src/client/common/broadcast_verification.rs
```diff
@@ -169,6 +205,33 @@ where
...
+				"Broadcast delivery gaps",
+			);
+		}
+	}
+
+	// Same as above, but with more info relevant to our node (prints exact nodes we couldn't
+	// reach).
+	let missed_us = reporters_missing_our_broadcast(&verification_messages, own_idx);
+	if !missed_us.is_empty() {
+		debug!(
```
j4m1ef0rd: nit: should this also be within the `if enabled!(Level::DEBUG)`?

j4m1ef0rd (approved):

utilities/src/with_std/task_scope.rs
```diff
@@ -760,7 +760,9 @@ impl<Error: Debug + Send + 'static> Drop for ScopeResultStream<Error> {
 
 /// Allows async code to run sync/blocking code without blocking the runtime.
 pub async fn without_blocking<C: FnOnce() -> R + Send + 'static, R: Send + 'static>(c: C) -> R {
-	match tokio::task::spawn_blocking(c).await {
+	let dispatch = tracing::dispatcher::get_default(|d| d.clone());
+	match tokio::task::spawn_blocking(move || tracing::dispatcher::with_default(&dispatch, c)).await
+	{
```
dandanlen: I was reading about what the dispatcher is for / where it comes from and the AI notice that it was also missing from here. 

In the `tracing` docs there was also a mention of a `set_global_dispatch` methods, i'll look into whether that's a better solution (it seems like it).

msgmaxim: Yeah, I saw that this was an option, initially decided against it (was a larger change in that it touches the logs that do already work), but I'm OK either way.

dandanlen (approved):

PR #6892, opened 2026-09-16
feat: range order minimum

j4m1ef0rd:
# Pull Request

Closes: PRO-3152

## Summary

Adds a minimum amount that must be in the range order when placing it or updating it. Similar to the limit order minimum that was introduced recently. But the range order one works a little different. Both the base and quote assets must have a minimum set for it to be enforced. If either the base or quote side is over the minimum for that asset, it will pass the check. The assets are not added together, so a sum that is larger than a minimum will still fail. 

### Extrinsics & Events

Add the new setting to the config gov extrinsic.

---

* The PR is complete:
  * [x] Reviewers are assigned.
  * Tests:


state-chain/pallets/cf-pools/src/lib.rs
```diff
@@ -438,6 +445,12 @@ pub mod pallet {
 	pub type MinimumLimitOrderAmount<T: Config> =
 		StorageMap<_, Twox64Concat, Asset, AssetAmount, ValueQuery>;
 
+	/// Minimum amount of an asset that a range order may hold. Set per asset by governance. A
+	/// value of `0` disables the check for that asset.
+	#[pallet::storage]
+	pub type MinimumRangeOrderAmount<T: Config> =
+		StorageMap<_, Twox64Concat, Asset, AssetAmount, ValueQuery>;
```
dandanlen, line 441: Does this really need to be a separate config item? Can we not use the same value for both limit and range orders?

j4m1ef0rd: We could, but I thought it better to keep it as a separate item because they have different weights for read/writes, so the minimum could be quite different. We could always just use one item and set it to the larger of the 2. No real strong opinion here so i'll change it. simple is probably better.

msgmaxim (approved):

state-chain/pallets/cf-pools/src/lib.rs
dandanlen: Yeah I think the benefit having only one config item outweighs the downside. If we decide later that a separate limit is necessary, we can add this.

dandanlen (approved):

PR #6909, opened 2026-09-21
fix: derive affiliate account id from the withdrawal address

dandanlen:
# Pull Request

Closes: PRO-3167

## Summary

Affiliate account ids were derived as `blake2_256(("chainflip/affiliate", broker_id, short_id))`. Since short ids are recycled when an affiliate is deregistered, a broker re-registering at a freed slot got back the *same* account id as the affiliate that previously occupied it — even when the new affiliate is a completely different party with a different withdrawal address.

This adds `withdrawal_address` to the preimage, so a re-registration produces a distinct account id unless it genuinely refers to the same withdrawal account. As a side benefit, the product now has a stable, unique account id per affiliate to work with.

### Backwards compatibility

Existing affiliate ids are unaffected and no migration is needed. The derivation runs only at registration time, and the result is persisted — every consumer resolves the id by lookup rather than recomputing it:

| Consumer | Source of the affiliate id |
|---|---|
| `chainflip/vault_swaps.rs` (short id → account) | `AffiliateIdMapping` |
| `custom-rpc/src/ingress_egress_tracker.rs` | `AffiliateIdMapping` |
| `runtime_apis/impl_api.rs` (`cf_affiliate_details`) | `AffiliateAccountDetails` |
| `AffiliateRegistry::get_short_id` / `reverse_mapping` | storage reads |
| broker-api `register_affiliate` | the `AffiliateRegistration` event |
| Benchmarks | `AffiliateIdMapping` |

Worth noting for the future: there is deliberately no extrinsic to update an affiliate's withdrawal address after registration, so the derivation inputs cannot drift out of sync with the stored id. If such an extrinsic is ever added, it will need to either re-derive and migrate the account or explicitly treat the id as a birth-time value.

### RPCS

None.

### Extrinsics & Events

No encoding changes. `register_affiliate` takes the same arguments and `AffiliateRegistration` has the same fields; only the *value* of the derived `affiliate_id` changes, and only for registrations made after the upgrade.

---

## *Checklist*

* The PR is complete:
  * [ ] Reviewers are assigned.
  * Tests:
    * [ ] Proptests — n/a.
    * [ ] Integration tests — n/a, no cross-pallet behaviour changes.
    * [ ] Bouncer tests — n/a, no external-chain behaviour changes.
  * [x] Benchmarks: unchanged; `register_affiliate` still registers distinct ids per short id.
  * [x] Migrations: none needed, see above.
  * [x] Bouncer typegen: no runtime type changes.
  * [x] Interfaces are clearly documented — the stale doc comment on `register_affiliate` is updated.

### Test

`reregistering_with_new_withdrawal_address_yields_distinct_id` registers and deregisters twice at the same recycled short id, and asserts that different withdrawal addresses yield different account ids — and that identical inputs still derive the same id. Verified as a genuine regression test: it fails on the old derivation (`left: 3815646062608304308, right: 3815646062608304308`) and passes with the fix.


MxmUrw (commented):
Since now the `withdrawal_address` of an affiliate should definitely not be changed, maybe let's add a WARNING comment above the field definition regarding this.

MxmUrw (approved):
Looks good, modulo above comment.

